PR #165, opened 2014-12-26
Implement vimperator-style hints

sebastiansturm:
Add hs.hints.style variable to control hinting style:

hs.hints.style = "vimperator" creates multi-character hints, with the
first character of each hint corresponding to the first character of the
parent application's title

hs.hints.style = "default" (or anything else) creates the usual
single-character hints, unless the available hintChars are exhausted. In
that case, Hammerspoon switches to multi-char hints (but without
application title prefixes)


sebastiansturm:
great, thanks for merging this!

PR #177, opened 2015-01-05
add isMainThread function to hs.crash

Habbie:
With this new function, you can add this to init.lua:

```lua
local function crashifnotmain(reason)
  print("crashifnotmain called with reason", reason) -- may want to remove this, very verbose otherwise
  if not hs.crash.isMainThread() then
    print("not in main thread, crashing")
    hs.crash.crash()
  end
end

debug.sethook(crashifnotmain, 'c')
```

allowing you to abort (causing backtraces in XCode, a Crashlytics report, etc.) when something is not happening on the main thread.

Pre 5d9130325f4635c7b66c72e0b5bec9257f04c447, this Lua code triggers it: `function mywifiwatcher() hs.alert.show(hs.wifi.currentNetwork()) end wifiwatcher = hs.wifi.watcher.new(mywifiwatcher) wifiwatcher:start()`

### DIFF
--- a/Hammerspoon/MJLua.m
+++ b/Hammerspoon/MJLua.m
@@ -3,6 +3,7 @@
 #import "MJUserNotificationManager.h"
 #import "MJConfigUtils.h"
 #import "variables.h"
+#import <pthread.h>
 
 static lua_State* MJLuaState;
 static int evalfn;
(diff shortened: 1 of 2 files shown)


Habbie:
Documentation added (in a different style than crash but consistent with other extensions), should be ready to merge if Travis passes.

Habbie:
Doc style now actually consistent. Again, ready to merge if Travis passes.

Issue #183, opened 2015-01-06
Missing apps with hs.hints


andrewhampton:
Calling hs.hints.windowHints doesn't display hints for all windows. Specifically, it's not working for HipChat and Mailbox. In IRC, Habbie and Ng mentioned they thought this was due to the apps not setting a title for the windows.

Could a potential fix be to discard apps based on window dimensions rather than window titles?

cmsj:
(note: I merged a slightly different thing to your PR, which works well for me, and doesn't seem to include any "unusual' windows it shouldn't)

andrewhampton:
Thanks!

trishume:
This is a real issue and I agree that check should be removed.

However as the original author of that code I remember why I wrote it and it is that every Google Chrome window has an extra hint in some weird corner because there's an invisible window with no title. Perhaps a more specific check would better screen out those annoying hints.

andrewhampton:
Are you still seeing the extra hint for Chrome @trishume? I don't see it.

trishume:
@andrewhampton oh that's interesting. Admittedly I haven't switched to Hammerspoon yet. I'm still on Mjolnir using my original hints code because I've been too lazy to update.

PR #186, opened 2015-01-08
Changing how app list is populated for hints

andrewhampton:
Possible fix for #183 

I'm not sure how to build Hammerspoon, so I've not tested this, but could this potentially be a fix for #183? 

Instead of just checking to see if the window has a title, I see if the window has a width and a height or a title.


cmsj:
(closing because I merged something similar, but different :)

PR #190, opened 2015-01-14
Fix hs.keycodes library creation, so the callbacklib __gc is actually called

cmsj:


cmsj:
I'd really like some more eyes on this ( @Habbie and @asmagill please!)
Right now, the __gc method isn't being called on the callback objects, which causes a crash after a config reload, if a callback function has been registered.
I've reworked the library setup function such that it doesn't crash, but I still don't have a fully formed understanding of these setup steps in my head, and we do it in so many different ways in different extensions, that I could well be missing something.

Thoughts? :)

PR #191, opened 2015-01-14
Reimagined hs.notify

asmagill:
The new hs.notify, ready to check the tires and kick the ashtrays...


cmsj:
I have:

hs.notify.show("Hammerspoon", "", "Config loaded", "")

at the bottom of my init.lua. If I run HS with this PR applied, click on the notification and then reload the config, I get a crash entering releaseNotification() (in the callchain down from notification_gc).

asmagill:
forgot to test withdraw... try it now.

PR #198, opened 2015-01-19
Don't leave an extra copy of app lying around

chdiza:
Move, rather than copy, the .app to the convenient location.  Otherwise, there are at least two copies of Hammerspoon.app lying around.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ PRODUCT_DIR = $(shell xcodebuild -workspace Hammerspoon.xcworkspace -scheme Rele
 $(APPFILE): build $(shell find Hammerspoon -type f)
 	rm -rf $@
 	xcodebuild -workspace Hammerspoon.xcworkspace -scheme Release clean build > build/release-build.log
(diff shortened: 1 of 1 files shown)


cmsj:
The concern I have here is that this is moving something Xcode is expecting to find there - e.g. does it still work if one wants to do an Analyze build (i.e. run HS under Instruments.app?)

chdiza:
I don't know what any of those things are, so I don't know.  Feel free to close.  Perhaps I'll look into how to change the location of the built app The Right Way.

asmagill:
I don't know much myself about Xcode, but I can confirm that moving the app out of the Derived Data folder instead of copying it won't fuss Xcode at all... it will just show that the Hammerspoon.app is missing (red) in the Products list and you'll need to recompile.

In fact, you can remove the entire Derived Data folder (as long as Xcode isn't running at the time) and all Xcode will do is re-scan the project at the next load -- it's slower, but it doesn't break anything.

cmsj:
I think I'm going to close this. I don't find the extra copy too annoying, and I'd rather not mess too much with the build system more than I have to. It already feels fragile!

PR #212, opened 2015-02-04
define hs.mouse.setAbsolutePosition

marshallbrekka:
the getAbsolutePosition was defined twice, so just fix the names/args


cmsj:
thanks!

PR #216, opened 2015-02-10
rewrite isWindow and push_element to use a common internal fn for checking the window status of an element

marshallbrekka:


tmandry:
looks good to me.

cmsj:
Thanks very much @marshallbrekka :)

PR #230, opened 2015-02-25
Update hs.drawing to include basic rectangles

ianclarksmith:
Following the format of the other drawing functions is basic rectangle functionality so people can play with it immediately. Terminology-wise, should *box*, *rect*, or *rectangle* preferred? I've used `hs.drawing.rectangle` to be distinct from `hs.geometry.rect` which is more primitive and appears in the hs.drawing docs.


cmsj:
I was going to go with box, but rectangle is more correct, so I'm persuaded that we should use that :)

Thanks for doing this!

Issue #232, opened 2015-02-25
Patch Lua to CLSLOG path of require()'d files


cmsj:

cmsj:
This would be covered by PR #234

PR #234, opened 2015-02-26
Wrap require() in such a way that we can get Crashlytics logs listing the C libraries loaded, and their on-disk paths

cmsj:


cmsj:
This is potentially *fantasically* dangerous :)

Accordingly, I at least want eyes and +1/-1 from @Habbie and I think it might be useful to have @asmagill too :)

The idea here is that we have a recurring crash report coming into Crashlytics which is from an unsymbolicated extension using lua_call(). Several releases ago, that extension was fixed to use lua_pcall(). My suspicion is that the user involved, has a local build of that extension hanging around somewhere, but the only way to tell, is with something like this patch in everyone's install.

cmsj:
This would close #232

cmsj:
I wonder if this would be safer if the bit that does the searching of the table and string matching, and suchlike, was a separate function that we pcall() out to, so there is more chance that the require() itself succeeds?

cmsj:
Probably also worth CLSLOG()'ing the location of the Hammerspoon.app bundle in setup.lua, now I come to think of it.

Habbie:
```
> x = require({})
...ducts/Debug/Hammerspoon.app/Contents/Resources/setup.lua:147: attempt to concatenate local 'modulename' (a table value)
```

Note this is outside of the part you propose to pcall. Pcall all the things!

Given enough pcall this is good to merge. Looks very useful.

I did do this locally because I couldn't get the output otherwise:
```
 static int crashLog(lua_State *L) {
     CLS_LOG("%s", lua_tostring(L, -1));
+    NSLog(@"%s", lua_tostring(L, -1));

     return 0;
```

Not sure if that means I did something else wrong.

PR #244, opened 2015-03-18
Remove alerts from playback controls for iTunes and Spotify

peterhajas:
I don't think Hammerspoon should show alerts for iTunes and Spotify playback commands.

`hs.alert` is something that config file authors are able to use for this functionality We can remove it from the base extensions.


cmsj:
Good point, thanks!

Issue #237, opened 2015-03-02
Window hints - error - attempt to index a nil value


ztomer:
I occasionally get the following error when a Steam window is opened after Hammerspoon (v0.9.24) was loaded:
(but not when steam is opened before Hammerspoon is loaded)

```lua
...poon.app/Contents/Resources/extensions/hs/hints/init.lua:80: attempt to index a nil value
stack traceback:
	...poon.app/Contents/Resources/extensions/hs/hints/init.lua:80: in function 'displayHintsForDict'
	...poon.app/Contents/Resources/extensions/hs/hints/init.lua:174: in function <...poon.app/Contents/Resources/extensions/hs/hints/init.lua:149>
	[C]: in function 'xpcall'
	...oon.app/Contents/Resources/extensions/hs/hotkey/init.lua:23: in function <...oon.app/Contents/Resources/extensions/hs/hotkey/init.lua:21>

**Comment by cmsj** [comment]
This is very odd, it looks, from the code, like either the Steam window has no :screen() value, or the screen has no :frame() value, neither of which should be possible!

How occasionally does it happen? I guess I need to install Steam and see if I can reproduce :)

**Comment by ztomer** [comment]
I'm not sure about the regularity - tried it now and it didn't misbehave, on other occasions, it crashes regularly.

More information: the steam window cannot be controlled using Hammerspoon - trying to move/resize it doesn't work (the window flickers, shows an image on the target location, and returns to the original position - it may or may not be related).

**Comment by cmsj** [comment]
This will be fixed when #256 lands :)

**Comment by trishume** [comment]
Now I know which app was causing my problems :P 

I had Steam running as well.

**Comment by tnorthcutt** [comment]
This seems to be biting me with PhpStorm. This code in `init.lua`:
```
hs.hotkey.bind(hyper, "Up", function()
  local win = hs.window.focusedWindow()
  local f = win:frame()
  local screen = win:screen()
  local max = screen:frame()

  f.x = max.x
  f.y = max.y
  f.w = max.w
  f.h = max.h
  win:setFrame(f)
end)
```

Results in this console output:
```
/Users/travis/.hammerspoon/init.lua:125: attempt to index a nil value (local 'win')
stack traceback:
	/Users/travis/.hammerspoon/init.lua:125: in function </Users/travis/.hammerspoon/init.lua:121>
stack traceback:
```

[...]

PR #256, opened 2015-04-02
Improve hs.hints: Fix bugs and add window title feature

trishume:
Hi, original author of `mjolnir.th.hints` here. I switched to Hammerspoon recently since it included my hints module (and improved it). 

Recently I encountered a bug triggered by some windows not having an associated screen that caused the hints operation to crash. This PR fixes that issue by checking if windows have a screen.

I also added an improved version of a feature that was in my original module which is showing titles in hints. When switching within an app and when using the vimperator mode the icon isn't always useful so showing the window title is very helpful for knowing which window to switch to. However, it gets really cluttered when there are too many hints on the screen.

Thus this PR adds a new variable `hints.showTitleThresh` (default 4) where if there are more than that many windows it won't show titles. To do this I added some auto-sizing frame logic to the Objective-C and some hint counting logic to the Lua.

I've tested this code with the vimperator mode on and off and checking different window counts are properly thresholded. It properly counts even the nested tables used in the vimperator mode.

Here's what titles look like:
screen shot 2015-04-02 at 4 05 40 pm

### DIFF
--- a/extensions/hints/init.lua
+++ b/extensions/hints/init.lua
@@ -31,6 +31,12 @@ hints.fontName = nil
 --- The size of font that should be used. A value of 0.0 will use the default size.
 hints.fontSize = 0.0
 
+--- hs.hints.showTitleThresh
+--- Variable
+--- If there are less than or equal to this many windows on screen their titles will be shown in the hints.
+--- The default is 4. Setting to 0 will disable this feature.
+hints.showTitleThresh = 4
+
 local openHints = {}
 local takenPositions = {}
 local hintDict = {}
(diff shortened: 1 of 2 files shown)


cmsj:
Looks good. One thing I'm curious about is the reasoning for showTitleThresh being <= 4 - if there are lots of windows, wouldn't the titles be more useful?

trishume:
@cmsj you're right they would be more useful, but they get cluttered all to hell and block other hints and your entire screen. Also getting the title requires sending and accessibility message to the application so it lags and the hints pop in slowly when there are too many.

If people really want this they can set `showTitleThresh` to 1000 or something.

I mostly wanted this feature for the case of switching windows within an app. I use multiple emacs windows (like the screenshot) for different projects and like to see the file names so I know which to switch to.

cmsj:
@trishume makes sense. thanks! :)

cmsj:
Closes #237

PR #259, opened 2015-04-03
grid: increase window size in all occasions

javigon:
Enable to make a window taller and wider from top and left (now only from bottom and right)


cmsj:
Thanks! Great idea :)

PR #260, opened 2015-04-04
Mjomatic refactor into Windowgram

JosephFerano:
Fixing bug with mjomatic where it would only resize the main window of an app instead of all. Added vim swp and suo files to gitignore.


cmsj:
Thanks! Let's loop in @Habbie since he wrote mjomatic

JosephFerano:
Sure. I saw that there's also some fixes that can go into it. I would really like to work with this module further because I feel like it has a really strong potential for an interesting window management base, where it would be easy to generate "spaces", and not just use it to have predefined areas for certain applications.

JosephFerano:
Hey guys, I decided that I like the mjomatic features but I felt it needed some work. First, I split the the "go" method into several methods. For now, these include;
```
getRects(wg)
getRatios(rects)
applyAppMapLayout(rects, map)
```
This way, the windowgram can be used to create a custom grid where windows can be moved to arbitrarily. Previously, it was only being used to apply a specific app to a rect. This is still possible, but now an app and all its windows wont have to be dedicated to a single screen.

Habbie:
Awesome! I'd been wanting to split up mjomatic for a while. Will try to look at your work asap.

Habbie:
There may be some inspiration in this PR against the original Mjolnir-mjomatic - https://github.com/Habbie/mjomatic/pull/3/files

JosephFerano:
@Habbie 
What do you mean? Do you have any recommendations/suggestions?

JosephFerano:
I think one of the next features I'll be working on is having the system of movement seen in the grid work with this. Might want to reconsider how all that stuff is organized.

Habbie:
No specific recommendations/suggestions - just wanted to make you aware of other unmerged work in case it's interesting or at all related :)

JosephFerano:
Cool. Yeah, I'll take a look at it. Thanks.

I do think I'm going to make the "getRects" method private. It just doesn't make sense in terms of a public API, but I think it's useful to keep it separate from the main function.

So if you can hold off on merging the pull request. I'd appreciate it.

extensions/windowgram/init.lua
```diff
@@ -0,0 +1,178 @@
+--- === hs.windowgram ===
+---
+--- auto-layout using ascii rows similar to tmuxomatic
+
+local windowgram = {}
+
+local alert = require 'hs.alert'
```
Habbie: this can go (wasn't using it except in commented code, that you rightfully removed)

extensions/windowgram/init.lua
```diff
@@ -0,0 +1,178 @@
...
+---
+--- ~~~lua
+--- local wg = {
+---   "AAAAAAAAAAAABBBBBBBBBBBB",
+---   "AAAAAAAAAAAABBBBBBBBBBBB",
+---   "AAAAAAAAAAAABBBBBBBBBBBB",
+---   "CCCCCCCCCCCCCCCCCCCCCCCC",
+---   "CCCCCCCCCCCCCCCCCCCCCCCC"
+--- }
+--- local windows = windowgram.getWindows(wg)
```
Habbie: this should be `.getRects` I think?

extensions/windowgram/init.lua
```diff
@@ -0,0 +1,178 @@
...
+        windows[char].w = column
+        windows[char].h = i
+      end
+    end
+  end
+  return windows
+
+end
+
+function removeWhitspace(wg)
```
Habbie: missing 'e' in Whitespace

extensions/windowgram/init.lua
```diff
@@ -0,0 +1,178 @@
...
+  local target = {}
+  for i,l in ipairs(wg) do
+    l = l:gsub("%s+","")
```
Habbie: The original code had
```lua
l = l:gsub('#.*','')        -- strip comments
l = l:gsub('%s*$','')       -- strip trailing whitespace
```

and I noticed this line from my config failing with windowgram:
```lua
  "CCCCCCCCCCCCCiiiiiiiiiii      # <-- The windowgram, it defines the shapes and positions of windows",
```

extensions/windowgram/LICENSE
```diff
@@ -1,6 +1,6 @@
-The MIT License (MIT)
+Released under MIT license.
 
-Copyright (c) 2014 Peter van Dijk
+Copyright (c) 2015 Joseph Ferano
```
Habbie: Would like having both names in here. Can be one line with 2015 and both names.

Habbie:
Commented on a few specific lines. Loving the work!

Habbie:
And, as said on IRC, I fully support the renaming to windowgram, and the splitting up of methods. Still pondering whether anything should happen for existing mjomatic users - perhaps an error referring to the docs, and some docs referring to the new three-step program that replaces `.go()`

extensions/windowgram/init.lua
```diff
@@ -0,0 +1,210 @@
+--- === hs.windowgram ===
+---
+--- region generation using ascii rows similar to tmuxomatic for window areas
```
cmsj: This line is what will be used as the module summary at http://www.hammerspoon.org/docs/ - it should try to sell itself in a single sentence. "region generation" is a bit too vague IMO. I think something more like the original description would be better here - the majority use case for end users, would be to use this for tuxomatic style window management.

extensions/windowgram/init.lua
```diff
@@ -0,0 +1,210 @@
...
+  end
+  return target
+end
+
+--- hs.windowgram.getRegionRatios(regions)
+--- Function
+--- Gets a table with the regions as percentages of the screen resolution
+---
+--- Parameters:
+---  * regions - The regions table returned by windowgram.convertToRegions(wg)
```
cmsj: Mentions of other API calls in the docstrings should be formatted like \`hs.windowgram.convertToRegions()\`

extensions/windowgram/init.lua
```diff
@@ -0,0 +1,210 @@
...
+---
+--- Returns:
+---  * Table with ratios. ex; { x=0, y=0.5, w=0.5, h=1 }
+---
+--- Notes:
+--- * These are numbers between 0 and 1 and represent two points; the top left
```
cmsj: Missing a space before the *

extensions/windowgram/init.lua
```diff
@@ -0,0 +1,210 @@
...
+
+--- hs.windowgram.getRegionRatios(regions)
+--- Function
+--- Gets a table with the regions as percentages of the screen resolution
+---
+--- Parameters:
+---  * regions - The regions table returned by windowgram.convertToRegions(wg)
+---
+--- Returns:
+---  * Table with ratios. ex; { x=0, y=0.5, w=0.5, h=1 }
```
cmsj: For consistency, this should be something like:
 * A table containing screen ratios. Each entry in the table is a table containing the following keys:
  * x
  * y
  * w
  * h

extensions/windowgram/init.lua
```diff
@@ -0,0 +1,210 @@
...
+---
+--- Notes:
+--- * These are numbers between 0 and 1 and represent two points; the top left
+--- corner and the bottom right corner.
+function windowgram.getRegionRatios(regions)
+  local ratios = {}
+  local screenSize = hs.screen.mainScreen():frame()
```
cmsj: Is this intentionally tied to the main screen? It's quite common to have more than one screen size (e.g. laptop + monitor)

extensions/windowgram/init.lua
```diff
@@ -0,0 +1,210 @@
...
+  for i,l in ipairs(wg) do
+    -- Remove comments and whitespace
+    l = l:gsub('#.*','')
+    l = l:gsub("%s+","")
+    table.insert(target, l)
+  end
+  return target
+end
+
+--- hs.windowgram.getRegionRatios(regions)
```
cmsj: This should have "-> table" at the end, we indicate return types in the function signature.

extensions/windowgram/init.lua
```diff
@@ -0,0 +1,210 @@
...
+    ratios[k] = {}
+    ratios[k].x = region.x ~= 0 and region.x / screenSize.w or 0
+    ratios[k].y = region.y ~= 0 and region.y / screenSize.h or 0
+    ratios[k].w = region.w ~= 0 and region.w / screenSize.w or 0
+    ratios[k].h = region.h ~= 0 and region.h / screenSize.h or 0
+  end
+  return ratios
+end
+
+--- hs.windowgram.convertToRegions(wg)
```
cmsj: This should also have "-> table" on the end. Also can we switch out "wg" for something that isn't cryptic, but also doesn't seem like an important keyword - maybe "wingram"?

extensions/windowgram/init.lua
```diff
@@ -0,0 +1,210 @@
...
+--- Function
+--- Gets a table with pixel coords describing a region, given a windowgram
```
cmsj: "Converts a windowgram string to a table of screen region co-ordinates" maybe?

extensions/windowgram/init.lua
```diff
@@ -0,0 +1,210 @@
...
+---
+--- Parameters:
+---  * wg - The windowgram: multiline string with the desired window layout
```
cmsj: * wingram - A multiline string describing a desired window layout

extensions/windowgram/init.lua
```diff
@@ -0,0 +1,210 @@
...
+---
+--- Returns:
+---  * Table with regions which contain an X, Y, W, and H values
```
cmsj: * A table containing screen regions representing the contents of the windowgram. Each table entry is a rect-table

extensions/windowgram/init.lua
```diff
@@ -0,0 +1,210 @@
...
+---
+--- Notes:
+---  * Regions are pixel based coordinates relative to the windowgram
```
cmsj: Not sure this is necessary.

extensions/windowgram/init.lua
```diff
@@ -0,0 +1,210 @@
...
+---
+--- On a 1920x1080 resolution, This will return the following table;
+--- ~~~lua
+--- print(wg["A"])
+--- -- prints { x=0, y=0, w=960, h=648 }
+--- print(wg["B"])
+--- -- prints { x=960, y=0, w=960, h=648 }
+--- print(wg["C"])
+--- -- prints { x=0, y=648, w=1920, h=432 }
+--- ~~~
```
cmsj: I kinda feel like this example, and ones for the other functions, should move up to the top level of the module, showing typical usage, broken down into its various stages.

extensions/windowgram/init.lua
```diff
@@ -0,0 +1,210 @@
...
+function windowgram.convertToRegions(wg)
+  local regions = {}
+  local totalWidth = 0
+  local totalHeight = 0
+  local screenSize = hs.screen.mainScreen():frame()
```
cmsj: (see other comment about mainScreen())

extensions/windowgram/init.lua
```diff
@@ -0,0 +1,210 @@
...
+--- hs.windowgram.applyAppMapLayout(regions, map)
+--- Function
+--- * Uses a key/app map table to layout the current windows based on the
+--- regions pixel coords
+---
+--- Parameters:
+---  * regions - The regions table returned by windowgram:convertToRegions(wg)
+---  * map - A key/value table where the key is the ascii character used in
+---  the windowgram and the value is the name of the map.
+---
```
cmsj: Returns:
 * None

extensions/windowgram/init.lua
```diff
@@ -0,0 +1,210 @@
...
+    regions[k].w = screenSize.w * widthRatio
+    regions[k].h = screenSize.h * heightRatio
+  end
+
+  return regions
+end
+
+--- hs.windowgram.applyAppMapLayout(regions, map)
+--- Function
+--- * Uses a key/app map table to layout the current windows based on the
```
cmsj: This needs to stay on a single line and not start with *
Also for the wording, maybe something like "Applies a windowgram layout using a map of applications"?

extensions/windowgram/init.lua
```diff
@@ -0,0 +1,210 @@
...
+--- regions pixel coords
+---
+--- Parameters:
+---  * regions - The regions table returned by windowgram:convertToRegions(wg)
```
cmsj: * regions - A table containing regions, as returned by `hs.windowgram.convertToRegions()`

extensions/windowgram/init.lua
```diff
@@ -0,0 +1,210 @@
...
+---  * map - A key/value table where the key is the ascii character used in
```
cmsj: * map - A table containing a mapping between characters in the windowgram string, and applications. Each entry in the table should consist of a string containing an ASCII character as the key, and a string containing the name of an application as the value

extensions/windowgram/init.lua
```diff
@@ -0,0 +1,210 @@
...
+---  the windowgram and the value is the name of the map.
+---
+--- Ex;
```
cmsj: Similarly, I think this should move up to the top as part of a unified example

extensions/windowgram/init.lua
```diff
@@ -0,0 +1,210 @@
...
+--- local map = {
+---   "A MacVim",
+---   "A Firefox",
+---   "B iTerm2",
+---   "B Skype",
+---   "C Xcode"
+--- }
+---  ~~~
+function resizeToRegion(window, region)
+  local screenSize = hs.screen.mainScreen():frame()
```
cmsj: See earlier mainScreen() comments

JosephFerano:
Per our conversation on IRC, it's probably going to be a better idea to move this stuff into "workspace" as a constructor.

PR #263, opened 2015-04-09
Modal key repeat function.

JosephFerano:
Adding ability for a modal key to have a repeat function. Improving the documentation for the method.


JosephFerano:
The method was just calling the equivalent hotkey.bind method so just had to pass the repeatfn function.

cmsj:
Thanks!

PR #273, opened 2015-04-25
Add hs.drawing:setTextFont() for changing the font of an hs.drawing.text object

peterhajas:
This change introduces a new function in `hs.drawing`, `hs.drawing:setTextFont()`. It accepts a string indicating the name of the font to use for the `hs.drawing.text` object. It will preserve the text size set in `hs.drawing:setTextSize()`

This also augments `hs.drawing:setTextSize()` to preserve the font that was set on the `hs.drawing` object.

Example configuration using `hs.drawing.setTextFont()`:

    text = hs.drawing.text(hs.geometry.rect(50, 50, 1000, 1000), "Zanch")
    text:setTextSize(60):sendToBack():show()
    text:setTextFont('Zapfino')

will produce:

screen shot 2015-04-24 at 10 22 23 pm

Thanks for your review!

### DIFF
--- a/extensions/drawing/internal.m
+++ b/extensions/drawing/internal.m
@@ -691,6 +691,35 @@ static int drawing_setText(lua_State *L) {
     return 1;
 }
 
+/// hs.drawing:setTextFont(fontname) -> drawingObject
+/// Method
+/// Sets the font of a drawing object
+///
+/// Parameters:
+///  * fontname - A string containing the name of the font to use
+///
+/// Returns:
+///  * The drawing object
+///
+/// Notes:
+///  * This method should only be used on text drawing objects
+static int drawing_setTextFont(lua_State *L) {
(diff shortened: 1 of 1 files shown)


cmsj:
Thanks!

cmsj:
Actually just one FYI kinda thing, I'm trying to switch is over to using luaL_checkstring instead of lua_tostring, when turning Lua function parameters into an NSString. It means we get a nice error in Hammerspoon's console window, rather than silently consuming a user mistake and initialising an NSString with a NULL pointer

I switched out the one in your PR in https://api.github.com/repos/Hammerspoon/hammerspoon/commits/bb252f3a40db38041deb885984f384d142c8bea3 :)

peterhajas:
I'm always a fan of type safety. :+1:  Thanks for merging!

PR #278, opened 2015-04-28
Improve documentation on uielement.watcher

tmandry:
Noticed that the summary on the documentation page doesn't really make sense, so I wanted to fix that.


cmsj:
Thanks :)

PR #279, opened 2015-04-28
Add winter extension

knl:
winter is a module for moving/resizing windows using a fluent interface,
which makes it really useful for creating succint keybindings.

Name comes from a portmanteu of win(dow) (mas)ter.

### DIFF
--- a/extensions/winter/init.lua
+++ b/extensions/winter/init.lua
@@ -1,0 +1,541 @@
+--- === hs.winter ===
+---
+--- A module for moving/resizing windows using a fluent interface (see Usage below).
+---
+--- The name winter is a portmanteu of **win**dow mas**ter**.
+---
+--- Usage:
+---
+---     local wintermod = require "hs.winter"
+---     local winter = wintermod.new()
+---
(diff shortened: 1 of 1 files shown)


knl:
I was to quick to submit this, sorry. Found a way to make it more succinct and I'm also working on improving the documentation.

cmsj:
FWIW, this is a pretty nice looking extension :)

I did wonder about the name though. I think this may be an ongoing source of friction for Hammerspoon - I want the module names to be generic and descriptive, but authors will naturally want a distinctive name that represents their work.

(E.g. My immediate reaction here would be to call this something like hs.window.chain, but let's not worry about that now. I'd rather have the functionality than not :)

knl:
Thanks! I pushed a new pull request (#280) for an updated version that is much cleaner and has the documentation.

Now, as for the name goes, I don't mind changing it, but the whole point of `winter` is to make you type less and have shorter, inspectable functions in `hotkey.bind`. That being said, `window.chain` might be a bit long :)

PR #280, opened 2015-04-28
Add the window.fluent extension

knl:
window.fluent is a module for moving/resizing windows using a fluent interface,
which makes it really useful for creating succinct key-bindings.

### DIFF
--- a/extensions/window/fluent.lua
+++ b/extensions/window/fluent.lua
@@ -1,0 +1,778 @@
+--- === hs.window.fluent ===
+---
+--- A module for moving/resizing windows using a fluent interface (see Usage below).
+---
+--- Usage:
+---
+---     local window = require "hs.window.fluent"
+---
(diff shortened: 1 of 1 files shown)


cmsj:
@knl apologies for the delay in getting to this review. We spoke about changing the name to something like hs.window.fluent, but I can't remember what we agreed on - do you?

knl:
Hi @cmsj that is right :) However, I haven't had time to do the change. Let me do it today, and then have let's have a look.

knl:
I've changed the name, please have a look.

cmsj:
So, after some discussion in #hammerspoon with @lowne, as much as I like hs.window.fluent, I think it probably is better if we update hs.window's methods to return the window object, so they can be chained.
It's not as good as full, optimised fluency, but it does mean a simpler API. Perhaps we can also figure out a smart way to make the current API more chainable, without duplicating it.

PR #282, opened 2015-04-28
Add the grille extension

knl:
grille is a module for moving and resizing windows along a virtual grid,
using a fluent interface. It depends on winter extension (#280).

Benefit of grille over hs.grid is that is allows changing grid dimensions
easily, and supports having multiple grids active at the same time.

### DIFF
--- a/extensions/grille/init.lua
+++ b/extensions/grille/init.lua
@@ -1,0 +1,184 @@
+--- === hs.grille ===
+---
+--- A module for moving/resizing your windows along a virtual grid (or more),
+--- using a fluent interface (see Usage below).
+---
+--- hs.grille was based on hs.sd.grid and hs.bg.grid modules, but went through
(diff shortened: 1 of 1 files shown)


knl:
This pull request depends on the extension included in #280, thus, once #280 gets merged, this one needs to be rebased.

cmsj:
@knl with #361 landed, does that address your intentions for this PR?

knl:
Hey Chris, sorry for long radio silence, it has been busy here. Yes, #361 seems to address most of my intentions, I'll see how I could move my [config](https://github.com/knl/dotskel/blob/master/hammerspoon/init.lua) to it.

cmsj:
@knl ok thanks. What I'll do is close out this PR, but if there is stuff you still need, please do re-open, or file an issue or something :)

PR #283, opened 2015-05-01
allow external readonly access to MJLuaState

asmagill:
allows module to include

     extern lua_State* MJGetActiveLuaState() ;

to get the current LuaState for comparison to stored values in callbacks... allows an extra check in case GC doesn't or isn't able to remove the callback during a restart or reload.


cmsj:
@asmagill thanks!

PR #286, opened 2015-05-02
hints title enhancements

echiu64:
* hints.titleMaxSize - set maximum size of a window title
 * hints.titleRegexSub - specify regex substitutions on titles

### DIFF
--- a/extensions/hints/init.lua
+++ b/extensions/hints/init.lua
@@ -37,6 +37,22 @@ hints.fontSize = 0.0
 --- The default is 4. Setting to 0 will disable this feature.
 hints.showTitleThresh = 4
 
+--- hs.hints.titleMaxSize
+--- Variable
+--- If the title is longer than maxSize, the string is truncated, -1 to disable, valid value is >= 6
+hints.titleMaxSize = -1
+
+--- hs.hints.titleRegex
+--- Variable
+--- List of regex substituions for window titles, in string.gsub style format:
+---    { key = { pat="<gsub_pattern>", sub="<gsub_replace>" }, ... }
+---
+---    Add as many keys as you like
+---
+--- e.g. { MSApps={pat="Microsoft%s*", sub=""}, Chrome={pat="Google Chrome", sub="Chrome"} }
+--- Only valid for vimperator style
+hints.titleRegexSub = {}
+
 local openHints = {}
 local takenPositions = {}
 local hintDict = {}
@@ -45,6 +61,25 @@ local selectionCallback = nil
 
 local bumpThresh = 40^2
 local bumpMove = 80
+
+function split(str, pat)
+   local t = {}  -- NOTE: use {n = 0} in Lua-5.0
+   local fpat = "(.-)" .. pat
+   local last_end = 1
+   local s, e, cap = str:find(fpat, 1)
+   while s do
+      if s ~= 1 or cap ~= "" then
+         table.insert(t,cap)
+      end
+      last_end = e+1
+      s, e, cap = str:find(fpat, last_end)
+   end
+   if last_end <= #str then
+      cap = str:sub(last_end)
+      table.insert(t, cap)
+   end
+   return t
+end
 
 function hints.bumpPos(x,y)
   for i, pos in ipairs(takenPositions) do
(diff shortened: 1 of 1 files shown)


extensions/hints/init.lua
```diff
@@ -37,6 +37,22 @@ hints.fontSize = 0.0
...
+--- hs.hints.titleRegex
+--- Variable
+--- List of regex substituions for window titles, in string.gsub style format:
+---    { key = { pat="<gsub_pattern>", sub="<gsub_replace>" }, ... }
+---
+---    Add as many keys as you like
+---
+--- e.g. { MSApps={pat="Microsoft%s*", sub=""}, Chrome={pat="Google Chrome", sub="Chrome"} }
+--- Only valid for vimperator style
+hints.titleRegexSub = {}
```
trishume: This should be made a list/array/integer table, there is no reason to have the keys here. Having them is confusing since it looks like they have meaning when in fact they have none.

extensions/hints/init.lua
```diff
@@ -46,6 +62,25 @@ local selectionCallback = nil
 local bumpThresh = 40^2
 local bumpMove = 80
 
+function split(str, pat)
```
trishume: This function doesn't look to be called from anywhere. Either remove it or use it somewhere and make it `local`.

extensions/hints/init.lua
```diff
@@ -114,7 +149,14 @@ function hints.displayHintsForDict(dict, prefixstring, showTitles)
         else
           local suffixString = ""
           if showTitles then
-            suffixString = ": "..win:title()
+            win_title = win:title()
+            if hints.titleMaxSize > 1 then
+                end_idx = hints.titleMaxSize - 6
+                if end_idx < 6 then end_idx = 6 end
+                win_title = string.sub(win_title, 1, end_idx)
+                if end_idx < #win_title then win_title = win_title .. "..." end
```
trishume: You already modified `win_title` in the previous line so won't this condition never activate?

extensions/hints/init.lua
```diff
@@ -114,7 +149,14 @@ function hints.displayHintsForDict(dict, prefixstring, showTitles)
         else
           local suffixString = ""
           if showTitles then
-            suffixString = ": "..win:title()
+            win_title = win:title()
+            if hints.titleMaxSize > 1 then
+                end_idx = hints.titleMaxSize - 6
+                if end_idx < 6 then end_idx = 6 end
```
trishume: Note that `end_idx` is based on a static preference. So this does `hints.titleMaxSize - 6 < 6` which is `hints.titleMaxSize < 12`. If the pref is less than twelve always trim to six characters? Why??

trishume:
Thanks for the effort but this code has some issues. Basically all the logic having to do with max title length is messed up. 

This might work better, but I haven't tested it:
```lua
if hints.maxTitleSize > 1 and #win_title > hints.maxTitleSize then
  local end_idx = math.min(0,hints.maxTitleSize-3)
  win_title = string.sub(win_title,1,end_idx) .. "..."
end
```
Much simpler, no arbitrary number 6, properly compensates for ellipsis.

extensions/hints/init.lua
```diff
@@ -114,7 +149,14 @@ function hints.displayHintsForDict(dict, prefixstring, showTitles)
         else
           local suffixString = ""
           if showTitles then
-            suffixString = ": "..win:title()
+            win_title = win:title()
```
trishume: should be `local win_title`

extensions/hints/init.lua
```diff
@@ -198,7 +240,13 @@ function hints.windowHints(windows, callback)
     if app and win:isStandard() then
       if hints.style == "vimperator" then
         if app and win:isStandard() then
-          local appchar = string.upper(string.sub(app:title(), 1, 1))
+          app_title = app:title()
```
trishume: should be `local app_title`

echiu64:
First time writing lua ;)
Basically wanted to ensure that for small titles, we don't add the extra "..." at the end, that was a last minute addition with minimal testing, i'll review your changes and resubmit a pull

trishume:
@echiu64 cool, do know that you can just fix this PR by committing new stuff to the branch that you created the pull request from. It will automatically mark my diff comments as resolved when you change something around them.

extensions/hints/init.lua
echiu64: I flipped math.min to math.max, I think that makes more sense to maximize the end_idx value. Everything I took your suggestions as is, thanks!

echiu64:
Oops nevermind, i didn't revisit the regex yet. Sorry about that.

cmsj:
@echiu64 it looks from your last comment like there is still some work to do here? Or is this ready to lane?

cmsj:
I'm going to close this out for now, please feel free to re-open it :)

echiu64:
Sorry stopped using it awhile back due to some OSX issues with it slowing down over time, it became unusable in general. I may revisit if that changes.

trishume:
Yah mine is also super slow and I don't use it for that reason. It has to accessibility message every app and I think some apps are just super slow to respond. I wonder if there is a faster way to do it.

PR #296, opened 2015-05-22
Better documentation for hs.grid

oskarols:
Poking around in the hs.grid I noticed the documentation was kind of lacking.

So this is a slight remedy.


cmsj:
Thanks! This has now been submitted to master, with a few changes along the way.

cmsj:
(see 5cc6c04 for the commit in question)

Issue #269, opened 2015-04-23
Onclick handlers for hs.drawing shapes


JosephFerano:
I've been creating a tabbing system for the window management and it currently generates a set of tabs using hs.drawing. It would be great if I could also attach an onclick handler to these shapes in order to shift the focus of the window in the region.

cmsj:
So, I originally had it in my head that hs.drawing wouldn't support any kind of mouse event, but perhaps a simple click event is ok. I suspect that it will lead to more though, and before long we'll have implemented a whole UI toolkit ;)

PR #300, opened 2015-05-27
Add hs.tabs module

trishume:
This module allows a certain kind of tabbing of the windows of any OSX app. I originally wrote it to tab Emacs frames but it might also be useful for other apps like Finder that don't have tabs but should.

Basically it organizes windows so that they are always all in the same position by listening to events. It then uses the drawing module to draw somewhat-nice looking tabs in the title bar with the window names when any window in the stack is focused.

It also exposes functions that allow for binding keys to switch between tabs and enumerate the tabbed windows and suchlike. It currently does not support clicking on the tab names since `hs.drawing` has no click event support.

screen shot 2015-04-19 at 5 51 11 pm

I've been using this module as a core part of my daily workflow for multiple weeks now so I'm fairly sure it holds up well. I think I've worked out all the bugs concerning apps crashing and restarting and so on.

### DIFF
--- a/extensions/tabs/LICENSE
+++ b/extensions/tabs/LICENSE
@@ -1,0 +1,21 @@
+The MIT License (MIT)
+
+Copyright (c) 2015 Tristan Hume
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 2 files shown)


jhgg:
Super cool.

oskarols:
Nice!

> (..) might also be useful for other apps like Finder that don't have tabs but should.

Finder actually has tabs in Yosemite; <kbd>⌘</kbd> + <kbd>T</kbd>

> It currently does not support clicking on the tab names since hs.drawing has no click event support.

If you really wanted to do this you could probably capture the mouse events via eventtap — return from the event handler and let them pass through to other applicaitons — and compare the clicked positions to the tabs and react accordingly.

cmsj:
@oskarols fwiw we have #269 to add proper click handlers to hs.drawing objects, which I suspect is probably a better way to go, than throwing eventtap into the mix.

trishume:
Yah my Finder example was bad (I don't actually use finder at all). Clicking tabs would be nice when the functionality becomes built in but for now I don't think it is necessary. They are still useful as a visual indicator of which windows are in a stack so you can directly press a numbered key without for example having to wait for a hints GUI.

Also cc @JosephFerano since he was also looking into tabbing.

trishume:
Is this good to be merged @cmsj? As I said I've tested it a bunch and it is super useful and works great.

cmsj:
@trishume Sorry, I know it's taken me a while to get to the review, I think it should be good, I just need to do a deeper check over of the docs.

cmsj:
I've made some slight tweaks to the docstrings and merged it into master. Thanks!

cmsj:
@trishume btw you might be interested in the recently added ability to add a click handler to hs.drawing objects. Will be in the next release :)

PR #316, opened 2015-06-15
Add hs.spaces.watcher for detecting space changes

tmandry:
OS X does have an official API for detecting space changes, and I needed it for a project, so here's an extension to watch for this event.

One problem with this code is it uses a deprecated API (since 10.8) to get the space number. I'm happy to take this out, as I don't rely on it, but it could be useful for other people.

Eventually I plan to add to the callback a list of window IDs on the current space, which can be used to detect which space we're on without using the deprecated space ID.

### DIFF
--- a/extensions/spaces/Makefile
+++ b/extensions/spaces/Makefile
@@ -1,0 +1,35 @@
+mkfile_path := $(abspath $(lastword $(MAKEFILE_LIST)))
+current_dir := $(notdir $(patsubst %/,%,$(dir $(mkfile_path))))
+
+MODULE = $(current_dir)
+
+PREFIX  ?= .
+OBJCFILE = watcher.m
+LUAFILE  = init.lua
+SOFILE  := $(OBJCFILE:m=so)
+
+CC=cc
+EXTRA_CFLAGS ?= -fobjc-arc
+DEBUG_CFLAGS ?= -g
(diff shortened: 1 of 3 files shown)


cmsj:
Thanks. I did a small amount of reworking (Makefile and docstrings mostly), but the rest seems fine as-is. I'd note that the space number always seems to be -1 for me (10.10.3, dual head, separate spaces for each display), but I'm not really bothered about that, this Spaces stuff is all a bit Wild West ;)

asmagill:
I look forward to testing this out... FWIW, every source I've found that tracks spaces in some way that works under 10.10 ends up doing something along the the lines of creating its own invisible window on each new space it visits and uses that to detect which space it is on when it revisits the space... and identifying spaces simply by a number is really not that useful anymore when you consider that the user may have "Displays have separate spaces" or "Automatically rearrange spaces based on most recent use" checked.

cmsj:
Agreed, I'm not expecting the space number thing to be tremendously useful. FWIW, I'm also doubtful that Apple is going to give us a Spaces API any time soon - I think the internal data model of multiple spaces on multiple displays that may be separated, that may re-order themselves magically, etc, etc. is just too much to expect an external Accessibility app to keep track of the state of.

tmandry:
I might take the number out then in a further PR. Using the invisible window is interesting. HS might benefit from having something that does that.

In my WM I'm using a heuristic to track the space (keep track of all windows on the space and decide on the space that has the most matching windows). Seems to be working fine so far, but I'm sure there are edge cases where it fails..

Issue #317, opened 2015-06-15
hs.eventtap limitations


asmagill:
In writing a potential replacement hotkey -- #313 -- came across a situation in which an eventtap seemed to only catch keyup and keydown for *some* keys, but not all, until I did a reload of the hammerspoon configuration, and then it worked flawlessly for all keys.

Digging, I think the reason is because I am starting the eventtap to capture keyup and keydown in my init.lua file.  When first starting, this means it is inserted during (but before it completes) [MJAppDelegate applicationDidFinishLaunching] and when reloading, it is inserted from the main runloop.

I'm guessing (but am a little uncertain how to verify) that there is an implicit eventtap inserted at the completion of an applications initiation... at first I thought it was because I had the console opening during startup, since most of the non-working keys involved things that a text input box might use, but keeping this closed during startup didn't seem to make a difference...

What did was adding this to the end of the new hotkey module:

    -- doesn't fire unless we save the timer (gc finalizer kicks in too quickly?)
    module.startEventWatcher = timer.doAfter(1, function()
        print("-- Starting hotkey eventtap")
        eventWatcher:start()
        module.startEventWatcher = nil -- we don't need to save the timer after it fires
    end)

Has anyone else noticed inconsistent eventtap behavior when started from init.lua (or a file it includes)?  Are there any other known work arounds?

`hs.eventtap` inserts its tap at the head of the `kCGSessionEventTap` (placed at the point where HID system and remote control events enter a login session) location... this means that it will be checked after root activated watchers at the `kCGHIDEventTap` (placed at the point where HID system events enter the window server.) location, but before any other eventtaps created *prior to starting Hammerspoon*.  Eventtaps created after ours may insert themselves before us, like we do, but that's always a problem -- the most recent created eventtap from any application *can* be put at the head of at least the `kCGSessionEventTap` location and override us...

If we want to be able to put events at the `kCGHIDEventTap` location, then we need a way to execute the relevant code as root... is it worth looking into this?  I'm leaning towards no, but I would also prefer a cleaner solution then the one I noted above (using a timer to activate the watcher).

asmagill:
Scratch the above...  apparently OS X disables an event tap if it takes too long or behaves oddly or looks at the moon wrong... the exact reasons are not well documented from what I could find.

[...]

PR #318, opened 2015-06-15
catch failure to create eventtap b/c of accesibility being off

asmagill:
if accessibility needs to be enabled, the previous version would crash... this catches the NULL return from CGEventTapCreate.

### DIFF
--- a/extensions/eventtap/internal.m
+++ b/extensions/eventtap/internal.m
@@ -272,7 +272,7 @@ static int checkKeyboardModifiers(lua_State* L) {
 ///
 /// Returns:
 ///  * Returns an array containing indicies starting from 1 up to the highest numbered button currently being pressed where the index is `true` if the button is currently pressed or `false` if it is not.
-///  * Special hash tag synonyms for `left` (button 1) and `right` (button 2) are also set to true if these buttons are currently being pressed.
+///  * Special hash tag synonyms for `left` (button 1), `right` (button 2), and `middle` (button 3) are also set to true if these buttons are currently being pressed.
 ///
 /// Notes:
 ///  * This is an instantaneous poll of the current buttons buttons, not a callback.  This is useful primarily in conjuction with other modules, such as `hs.menubar` where a callback is already in progress and waiting for an event callback is not practical or possible.
@@ -290,6 +290,9 @@ static int checkMouseButtons(lua_State* L) {
             } else if (i == 1) {
                 lua_pushboolean(L, TRUE) ;
                 lua_setfield(L, -2, "right") ;
+            } else if (i == 2) {
+                lua_pushboolean(L, TRUE) ;
+                lua_setfield(L, -2, "middle") ;
             }
         }
         lua_pushinteger(L, i + 1) ;
--- a/extensions/eventtap/internal.m
+++ b/extensions/eventtap/internal.m
@@ -5,7 +5,6 @@
 
 typedef struct _eventtap_t {
     lua_State* L;
-    bool running;
     int fn;
     int self;
     CGEventMask mask;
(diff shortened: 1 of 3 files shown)


extensions/eventtap/internal.m
```diff
@@ -173,12 +172,18 @@ static int eventtap_start(lua_State* L) {
...
-    return 1;
+        CGEventTapEnable(e->tap, true);
+        e->runloopsrc = CFMachPortCreateRunLoopSource(NULL, e->tap, 0);
+        CFRunLoopAddSource(CFRunLoopGetMain(), e->runloopsrc, kCFRunLoopCommonModes);
+
+        lua_settop(L,1);
+        return 1;
+    } else {
+        showError(L, "Unable to create eventtap.  Is Accessibility enabled?");
+    }
```
cmsj: This is missing a return statement. We should probably have the lua_settop(L, 1); return; here. This method is supposed to return the eventtap object.
We could also arguably return nil and document that that happens in the event of an error.

asmagill: Sorry -- the examples I glanced at when trying to deciding wether to use showError or luaL_error didn't immediately follow showError with a return, and luaL_error *doesn't* return... perhaps the examples I perused fell through to an eventual return that I missed.  Returning nil would probably be appropriate.

I actually had a question sort of forming in my mind about that and hadn't decided if I should put it to the group yet or not... my current practice is that, except for specific getter/setter functions, class methods that I write that act on an object return that object... it allows things like `object:method1():method2():...:methodN()`

I noticed that back when I ported eventtap, I return nil for both start and stop if it is already in that state...  while I'm not sure that it is likely that eventtap would be chained in this manner, should we consider making sure all modules are consistent across the board in this respect? 

On the one hand, deciding yes would mean a more consistent experience for the end-user/local-developer... on the other hand, it would mean being more dogmatic about module requirements when  developers propose new core modules...  I'm not sure where I fall on this, but it seemed as good a time as any to bring it up.

cmsj: Yeah I think we should always try to return the object if possible, being able to chain the method calls is very pleasing. I would welcome an audit of the existing modules, and would definitely help out with such an effort :D

asmagill:
Ok, removed internal running flag and now actually check for tap state directly.  Consistent returns of object... even if accessibility fails so can be retried after it is toggled.  Use isEnabled to verify status if necessary.

extensions/eventtap/internal.m
```diff
@@ -189,33 +191,45 @@ static int eventtap_start(lua_State* L) {
...
-    remove_event(L, e->self);
-    e->self = LUA_NOREF;
-    e->running = false;
-
-    CGEventTapEnable(e->tap, false);
-    CFMachPortInvalidate(e->tap);
-    CFRunLoopRemoveSource(CFRunLoopGetMain(), e->runloopsrc, kCFRunLoopCommonModes);
-    CFRelease(e->runloopsrc);
-    CFRelease(e->tap);
+    if (e->tap && CGEventTapIsEnabled(e->tap)) {
```
cmsj: Is there a situation where we might find that the tap is not enabled, but e->tap still exists? If so, we'd want to clean up some of the resources there too.

asmagill: I presume you mean other than when the system disables the tap and issues a disabledByUserInput or disabledByTimeout event (because in this case, e->tap does exist and that's why the CGEventTapIsEnabled is called to check its status out... we don't want to _gc the object then because likely it will be restarted by the program (see #317)...

To my knowledge no, but perhaps it should be explicitly set to null in hs.evettap.new(...) like it is in hs.eventtap.stop().

Stop nulls it, and _gc erases it, and Accessibility being disabled sets it to NULL, so... the only assumption that is being made, I think, is when new() is called. Unless I'm missing your meaning.

I can resubmit later tonight.

asmagill:
camelCaps addition will print warnings for use of original versions of keys (i.e. existing code will continue to work, but with warnings printed to console).

also includes label for middle in hs.eventtap.checkMouseButtons()

asmagill:
hs.eventtap.checkMouseButtons indexes buttons starting at 1 so that you can use lua's for and/or ipairs and ignore the shortcut labels (left,right, and middle) if you want to check more complex button combinations.

hs.eventtap.event.getButtonState(button) starts at 0 (i.e. left = 0, right = 1, etc.)

I think all of the other references to buttons are by label, unless I missed something, so they should be ok as is.

The two functions/methods are very different in use and purpose, but... should we make the numbering consistent?  checkMouseButtons is more lua-friendly, but getButtonState is closer to low-level specifics... I can't decide... I am ok leaving it as is, unless anyone else has a preference.

PR #324, opened 2015-06-19
updated to 5.3.1, plus more

asmagill:
Core functions updated to official Lua 5.3.1 code -- minor tweaks to handle overrun, convert strings to const, etc.  Nothing earth-shattering.

I changed the codePoint function back to the non-camel codepoint because almost every reference I find to the phrase "Unicode codepoint" online and in other programs treats it as one, all lowercase word, although a handful of references do treat it as two words.  I'm not wedded to either, but I suspect programmers will be more familiar with thinking of it as one word unit and not two.  Change it back if you disagree.

I also added support for registering keys and direct codepoint to UTF8 conversion for use with print and hs.drawing.text

### DIFF
--- a/extensions/utf8_53/init.lua
+++ b/extensions/utf8_53/init.lua
@@ -1,9 +1,9 @@
 --- === hs.utf8_53 ===
 ---
 --- Functions providing basic support for UTF-8 encodings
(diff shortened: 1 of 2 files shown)


asmagill:
Kept the name `utf8_53` for now... on the one hand, I agree `hs.utf8` is easier to type, on the other hand, it's just a name and a simple `hs.utf8 = hs.utf8_53` makes all but the docs interchangeable... I think only `init.lua` would need changing, and a global search and replace would probably be sufficient.

PR #340, opened 2015-06-24
Add some basic application bundle info to Hammerspoons lua environment

asmagill:
Primarily, I want to be able to do this:

     relaunch = function()
         os.execute([[ (while ps -p ]]..hs._hammerspoonAppInfo.processID..[[ > /dev/null ; do sleep 1 ; done ; open -a "]]..hs._hammerspoonAppInfo.bundlePath..[[" ) & ]])
         hs._exit(true, true)
     end

but the other info may be useful to others as well.  The bundle ID of Hammerspoon is also in hs.settings... should it be removed, since this is a more logical place for it?

### DIFF
--- a/Hammerspoon/MJLua.m
+++ b/Hammerspoon/MJLua.m
@@ -39,6 +39,28 @@ static int core_reload(lua_State* L) {
     return 0;
 }
 
+/// hs._hammerspoonAppInfo
+/// Variable
+/// A table containing read-only information about the Hammerspoon application instance currently running.
+static int push_hammerAppInfo(lua_State* L) {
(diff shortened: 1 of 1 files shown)


asmagill:
Forgot I'm going to want this function to check when Accessibility state changes as well, to trigger the relaunch.

cmsj:
Excellent idea, although I'm not sure about "hs._hammerspoonAppInfo". How about "hs.processInfo"?

asmagill:
The name change is fine... do I need to update the pull, or can you do that?

cmsj:
I'll do it :) thanks!

PR #347, opened 2015-06-26
Add hs.delayed

lowne:
Two functions: `hs.delayed.doAfter(previous_id, delay, fn, ...) -> id` and `hs.delayed.cancel(id)`.
Alternatively, this could go into `hs.timer`, "overriding" `hs.timer.doAfter` (API is compatible) to return a "delayed" object with a `:cancel()` method.

### DIFF
--- a/extensions/delayed/init.lua
+++ b/extensions/delayed/init.lua
@@ -1,0 +1,161 @@
+--- === hs.delayed ===
+---
+--- Simple helper for delayed, cancellable callbacks
+
+-- * Needs a native gettime(): https://developer.apple.com/library/mac/documentation/Cocoa/Reference/Foundation/Classes/NSDate_Class/index.html#//apple_ref/occ/instp/NSDate/timeIntervalSince1970
+-- * Debugging can get difficult; ideally this should be fixed (either via explicit params, or better via debug.getinfo)
+
+local gettime=require'socket'.gettime --FIXME need a native hook
+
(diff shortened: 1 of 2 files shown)


asmagill:
Haven't really looked at this yet to see what it brings that is new, but just in case it isn't clear from the docs, hs.timer.doAfter(...) is cancelable...

From the console:

      a = hs.timer.doAfter(5,function() print("yay") end)

and as long as you type `a:stop()` within the 5 second window, it will cancel the action.

Just want to make sure this isn't addressing a non-need -- if it does more great, and I'll look closer at it this evening when I get a chance.

lowne:
Hah, I've never known that!
screen shot 2015-06-26 at 16 48 40

Well, never mind then :) - I suppose I'll just have a thin wrapper user-side to avoid repeated boilerplate:

```lua
local function doAfter(prev_timer,delay,fn,...)
  if prev_timer then prev_timer:stop() end
  return hs.timer.doAfter(delay,function()fn(...)end)
end
```

Is there any reason to worry about the overhead of several (possibly dozens of) timer instances

asmagill:
I'll update the docs tonight... that slipped through and is misleading.  As to the impact, timers are all queued in the main application loop, so I doubt anything in lua would be able to handle them faster... we probably do have the issue of lua being single threaded, even with the use of coroutines, so it can only handle one callback at a time, but that would affect any implementation we chose that didn't involve multiple independant lua threads (and thus separate variable spaces).

PR #348, opened 2015-06-26
[WIP] Migrate from Lua cocoapod to using LuaSkin.framework

cmsj:
This is WIP for now, but I'm throwing it up here as a PR so others can look at it. The deal is that I want to move away from the Lua cocoapod to our own thin framework that wraps lua_State management in objc. I'd also like to, over the longer term, gather up the various lua_State operations we do a lot, and abstract/unify them into LuaSkin.
(Then further out from that, start gathering Hammerspoon-specific things into a Hammerspoon.framework that can be re-used throughout the extensions - stuff like common callback registration/dispatch patterns, use of LuaSkin, etc)


cmsj:
(rebased on top of current master & convert_number_to_integer)

asmagill:
What would this change in how external modules are built or coded?  Would existing ones need to be updated?

cmsj:
@Asmagil technically, no. LuaSkin is just a helper wrapper around lua_State. Anything compiled against the right version of Lua should load fine (like you can compile against a system Lua now and it works with our Cocoapods lua). Having said that, the framework exists to collect core Lua tasks that extensions use a lot, so it would certainly help developers to adopt it :)

It'll be Hammerspoon.framework that's potentially more invasive.

PR #354, opened 2015-06-30
Fix chained actions on hs.window

lowne:
Workaround for issues with chained actions when hs.window.animationDuration>0.
(EDIT: AFAICT (and that's not very far at all) the problem seems to be that `currentAnimation.currentProgress = 1.0;` does *not* synchronously call `setCurrentProgress:(NSAnimationProgress)progress`, so the window attributes do not get immediately set. There's probably an obvious fix for that - which is beyond my non-existing ObjC expertise - that should result in *correct* (albeit still "weird" (windows jumping around)) behaviour)

Ideally the (ugly) hs.timer should be removed, and the corresponding cleanup (`init.lua` row 98) should instead happen in a function that in turn gets called from animationDidStop/End in `internal.m`. I don't know if that's possible; if it is, I surely have no idea how.

Example for testing:
```lua
local w=hs.window.focusedWindow()
w:moveToScreen(w:screen():next()):moveToScreen(w:screen():previous()):moveToUnit({x=0,y=0,w=0.5,h=0.5})
```
The `moveToScreen`s should cancel out and the window should just move to the top left quadrant of its current screen, if not already there.

-----

The changes seem potentially dangerous wrt GC (see comment) *and* lack thereof if the weak tables don't work as I expect (I don't use them very often), so close scrutiny/testing is warranted.

(@cmsj suggested on IRC as a possible alternative to strongly advise, in the documentation, to set `animationDuration` to 0 if any chaining might occur)

### DIFF
--- a/extensions/timer/internal.m
+++ b/extensions/timer/internal.m
@@ -202,7 +202,7 @@ static int meta_gc(lua_State* __unused L) {
     return 0;
 }
 
-/// hs.timer.time() -> sec
+/// hs.timer.secondsSinceEpoch() -> sec
 /// Function
 /// Gets the number of seconds since the epoch, including the fractional part; this has much better precision than `os.time()`, which is limited to whole seconds.
 ///
@@ -211,7 +211,7 @@ static int meta_gc(lua_State* __unused L) {
 ///
 /// Returns:
 ///  * The number of seconds since the epoch
-static int timer_gettime(lua_State *L)
+static int timer_getSecondsSinceEpoch(lua_State *L)
 {
     struct timeval v;
     gettimeofday(&v, (struct timezone *) NULL);
@@ -234,7 +234,7 @@ static const luaL_Reg timerLib[] = {
     {"doAfter",    timer_doAfter},
     {"new",        timer_new},
     {"usleep",     timer_usleep},
-    {"time",       timer_gettime},
+    {"secondsSinceEpoch",       timer_getSecondsSinceEpoch},
     {NULL,          NULL}
 };
 
--- a/extensions/timer/internal.m
+++ b/extensions/timer/internal.m
@@ -1,4 +1,5 @@
 #import <Cocoa/Cocoa.h>
+#import <sys/time.h>
 #import <lua/lauxlib.h>
 #import "../hammerspoon.h"
 
@@ -201,6 +202,25 @@ static int meta_gc(lua_State* __unused L) {
     return 0;
 }
 
+/// hs.timer.time() -> sec
+/// Function
+/// Gets the number of seconds since the epoch, including the fractional part; this has much better precision than `os.time()`, which is limited to whole seconds.
+///
+/// Parameters:
+///  * None
+///
+/// Returns:
+///  * The number of seconds since the epoch
+static int timer_gettime(lua_State *L)
+{
+    struct timeval v;
+    gettimeofday(&v, (struct timezone *) NULL);
+    /* Unix Epoch time (time since January 1, 1970 (UTC)) */
+    lua_pushnumber(L, v.tv_sec + v.tv_usec/1.0e6);
+    return 1;
+}
+
+
 // Metatable for created objects when _new invoked
 static const luaL_Reg timer_metalib[] = {
     {"start",   timer_start},
@@ -214,6 +234,7 @@ static const luaL_Reg timerLib[] = {
     {"doAfter",    timer_doAfter},
     {"new",        timer_new},
     {"usleep",     timer_usleep},
+    {"time",       timer_gettime},
     {NULL,          NULL}
 };
 
--- a/extensions/window/init.lua
+++ b/extensions/window/init.lua
@@ -11,6 +11,9 @@ local application = require "hs.application.internal"
 local fnutils = require "hs.fnutils"
 local geometry = require "hs.geometry"
 local hs_screen = require "hs.screen"
+local timer = require "hs.timer"
+local cachedFrames = setmetatable({},{__mode="k"}) -- weak table
+local cacheTimers = setmetatable({},{__mode="k"}) -- weak table
 
 --- hs.window.animationDuration (integer)
 --- Variable
@@ -69,6 +72,7 @@ end
 --- Returns:
 ---  * A rect-table containing the co-ordinates of the top left corner of the window, and it's width and height
 function window:frame()
+  if cachedFrames[self] then print('cachedframe:',hs.inspect(cachedFrames[self])) return cachedFrames[self] end
   local s = self:size()
   local tl = self:topLeft()
   return {x = tl.x, y = tl.y, w = s.w, h = s.h}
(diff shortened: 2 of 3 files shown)


cmsj:
So according to the docs, setting currentAnimation.duration to 0 ought to immediately finish the animation. Do you fancy testing with that instead of the timer stuff?

lowne:
Ok, it took a while but i tested *all* permutations.
- not calling stopAnimation, as expected, leaks (or something) and HS soon crashes
- .duration=0 has no additional effect over the (required) stopAnimation
- .currentProgress=1 is what "updates" the window frame to the end location; but that seems to happen asynchronously 1 frame later or something like that, so chaining is broken (*not* setting it also results in broken chaining, of course)

As for my last commit: I noticed that the `self` userdata in hs.window methods varies across calls (for the same window), so can't use that as cache table key, so I switched to the window `:id()` - I hadn't done that from the get go because I wasn't sure that :id() would always succeed, but a look at the code seems to indicate otherwise (so there are no checks on that).
A side effect of that is that there should be no worries wrt GC, weak tables, leaks etc. The userdata isn't referenced anywhere and the timer guarantees cleanup. I tried a bunch of additional chaining tests and it all works nicely.

(Alas it's still using a timer, and that still seems inelegant. Naturally if I could be trusted around ObjC I would've tried doing the caching thing in `internal.m` where it belongs, via a private field (please translate to appropriate ObjC lingo) somewhere (I have no idea what's going on with the varying userdata, so I don't know what'd be an appropriate place) holding the cached frame (or topleft+size) - the 5 affected functions would be window_(set)topleft, _(set)size, _transform, mirroring what I did Lua side.)

cmsj:
@lowne I *think* I may have a decent obj-C fix for this. I'm trying desperately to avoid timers, because I think it will be cleaner in the long run. The approach I've taken is to modify various functions to behave differently if an animation is running - things that query the current window frame/position/size will return the target value for the animation, and things that start an animation will now modify an existing animation in preference to killing it and starting a new one.

I've pushed the change to cmsj:fix-chained-animations and it passes all of the tests I had previously, plus the one here and the one in #369. I'd be very grateful if you could give it a try and see if you have any tests that break it (if so, please post them so I can figure out what I missed) :)

lowne:
My informal testing code:
```lua
hs.hotkey.bind(HYPER, 'l', function()
  local w=hs.window.focusedWindow()
  w:moveToScreen(w:screen():next()):moveToScreen(w:screen():previous()):moveToUnit({x=0,y=0,w=0.5,h=0.5})
end)
hs.hotkey.bind(HYPER, 'k', function()
  local w=hs.window.focusedWindow()
  hs.grid.pushWindowLeft().pushWindowRight().pushWindowUp().pushWindowDown()
end)
hs.hotkey.bind(HYPER, 'j', function()
  local w=hs.window.focusedWindow()
  hs.grid.pushWindowLeft().pushWindowUp().pushWindowDown().pushWindowRight()
end)
hs.hotkey.bind(HYPER, 'p', function()
  local w=hs.window.focusedWindow()
  print('P:',w:frame().x)
  hs.grid.pushWindowRight().pushWindowRight().pushWindowDown().pushWindowRight()
  print(w:frame().x,'-------')
end)
hs.hotkey.bind(HYPER, 'o', function()
  local w=hs.window.focusedWindow()
  print('O:',w:frame().x)
  hs.grid.pushWindowRight()
  print('O:',w:frame().x)
  hs.grid.pushWindowRight()
  print('O:',w:frame().x)
  hs.grid.pushWindowDown()
  hs.grid.pushWindowRight()
  print(w:frame().x,'-------')
end)

hs.grid.GRIDWIDTH = 6
hs.grid.GRIDHEIGHT = 6
hs.grid.MARGINX = 10
hs.grid.MARGINY = 10
hs.hotkey.bind(hyper, 'a', hs.grid.pushWindowLeft)
hs.hotkey.bind(hyper, 'd', hs.grid.pushWindowRight)
hs.hotkey.bind(hyper, 'w', hs.grid.pushWindowUp)
hs.hotkey.bind(hyper, 's', hs.grid.pushWindowDown)

hs.hotkey.bind(hyper,'f', hs.grid.resizeWindowThinner)
hs.hotkey.bind(hyper,'h', hs.grid.pushWindowNextScreen)
hs.hotkey.bind(hyper,'g', hs.grid.resizeWindowShorter)
hs.hotkey.bind(hyper,'t', hs.grid.pushWindowPrevScreen)
```

cmsj:
Unless there are any last minute objections, today I'm going to merge my fix-chained-animations branch. We're not perfect, but it should cover most things.

lowne:
I wanted to have one last look at this today, could you make it wait till the evening?

cmsj:
Sure :)

lowne:
As per IRC discussion after the long bug hunt and the discovery of an intractable (albeit not deal breaking) problem with Lua GC messing with the ObjC side, I'll extend this PR to do animations wholly in Lua.

lowne:
Uh, apparently I wasn't supposed to cherry-pick the timer commits. Or at least that's my best guess on what the conflict is  ¯\_(ツ)_/¯ git noobness...

cmsj:
I got a traceback just now, which looks like it's from a window with no :id() or a nil id?

/Users/cmsj/.hammerspoon/init.lua:470: attempt to index a nil value
stack traceback:
	/Users/cmsj/.hammerspoon/init.lua:470: in function </Users/cmsj/.hammerspoon/init.lua:470>
...oon.app/Contents/Resources/extensions/hs/window/init.lua:157: table index is nil
stack traceback:
	...oon.app/Contents/Resources/extensions/hs/window/init.lua:157: in function 'hs.window.internal.setFrame'
	...oon.app/Contents/Resources/extensions/hs/window/init.lua:512: in function 'hs.window.internal.moveToUnit'
	...oon.app/Contents/Resources/extensions/hs/layout/init.lua:186: in function 'hs.layout.apply'
	/Users/cmsj/.hammerspoon/init.lua:476: in function </Users/cmsj/.hammerspoon/init.lua:476>

lowne:
yeah, should've seen that coming, sorry. I'm assuming _setsize & _settopleft won't explode when :id() is nil, or they would've crashed all along.

cmsj:
Manually de-conflicted and merged to master. Thanks!

lowne:
OT, what's the etiquette in these cases? Github doesn't look very happy with what happened, so I imagine I was supposed to fix the merge conflict; but how? I can think of

1. "reverse engineer" the conflict (which was due to shuffled `#import` order) and craft+push a specific commit to remove it
2. cherry-pick the problematic commit from master (in this case the LuaSkin one)
3. merge master into the PR branch
4. merge master into branch, `git rebase -i` and `edit` to discard all the extraneous changes

But 1 and 2 seem awfully ad-hoc, 3 sounds like the atomic option and 4 is a lot of work; isn't there a magic `git` incantation to better take care of these situations?

cmsj:
I usually do something like 4. With the PR branch checked out:

git fetch origin master
git rebase -i master (at this point I tend to use the "fixup" option to squash the branch down to a single commit)
git push -f cmsj

The rebase throws up actual text conflicts if there are any, which I then figure out and fix. The final step overwrite the contents of the PR branch in my user repo of Hammerspoon, leaving me with a branch that can merge cleanly :)

lowne:
Ah, you rewrite *remote* history! (with `-f`, which indeed I do all the time on my "meant for myself only" repos)
I thought github would complain *even more* if I did that... I've read everywhere "*rebase locally all you want, but __never__ rebase remote branches, especially not when they're __public__*" - such as open PR branches, I'd say - "since they might be in use by someone else and it'll all go down in flames" - I suppose for situations like a 3rd person branching a "fix this detail" branch off a PR branch (like what I did with your fix-chained-animations branch... suppose your branch had an open PR here, I branch off to explore further, some commit in master introduces a conflict, you `push -f` a rebase on your branch; what would happen in the PR? What would happen to my repo at my next `git fetch cmsj`?)
Then again if gh is fine with that, and all there is to worry about is a rebase domino effect, I guess that's only a concern for repos with tens of active contributors, so the pragmatic solution wins :)

lowne:
Ok, I held my breath and tried a `git amend` and `git push -f` on another PR branch, and github seemed a bit confused but after a refresh all was fine again; so now I know :smiling_imp:

Issue #355, opened 2015-06-30
hs.hints gc causes crash on reload?


asmagill:
I suspect you're going to ask me to install a point release instead of my development build, but I'll start with this:

What I'm seeing is that if I use the hs.hints module (actually invoke it via hotkey) and select an icon, everything works fine... I select an app, it comes to focus, all is good.  Until I reload my configuration later... then I get a crash -- note that the earlier issue I posted where icons didn't disappear if a reload occurred before selecting an application has been addressed -- I can reload successfully, as long as icons from hints are visible.  After they have been shown at least once and go away because I select a target application/window, then if I reload, I get the crash.  The console log shows the following:

~~~
6/30/15 4:40:31.154 PM Hammerspoon[26209]: objc[26209]: Method cache corrupted. This may be a message to an invalid object, or a memory error somewhere else.
6/30/15 4:40:31.154 PM Hammerspoon[26209]: objc[26209]: receiver 0x6000003e7400, SEL 0x7fff83b1c008, isa 0x600000058910, cache 0x600000058920, buckets 0x6000001bbc80, mask 0x1, occupied 0x0
6/30/15 4:40:31.154 PM Hammerspoon[26209]: objc[26209]: receiver 256 bytes, buckets 224 bytes
6/30/15 4:40:31.154 PM Hammerspoon[26209]: objc[26209]: selector 'retain'
6/30/15 4:40:32.463 PM com.apple.xpc.launchd[1]: (org.hammerspoon.Hammerspoon.336004[26209]) Service exited due to signal: Segmentation fault: 11
6/30/15 4:40:32.536 PM ReportCrash[32151]: Saved crash report for Hammerspoon[26209] version 0.9.32 (0.9.32) to ...
~~~

and the head of the crash file looks like this:

~~~
Crashed Thread:        0  Dispatch queue: com.apple.main-thread

Exception Type:        EXC_BAD_ACCESS (SIGSEGV)
Exception Codes:       KERN_INVALID_ADDRESS at 0x0000000000000008

VM Regions Near 0x8:
--> 
    __TEXT                 0000000106171000-0000000106203000 [  584K] r-x/rwx SM=COW  /Volumes/VOLUME/*/Hammerspoon.app/Contents/MacOS/Hammerspoon

Application Specific Information:
objc[26209]: Method cache corrupted. This may be a message to an invalid object, or a memory error somewhere else.
[...]

PR #356, opened 2015-06-30
remove metatable from hint when closing hint object

asmagill:
Prevents double calls to hints_close.  Addresses #355.


cmsj:
Thanks very much for this. I suspect a rewrite of hs.hints is due at some point, since it should mostly be Lua calling hs.drawing, which would likely help eliminate whatever is really going on here!

PR #359, opened 2015-07-01
Make hs.screen:toEast etc stricter and more sensible

lowne:
I have 4 screens placed horizontally, with different pixel heights. Currently hs.screen:toNorth()/toSouth() can jump "laterally" across screens due to the difference in height (as long as the geometric centres are horizontally misaligned, the target screen is considered to be "north" or "south"); this doesn't seem to make much sense.
The fix takes care of this by discarding screens which have (in the North/South case) *no* horizontal overlap.

Note that this means that an arrangement like this
screen shot 2015-07-01 at 15 52 01
would probably make the rightmost screen "unreachable" by either toEast or toNorth; but I don't think anyone would ever have such an arrangement.

### DIFF
--- a/extensions/screen/init.lua
+++ b/extensions/screen/init.lua
@@ -44,6 +44,7 @@ function screen:next()
   return screens[i]
 end
 
+
 --- hs.screen:previous() -> screen
 --- Method
 --- Returns the screen 'before' this one (I have no idea how they're ordered though); this method wraps around to the last screen.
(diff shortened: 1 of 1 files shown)


cmsj:
Rather than removing the poor choices from the table, what would happen if we put them at the end of the table? That ought to render them unlikely to be chosen except when there is no good choice, right?

lowne:
Well, using my layout as example (screenshot above, but rightmost screen is also aligned horizontally), I never want any screen:toNorth() to return *anything*. If `s` is the leftmost screen, `s:toNorth()` would create a list of 1 screen (the main one, since it's taller); then we'd put the main screen (a poor choice) at the end, but we still have the same list of 1 screen, so toNorth() would still return it.
If you're worried about that crazy corner case, it's easy enough to check for it specifically.

asmagill:
I can imagine cases with windows and screens where I want a "most" North (i.e. where I want anything, but ordered as to what is most likely), and other cases where I want an "absolutely" North and nothing that isn't... what about a flag to indicate if the far outliers should be added as @cmsj suggests?

cmsj:
I like the idea of making it configurable. I'm undecided whether I think it would be better as an optional method argument, or as a module variable (or indeed both, as animationDuration is). Any thoughts?

lowne:
Added both way to configure strict mode.
Also added an optional "starting point/rect"; the use case is this: suppose a screen layout like this (3 identical monitors):
```
    |------|
    |      |
    |------|
|------||------|
|      ||      |
|------||------|
```
If i'm `hs.grid.pushWindowDown()`ing from the top screen (smack in the centre above the lower two), if the window is at the bottom left, it'll jump to the left screen; if bottom right, right screen.
Please note that I haven't actually tested this code yet, will update when I do.

As a side note, I think `hs.geometry.intersectionRect` should behave like the `projection` function in the commit. It's easy to establish when there's no actual intersection (.w *or* .h are 0) - having all fields 0 loses useful information.

lowne:
Tested informally, works.

PR #360, opened 2015-07-02
Hotkey fixes

### LINKED ISSUE #358: obscure hs.hotkey.modal error with first bound key's repeat function
Switching back temporarily from my WIP eventap based hotkey to determine if random temporary key lockup is due to my code or something else, and I discovered something else very odd with hs,hotkey (the one currently in core)...

If you have a modal key setup, something like:

~~~
test = hs.hotkey.modal.new({"cmd","alt","ctrl"},"f")
function test:entered() hs.alert.show("entered modal mode") end
test:bind({"cmd"},"c", function() print("cmd-c down") end, nil, function() print("cmd-c repeat") end)
test:bind({},"escape", function() test:exit() end)
function test:exited() hs.alert.show("exiting modal mode") end
~~~

Now, using the core hs.hotkey, if you enter the modal mode by hitting the `cmd-alt-ctrl f`, the repeat function for the first bound key gets repeatedly triggered after test:entered() is executed.  Without pressing any key at all.  Press any key, and this repeating stops, but...

Exit and re-enter the modal -- it starts up again.

I doubt the specific modifiers matter -- the specific keys don't -- but this is pretty close to the code I'm using that brought this to my attention.  And if the repeat function for the first binding is nil, then this doesn't happen (or maybe it does, but being nil, has no visible effect.)

I don't know the hs.hotkey repeat code at all, so I throw this out in case someone has an "A-Ha!" thought... otherwise, I'll look into it tomorrow.

asmagill:
only trigger timer if hotkey definition has a repeatfn defined
call modal:entered and modal:exited as methods, rather than functions because that's how we define them (i.e. with an implicit self)

Fixes #358

### DIFF
--- a/extensions/hotkey/init.lua
+++ b/extensions/hotkey/init.lua
@@ -161,7 +161,7 @@ function hotkey.modal:enter()
     self.k:disable()
   end
   fnutils.each(self.keys, hotkey.enable)
-  self.entered()
+  self:entered()
   return self
 end
 
@@ -182,7 +182,7 @@ function hotkey.modal:exit()
   if (self.k) then
     self.k:enable()
   end
-  self.exited()
+  self:exited()
   return self
 end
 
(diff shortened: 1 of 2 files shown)


cmsj:
Thanks! I just tweaked it a little with 2d4c3a6 to collapse all the conditionals into the single if().

PR #361, opened 2015-07-02
improved hs.grid

lowne:
- added separate grids per screen
- added interactive resizing mode
- improved .pushWindowAround to cross screen boundaries (with #354 and #359 it's a lot of fun)
- fixed a couple minor bugs+docstrings in the original
- fully backward compatible

### DIFF
--- a/extensions/grid/init.lua
+++ b/extensions/grid/init.lua
@@ -187,6 +187,17 @@ end
 ---  * In the modal interface, press the arrow keys to jump to adjacent screens; spacebar to maximize/unmaximize; esc to quit without any effect
 ---  * The keyboard hints assume a QWERTY layout; if you use a different layout, change `hs.grid.HINTS` accordingly
 
+--- hs.grid.hide()
+--- Function
+--- Hides the grid, if visible, and exits the modal resizing mode.
+--- Call this function if you need to make sure the modal is exited without waiting for the user to press `esc`.
+---
+--- Parameters:
+---  * None
+---
+--- Returns:
+---  * None
+
 --- hs.grid.HINTS
 --- Variable
 --- A bidimensional array (table of tables of strings) holding the keyboard hints (as per `hs.keycodes.map`) to be used for the interactive resizing interface.
@@ -740,7 +751,9 @@ function grid.show()
   resizing:enter()
 end
 
-
+function grid.hide()
+  if initialized then resizing:exit() end
+end
 
 
 -- Legacy stuff below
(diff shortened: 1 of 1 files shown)


lowne:
@knl I'd love to have your input on this PR wrt #282. 
Afaict I'm missing some additional functions (such as the "most"s) and, importantly, the ability to have multiple grids per screen; the former is easily addressed, and as for the latter I tried to prepare the code for accommodating that with minimal disruption.
However I stopped short of actually adding all that because I wasn't clear on the purpose, which informs api design; I suspect these features are meant to be used for layout-like purposes (vs "manual" window management by the user) - can you confirm? Or more specifically, what's your use case (and/or init.lua) for grille?

cmsj:
I've been testing this out today, it seems to work very well. One thing I did notice, is that doing hs.grid.show() on a fullscreen window and hitting space, pops the window out of fullscreen, but doesn't hide the overlay. Was that intentional?

lowne:
Yes, the idea was that you might want to place the window somewhere else right away - if not you'd just press `esc`. However upon revisiting it now I thought that resizing fullscreen windows, while kind of cool, doesn't make much sense (it'll just leave black empty space); so with 6858cca when `show()` is called on a fullscreen window, it's immediately popped back (and the grid is still shown). But I'm not 100% sure it's the right behaviour - an argument can be made that it doesn't respect the (likely) user expectation.
(OT: #402 - which I had already fixed - caused the merge conflict; I fearlessly `rebase`d and `push -f`ed and indeed nothing exploded on github :dancer: )

cmsj:
I did also notice that the resizeWindow{Taller/Thinner/etc} functions don't behave quite the same way. I'm fine with them taking a window argument, it makes a lot of sense, but it would be more backwards compatible (and perhaps convenient) if they would default to using `hs.window.focusedWindow()` if no argument has been supplied.

cmsj:
(I may just hack that in myself, since I want to get 0.9.34 out the door tomorrow, and I really want this improved hs.grid implementation to be in it :)

cmsj:
Ah, I see what the cause of that was. I'm using hs.urlevent for my grid operations. Previously I was just binding directly to hs.grid.resizeWindowFoo(), which were unconditionally using focusedWindow(). Now that there is more going on in those functions, they are actually paying attention to the arguments given to them, which in the case of a urlevent callback, is two strings.

I don't think it's a good idea to build logic into hs.grid to specifically detect that it has been passed a string instead of a window, so I'll just document it in the release notes as an API change.

justmytwospence:
I used to be setting margin to 0 in my config like this:

```
grid.MARGINX = 0
grid.MARGINY = 0
```

Whats the right way to do that now?

lowne:
@justmytwospence that still works, or `hs.grid.setMargins'0x0'`

PR #367, opened 2015-07-05
add high precision hs.timer.time

lowne:


cmsj:
I'm surprised that Lua doesn't have that!

How about we can it hs.timer.secondsSinceEpoch()? .time() is quite general

lowne:
I called it `time` to mirror `os.time()`, I was going with `gettime` originally. Of course feel free to rename it to whatever, I just need the function to be there - `secondsSinceEpoch`, while not very concise, is nicely descriptive.

Issue #254, opened 2015-04-02
Add a global window event watcher


cmsj:
We have hs.application.watcher for global application events, but no equivalent for window events (creation, destruction, moving, resizing, etc)

AFAICS there's no global window event mechanism in the OS X SDK, so instead this would have to operate at the application level, creating hs.uielement.watchers for each app, to monitor its windows, and then dispatch those events to subscribers.

(While this could be implemented in objc, it would just be repeating what's already in hs.application.watcher and hs.uielement.watcher, so this would probably make the most sense to be done in pure Lua, building on those extensions)

cmsj:
Something like https://gist.github.com/cmsj/591624ef07124ad80a1c might well do here (that gist is forked from @tmandry 's gist repo)

cmsj:
Anyone fancy getting that in shape to be hs.window.watcher? :D

tmandry:
@cmsj I have updated the code since that gist, here's the version I have now: https://gist.github.com/tmandry/43ca021fbd3fb8b42a20

Can you tell if anything further needs to be done on it? Or is it pretty much in good shape to be incorporated into HS?

cmsj:
This is covered by #374

PR #372, opened 2015-07-07
Add windowtracker extension

tmandry:
Here's my response to #254. I personally think this is a simple and effective approach to dealing with window events. Happy to discuss and make revisions.

I've toyed with the idea of giving it criteria for which windows to watch, but not sure on the interface or what the use case is. (For the interface, it could accept a list of app bundle IDs or something.) What do you think?

Docs
---

Track all windows on the screen. windowtracker abstracts away applications and treats all
windows the same, subscribing to all events on all windows.

You can watch for the following events:
* hs.windowtracker.windowCreated: A window was created.
* hs.windowtracker.windowDestroyed: A window was destroyed.
* hs.windowtracker.mainWindowChanged: The main window was changed. This is usually the same as
  the focused window (except for helper dialog boxes like file pickers, which are not reported
  by this event). Note that switching applications triggers this event, unlike the OS X
  accessibility API.
* hs.windowtracker.windowMoved: A window was moved.
* hs.windowtracker.windowResized: A window was resized.
* hs.windowtracker.windowMinimized: A window was minimized.
* hs.windowtracker.windowUnminimized: A window was unminimized.

Note that Hammerspoon windows (the console) are ignored by windowtracker. This is because the
console pops after an error, which can cause infinite exception loops, rendering the computer
unusable.

Caveats
---
Calling `start()` can hang for awhile because of asking every app for all of its windows (the same way that `hs.window.allWindows()` works).

### DIFF
--- a/extensions/windowtracker/LICENSE
+++ b/extensions/windowtracker/LICENSE
@@ -1,0 +1,21 @@
+The MIT License (MIT)
+
+Copyright (c) 2014 Various contributors (see git history)
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
--- a/extensions/windowtracker/init.lua
+++ b/extensions/windowtracker/init.lua
@@ -1,0 +1,176 @@
+--- === hs.windowtracker ===
+---
+--- Track all windows on the screen. windowtracker abstracts away applications and treats all
+--- windows the same, subscribing to all events on all windows.
+---
+--- You can watch for the following events:
+--- * hs.windowtracker.windowCreated: A window was created.
+--- * hs.windowtracker.windowDestroyed: A window was destroyed.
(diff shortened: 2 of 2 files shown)


cmsj:
aha, @lowne also has a windowtracker module. I now have no idea which one to pick ;)

(perhaps you guys could look at each other's windowtracker module and see if anything jumps out, feature/bug/size/quality wise?)

lowne:
I'll open a PR with mine later today

tmandry:
Had a quick look. Really interesting what yours is doing, Mark. There is a
lot of sophistication in there, some of which (app filters) I'd like to add
to mine.

Right now I'm unclear on what much of the rest is for, but will have to dig
deeper when I have time. I saw som deferred/timer code, which makes me a
little uneasy, but like I said I haven't really taken the time to grok it :)

For context, this is the code I've been using for tmandry/whammy and I just
now got around to polishing it up for HS inclusion. I've been really happy
with it so far. Curious to see what Mark is doing with his, maybe there are
some things I hadn't thought of.

On Monday, July 6, 2015, Mark Lowne <notifications@github.com> wrote:

> I'll open a PR with mine later today
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Hammerspoon/hammerspoon/pull/372#issuecomment-119093574>
> .
>


-- 
Tyler Mandry

lowne:
I finally landed my PR at #374 (but the conversation started here, so here it shall continue).

First, the story about my take: as I started writing modules for my own use, there were two things that I was doing over and over on each module:
1. limit the scope of whatever the module was doing to specific windows
2. dance the dance with watchers for getting events (that got old *real* quick)
And so I made windowfilter and windowwatcher, respectively; the latter basically started exactly as this one, which I was unaware of; this "independent invention", along with #254 (which I also just recently learned of) tells me that I'm not alone in dreading OS-level event watchers :)
I made windowwatcher depend on windowfilter because I *always* want at least basic filtering (which is what `hs.windowfilter.default` is about); during revisions I integrated them more and more, until I decided it made sense to just merge them - and the "windowfilter" name might not be the best choice for the resulting module, it just stuck (and is ofc amenable to improvement).

As its sheer size suggests, windowfilter does quite a bit more than windowtracker (which is not necessarily a good thing by itself). These imho are the more interesting additions:
* the whole filtering business, which is half the module
* in the "watcher" part alone, for a fairer comparison, windowfilter keeps its own internal data structure tracking all the apps and windows; incoming events act on it, and then windowfilter events are emitted; this allows it to do a lot of cleanup on the incoming event stream, making the resulting stream *much* saner (see comments at the top of the file for examples); the end user needn't worry about OS X quirks (which bit me many many times) 
* the :notify method is very convenient for otherwise quite convoluted use cases - see the "timewaster" example in my PR, which doesn't really make it justice but hopefully gets the idea across.

lowne:
Also fwiw the timers simply act as flood protection for windowMoved/windowTitleChanged events, to avoid having the user callbacks run 200 times when windows are dragged around.

tmandry:
@lowne Sorry I haven't been able to look over it yet, I will try to do it tonight. From what you said I think your additional layer of abstraction could be useful.. Anyway, I think if we put our heads together, whatever we end up with will be great :)

cmsj:
@tmandry so I decided to land @lowne 's windowfilter module, which I think pretty much covers this PR. Before I go ahead and close it, is there anything that we're missing?

PR #374, opened 2015-07-08
Add hs.windowfilter

lowne:
Main features:
* set the exact scope of what you're interested in:
  * `wf_terminal=hs.windowfilter.new'Terminal'` - Terminal windows
  * `wf_timewaster=hs.windowfilter.new(false):setAppFilter('Safari','reddit')` - any Safari windows with "reddit" anywhere in the title
  * `wf_bigwindows=hs.windowfilter.new(function(w)return w:frame().w*w:frame().h>3000000 end)` - only very large windows
* naturally all the filters are dynamically updated
* easily subscribe to (sane) events:
  * `wf_terminal:subscribe(hs.windowfilter.windowFocused,some_fn)` - run a function whenever a terminal window is focused
  * `wf_timewaster:notify(function(wins)if #wins>0 then startAnnoyingMe() else stopAnnoyingMe() end end)` - fight procrastination :)
* never go crazy again trying to have application, window, uielement watchers waltz together

(EDIT: hi #254)

### DIFF
--- a/extensions/windowfilter/init.lua
+++ b/extensions/windowfilter/init.lua
@@ -627,19 +627,19 @@ end
 function App:activated()
   local prevactive=global.active
   if self==prevactive then return log.df('App %s already active; skipping',self.name) end
-  if prevactive then prevactive:deactivated() end
+  if prevactive then prevactive:deactivated(true) end
   log.vf('App %s activated',self.name)
   global.active=self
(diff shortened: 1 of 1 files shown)


lowne:
Whops, I just realised that the addition of :notify caused a huge regression - will fix as soon as I can.

PR #377, opened 2015-07-09
eventtap updates

asmagill:
* Add additional NSEvent based events for watching
* add getRawEventData method to examine event data in more detail (really only useful to developers in determining what helper methods might be necessary, but read only, so harmless)
* Clean up types and properties documentation
* add systemKey method to get information about brightness, contrast, media, power key etc.
* add getCharacters method so eventtap can be used as a key-input capture method without having to understand every locales specific keycode arrangments
* made setKeyCode and setFlags chainable
* getProperties and setProperties know when to use integer versus number
* __tostring metamethods for properties and types tables
* auto-restart eventtaps when disabledByTimeOut or disabledByUserInput occurs
* cleaner shutdown/restart of eventtap, runloopsrc and machport

### DIFF
--- a/extensions/eventtap/event.m
+++ b/extensions/eventtap/event.m
@@ -1,4 +1,6 @@
 #import "eventtap_event.h"
+#import <IOKit/hidsystem/ev_keymap.h>
+
 #import "../hammerspoon.h"
 
 static int eventtap_event_gc(lua_State* L) {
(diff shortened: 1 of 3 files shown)


asmagill:
Travis doesn't have the 10.10.3 Xcode API Updates... should I remove NSEventTypePressure for now?

asmagill:
Maybe related, but (a) I don't know much about travis, and (b) it's not like I know what the event is useful for yet -- I just want it available so someone who does might think of what needs to be added for full support:   http://blog.travis-ci.com/2015-07-08-xcode-6-4-now-available/

cmsj:
NSEventTypePressure relates to force touch trackpads. I doubt it's relevant to us at this stage :)

cmsj:
thanks!

PR #384, opened 2015-07-11
support builtin system images with hs.drawing.image

asmagill:
Allows you specify builtin system images by name with hs.drawing.image... see picture for example

image

or try this code:
~~~lua
a = {
  hs.drawing.rectangle{x=10, y=40, w=1420, h=720}
      :setRoundedRectRadii(20,20):setStroke(true):setStrokeWidth(10)
      :setFill(true):setFillColor{red=1, blue=1, green = 1, alpha = 1}:show()
}
for i,v in ipairs(hs.drawing.systemImageNames) do
    table.insert(a, hs.drawing.image({
          x=20 + ((i - 1) % 14) * 100,
          y=60 + math.floor((i-1)/14) * 140,
          h=100, w=100}, v):show())
    table.insert(a, hs.drawing.text({
          x=20 + ((i - 1) % 14) * 100,
          y=160 + math.floor((i-1)/14) * 140,
          h=40, w=100}, v)
          :setTextSize(10):setTextColor{red = 0, blue = 0, green = 0, alpha=1}
          :setTextFont("Menlo"):show())
end

-- hs.fnutils.map(a, function(a) a:delete() end) -- use this to clear from screen
~~~

### DIFF
--- a/extensions/drawing/init.lua
+++ b/extensions/drawing/init.lua
@@ -29,8 +29,18 @@ local __tostring_for_tables = function(self)
     return result
 end
 
-module.fontTraits      = setmetatable(module.fontTraits,      { __tostring = __tostring_for_tables })
-module.windowBehaviors = setmetatable(module.windowBehaviors, { __tostring = __tostring_for_tables })
+local __tostring_for_arrays = function(self)
+    local result = ""
+    for i,v in ipairs(self) do
+        result = result..v.."\n"
+    end
+    return result
+end
+
+table.sort(module.systemImageNames)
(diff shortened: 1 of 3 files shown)


asmagill:
I went back and forth about adding the table of image names... it might not (in fact probably doesn't) contain every possible image name that will work with [NSImage imageNamed:_x_], and system updates may remove or add names.

But I would recommend keeping at least the changes to the image loading code, so that they can be referenced, even if we have to look up the name to use elsewhere.

lowne:
This is quite useful! I'd say keep the names and just add a caveat in the docs to the effect that some names might stop working after OS updates etc.

asmagill:
superseded

PR #386, opened 2015-07-11
[WIP] add optional alert to hs.hotkey.bind

lowne:
(EDIT: #409 is a superset of this one; but read on, there's useful discussion here)

Fully backward compatible.
Imho visual feedback can help greatly with creating 'muscle memory', especially right after the initial user configuration setup, especially if said configuration has a lot of hotkeys, and especially if the resulting actions might not always be obvious as the user gets acquainted with HS - so maybe we could encourage this by making it a little bit easier on new users, which would hopefully lead to higher "customer *sat*", which would lead to higher retention, which would lead to more users.

(Also fixed up docstrings a little)
(The concave diamond thing is meant for those of us who map a HYPER modifier via Karabiner - I doubt anyone would actually want to use the physical-world `ctrl-alt-cmd-shift` as modifier)

### DIFF
--- a/extensions/hotkey/init.lua
+++ b/extensions/hotkey/init.lua
@@ -6,7 +6,7 @@ local hotkey = require "hs.hotkey.internal"
 local keycodes = require "hs.keycodes"
 local fnutils = require "hs.fnutils"
 
---- hs.hotkey.new(mods, key, pressedfn[, releasedfn, repeatfn]) -> hs.hotkey
+--- hs.hotkey.new(mods, key, pressedfn, releasedfn, repeatfn, message, duration) -> hs.hotkey
 --- Constructor
 --- Creates a new hotkey
 ---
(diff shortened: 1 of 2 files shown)


cmsj:
This is a nice idea, although I have a couple of comments:

 * I failed to update the docs properly when I did some recent work on hs.hotkey. pressedfn is also optional, so you should probably take that into account (and either fix up the docs for pressedfn, or I can do that)
 * I think the functionality should be moved to hs.hotkey.new, rather than .bind, so people who are constructing hotkeys manually can still take advantage of this.

lowne:
Should be fixed now

cmsj:
I found another issue here - it's looking up the mods in hs.utf8.registeredKeys, which only lists the modifier keys by name, but it's valid to bind with the UTF8 values to (e.g. I have: `local hyper = {"⌘", "⌥", "⌃", "⇧"}` in my init.lua)

asmagill:
@cmsj re your comment about hs.utf8.registeredKeys, this is the nature of what registeredKeys was designed for... to give a textual label to un-typable (or at least hard to type) keys on average keyboards that makes sense to someone just reading the code... should a method be added to go the other direction (i.e. return the label, if it exists, for a unicode character?  what should it return if the label doesn't exist?)

cmsj:
I'm looking at this again and I'm more worried than last time about the type dance in hotkey.new().

I can't help but wonder if this should actually be new API - hs.hotkey.newWithAlert(mods, key, message, duration, pressedfn, releasedfn, repeatfn) and hs.hotkey.bindWithAlert(mods, key, message, duration, pressedfn, releasedfn, repeatfn).

What do people think?

lowne:
@cmsj see convo at #409 (TL;DR: I (fwiw) disagree (unless there are compelling reasons to ditch the type dance) but @asmagill agrees with you)

cmsj:
@lowne I think the reason is the complexity it adds. I suspect it might be a lot simpler if we do two things:
 * drop the duration argument and make that a variable on the module
 * move the message argument to before the callbacks, that way we only need to do the type checking on a single argument, and we immediately know where we are.

lowne:
Having the message/label before the callbacks results in *much* better readability in the init.lua, so I'm all for it; originally I discarded the possibility as I just defaulted to "maintain backward compatibility=add new arguments at the end" without much thought. I'm more ambivalent about dropping the duration argument, probably there will be a 1% of cases where users want a different duration - then again, they can call have the alert in their own callback for those cases. 
So: :+1: - I'll do both things in #409 (superset of this PR) and re-implement here (the branches diverged significantly) if that one is rejected and this one is not.

lexinator:
sorry this is a bit out of the blue but I was just thinking today that it'd be nice to display a help page to show all the key bindings.  I often know that I have a hotkey for some task, but I don't always remember which exact key it is, so off I go to read the hammerspoon dotfile.

To solve my problem I would need to associate a short description to each keybinding, then I could iterate over all the keybindings(I've been assuming I can get all the bounded keys) and include the description to show as a help page.  With this pull request, the """message""" could be the description field.  If duration is 0, then it would not be an alert (or >0 could be used for both)

Does this sound plausible?

lowne:
@lexinator see #409, it's already there, for exactly the reason you stated, and implemented exactly as you described :) - I'll also add another method to just get the plain table for specialised consumption as per conversation there.

lexinator:
nice!  The title of #409 didn't give any hint that it would solve my problem =]  thanks.

lowne:
@Ng: Turns Out™ there's a slight problem: if the signature is `hs.hotkey.bind(mods,key,[msg,]pressedfn,releasedfn,repeatfn)`, where msg can be omitted (as opposed to just nil) for backward compatibility, in `hs.hotkey.bind('cmd','=',nil,function()print'gotcha!'end)` is that callback pressedfn or releasedfn?

A possible solution is to make `msg` the first parameter, but then it can't be just omitted in the "new syntax" in 409 (which takes a much more convenient freeform string for `mods`, and that's too good to give up), so if I don't want a message I can do 
- `bind({'cmd'},...) -- old school` or
- `bind(nil,'cmd',...)`

and I'm not particularly fond of either, but I can live with the latter.

----

...Unrelatedly, why is the type dance so awful? It *still* feels innocent enough to me:
```lua
  if type(releasedfn)=='string' then duration=repeatfn message=releasedfn repeatfn=nil releasedfn=nil
  elseif type(repeatfn)=='string' then duration=message message=repeatfn repeatfn=nil end
```

lowne:
...Nope, I don't quite like msg as the first parameter. Ok, last offer: msg is third parameter where it belongs, the gotcha from my previous comment is solved by assuming the old syntax (where msg doesn't exist) if `mods` is a table, and the new one (where msg must be a string or explicitly nil) if `mods` is a string, which I think is reasonable since the new docs never mention tables for `mods`. Sounds good?

lowne:
As per @cmsj suggestion on IRC, solved the gotcha another way: if the user doesn't want an alert, `message` (in 3rd position) must be *actually* omitted; a `nil` there will be interpreted as belonging to `pressedfn` (this is made explicit in the docs).
@cmsj: the functionality to take a string for `mods` is deeply embedded within the whole alert thing to begin with (mods need parsing anyway to generate a string for the keyboard combination). So it cannot (semantically) be its own commit, separate from the alert infrastructure. The way to remove this capability (and force mods to be a table like it was previously) is to simply add `if type(mods)~='table' then error(...) end`, and then revert that to have an "atomic change"; but that seems rather silly...

NOTE: the *actual* code is all in #409 as per notices above:
> re-implement here (the branches diverged significantly) if that one is rejected and this one is not.

lowne:
Todo: if #409 lands, nothing; this PR can be closed. Otherwise:
- [ ] backport just the label/alert stuff from #409, remove the nesting infrastructure

lowne:
Superseded by #409, closing.

PR #388, opened 2015-07-11
add hs.drawing.imageFromWindowID

asmagill:
Create hs.drawing.image from application window by id (hs.window:id())

Caveat -- doesn't like hidden/minimized windows... sometimes these have nil for the ID so it makes sense, other times (specifically System Preferences) retains their ID, and "screencapture -l x filename" works, but... I can't seem to figure out why this doesn't.  "Wisdom" on the net from my initial research says "it can't be done" because the image is released from memory, but... apparently screencapture knows something else...

At any rate, it's a good start, and if anyone knows of a way around this limitation, I'm all ears!

### DIFF
--- a/extensions/drawing/internal.m
+++ b/extensions/drawing/internal.m
@@ -332,6 +332,33 @@ NSMutableArray *drawingWindows;
 
     return;
 }
+
+- (void)setImageFromWindowID:(NSInteger)windowID {
+    CGImageRef windowImage = CGWindowListCreateImage(
+          CGRectNull,
+          kCGWindowListOptionIncludingWindow,
+          windowID,
+          kCGWindowImageBoundsIgnoreFraming | kCGWindowImageShouldBeOpaque);
+
+    if (!windowImage) {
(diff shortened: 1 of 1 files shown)


asmagill:
In thinking more about this... would we be better off creating a NSImage userdata and allowing other modules like hs.screen (for screenshot) and hs.window (for windows) and others not thought of yet to be able to create an image anyway it wants, as long as it ends up an NSImage type, and then have hs.drawing handle this type from any source?  Not sure what implications this might have for GC as one module will be creating a chunk of data which then becomes the provence of another, but... it scales better with image generation ideas we haven't even envisioned yet...  thoughts?

cmsj:
I think the time has definitely come for an hs.image object.

As for lifecycle, I think the safest option is to clearly state that you can create and manipulate an image object, but when you use it, the thing you used it in, makes an immutable copy (which is trivial in C because either a) the Apple API explicitly makes a copy, or b) we make a copy using the NSCopying pattern).

asmagill:
Hold off on this and #384... I've got an idea for a start to hs.image that would merge both of these...

asmagill:
superseded

PR #392, opened 2015-07-13
hs.image module

asmagill:
Suggested base hs.image module which include the systemImageNames from #384 and addition to hs.window for window snapshots from #388.  

Change to drawing to support the hs.image (hs.drawing.hsImage) type.  Change to hammerspoon.h adds 2 functions for getting and creating an hs.image object from an NSImage object in other modules.

I envision adding functions/methods for building composite images from multiple sources, etc. in hs.image, so ideally anything that uses an image (drawing and menubar) could benefit from an image from any source, and anything that generates an image (window snapshot, maybe port application Icon and screenshot from screen) can be one of those sources.

### DIFF
--- a/extensions/drawing/internal.m
+++ b/extensions/drawing/internal.m
@@ -332,6 +332,16 @@ NSMutableArray *drawingWindows;
 
     return;
 }
+
+- (void)setImageFromHSImage:(NSImage *)newImage {
+
+    self.HSImageView.image = newImage;
+    self.HSImage = newImage;
+
+    self.needsDisplay = true;
+
+    return;
+}
 @end
 
 // Lua API implementation
@@ -692,7 +702,7 @@ static int drawing_newImage(lua_State *L) {
 
             break;
         default:
-            CLS_NSLOG(@"ERROR: Unexpected type passed to hs.drawing.text(): %d", lua_type(L, 1));
+            CLS_NSLOG(@"ERROR: Unexpected type passed to hs.drawing.image(): %d", lua_type(L, 1));
             lua_pushnil(L);
             return 1;
             break;
@@ -756,7 +766,7 @@ static int drawing_newAppImage(lua_State *L) {
 
             break;
         default:
-            CLS_NSLOG(@"ERROR: Unexpected type passed to hs.drawing.text(): %d", lua_type(L, 1));
+            CLS_NSLOG(@"ERROR: Unexpected type passed to hs.drawing.appImage(): %d", lua_type(L, 1));
             lua_pushnil(L);
             return 1;
             break;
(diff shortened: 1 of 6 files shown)


cmsj:
Looking good so far. I think we should slim down on the C side a bit, and have more of the hs.drawing image creation stuff live in Lua and call through to the underlying hsImage implementations. I can work on that after landing this though :)

cmsj:
I'm about to do some reworking on this, sorry, should really have merged it into a feature branch and then worked on it and then landed on master.

asmagill:
No prob... I'm curious to see what you want to move into Lua... I do see how adopting something like this module suggests some rework to drawing and maybe a few others, but I think in the long run it will be easier to extend this way.

BTW, if you know of an easy way to extend saveToFile so that it can render a larger version of an image to a file cleanly like when rendering one of the system icons to the screen, I'd love to see it... pointers to code or adding it yourself is fine -- the examples I've seen so far were ... not pretty!  Noticed this when testing with writing a file from hs.image.systemImageNames.TrashFull (for example), but figured I should get this out and let others more knowledgable see it and make suggestions.

cmsj:
See 0bcd26a for the changes. I've basically converted all of hs.drawing over to using hs.image, and fleshed out the constructors in hs.image with some more of the things that used to be in hs.drawing.

cmsj:
(I did also slightly tweak some API names, most notable hs.image.getImageByName becomes hs.image.imageFromName)

asmagill:
Developer note in docs no longer necessary (see hammerspoon.h).  I'll update it in a later pull if someone doesn't first -- I just want to make it clear in case someone starts to use it before then...

>DEVELOPER NOTE: If you need to create an hs.image object from within another module, make sure to include this module in the requirements to ensure that the `hs.image` userdata type exists before you create a new image.

lowne:
@asmagill here `hs.window:snapshot()` returns a valid 1x1 transparent image instead of `nil` when called on non-visible windows - is that the intended behaviour?
(also minor nitpick, the docstring says `Function` instead of `Method`)

asmagill:
That is something I've been trying to work around but haven't found a way to yet... everything I find online says there is no way to do it, but screencapture manages it, so I'm not sure how it gets around this.

So, it's expected because I'm not sure how to fix it yet, but it's not wanted, if anyone has any thoughts!

As to the doc, It's fixed in my working tree as I try out things to expand hs.image, but I haven't pushed it yet... Thanks for the reminder!

cmsj:
@asmagill can you give me an example of a screencapture command that does that? (are you just going by window id?)

cmsj:
https://gist.github.com/chinmaygarde/4647506 is an interesting method of getting a screen capture, which also looks like it would work for a single window. I wonder if it works with non-visible ones. 

I also note that WebKit seems to use CGSHWCaptureWindowList()

asmagill:
The one I test with is the System Preferences application because it does have a window id even when the window is minimized or hidden -- not all "hidden" windows do, and without a window id, our code doesn't work at all.

Open System Preferences, then minimize the window (or hide it with Command - H... both show this behavior)

From the HS console, enter the following:  `hs.appfinder.appFromName("System Preferences"):allWindows()[1]:id()`

Note the number, and from Terminal, type `screencapture -l # test.png`. Open the image to verify that it shows up.

Now, from the Hammerspoon console, type `a = hs.drawing.image({x=100,y=100,h=500,w=500},hs.appfinder.appFromName("System Preferences"):allWindows()[1]:snapshot()):show()`

Or write it to a file... doesn't matter... either way, screencapture gets the hidden window, while our code doesn't.  Looking online, the consensus is that this is because hiding or minimizing a window causes its bitmap representation to be released and therefor "not available", but nothing I've seen so far explains how screencapture is able to get an image anyway.

asmagill:
This is a little different than our code -- it stays within CoreGraphics to generate the image data, while our code uses NSImage to do the actual generation... I'll take a look at it this afternoon when I have some time.

cmsj:
@asmagill definitely try the CGSHWCaptureWindowList() method, I believe that's what screencapture uses.

asmagill:
@cmsj, What should I import to define CGSHWCaptureWindowList?  ack CGSHWCaptureWindowList in /System/Library/Frameworks turns up nothing, and I keep getting "implicit declaration..." warnings and/or errors in trying to adapt this code.  (BTW, the Gist produced the same results we currently get, so the webkit approach may be the way to go if i can figure out what to import or a docset that for Dash that includes these...)

cmsj:
@asmagil hmm, it's private API. The defines are in https://github.com/WebKit/webkit/blob/master/Source/WebCore/platform/spi/cg/CoreGraphicsSPI.h

I'm not too concerned about it being private, if Apple is using it in public WebKit sources

asmagill:
I'll take a look at that, but I think I've worked around it... all of the examples I've seen online use `CGWindowListCreateImage` to create the image... makes sense, as it takes a CGWindowID as one of the parameters and allows you to render things above and under wrt to the window in question...

However, a line in the description for `CGWindowListCreateImageFromArray` caught my eye:  *"This function ignores the onscreen ordering of the windows and instead composites them using the order specified in the windowArray parameter"* -- if it's rendering them out of display order, then it must be actually rendering the contents at the time of the snapshot...

Initial tests look promising, and I hope to have a pull request up shortly...

asmagill:
Followup in pull #413

PR #393, opened 2015-07-13
add __tostring methods to result tables for easier use in console

asmagill:
Doing some cleanup and adding methods to make data more easily identifiable and useful in console or in printed output.  Probably more to follow, but I want to keep this minor stuff separate from new modules or significant changes to anything.

So far:  hs.battery.getAll(), hs.host.vmStats(), and hs.host.cpuUsage()[x] all output to the console or via print without need for hs.inspect.

### DIFF
--- a/extensions/battery/init.lua
+++ b/extensions/battery/init.lua
@@ -7,11 +7,26 @@
 
 
 local module = require("hs.battery.internal")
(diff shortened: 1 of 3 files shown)


cmsj:
Awesome, thanks. I was just thinking that we should do this :D

PR #396, opened 2015-07-14
hs.menubar uses hs.image for setIcon

asmagill:


asmagill:
BTW, re the //Fixme lines... to the asker:  the lua_settop(L, 1) line only make sense if the method is a chain-able one -- it makes sure that, whatever else the method may have done, the return value is the object the method was called on.  Since the methods in question aren't chain-able (their return value indicates success/failure) you can safely remove them.

cmsj:
Thanks, I completely forgot to add that conversion last night. I think that's us completely converted to hs.image! woo :)

asmagill:
I think we still have hs.screen.shotAsJPG and hs.screen.shotAsPNG to convert... divesting the screencapture code from the write to file code and using hs.image.saveToFile would help with #397, I think

cmsj:
@asmagill yup, I'm working on that right now :)

PR #409, opened 2015-07-22
[READY] Hotkey nesting

### LINKED ISSUE #184: inconsistent behaviour when doubly registering hotkeys
When I do:
```lua
h1=hs.hotkey.bind({"cmd", "ctrl", "alt"}, 'y', function() hs.alert.show(1) end) h2=hs.hotkey.bind({"cmd", "ctrl", "alt"}, 'y', function() hs.alert.show(2) end) h1:disable()
```

I expect the hotkey to alert '2', but instead nothing happens at all.

If I then do
```lua
h2:disable() h2:enable()
```

The '2' starts to work.

Also, when I do `h1=nil` while `h1` is enabled, the hotkey still triggers a '1' and I can't do anything about it other than reloading/restarting Hammerspoon.

@cmsj mentioned "we don't check the return values of the OS X register/unregister hockey calls, which would probably help there."

lowne:
This branch continues from #386, but different goal so separate PR.

- Fixed #184 and similar issues (potentially more troubling than the examples in that issue is the fact that modal hotkeys use the same global pool - one wouldn't expect conflicts there) by allowing multiple hotkeys to be registered for the same keyboard combo following nested shadowing logic; when a hotkey is enabled "over" of a previous one, the latter is disabled, and automatically reenabled should the former get disabled/deleted. (There's an undocumented parameter in `hotkey:enable` to override the nesting rule, so that an already shadowed hotkey can be forced back on top - no use case in mind, it's there just in case it turns out to be needed for something)
- Changed the `mods` parameter to string, like `'cmd-alt'` or `'⌘⌥'`  (`'controloptionshiftcommand'`will also work) + backward compatible with the old tables
- Added some basic logging (should help a bit when users are trying to wrap their head around modal hotkeys for the first time)
- Added `hotkey.showHotkeys` to (very crudely for now) list all active hotkeys (a-la cheatsheet) to help memory with new and/or huge configurations

### DIFF
--- a/extensions/alert/internal.m
+++ b/extensions/alert/internal.m
@@ -178,7 +178,8 @@ static int alert_show(lua_State* L) {
     if (lua_isnumber(L, 2))
         duration = lua_tonumber(L, 2);
 
-    HSShowAlert(str, duration);
+    if (duration>0.0)
+        HSShowAlert(str, duration);
 
     return 0;
 }
--- a/extensions/hotkey/init.lua
+++ b/extensions/hotkey/init.lua
@@ -217,6 +217,43 @@ function hotkey.deleteAll(mods,key)
   hotkeys[idx]=nil
 end
 
+--- hs.hotkey.showHotkeys(mods, key)
+--- Function
+--- Creates (and enables) a hotkey that shows all currently enabled hotkeys while pressed
+---
+--- Parameters:
+---  * mods - (optional) A string containing (as substrings, with any separator) the keyboard modifiers required, which should be zero or more of the following:
+---   * "cmd", "command" or "⌘"
+---   * "ctrl", "control" or "⌃"
+---   * "alt", "option" or "⌥"
+---   * "shift" or "⇧"
+---  * key - A string containing the name of a keyboard key (as found in [hs.keycodes.map](hs.keycodes.html#map) ), or if the string begins with a `#` symbol, the remainder of the string will be treated as a raw keycode number
+---
+--- Returns:
+---  * The new `hs.hotkey` object
+
+local helpHotkey
+local function showHelp()
+  local t={}
+  for idx,hks in pairs(hotkeys) do
+    for i=#hks,1,-1 do
+      if hks[i].enabled and hks[i]~=helpHotkey then
+        t[#t+1] = hks[i]
+        break
+      end
+    end
+  end
+  table.sort(t,function(a,b)if#a.idx==#b.idx then return a.idx<b.idx else return #a.idx<#b.idx end end)
+  local s=''
+  for _,hk in ipairs(t) do s=s..hk.msg..'\n' end
+  --  hs.alert(s,math.min(15,math.max(#s/10),3))
+  hs.alert(string.sub(s,1,-2),3600)
+end
+function hotkey.showHotkeys(mods,key)
+  if helpHotkey then delete(helpHotkey) end
+  helpHotkey = hotkey.bind(mods,key,showHelp,hs.alert.closeAll,'Show enabled hotkeys',3600)
+  return helpHotkey
+end
 --- hs.hotkey.bind(mods, key, pressedfn, releasedfn, repeatfn, message, duration) -> hs.hotkey
 --- Constructor
 --- Creates a hotkey and enables it immediately
--- a/extensions/hotkey/init.lua
+++ b/extensions/hotkey/init.lua
@@ -216,9 +216,24 @@ function hotkey.deleteAll(mods,key)
   hotkeys[idx]=nil
 end
 
---- hs.hotkey.showHotkeys(mods, key)
+--- hs.hotkey.getHotkeys() -> table
 --- Function
---- Creates (and enables) a hotkey that shows all currently enabled hotkeys while pressed
+--- Returns a list of all currently active hotkeys
+---
+--- Parameters:
+---  * None
+---
+--- Returns:
+---  * A table containing the hotkeys that are active, i.e. enabled and not "shadowed", in the current context
+---    (usually, the global hotkey context, but it could be a modal hotkey context). Every element in the list
+---    is a table with two fields:
+---    * idx - a string describing the keyboard combination for the hotkey
+---    * msg - the hotkey message, if provided when the hotkey was created (prefixed with the keyboard combination)
+
+--- hs.hotkey.showHotkeys(mods, key) -> hs.hotkey
+--- Function
+--- Creates (and enables) a hotkey that shows all currently active hotkeys (i.e. enabled and not "shadowed"
+--- in the current context) while pressed
 ---
 --- Parameters:
 ---  * mods - (optional) A string containing (as substrings, with any separator) the keyboard modifiers required, which should be zero or more of the following:
@@ -232,7 +247,7 @@ end
 ---  * The new `hs.hotkey` object
 
 local helpHotkey
-local function showHelp()
+function hotkey.getHotkeys()
   local t={}
   for idx,hks in pairs(hotkeys) do
     for i=#hks,1,-1 do
@@ -243,6 +258,11 @@ local function showHelp()
     end
   end
   table.sort(t,function(a,b)if#a.idx==#b.idx then return a.idx<b.idx else return #a.idx<#b.idx end end)
+  return t
+end
+
+local function showHelp()
+  local t=hotkey.getHotkeys()
   local s=''
   for _,hk in ipairs(t) do s=s..hk.msg..'\n' end
   hs.alert(string.sub(s,1,-2),3600)
--- a/extensions/hotkey/init.lua
+++ b/extensions/hotkey/init.lua
@@ -5,17 +5,107 @@
 local hotkey = require "hs.hotkey.internal"
 local keycodes = require "hs.keycodes"
 local fnutils = require "hs.fnutils"
+local alert = require'hs.alert'
+local tonumber,pairs,ipairs,type,tremove,tinsert,tconcat = tonumber,pairs,ipairs,type,table.remove,table.insert,table.concat
+local supper,slower,sfind=string.upper,string.lower,string.find
 
+local function getKeycode(s)
+  if (s:sub(1, 1) == '#') then return tonumber(s:sub(2))
+  else return keycodes.map[s:lower()] end
+end
+
+local hotkeys,hkmap = {},{}
+
+--- hs.hotkey:enable() -> hs.hotkey
+--- Method
+--- Enables a hotkey object
+---
+--- Parameters:
+---  * None
+---
+--- Returns:
+---  * The `hs.hotkey` object
+
+local function enable(self,force)
+  if not force and self.enabled then return self end --this ensures "nested shadowing" behaviour
+  local idx = hkmap[self]
+  if not idx or not hotkey[idx] then error('Internal error!') end
+  local i = fnutils.indexOf(hotkey[idx],self)
+  if not i then error('Internal error!') end
+  tremove(hotkey[idx],i)
+  for _,hk in ipairs(hotkey[idx]) do hk._hk:disable() end
+  self.enabled = true
+  self._hk:enable() --objc
+  tinsert(hotkey[idx],self) -- bring to end
+  return self
+end
+
+--- hs.hotkey:disable() -> hs.hotkey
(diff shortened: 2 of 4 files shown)


lowne:
showhotkeys

asmagill:
In the past, there has been some support for the notion that doubly assigning a currently defined and enabled hotkey should return an error (or at least a warning) and leave the earliest enabled one still active, under the assumption that the new key is an error or mistake in programming and we don't want the "basic" hotkeys changing unexpectedly on the user... is the consensus changing? TBH, I can see both sides of this, so I don't really care as well as whatever the behavior, it is well documented and consistent.

As you note, this is somewhat problematic for nested modal key sets, but it is workable if you're careful (see https://github.com/asmagill/hammerspoon-config/blob/master/utils/_keys/fonts.lua for one way of handling it, though it really only works because the second modal can *only* be invoked from the first, so they "know" about each other)

Maybe a better approach is to allow a hotkey to be over-ridable with a flag -- by default, no, for hs.hotkey.bind, but by default yes for hotkey.modal:bind... or does that get too arbitrary?

And I'm not sure about the addition from #386 about adding an alert option to every hotkey... you can easily do that by creating your own wrapper to hs.hotey.bind/new or by sticking it into the action function... do we really need to add it to the basic function?  Even creating new methods in hs.hotkey such as bindWithAlert and newWithAlert seems cleaner (to me at least) then changing the existing basic function(s) which are lean and specific.  Just my opinion...

I'd also modify the showHotkeys function so that it returned the definitions as a table (one key per entry)... you don't know up front if I want to simply stick it into an alert -- `hs.alert.show(table.concat(resultTable,"\n"))` or maybe do something with hs.drawing.text to make it more presentable/persistent.

lowne:
@asmagill I very much appreciate the feedback, thanks!

Alert: it's true - then again *any* Lua-only addition can by definition be implemented by the user.  *Actual* functional additions require low level wizardry (see e.g. all the awesome work you've done recently that will let me build stuff on top of it) and I keep away from that as I don't trust myself around ObjC :smile: 
All my PRs are Lua-only and strictly speaking they're nothing more than *conveniences*, as in literally just saving the user the hassle to implement these things for herself in init.lua - the basic idea being that I'm not the only one who would find use for said functionality (if that's not the case, then these additions don't belong in HS but in my personal init.lua). As for the alerts specifically, hotkey.new is fully backwards compatible, and behaves exactly the same way as the old one if the optional message is omitted (but, in a "two birds with one stone" outcome, .showHotkeys and other potential future additions don't make much sense if all hotkeys have no messages - granting that not *all* hotkeys should necessarily benefit from having an associated message/alert, which is why it's optional in the first place).

Speaking of .showHotkeys, using hs.alert is admittedly (as per OP) a quick, ugly and ideally temporary hack; I agree that it might be useful to just return the table (though it should be noted that it's meant to be contextually dynamic - see below), but I'll also say that a more suitable presentation via hs.drawing, if possible, belongs in here HS as opposed to the enterprising user's init.lua, since anyone using .showHotkeys would benefit from it. Such an improvement is not my top priority for now, but it should definitely be in the nice-to-have todo.

About the shadowing business, I was under the impression that printing errors was simply the "easier fix" to the double-enable issue - although I might very well be wrong. At any rate, I **strongly** disagree that that should be the default behaviour. I concur that there might be cases of genuine user error (i.e. typo) resulting in unexpected double-enables, and that's why I added logging (to help the user self-diagnose such things - speaking of, if hs.logger ever becomes a "thing", I'd propose to add a `hs.setLogLevel` global function to act on all loaded modules); however I'd venture to say that in almost all cases such double-enable errors happen because the user was trying *in the first place* to have the kind of functionality that this PR enables (e.g. hotkey-X does Y, but if window W is focused it does Z) without the added complexity of hs.hotkey.modals - which IMO are meant, and are a much better fit, for other use cases such as the new hs.grid.show().

asmagill:
Yeah, I kind of blurred the alert and label thing, didn't I?  I was working on a hotkey replacement using a different method of capturing the keys (it's on hold right now because I can't get past it dropping keys under load... not good when you're writing code or a document!) and also added a way to add labels to a key, but through a slightly different mechanism -- through a method which could be "chained" to the hotkey binding... by that, I mean you would type something like  `a = hs.hotkey.bind({mods},"key",fn1):desc("the description")` or, later, `a:desc("the description")`... If you didn't assign a description, then it defaulted to "hotkey for &lt;textual-representation of mods-key&gt;"

I personally like this approach because it doesn't change any of the core functions as they currently are defined, and you don't have to add logic that essentially tries to figure out what parameter 3-6 are based upon parameter type passed in.

I guess, yes, I like giving hotkey labels, yes, there should be some method of displaying them in a help screen or writing them to a text file for printing out (there's an idea... hs.print... have to think about that...) , but I'm not sold on the syntax and would like to hear what others have to say before committing to an approach.

lowne:
Yes, I had misinterpreted; I've now seen your other gh issue about hotkey (I'm *really* bad at keeping track of all the issues, present and past) and if anything your support of labels predates mine :)
As for syntax:
- `hs.hotkey.bind(mods,key,fn1,[fn2,fn3]):desc("description")` vs 
- `hs.hotkey.bind(mods,key,fn1,[fn2,fn3,]"description")`

I favour the latter mainly because it clearly conveys (being constructor-based) that the label is "set in stone", as opposed to something that can be dynamically altered; the former kind of hints "you can have if-then-else logic inside your callbacks, and change the label accordingly" whereas I'd want to encourage the shadowing approach "if you need two behaviours, you create two hotkeys; the if-then-else merely enables/disables them".

But you have a point that trying to shoehorn function overloading in Lua results in parameter "parsing" logic that isn't particularly elegant, to say the least. (Then again, as long as it works...)

lowne:
Hotkey shadowing example taken from my init.lua:
```lua
--- system dialogs/annoyances that can't be directly dealt with from the keyboard
local mdwf=hs.windowfilter.new({'UserNotificationCenter','SecurityAgent'},'md-wf',5)
bind(HYPER,'m',function()local w=mdwf:getWindows()[1] if w then w:focus()end end,'Focus modal dialog')
local ncwf=hs.windowfilter.new(false,'nc-wf',5):setAppFilter('Notification Center',nil,nil,{'AXNotificationCenterAlert','AXNotificationCenterBanner'})
local ncwindow
local nchk=hs.hotkey.new(HYPER,'m',function()
  local f=ncwindow:frame()
  hs.eventtap.leftClick{x=f.x+f.w-5,y=f.y+5}
end,'Close NC alert')
ncwf:notify(function(wins)
  ncwindow=wins[1]
  nchk[#wins>0 and 'enable' or 'disable'](nchk)
end)
```
So when I press hyper-m:
- if there's a NC alert, dismiss it
- otherwise if there's a password prompt etc, focus it
- otherwise do nothing

cmsj:
@lowne that's a very interesting example. I think I'm convinced that multiple hotkeys could be *defined*, but I strongly feel that we should make it impossible to *enable* multiple identical hotkeys at a time. I'm pretty much ambivalent as to whether calling :enable() on a second identical hotkey should either fail or disable the first. I think it might make sense to have them fail, but add a "force" argument, that will disable the previously enabled hotkey, but I don't imagine any answer here is going to fit all scenarios.

lowne:
@cmsj I quote myself from earlier, then elaborate further:
> About the shadowing business, I was under the impression that printing errors was simply the "easier fix" to the double-enable issue - although I might very well be wrong. At any rate, I strongly disagree that that should be the default behaviour. I concur that there might be cases of genuine user error (i.e. typo) resulting in unexpected double-enables, and that's why I added logging (to help the user self-diagnose such things - speaking of, if hs.logger ever becomes a "thing", I'd propose to add a hs.setLogLevel global function to act on all loaded modules); however I'd venture to say that in almost all cases such double-enable errors happen because the user was trying in the first place to have the kind of functionality that this PR enables (e.g. hotkey-X does Y, but if window W is focused it does Z) without the added complexity of hs.hotkey.modals - which IMO are meant, and are a much better fit, for other use cases such as the new hs.grid.show().

So, here are afaict the the situations where a double-enable might arise:

1. the user sets a global hotkey in their `init.lua`, and (say, as worst case scenario) months later, adds another one for the same key combo. In order of decreasing likelihood:
  - it's something like my example above; the user consciously *wants* shadowing: **1-0**
  - they forgot about the previous hotkey and haven't used it for a long while; if it gets shadowed, they won't even notice nor care; if there's an error or the new hotkey is silently discarded, they will have to fix their config: **2-0**
  - they were actively using the previous hotkey, and the new one was just a typo/oversight; either way (error/shadowing/silent discard) the user will notice the problem soon enough (either the old one or the new one won't work, but they want both), even without checking the console logs, and fix the config: still 2-0
2. a module adds a global hotkey when loaded, on top of a user-defined one (or vice versa); modules should *never* do this imo, regardless of the current discussion. "Opinionated" modules (as per discussion somewhere else) that ship with "default" keybindings *must* be explicitly allowed to do so by the user; and even *then*, the unwanted outcome to double hotkeys will be pretty much random either way, depending on what's called first, so there's no winner: still 2-0
3. the user has set up global hotkeys (say, the arrow keys, or ctrl-e) for their own use, and a module (say, hs.grid.show, for moving the window across screens, or selecting a cell after #426 is fixed) tries to enable those same hotkeys *when entering a modal*
  - with shadowing, everything works as intended: **3-0**
  - if there's an error as soon as the user calls hs.grid.show, or if they can't move the window to their next screen/select that cell despite the onscreen indications, they will be confused for a loooong time, until they understand that they must disable their own hotkeys (!) if they want to use hs.grid.show: **10-0** :smile_cat: 

Maybe "shadowing" isn't the best term, but this is the (user-side) idea that it should convey: "I want this key combination to cause this behaviour **now**; and *of course* I realise that any previous behaviour will cease, while the new one is in force".
 
In short, we should **at least** have a `force` argument; but I see no situation where going even further with plain always-on shadowing would be *worse* than having to specify the force parameter, so I think it'd be redundant at best. If I overlooked something (which is likely) please let me know!

lowne:
As a possibly interesting use case for this, I set it up on my living room Mac mini; I use a Harmony remote that can be configured to send certain keypresses (arrows, return, esc, f1..f12 and a few more); then these keys are bound to a global/"default" action, and some are dynamically :enabled/:disabled (shadowing the default action) according to whichever app is focused (Plex, Spotify, iTunes, YouTube,...).
The best part is that the "Guide" button will show a context-sensitive list of the *current* bindings (via .getHotkeys) - which is helpful even for myself since I keep forgetting the buttons for the more esoteric/rarely used functions.

cmsj:
So, I've looked at this PR a few times and I mostly like it, but I'm really not sure about changing the mods thing from table to string. I'm fairly sure that most users have pre-defined tables with the mods they use, and are just passing that around.

lowne:
@cmsj you mean in the docstring? Passing tables still works.

cmsj:
@lowne partly the docstring, but also that it exists at all. I'm just not sure I see the need for it.

lowne:
@cmsj well, code-wise the difference is just the docstring; the hotkey indices need to merge equivalent (possibly differently ordered) lists into the same set and become strings anyway (for alerts, showHotkeys, etc); the code for only allowing tables is the same (well, instead of linearly scanning a string, it'd linearly iterate on a table, but same difference). 

API-wise, `'ctrl alt'` is simply easier, faster, and more natural to type than `{'ctrl','alt'}` for that minority of users without premade tables in their configs, which anecdotally includes, FWIW, myself - sure I have HYPER, but I'm defining hotkeys for testing all the time, and 1. I didn't bother to preset all possible mod combos in my `init.lua` since they're not used in "production" and 2. even if I did, it wouldn't do me any good (because globals are evil) when I want to `.bind` on the fly from the console (and I do this a lot).

But setting aside my peculiar (and irrelevant) needs, as you said (and I agree) most users have likely defined premade tables (probably after having typed a large enough number of tables to grow tired of it); having strings allows **new** users to not worry about that at all, since `CTRLALT` is arguably not an improvement over `'ctrl alt'` anyway.

cmsj:
@lowne If the docstrings can be updated to say that either people can either use a string containing any of the supported options, or a table containing any of the supported options, then I'll land this :)

lowne:
@cmsj done

cmsj:
Thanks :)

PR #413, opened 2015-07-23
Imageplaying

asmagill:
hs.window.snapshot can now get images of any window with an ID, even if it is minimized or hidden.
hs.window.snapshot takes an optional parameter indicating whether or not to create image opaquely (default) or with window's transparency in effect (pass boolean value of true).

I think there is a possible memory leak in hs.image because any image which is given a name (:setImageName method) can still be called up via hs.image.imageFromName(oldname) after reloading your hammerspoon config -- this is supposed to invoke all garbage collection calls, and I have verified that they do in fact execute, but apparently I'm missing something in __gc... any thoughts or help would be appreciated!  (actually, I wouldn't mind keeping images around through reloads, as long as it was toggle-able, so it's maybe sortof a feature!)

documentation fixes for hs.window.snapshot

hs.dockicon.setBadge now accepts nil to clear the badge.


cmsj:
@asmagill According to the docs for NSImage, naming an image adds it to the AppKit image repository, which likely means it's copying it and we're not technically leaking. However, I do think it's a good idea to clean them up in __gc, so it looks like we should `setName` the image to `nil` (https://developer.apple.com/library/mac/documentation/Cocoa/Reference/ApplicationKit/Classes/NSImage_Class/index.html#//apple_ref/occ/instm/NSImage/setName:)

asmagill:
Doh!  I knew those docs were good for something...

lowne:
@asmagill it can actually capture hidden windows "without" id, as long as it has some way of knowing the id anyway; the problem is just that `_AXUIElementGetWindow(win, &windowID)` fails for hidden windows.
However `window:id()` seems to work fine for hidden windows; it does `lua_getfield(L, -1, "id")`, but I have no idea how to juggle the lua stack in order fix :snapshot myself...

asmagill:
@lowne -- to make sure I understand... (1) window id's never change and (2) if we can cache this id from whenever we can sneak a peak at it, that should be sufficient for the snapshot?

If so... well, part of the problem will be that the userdata value (address) itself isn't unique, so anything stored in it is only stored in the single instance:

~~~
> hs.window.focusedWindow(), hs.window.focusedWindow()
-- Loading extension: window
userdata: 0x60000024af48	userdata: 0x60000025d878
~~~

So unless we're always using the exact same captured userdata we'll need some way of either addressing this or some way to tie them together separate from the userdata address... I'll mull this over a bit -- I've been considering ways to compare userdata objects directly to each other and this pushes the need to a higher priority.

lowne:
@asmagill I'm not entirely clear on the internal dynamics of hs.window so the following paragraph is a bit of guesswork, but see window.h: the id is saved in new_window via lua_setfield, and retrieved in `window:id()` via lua_getfield; the userdata object is the first parameter in both `:id()` and `:snapshot()` (a.k.a `self`) and the different userdata references (as instanced anew with each call to e.g. `.focusedWindow()`) 'wrap' the same AXUIElementRef.

So:

- `hs.fnutils.find(hs.window.allWindows(),function(w)return w:isMinimized()end):id()` might break (nonetheless, it (usually?) doesn't anyway)
- `w=hs.window.focusedWindow() w:minimize() w:id()` will *certainly* return the id as I'm retaining the userdata (w) created when the window surely has an id

Bottom line: 1) window ids never change and 2) they're already "cached" in the userdata as opposed to queried anew on every call to `:id()`. I think that just retrieving the id via lua_getfield - just like :id() does - will let :snapshot() capture hidden windows most (if not all of) the time.

asmagill:
When an obj-c window object is returned from a c function (*not* retrieved from a passed in userdata from lua), a brand-new userdata object is created and the id is added to it's structure if `_AXUIElementGetWindow(win, &winid)` is defined *at that instant in time*.  When a window is hidden or minimized, the window id isn't always available to retrieve (seems to be application specific), so cached userdata values will have it, but the one being dealt with *right now* might not.

The only real solution is in the C-Code, check to see if a userdata object already exists for a specific window, before creating a new one (and also every time a C-function uses a userdata, check for an id and pid if they're not currently defined)... Not a trivial task, but certainly a doable one.  If no one claims the task, I'll take a crack at it next week.

lowne:
Yes, that's precisely what I meant in the first paragraph of my previous comment; I forgot to clarify at the end
>  just retrieving the id via lua_getfield - just like :id() does - will let :snapshot() capture hidden windows most (if not all of) the time **as long as the use case is the second one** (in the bulleted list)

... which should be a simple fix and should cover the majority of real word usages (or at least it covers mine, in hs.expose - I keep around the userdata objects); but yes, it won't be a 100% solution, since as you said some apps won't have ids for hidden windows for the first (in the bulleted list) use.

Then again, if it's for :snapshot() alone, I'd say your proposed 100% solution is too much work for too little (pratical) benefit; but it seems to be worth it overall, because it just makes sense to avoid allocating new userdata for the same windows over and over again.

(But either way requires patching :snapshot() to get the "cached" id field, if it's there - if not, same `err`)

asmagill:
If window:snapshot could take an actual integer ID as a parameter and use that (if provided) instead of trying to find out itself, would it help? That I can probably get out later tonight.

The bigger issue touches on a larger one...

It's not really an issue, per se, but more of a "best practices approach" in that we're wasting memory in most (if not all) modules... not much, a few bytes here, a few bytes there... usually just the size of a pointer, but sometimes an actual structure of multiple data pieces, by having multiple different userdata objects refer to the identical "real" items.

I've been thinking about this for a bit, and if a generalized solution can be made to apply easily to all modules, it'll be worth the time.  I'm also envisioning adding more metamethods, so we can do things like window1 < window2 to determine which one is in front of the other without having to find their position in allWindows or the like... so, since doing this for window is part of something larger I'd like to see/do, I don't see the effort as too much... but it's also not my highest priority!

lowne:
I've already cobbled the integer param together (literally copy-pasting snippets of code, otherwise I'd absolutely forget all the semicolons) in my local repo for testing with hidden windows (and it works);  but let me be clear, there's no hurry for either one of the "better" solutions - worst case hidden windows will have no thumbnail (well, the actual *worst* case is I'll finally read up on part IV of Programming In Lua and might eventually be tempted to push it's-so-horrible-you-cannot-unsee-it C code onto you guys :laughing:)

About the larger issue, I very much agree (better avoid useless allocations, it's the Right Thing) - but afaict there's also a general philosophy at play here of *not* trusting the OS (e.g. stuff might be going away unannounced, and all your "caches" are ~~belong to us~~ now invalid and you don't even know it) which seems very reasonable; so care must be taken into identifying and affecting only *true* (as in, they cannot *ever* not be) "duplicates". I'm super aggressive with caching on the Lua side because at worst it's a stacktrace, but subtle heisenbugs on the foundation are the silent killer...

PR #420, opened 2015-07-25
add pasteboardTypes and contentTypes

asmagill:
Added a way to get information about the sender and intent of the clipboard contents via pasteboardTypes and contentTypes.

In reference to https://groups.google.com/d/msgid/hammerspoon/9F87511F-8858-4034-9336-75778DDF4F82%40tenshu.net?utm_medium=email&utm_source=footer

See also http://nspasteboard.org.

Sample of the results from 1Password placing a password into the clipboard:

~~~
> inspect(hs.pasteboard.pasteboardTypes())
{ "public.utf8-plain-text", "NSStringPboardType", "com.agilebits.onepassword", "org.nspasteboard.ConcealedType" }

> inspect(hs.pasteboard.contentTypes())
{ "public.utf8-plain-text", "com.agilebits.onepassword", "org.nspasteboard.ConcealedType" }
~~~

Note the "org.nspasteboard.ConcealedType", which indicates that we should not store or save this data.

Currently contentTypes only returns the information about the 1st item on the pasteboard, but since that's the only one we can manipulate at present, that's not a problem.

### DIFF
--- a/extensions/pasteboard/internal.m
+++ b/extensions/pasteboard/internal.m
@@ -61,6 +61,49 @@ static int pasteboard_clearContents(lua_State* L) {
     return 0;
 }
 
+/// hs.pasteboard.pasteboardTypes([name]) -> boolean
+/// Function
+/// Return the pasteboard type identifier strings for the specified pasteboard.
+///
+/// Parameters:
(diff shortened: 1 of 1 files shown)


cmsj:
Awesome, lgtm :)

Issue #418, opened 2015-07-24
Allow customization of hs.grid.show()


jfchevrette:
It would be nice to be able to customize the grid display that is shown when using hs.grid.show()

- grid lines/cell background colors
- cell transparency
- font size/zolor

CC @lowne

Issue #417, opened 2015-07-24
Enable animations for hs.grid.show() operations


cmsj:
I think it would be nice if the movements coming out of hs.grid.show() weren't hard-coded to an animation duration of 0.

We could leave it up to the user's default animationDuration, or accept an optional parameter in grid.show().

@lowne - thoughts? :)

PR #422, opened 2015-07-25
Grid fixes

### LINKED ISSUE #416: hs.grid.set()'s screen parameter should be optional
(so people don't need to pass it if they just want to move the window on its current screen)

Any objections @lowne ?

lowne:
fix #416, #417, #418


jfchevrette:
You guys are the best :-D

PR #423, opened 2015-07-25
Update internal.m

asmagill:
Fix documentation result type mistakes


cmsj:
For stuff like this, I'd say just push, no need to PR :)

PR #424, opened 2015-07-26
hide grid when switching spaces

lowne:
In a just world, there'd be a way of "dragging" the window along to the new space. Maybe a hack could be put in place with creative use of something like :setBehavior from the revamped hs.drawing module, but I doubt it'd be possible to do that on foreign windows anyway.


cmsj:
Agreed that it's better to play this one safely

PR #427, opened 2015-07-28
add hs.host.operatingSystemVersion and hs.host.operatingSystemVersionString

asmagill:
Getting the exact version as major, minor, and patch turned out not to be as simple as assumed... 10.10 and later provide a selector for this, but prior versions can only be approximated... the method I chose was to compare the NSAppKitVersionNumber value against the definitions in NSApplication.h.

I've only been able to check this against 10.10.4 so far, but will be able to test under 10.8 later this week, unless someone else wants to try before then.

### DIFF
--- a/extensions/host/internal.m
+++ b/extensions/host/internal.m
@@ -314,6 +314,24 @@ static int hs_operatingSystemVersion(lua_State *L) {
         lua_pushinteger(L, OSV.patchVersion) ; lua_setfield(L, -2, "patch") ;
         lua_pushboolean(L, YES)              ; lua_setfield(L, -2, "exact") ;
     } else {
+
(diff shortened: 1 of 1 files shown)


cmsj:
I think that things after NSAppKitVersionNumber10_8 need some #ifdef guards

asmagill:
good point... I just copy and pasted most of that from various places, then cleaned it up a bit and made it lua friendly, but since we "allow" 10.8 as the unsupported lowest,  and people can always try to compile it with whatever they have, we should put some basic checks in there...

I'll adjust it tomorrow (well, later today, as I was just about to try and get some sleep...) and update the pull request, then probably apply it myself, unless you see a need for more/additional changes.

cmsj:
+1

PR #430, opened 2015-07-28
Utf8

asmagill:
add hs.utf8.registeredLabels as opposite member to hs.utf8.registeredKeys -- both can be accessed as function or array (though each is implemented in actuality as whatever was easiest) to make code which uses both easier (less picky)

### DIFF
--- a/extensions/utf8/init.lua
+++ b/extensions/utf8/init.lua
@@ -39,8 +39,35 @@ module.registeredKeys = setmetatable({}, { __tostring = function(object)
                 output = output..string.format("(U+%04X) %-15s  %s\n", utf8.codepoint(v), i, v)
             end
             return output
-    end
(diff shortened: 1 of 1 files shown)


asmagill:
@cmsj, considering a comment you left in #386, would this function help at all?  I'm of two minds about it because I don't really see a common use case for it, though if we can find a memory friendly/fast way to include a list of all of the "official" names for every UTF8 character, it might be useful to extend registeredKeys and registeredLabels to reference that as well... then bi-directional support makes more sense to me, but I'm not against this if you or someone else think it's useful as it stands right now...

cmsj:
@asmagill so this would help with things that want to support the UTF8 versions of keyboard modifiers, if nothing else, so I think it's ok to go in.

PR #431, opened 2015-07-29
[hs.grid] Increase grid size limit to 100x100

russellhancox:
Increases the size limit of a grid to 100x100. I've been using this grid size for a while without any trouble and it's useful being able to size a window to a percentage of overall screen size.

### DIFF
--- a/extensions/grid/init.lua
+++ b/extensions/grid/init.lua
@@ -90,7 +90,7 @@ function grid.setGrid(gr,scr)
     if not scr then error('Invalid screen or geometry',2) return end
   else scr=1 end
   if type(scr)~='number' then scr=toKey(scr) end
-  gr.w=min(gr.w,50) gr.h=min(gr.h,50) -- cap grid to 50x50, just in case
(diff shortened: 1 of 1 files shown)


lowne:
:+1: afair there's no reason this should break anything

asmagill:
I haven't looked into the hs.grid code closely enough to come up with a hard and fast answer for it yet, but I have noticed that if the cell spacing (i.e. difference in screen points between two adjacent rows or columns) is close to that of the margin size, then at least some windows (the Hammerspoon conole being one of them) won't resize by one increment at a time (e.g, the grid.resizeWindowX functions)

If it could be reduced to a simple formula or comparison, then putting an "in effect" max on the margins in calculations should solve it.

lowne:
That could be (from memory) due to the fact that `hs.grid.get` on margin-applied windows will return the "wrong" (shrunken) cell, since (naturally) it doesn't take margins into account. At any rate, if one has a 100x100 (or anyway large) grid, they won't be doing unit-increments on windows manually; but if they *really* want to for some reason, they can live with setting a reasonably smaller margin themselves imo.

PR #433, opened 2015-07-29
Add support for scroll wheel

bbarkley:
**Full Disclosure** - This is my first time working with OS X or the Lua C APIs and was hacked together through a combination of reference docs and Google results. Please review with care, and I welcome suggestions for improvement.

This adds support for creating and executing scroll wheel events. I figured it belonged in the event module next to the `newMouseEvent` support for clicking rather than the mouse module which seems to be not event based and just for movement.

### DIFF
--- a/extensions/eventtap/event.m
+++ b/extensions/eventtap/event.m
@@ -406,6 +406,60 @@ static int eventtap_event_newKeyEvent(lua_State* L) {
     return 1;
 }
 
+/// hs.eventtap.event.newScrollWheelEvent(offsets, mods, unit) -> event
+/// Constructor
+/// Creates a scroll wheel event
+///
+/// Parameters:
+///  * offsets - A table containing the {horizontal, vertical} amount to scroll. Positive values scroll up or left, negative values scroll down or right.
+///  * mods - A table containing zero or more of the following:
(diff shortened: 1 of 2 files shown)


extensions/eventtap/event.m
```diff
@@ -406,6 +406,60 @@ static int eventtap_event_newKeyEvent(lua_State* L) {
     return 1;
 }
 
+/// hs.eventtap.event.newScrollWheelEvent(offsets, mods, unit) -> event
+/// Constructor
+/// Creates a scroll wheel event
+///
+/// Parameters:
+///  * offsets - A table containing the {vertical, horizontal} amount to scroll
```
cmsj: Even though vertical scrolling is much more common, I think it'd be better to follow the convention of having the offsets be {horizontal,vertical}.
I also think it would be useful to indicate which directions the positive/negative variants of each offset, scroll in.

cmsj:
Thanks, this is a great idea. I added two in-line comments, but apart from those, this is good to land.

bbarkley:
Updated docs and reversed the axes as suggested. Thanks for the feedback.

cmsj:
Thanks!

PR #435, opened 2015-07-30
add support for creating pop-up menus

asmagill:
hs.menubar changes:

* added `hs.menubar:popupMenu(point)` to allow drawing the menu at any point on the screen as a popup menu.
* modified all methods to return menubaritem to facilitate chaning
* added `hs.menubar.removeFromMenuBar` and `hs.menubar.returnToMenuBar` so that a menu intended to be used as a popup menu doesn't take space on the system menubar.
* added `menu` key to setMenu docs describing submenus.
* hs.menubar.new now takes an optional argument to cause a menubaritem to start out off of the menubar.  The default behavior, if the argument is left out, is still to put it into the statusbar area of the menubar.

The only API change this introduces is in regards to setIcon -- previously it returned true on success and nil on failure... now it returns the menubaritem on success and nil on failure.  Since lua treats anything that is not nil or false as `true`, this should only affect code actually doing a result check like `if result == true then...`, rather than `if result then...`

Like hs.image, this module will also now display a customized label for its userdata in the console: e.g.`hs.menubar: (0x600000877fe8)` or `hs.menubar:Apps (0x600000877fe8)`, if it has a title.

Is this something worth eventually applying to all userdata objects in all modules?

### DIFF
--- a/extensions/menubar/internal.m
+++ b/extensions/menubar/internal.m
@@ -72,7 +72,8 @@ void erase_menu_items(lua_State *L, NSMenu *menu);
 typedef struct _menubaritem_t {
     void *menuBarItemObject;
     void *click_callback;
-    int click_fn;
+    int  click_fn;
+    BOOL removed ;
 } menubaritem_t;
 
 // Define an array to track delegates for dynamic menu objects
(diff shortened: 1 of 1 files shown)


cmsj:
Nice! It hadn't even occurred to me that we could do this, but it seems like it could be useful in all sorts of ways.

As for adding __tostring for other userdata objects, +5555555

PR #436, opened 2015-07-30
[READY] add hs.expose (needs #513)

lowne:
Requires windowfilter in #374


cmsj:
Deferring this until 0.9.37, so hs.windowfilter can get a few trips around the block first.

lowne:
Todo:
- [ ] wait for windowfilter -> window.filter renaming

PR #438, opened 2015-07-30
Remove unused udhandlers, they are mostly cargo culted and unused.

cmsj:
Only hs.hotkey and hs.notify actually use them.

I'm running this branch for now, to see if anything weird pops up, but I find it very hard to believe that anything will, all of this seems to just be unused cruft.

@asmagill any thoughts on this? Fancy testing it out? :)

### DIFF
--- a/extensions/application/watcher.m
+++ b/extensions/application/watcher.m
@@ -43,30 +43,14 @@
 
 #define USERDATA_TAG "hs.application.watcher"
 
-static int store_udhandler(lua_State* L, NSMutableIndexSet* theHandler, int idx) {
-    lua_pushvalue(L, idx);
-    int x = luaL_ref(L, LUA_REGISTRYINDEX);
-    [theHandler addIndex: x];
-    return x;
-}
-
-static int remove_udhandler(lua_State* L, NSMutableIndexSet* theHandler, int x) {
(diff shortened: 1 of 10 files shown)


asmagill:
If it compiles without them, then it's probably safe -- the cases where it's needed would have stood out to you because they would have involved more than just the udhandler add and udhandler remove code.  I'll take a close look, though, later and see if I notice anything missing, though.

WRT to hs.window (and maybe a couple of others, have to see), I may add something like it back because with hs.window, we have multiple ways to get at a window (via a watcher, from hs.application, from hs.appfinder, etc.) and not everything (mainly the ID, but other data as well?) is always available when a given request for the object comes through -- I want to arrange hs.window so that it always returns the same userdata, if one has already been generated for the given window, so that we can capture things like the ID when they are determinable and present them later, even if they aren't atm.

But that's a specific use case, not a general copy-and-paste-but-not-understand type of mistake.

cmsj:
So, I think the only way to get this wider testing is to land it. @asmagill @lowne @Habbie if y'all are running local builds, please keep an eye out for strange changes in behaviour - things that disappear/stop working, that might be caused by things being garbage collected early, or suchlike.

PR #443, opened 2015-07-31
popupMenu can handle clickCallback functions

asmagill:
the position sent into hs.menubar:popupMenu() is effectively ignored as we can't affect the functions visual effects, if any, but it does work.


asmagill:
hold on this... want to add something else to the pull.

PR #447, opened 2015-07-31
hs.image: add _eq for testing if 2 userdata objects refer to the same NSImage

asmagill:


cmsj:
@asmagill how come the #import changes from <> to ""?

asmagill:
So it would build outside of the core directory with only a symlink to the proper location.   Menubar was set up this way as well, so I assumed it wasn't a problem.

PR #451, opened 2015-07-31
add javascript for automation extension based on hs.applescript

heptal:
Ported hs.applescript extension to work with JavaScript using `osacompile`, which has a drawback of needing temporary files since OSAKit is completely undocumented/immature.

Still, it's much easier to write `Application("iTunes").playpause()` instead of `tell application "iTunes" to playpause`, etc.


asmagill:
+1

Looks useful, but for temporary files, it'll probably be better (certainly more "Apple approved") to use NSTemporaryDirectory instead of mktemp (a glance at http://nshipster.com/nstemporarydirectory/ suggests it might be clear enough for your purposes, but I'l admit I haven't actually had to do this for anything yet.)

asmagill:
Is there enough interest in this that I should try and apply the proposed changes?  I like that our pulls are in the single digits and I'd like to stay there!

lowne:
+1 for me - personally it's just theoretical, in that *anything* is better than applescript but (lack of) docs and Script Editor amply succeeded (when 10.10 came out) in cooling my enthusiasm and convincing me to keep staying away from OSA; but I suppose others would like to have the option.

cmsj:
I have reflexes against doing system() type calls, so I wonder if we can use OSAKit even though it's completely undocumented.

http://boredzo.org/blog/archives/2008-11-09/framework-friday-osakit suggests its API is actually almost identical to NSAppleScript. I've not managed to find any code using it yet on github, but it seems like it'd be worth experimenting with - we might be able to make hs.applescript just magically work with Javascript too.

cmsj:
It kinda looks like we'd just need to do something like:

```
    OSAScript *osa = [[OSAScript alloc] initWithSource:script language:[OSALanguage languageForName:@"JavaScript"]];
```

and either use `JavaScript` or `AppleScript` and then the `OSAScript` object can use the same `executeAndReturnError` that we use on `NSAppleScript`.

cmsj:
https://github.com/curtclifton/OmniFocusScripts/blob/8e5f0f00184b9858318f98292eb0290a94064d88/OFKanban/OFKanban/OFKanban/OSAScript%2BCCKAppleScript.m uses OSAKit for AppleScript, https://github.com/ccgus/CocoaScript/blob/master/src/editor/JSTDocument.m uses it for JavaScript.

It's not very commonly used, it would seem, but it is public API.

bradparks:
For what it's worth, this seems like a great idea to me ;-)

heptal:
I have updated the code to use OSAKit instead of the dirty stuff from before. The code is a lot smaller and cleaner and uses LuaSkin as hs.applescript now does. Some minimal details on OSAKit are found [here](https://developer.apple.com/library/mac/releasenotes/General/APIDiffsMacOSX10_11/Objective-C/OSAKit.html). 

I did find it a lot harder to get a new extension working with the latest code. I have not included my Hammerspoon.xcodeproj/project.pbxproj file because I'm not sure this is the correct way, but what I did was duplicate the `applescript` target, edit the name and source reference, then, in Build Phases, add `javascript` to 'Target Dependencies' and to `HS_MODULES` within 'Copy extensions to bundle'.

asmagill:
That process is basically what I do when I add new modules, so you're instincts were good.  After #620 lands (hopefully tomorrow, if no one has any suggestions/objections), I'll add this and make the necessary project changes then.

Issue #450, opened 2015-07-31
hs.drawing and mouse clicks


asmagill:
@cmsj: I've just started looking at this, so I apologize in advance if any of these will become obvious as I dig into it, but a couple of questions about the hs.drawing.setClickCallback...

* any particular reason it acts on mouseUp and not mouseDown or both?
* off the top of your head, can it made so that Hammerspoon doesn't become the active application (or at least so that it doesn't bring the console forward, if it's open) when accepting a click?
* what about right clicks?

Like I said, just beginning my looking into it, but I thought I'd see if there were specific reasons or roadblocks you encountered that led to the current implementation.

asmagill:
Ok, as to my first question, it's easy enough to support mouse down, but I'm guessing you chose mouseUp because most items will be thought of us "button like" in which the action doesn't occur until the mouse is released, and not "menu-like"

I'd like to add support for both, but because of the nature of hs.drawing's objects, our standard approach of list functions in "down, up, repeat" order doesn't make as much sense -- the current behavior *is* the most likely expected behavior and I only noticed it because I'm working on providing popup menus...

So as a point of syntax, my first two thoughts are
* modify setClickCallback to take up to two (three?  will have to see if we get repeated downs) functions, but keep mouseUp as the first function and make the second mouseDown.  Either could be nil of course.
* add a boolean to setClickCallback which switches from the mouseUp behavior to the mouseDown behavior... loses repeat functionality, if its even possible.

Right clicks should also be doable, but because of the default behavior of providing a contextual menu in NSView for right menus, it will require a few extra steps, but I think I know what to do.

Still not sure about the issue with bringing the console forward, though.

cmsj:
I'm also not sure about the bringing of the console forward. I don't see an NSWindow thing that lets it avoid getting focus, or an NSApplication thing. I landed your PR, it looks like a useful addition - I also like that the mouse button detection is bounced up for Lua to care about. Repeating seems unnecessary to me, so let's see if anyone ever asks for that :)

asmagill:
FWIW, I've added an application watcher to close the HS console window when I leave for another application... it mitigates the issue for now, but I'll keep my eyes open.

If anyone is interested, a variant of this is what I'm using: https://gist.github.com/asmagill/251f8ea70b61a177a205

PR #452, opened 2015-08-01
hs.drawing.setClickCallback supports up and down mouse events

asmagill:
add mouseDown function the hs.drawing.setClickCallback and capture any button, not just the left mouse button.

Still looking for a way to not bring the console, if open, forward, but for now, I'll just keep it closed except when using it :-)

To capture a held mouse down similar to hotkey's repeatFn will require a similar timer to be added.  Worth it?  I'm not going to want it any time soon, but that's just me...

### DIFF
--- a/extensions/drawing/internal.m
+++ b/extensions/drawing/internal.m
@@ -26,6 +26,7 @@ NSMutableArray *drawingWindows;
     lua_State *L;
 }
 @property int mouseUpCallbackRef;
+@property int mouseDownCallbackRef;
 @property BOOL HSFill;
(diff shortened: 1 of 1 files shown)


asmagill:
partially addresses #450

asmagill:
FWIW, I've added an application watcher to close the HS console window when I leave for another application... it mitigates the issue for now, but I'll keep my eyes open.

If anyone is interested, a variant of this is what I'm using: https://gist.github.com/asmagill/251f8ea70b61a177a205

PR #453, opened 2015-08-01
Documentation improvements

oskarols:
Hey, so I really like what's happened with the documentation pages. One thing I'm missing from the old ones however, was that they were significantly more condensed and easier to survey on my laptop screen. So I've dialed back the whitespace a bit, and removed some things that just clutter up the docs.

- Significantly decreased the amount of unnecessary whitespace on
  header elements and the body to make the documentation easier
  to browse on laptops
- Fixed bug where each table cell on the documentation startpage was
  getting extra bottom padding
- Removed superflous border on body
- Removed superflous border on each function section; instead using
  whitespace and proximity to guide the eye.
- Decreased all but the top header (h1) sizes on the module pages
  since h1s and h2s are so similar in size.

### DIFF
--- a/scripts/docs/templates/docs.css
+++ b/scripts/docs/templates/docs.css
@@ -7,13 +7,12 @@
 body {
   color: #333;
   font-family: system, -apple-system, ".SFNSDisplay-Regular", HelveticaNeue, LucidaGrande, sans-serif;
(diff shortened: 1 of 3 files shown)


lowne:
+1 for more compact docs

PR #455, opened 2015-08-02
fix iterations in fnutils

lowne:
Some fnutils functions are used in contexts where in-order processing should be maintained (e.g. window list operations/filtering on orderedWindows()); now, while it might very well be the case that `next` - and thus `pairs` - will always iterate in order for tables built just-so in c, the docs say that next iterates in arbitrary order; there might be a chance that long-lived tables might undergo transformations that result in out-of-order iteration.
It seems to me that the original implementation is betting (or banking) on `next` maintaining order while still allowing generality for use with non-pure-array tables; but the resulting tables are array-only anyway and all the hash keys are lost, which *might* be the desired behaviour (but probably isn't) for all "internal" uses but seems a bit excessive (if not just wrong) for a general purpose library.

Therefore I propose to:

1. add proper array-only functions (`imap`, `ifilter`, etc.) that use `ipairs` to avoid surprises (such as the fact that legitimate holes are silently "filled in" in the original implementation)
2. change `map` etc. to be truly general purpose wrt hash or mixed tables, while maintaining the old behaviour for array-only tables
3. over time, change all `fnutils.FN` in HS modules to `fnutils.iFN` where it makes sense (which is, I bet, everywhere)
4. optionally (if deemed useful), add `.mapKeys()` or `.kmap()` etc. functions that act on keys instead of values. 
5. EDIT: eventually (or even immediately, depending on considerations about whether it was the intended use in the first place) make `map` etc. "pure" by ditching the backward-compatbility "fill the holes" aspect, i.e. always assigning onto the same key (so any holes in the source table remain in the result table) - note that `filter` needs the "fill the holes" aspect until all calls are replaced with `ifilter`, and then making it pure will still break user configurations using it; a possible (and sensible) solution is to engineer `filter` so that *existing* holes in the array part are preserved, but new ones created by `filter` itself are filled in for the current "index block".

I worked on the first 3 functions (EDIT: not on point 5 thought, pending agreement; the as-much-as-possible backward compatibility workarounds are in place) - see the diff. If my proposal is accepted, I'll proceed with the rest of the module.

Backwards compatibility is (necessarily) broken for hash and mixed tables. E.g.:
```lua
t={'a','b','c',key='value'}
hs.inspect(hs.fnutils.map(t,function(x)return x end))
-- old: { "a", "b", "c", "value" }
-- new: { "a", "b", "c", key = "value"}

hs.inspect(hs.fnutils.imap(t,function(x)return x end))
-- new: { "a", "b", "c"}
```
but the old behaviour doesn't make sense imo, and as I said above, I doubt there's any code in HS that uses fnutils for intentionally mixed tables (EDIT2: a new `flatten` function could take care of these cases, if any), and where it's not intentional, the current implementation results in bugs.

Additionally, there is (what I believe is) a bug in `map` (`table.insert(nt, fn(v) or nil)`) that discards `false` values (which I, for example, use quite extensively); I have fixed it, but that's also a potentially breaking change, if there's somewhere a module "relying" on this bug when using a predicate `map` `filter`-like, instead of properly returning `nil` - then again, I'd regard such code as needing fixing itself, regardless of this PR.

### DIFF
--- a/extensions/fnutils/init.lua
+++ b/extensions/fnutils/init.lua
@@ -4,55 +4,130 @@
 
 local fnutils = {}
 
+local pairs,ipairs = pairs,ipairs
+local floor = math.floor
 
---- hs.fnutils.map(table, fn) -> table
+--- hs.fnutils.imap(list, fn) -> list
 --- Function
---- Execute a function across a table and collect the results
+--- Execute a function across a list-like table in order, and collect the results
 ---
 --- Parameters:
----  * table - A table containing some sort of data
----  * fn - A function that accepts a single parameter. Whatever this function returns, will be collected and returned
+---  * list - A list-like table, i.e. one whose keys are sequential integers starting from 1
+---  * fn - A function that accepts a single parameter (a table element). The values returned from this function
+---    will be collected into the result list; when `nil` is returned the relevant element is discarded - the
+---    result list won't have any "holes".
 ---
 --- Returns:
----  * A table containing the results of calling the function on every element in the table
-function fnutils.map(t, fn)
+---  * A list-like table containing the results of calling the function on every element in the table
+---
+--- Notes:
+---  * If `list` has "holes", all elements after the first hole will be lost, as the table is iterated over with `ipairs`;
+---    use `hs.fnutils.map()` if your table has holes
+function fnutils.imap(t, fn)
   local nt = {}
-  for k, v in pairs(t) do
-    table.insert(nt, fn(v) or nil)
+  for _, v in ipairs(t) do
+    nt[#nt+1] = fn(v) -- or nil < removed, as this precludes inserting false!
   end
   return nt
 end
 
---- hs.fnutils.each(table, fn)
+local function isListIndex(k)
+  return type(k)=='number' and k>=1 and floor(k)==k -- not using 5.3 syntax (k//1==k), as you never know
+end
+--- hs.fnutils.map(table, fn) -> table
 --- Function
---- Execute a function across a table and discard the results
+--- Execute a function across a table (in arbitrary order) and collect the results
 ---
 --- Parameters:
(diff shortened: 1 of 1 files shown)


extensions/fnutils/init.lua
```diff
@@ -4,55 +4,130 @@
 
 local fnutils = {}
 
+local pairs,ipairs = pairs,ipairs
```
cmsj: I'm curious if this actually gets us anything? Isn't it just saving a quick lookup in local scope, before going to global scope?

cmsj:
I tend not to use fnutils very deeply, to I'm not sure if I should care or not about switching the current behaviours around, but since we're planning a pretty long dev cycle before 0.9.39, to bed down the huge build system changes that have just landed, I'm prepared to let this go in and we'll see if anything breaks.

Re 4. I'm fine with adding more functions to fnutils. I don't see myself using them much, but I imagine they would be useful to some.

Re 5. Let's see what happens with the changes as they are so far. I'm not sure that any of us around at the moment, can vouch for the original intentions of fnutils (maybe @asmagill can?), so I'd like to be a little cautious here.

lowne:
re "localing" globals, yes, 'more' local lookups are a bit faster (although I doubt the difference is noticeable on powerful x86 CPUs); performance-wise, the main thing is that global lookups are actually *table* lookups instead of stack, and the difference there is more marked. With LuaJIT it probably wouldn't be worth it, and it may not be worth it anyway since we're not a performance-constrained platform/environment. But there's another *very* good reason to avoid globals: a good IDE will highlight them in the code because it's easy for typos (that cause global lookups that go to nil) to create hard-to-trace bugs; so avoiding globals helps in finding these typos just by glancing at the code (the only global I usually allow is `error`).

PR #456, opened 2015-08-02
improve window.focusWindow.. and window.windowsTo.. methods

lowne:
Added `candidateWindows`, `frontmost`, `strict` parameters; removed `sameApp` parameter (but backward compatibility is maintained) because I suspect it's not a general use one, i.e. it's only likely to be used for apps like terminals, editors and browsers; for those cases, better to encourage specialised windowfilters, now that it's a thing.

I also want to add module-level variables for `frontmost` and `strict`.

### DIFF
--- a/extensions/window/init.lua
+++ b/extensions/window/init.lua
@@ -12,8 +12,10 @@ local fnutils = require "hs.fnutils"
 local geometry = require "hs.geometry"
 local hs_screen = require "hs.screen"
 local timer = require "hs.timer"
-local pairs,next,min,max,type = pairs,next,math.min,math.max,type
-
+local pairs,ipairs,next,min,max,type = pairs,ipairs,next,math.min,math.max,type
+local tinsert,tsort = table.insert,table.sort
+local intersection,hypot,rectMidPoint,rotateCCW = geometry.intersectionRect,geometry.hypot,geometry.rectMidPoint,geometry.rotateCCW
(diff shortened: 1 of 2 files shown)


lowne:
I'll gladly take pointers on how to diagnose errors for `make docs`; from the output I have no idea what's going on...

asmagill:
You're replicating method/function names at around 1386 in the windowfilter init.lua of the format `hs.window:focusWindow...` not looking too closely, I can't say for certain, but do you mean `hs.windowfilter:focusWindow...` here?  Otherwise, they're colliding with the definitions in the hs.window init.lua.

lowne:
Thanks!

lowne:
Pasting from #462:
> I added (without publicising it) hs.window.frontmostWindow() in #456 - the plan is to substitute focusedWindow with frontmostWindow everywhere there's an optional window argument that defaults to the focused window, or things like hs.grid.show that just go for focusedWindow without any parameter; it's meant for those (imho) unintuitive cases where e.g. you hit cmd-w on your last terminal window, the previously obscured underlying Safari window pops up and you want to act on it, but it's not the focused window until you cmd-tab to it.

asmagill:
re #462, I'll try and take a look at it today... Qupzilla was giving me grief in other ways as well, so it may or may not become a mainstay of my work flow, but as apps like it are out there, knowing how/why they work/fail is useful in any case.

PR #458, opened 2015-08-03
Updates to allow core to edit application Preferences from lua (console)

asmagill:
Added:

    hs.openPreferences() -- opens Hammerspoon preferences panel
    hs.openAbout() -- open Hammerspoon About panel
    hs.autoLaunch([boolean]) -- set/retrieve autolaunch status
    hs.menuIcon([boolean]) -- set/retrieve display of menu icon
    hs.dockIcon([boolean]) -- set/retrieve display of dock icon (actually just a wrapper to hs.dockicon module)
    hs.consoleOnTop([boolean]) -- set/retrieve whether the console should always be "on top" of other windows

Since I usually run a development version, I can't really test functions to set/check/do an update, so I have left those out for now.  Let me know if there is interest and I can try, but I'll need others to test it for me.

Also updated:

    hs.openConsole can take an optional boolean value which, if provided and set to false, will cause the console to be displayed, but not brought to the front (i.e. just revealed)
    an observer is also setup so that when defaults change via any method, the settings on the preferences pane (if it exists) reflect the current, not cached, values

I added these because I wanted a way to get rid of all of the default Hammerspoon visual elements but yet not lose an easy way to change any of them, should I need/want to.

### DIFF
--- a/Hammerspoon/MJLua.m
+++ b/Hammerspoon/MJLua.m
@@ -8,6 +8,7 @@
 #import "../extensions/hammerspoon.h"
 #import "MJMenuIcon.h"
 #import "MJPreferencesWindowController.h"
+#import "MJConsoleWindowController.h"
 #import "MJAutoLaunch.h"
 
 static LuaSkin* MJLuaState;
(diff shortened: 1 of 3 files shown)


asmagill:
some of these would allow a user to bypass the warning if both the menu and dock icons were hidden, but we can now also pull up the preferences panel via hs or the console, so...

go ahead, or modify/remove some of these?

cmsj:
+1

Re checking for updates, that would likely need to be done with weak references, since a local build won't even link Sparkle.framework.

PR #459, opened 2015-08-03
hs.host additions

asmagill:
The first three are often used in combination to create a temporary file or directory in an "approved Apple way".  The fourth is to serve as a flag if you want to offer different drawing elements (or at least different colors) depending upon the interface style.

hs.host.temporaryDirectory() -- returns the users system recommended "temporary" directory for the current user.

hs.host.globallyUniqueString() -- returns a unique identifier based on host, time, etc. guaranteed to be unique on the local network

hs.host.uuid() -- returns a new UUID as a string

hs.host.interfaceStyle() -- returns the value of the NSGlobalDomain's AppleInterfaceStyle key, which will contain `Dark` if the OS X interface style is set to Dark, or nil if it is not.

### DIFF
--- a/extensions/fs/internal.m
+++ b/extensions/fs/internal.m
@@ -323,7 +323,7 @@ static int lfs_lock_dir(lua_State *L) {
 }
 static int lfs_unlock_dir(lua_State *L) {
   lfs_Lock *lock = (lfs_Lock *)luaL_checkudata(L, 1, LOCK_METATABLE);
-  if(lock->fd != INVALID_HANDLE_VALUE) {    
+  if(lock->fd != INVALID_HANDLE_VALUE) {
     CloseHandle(lock->fd);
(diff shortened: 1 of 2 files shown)


cmsj:
Should temporaryDirectory() go in hs.fs?

asmagill:
Since hs.fs is a port of luafilesystem, I am thinking probably not... even though it's 3rd party and not core, it's not "ours" and I think should probably stay as close to its roots as possible... at least that's my first thought.

However, I'm not wedded to it, and I'll go with the consensus.

cmsj:
@asmagill at this point hs.fs is really a superset of luafilesystem, e.g. we added the Finder tag support already. I think adding a 4 line function for temporaryDirectory() is perfectly fine, and it feels like the most relevant location, IMO.

PR #464, opened 2015-08-04
add screen.primaryScreen(), .screenPositions(), :position()

lowne:


asmagill:
How does screen.primaryScreen differ from screen.mainScreen?

lowne:
screen.mainScreen doesn’t do what one would expect (in its defense, it calls [NSScreen mainScreen] indeed): 
/// Returns the 'main' screen, i.e. the one containing the currently focused window.

asmagill:
... the doc for primaryScreen says "where the menubar is"... is this a 10.11 change, or a difference when "Displays have separate spaces" is unchecked? (it's checked on my machine)

I can see that they use different processes... I'm trying to figure out (without the second monitor right now, I admit) when one would return different results than the other...  and maybe suggest that the docs for both be updated to make it clear why two seemingly similar functions are really different.

lowne:
To simplify, mainScreen does `return hs.window.focusedWindow():screen()` (sort of), so it depends on what window currently has focus.

The primary screen is the one with the thin white bar at the top in System Preferences/Displays/Arrangement, hence the docstring. tbh idk if the menubar can jump around screens in 10.11 and/or with ‘Displays have separate spaces’ checked.

cmsj:
In 10.11, at least with the separate spaces option checked, *all* screens get a menubar, but it is slightly greyed out on screens that don't have focus. It doesn't really change the meaning of the primary screen though, there is still one which has the white strip in Displays->Arrangement, and that is the one with 0,0 as its top left pixel.

cmsj:
Actually 10.10 does the same thing with the menubars, AFAIR

PR #465, opened 2015-08-04
add new shorthands for delayed and repeated functions

asmagill:
* `hs.timer.doEvery(...)`
   shortcut for hs.timer.new(...):start()

* `hs.timer.doUntil(predicateFn, actionFn [, interval])`
   repeat actionFn until predicateFn returns true

* `hs.timer.doWhile(predicateFn, actionFn [, interval])`
   repeat actionFn while predicateFn returns true

* `hs.timer.waitUntil(predicateFn, actionFn [, interval])`
   delay actionFn until predicateFn return true

* `hs.timer.waitWhile(predicateFn, actionFn [, interval])`
   delay actionFn while predicateFn returns true

* `hs.timer:running()`
   returns true if timer is running (active) or false if stopped.

also added descriptive __tostring to hs.timer object

### DIFF
--- a/extensions/timer/init.lua
+++ b/extensions/timer/init.lua
@@ -56,6 +56,117 @@ function module.days(n)    return 60 * 60 * 24 * n end
 ---  * The number of seconds in n weeks
 function module.weeks(n)   return 60 * 60 * 24 * 7 * n end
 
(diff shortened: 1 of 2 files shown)


cmsj:
Very nice!

PR #466, opened 2015-08-04
Mouse

asmagill:
Deprecate (with warnings)  hs.mouse.get and hs.mouse.set, and add wrapper to checkMouseButtons in hs.eventtap as I keep looking for it here!

Also... anyone know what the local array check_list is for in the init.lua file?  It appears as far back as my git logs go but is never used.


cmsj:
check_list is cargo culting from hs.battery. I'll remove it.

PR #471, opened 2015-08-06
[READY] hs.geometry overhaul

lowne:
> 17:47 lowne: > g'0 0 100 100'..'80 80 50x50'
17:47 lowne: hs.geometry.rect(0,0,130,130)
17:47 lowne: > g'0 0 100 100'^'80 80 50x50'
17:47 lowne: hs.geometry.rect(80,80,20,20)
17:48 lowne: > g'5 5'<'0 0 10 10'
17:48 lowne: true
17:49 lowne: fun!

But:
- there might be bugs, I didn't test everything yet
- line segments (let alone orientated ones) aren't supported; would that be useful?
- no fancy stuff for line intersections etc., see item above
- according to the module docstring promise, stuff like :setFrame(f) in windows and drawings should wrap the parameter in `hs.geometry(arg)` - will do if this is deemed mergeable


cmsj:
I'm not immediately sure what I'm looking at. What's the aim here?

lowne:
1. to make life easier with frames - e.g. all the arg parsing code in `hs.grid.setGrid()` is replaced by a single `hs.geometry(...)` call
2. this (from hs.window):
```lua
    local frame = win:frame()
    local p2 = rectMidPoint(frame)
    p2 = rotateCCW(p2, p1, numrotations)
    local delta = {x=p2.x-p1.x,y=p2.y-p1.y}
    if delta.x > (strict and abs(delta.y) or 0) then
      local angle = atan(delta.y, delta.x)
      local distance = (delta.x^2+delta.y^2)^0.5
      local score = (distance/cos(angle/2))+z
      tinsert(wins,{win=win,score=score,z=z,frame=frame})
    end
```
becomes
```lua
local p2 = geometry(win:frame()).center:rotateCCW(p1,numrotations)
local delta=p2-p1
if delta.x > (strict and abs(delta.y) or 0) then
  local score = (#delta/cos(delta:angle()/2))+z
  tinsert(wins,{win=win,score=score,z=z,frame=frame})
end
```

extensions/geometry/init.lua
```diff
@@ -1,44 +1,494 @@
...
+      w,h=t.w or t.x2-x,t.h or t.y2-y
+    end
+  end
+  local t={_x=x,_y=y,_w=w,_h=h}
+  for k,v in pairs(t) do if type(v)~='number' then t[k]=nil end end
+  norm(t)
+  if not gettype(t) then error('Cannot create geometry object, wrong arguments',2) end
+  return setmetatable(t,geometry)
+end
+--- hs.geometry.new(...) -> hs.geometry
```
cmsj: I'm not a fan of having "hs.geometry" as the return type. We're not very consistent about this yet, but I think the word "object" needs to be in there - maybe "geometry object"?

lowne:
@cmsj You're right, it can be ambiguous as sometimes the return value is the module itself rather than a module object. You favour "geometry object" over "hs.geometry object"? (And similarly for hs.window etc.)

asmagill:
I haven't looked at all of the various combinations your parser handles, but one that would be handy for me is the ability to pass in a point ({x=#,y=#}, note: no h or w) and a size ({h=#,w=#}, note: no x or y)  and get back a full geometry object of all 4 for passing in to things like hs.drawing, etc.  Does this or could this be done with the parser?

lowne:
@asmagill not yet, or not directly with tables. You can build a point and later set its .w and .h fields separately (or vice versa, build a size etc.). I will add a .size field as a shortcut. Similarly there is .center, but I suppose you'd want .topleft, however I'm not 100% sure about that as I was trying to remain coordinate system agnostic...

cmsj:
@lowne I think I prefer "hs.geometry object"

lowne:
@asmagill I think I had misunderstood your previous comment - on rereading, I think you want that immediately on the constructor, right? With the latest commit you can do `myrect=geometry({x=0,y=0},{w=100,h=100})` (or, fwiw, the string equivalent `myrect=geometry('0,0','100x100')`)

asmagill:
That will likely work for the cases I have in mind -- specifically with hs.drawing, its constructors take a full rect as the initial parameter, but often (at least in my code) I'm generating the size and position independently... I've been getting around this by putting {} into the initial constructor and following it with `:setSize(table1):setTopLeft(table2)` but using a geometry object in the initial constructor may be "cleaner" in some cases.

I'll take a look later and let you know!

PR #473, opened 2015-08-06
add lvl, terminate parameters to hs.showError()

lowne:


Hammerspoon/setup.lua
```diff
@@ -26,12 +26,14 @@ hs.configdir = configdir
...
 hs.docstrings_json_file = docstringspath
 
---- hs.showError(err)
+--- hs.showError(err,lvl,terminate)
 --- Function
 --- Shows an error to the user, using Hammerspoon's Console
 ---
 --- Parameters:
 ---  * err - A string containing an error message
+---  * lvl - A number containing the level in the call stack for debug.traceback
```
cmsj: lvl and terminate should say that they're optional, and what their defaults are :)

Hammerspoon/setup.lua
```diff
@@ -26,12 +26,14 @@ hs.configdir = configdir
...
+---  * terminate - boolean, if true terminates the last protected call
```
cmsj: I think "the last protected call" could do with a bit more explanation, or rewording to be clear about what it means.

asmagill:
I understand tweaking and better explaining showError... but should we really be modifying error as well?

From the docs:

>`error (message [, level])`
>
>Terminates the last protected function called and returns `message` as the error object. Function `error` never returns.

>Usually, `error` adds some information about the error position at the beginning of the message, if the message is a string. The `level` argument specifies how to get the error position. With level 1 (the default), the error position is where the `error` function was called. Level 2 points the error to where the function that called `error` was called; and so on. Passing a level 0 avoids the addition of error position information to the message.

Among other things, `error()` never returns, and hs.showError does (we get around this in most of the C code by using it as the message handler to lua_xpcall).

Try `for i = 1, 5, +1 do hs.showError("boo!") end` versus `for i = 1, 5, +1 do error("boo!") end` in the console.

asmagill:
Ok, closer look and I see that you're addressing that, but still... I may wrap my own function calls with pcall, in which case I'm expecting a regular one-line error message, not a full traceback, which this would change.

lowne:
@asmagill very good point, fixed now; it's on me to make a local alias with the desired behaviour (which is, have a traceback *and* terminate) rather than force others to use hs.rawerror; I guess I was subconsciously attached to having `error` be the only global in a module (therefore easy to spot in a code-highlighting IDE).
@cmsj did my best to fix the docs, but I don't really know how to explain "last protected call" other than pasting the whole Lua docstring for pcall/xpcall...

cmsj:
@lowne thanks, I think the docstrings can live, this isn't something that most people are likely to use in their init.lua anyway :)

Issue #477, opened 2015-08-09
C code and luaL_error versus showError


asmagill:
#473 got me thinking...

Let me just start by saying up front, I know I am guilty of this... maybe I'm the only one, but if not, I put this forward as a suggested best practice.  Unless the general consensus is strongly otherwise, I plan to change my habits and adjust it as I notice it in other code...

When C code encounters an error -- say the wrong variable type is passed, or a required argument is missing, etc., then I believe we should be terminating the function/method with `return luaL_error(L, ...)`  This terminates the function/method and allows for a formatted error string with a variable number of arguments in the error.

When a lua callback is invoked from something that C code created (timer, eventtap, etc.), it should return errors via the showError function as defined in Hammerspoon.h.  But _only_ callbacks should handle errors this way.

Any other method/function is invoked within the context of other lua or c code, and it is up to that code to determine if it should wrap its calls in `pcall` or `xpcall` (if lua) or lua_pcall (if c) to catch and handle errors, or let the error terminate execution of whatever is going on.

As I said, I know I'm guilty of sticking showError in in places it doesn't really belong, but I plan to stop.

Thoughts?  Objections?

cmsj:
I am sure I am guilty of the same and worse - sometimes in a rush I've just made C code call NSLog and return!

This seems to me like a prime candidate for objc methods in LuaSkin. We could have something like functionError and callbackError that take nice NSString formats&values, and functionErrorWithCString and callbackErrorWithCString that take C string formats&values.

Internally they can do all the things we want - maybe they log to Crashlytics, etc.

cmsj:
Oh, and I'd note that LuaSkin already has a method that does a pcall with debug.traceback as its error handler, which we could extend to have error handling options.

cmsj:
tl;dr, LuaSkin exists to provide convenient wrappers to our common lua API patterns :)

cmsj:
This might be complicated by the fact that sometimes our methods make specific promises about their ability to return things - i.e. we sometimes still return self even in an error (I think, I haven't checked every return point).

I suppose that's just another thing for LuaSkin to think about for its API in this area - does it want to do something that terminates the current execution, or not.

asmagill:
There is certainly a case to be made for certain methods having a default fallback if their arguments make no sense or are absent... I agree that it's case by case, but anything that could be "fixed" by an attentive program will need to capture the error and probably doesn't want the visual display that showError causes...

[...]

PR #480, opened 2015-08-09
Convert almost all uses of lua_pcall to LuaSkin's protectedCallAndTraceback

cmsj:
Reasonable reduction in code, dramatic reduction in complexity/fragility (IMO). Relates to #477 in some ways.

I'm testing this branch now, and I'll land it in the next day or so unless I hear any objections or find any issues (or discover there is no usable Internet at the place I'm travelling to tomorrow :)

### DIFF
--- a/Hammerspoon.xcodeproj/project.pbxproj
+++ b/Hammerspoon.xcodeproj/project.pbxproj
@@ -60,6 +60,7 @@
 		4F4CB4571B73B15F000EA9B6 /* AppKit.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 9445CA1119083251002568BB /* AppKit.framework */; };
 		4F4CB4581B73B1B1000EA9B6 /* Fabric.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 4F0F97561ADE9F8600F2A199 /* Fabric.framework */; };
 		4F6345EA1B7789E30053B024 /* Fabric.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 4F0F97561ADE9F8600F2A199 /* Fabric.framework */; };
+		4F6345EB1B7795AA0053B024 /* Fabric.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 4F0F97561ADE9F8600F2A199 /* Fabric.framework */; };
 		4F64D8BF1B724C4B00FC4C65 /* internal.m in Sources */ = {isa = PBXBuildFile; fileRef = 4F64D8BE1B724C4B00FC4C65 /* internal.m */; };
 		4F6B3EA21B73D6D300B2A4DE /* Cocoa.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 9445CA0E19083251002568BB /* Cocoa.framework */; };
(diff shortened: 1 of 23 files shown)


asmagill:
Lots o changes! I'll merge the pull locally as well and let you know if I find anything.

cmsj:
@asmagill pushed another commit that also removes the lua_State pointers from the various userdata structs.

Note that I skipped hs.keycodes because I'm not touching that module any more than I have to, until it's had an ARC conversion (our last outstanding one). I also skipped hs.notify because the relevant section of its code, terrifies me ;)

asmagill:
I'll take a look at the two... yeah, I've progressed a bit since notify and it probably could stand an overhaul.

asmagill:
WRT to keycodes, looking at its history in the log file, I'm not entirely certain I see the hesitation with bringing it into line with ARC... As I understand it we basically want to use __bridge_retained when storing a retainable type in the userdata (the observer in this case) and then __bridge_transfer during __gc (and maybe stop, case depending) when removing it from the userdata... all other times use __bridge.  That's simplified a bit, but that's the gist of what I understand we need to be doing...

Did this cause crashes before?  Or compile errors?  I'll give it a shot later tonight, but if there was specific behavior to look for, it would help in testing.

asmagill:
Ok, I don't know what to do within Xcode to ensure this compiles with ARC enabled, but if you want to take a look at my initial take on keycodes, it's at: https://github.com/asmagill/hammerspoon/tree/480plus

I don't normally change layouts, but I added a few, added a watcher function, switched between them, and reloaded/restarted a few times all making sure that it was loading from my locally installed copy (sometimes the CLS barfing of every require to the log comes in handy  😜)... what else should be checked?  (Oh, and take a look at the changes I made to use your protect skin method and skin lua_state... correct? -- it worked, but it's always good to have a second opinion.)

cmsj:
@asmagil lgtm. To enable ARC in Xcode, go to the Hammerspoon project at the very top of the project explorer (cmd 1), in the right pane scroll to the keycodes target, in its build phases tab open the Compile Sources section and remove -fno-objc-arc from the compiler options column. Simple ;)

(Fortunately we don't use that per-file compiler options thing for any other files)

cmsj:
@asmagill Thanks for doing keycodes, I cherry-picked it onto the branch.

How do you feel about landing this? I think I'm ready :)

(as for notify, I hope I didn't come across as rude, I very much appreciate the work you did on it, but I think we do need to rewrite it. There is something very wrong with the lifecycle of the various objects, and I had to do terribly hacky things to keep it from crashing all the time)

asmagill:
Agreed, and I understand --  it was my first real attempt at doing something in Objective-C for OS X that wasn't just porting and tweaking or minor additions, and it shows.

I agree it needs an overhaul, and I even have some thoughts on moving some of the logic back into lua and *maybe* even making it handle responding to notifications whose clickbacks trigger the relaunching of Hammerspoon... but that is a little further off, as it will also require (hopefully minor) changes to the core app and I tread lightly there.

Like you, I'm a little hesitant to touch it and almost want to redo it from scratch.

It's on my radar, though, and I plan to tackle it this weekend.

As for the rest of pull 480's changes, go for it. I've been using it for a couple of days now and haven't seen any problems that I didn't create myself!

PR #487, opened 2015-08-12
[WIP] Switch module registration to LuaSkin's helper functions

cmsj:
Again, I have covered almost everything except the terrible trifecta that is hs.application/hs.window/hs.uielement, and hs.notify.

In most cases, this dramatically simplifies the luaopen_hs_foo_internal() function. In some cases it makes it slightly longer, but I'm prepared to pay that price in the name of consistency.

It also meant cleaning up some of the luaL_Reg arrays, so they actually make sense. I didn't rename most of them though, which I may come back to later, again for consistency.

### DIFF
--- a/Hammerspoon/MJLua.m
+++ b/Hammerspoon/MJLua.m
@@ -246,6 +246,20 @@ static int core_focus(lua_State* L) {
     return 0;
 }
 
+/// hs.getObjectMetatable(name) -> table or nil
+/// Function
+/// Fetches the Lua metatable for objects produced by an extension
+///
+/// Parameters:
+///  * name - A string containing the name of a module to fetch object metadata for (e.g. `"hs.screen"`)
+///
+/// Returns:
+///  * The extension's object metatable, or nil if an error occurred
+static int core_getObjectMetatable(lua_State *L) {
+    luaL_getmetatable(L, lua_tostring(L,1));
+    return 1;
+}
+
 static int core_exit(lua_State* L) {
     if (lua_toboolean(L, 2))
         lua_close(L);
@@ -287,6 +301,7 @@ static luaL_Reg corelib[] = {
     {"reload", core_reload},
     {"focus", core_focus},
     {"accessibilityState", core_accessibilityState},
+    {"getObjectMetatable", core_getObjectMetatable},
     {"_exit", core_exit},
     {"_logmessage", core_logmessage},
     {"_notify", core_notify},
--- a/extensions/alert/internal.m
+++ b/extensions/alert/internal.m
@@ -217,11 +217,9 @@ static const luaL_Reg metalib[] = {
     {NULL, NULL}
 };
 
-int luaopen_hs_alert_internal(lua_State* L) {
-    luaL_newlib(L, alertlib);
-
-    luaL_newlib(L, metalib);
-    lua_setmetatable(L, -2);
+int luaopen_hs_alert_internal(lua_State* L __unused) {
+    LuaSkin *skin = [LuaSkin shared];
+    [skin registerLibrary:alertlib metaFunctions:metalib];
 
     return 1;
 }
--- a/extensions/applescript/internal.m
+++ b/extensions/applescript/internal.m
@@ -2,8 +2,6 @@
 #import <LuaSkin/LuaSkin.h>
 #import "../hammerspoon.h"
 
-// Check out NSScriptClassDescription for expanding obj return values...
-
 /// hs.applescript._applescript(string) -> bool, result
 /// Function
 /// Runs AppleScript code
@@ -29,13 +27,10 @@ static int runapplescript(lua_State* L) {
     NSAppleEventDescriptor* result = [script executeAndReturnError:&error];
 
     lua_pushboolean(L, (result != nil));
-    if (result == nil)
-//         mjolnir_push_luavalue_for_nsobject(L, (NSArray *)error); // I don't think this ever worked, but it is what was in Hydra
+    if (result == nil) {
         lua_pushstring(L, [[NSString stringWithFormat:@"%@", error] UTF8String]);
-    else {
-//         lua_pushstring(L, [[result stringValue] UTF8String]); // worked only for string results...
+    } else {
         lua_pushstring(L, [[NSString stringWithFormat:@"%@", result] UTF8String]); // ugly, but parseable in Lua, sorta...
-//         mjolnir_push_luavalue_for_nsobject(L, arrayFromDescriptor(result)); // my pipe dream, but not yet...
     }
     return 2;
 }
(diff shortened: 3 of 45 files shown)


cmsj:
(I think longer term, I may want to make each module's objc class be a subclass of LuaSkin, which would make this all even simpler and more consistent - e.g. moving the luaL_Reg arrays into the object, with standard naming)

cmsj:
Marking this as WIP for now, because there are some breakages

cmsj:
Hmm, so this is breaking horribly where we need to access object methods from Lua, because I've been studiously moving them all out of the library functions and into their "proper" object metatables.

@asmagill I'd be curious to get your thoughts here - should we keep the old hacky solution of having some objects keep some/all of their object methods in the main library method table, or maybe instead look at a generic mechanism for fetching a library's object metatable?

cmsj:
(I'm tempted to say the latter)

asmagill:
You mean something like this?

~~~lua
static int extras_getmetatable(lua_State *L) {
    luaL_getmetatable(L, lua_tostring(L,1));
    return 1 ;
}
~~~

I'm curious the impetus behind this change -- seems like a lot of work for admittedly some benefit, but not as much as it would seem to merit -- at lease imho.  I'm all for standardization, but what does this gain us?

Just curious.

asmagill:
(BTW the luaL_getmetatable function above is cleaner then a trick I do in a couple of the init.lua files by creating an empty new object, adding to the meta-table and then deleting the created object... I recommend it be added like the constant making function to wherever you think they belong so every init.lua file can access them)

cmsj:
@Asmagill the impetus is that I find the module registration code complex, inconsistent and ugly :)

asmagill:
Fair answer :-)

cmsj:
I'm pretty sure there is going to be more fallout from this. I don't think I'd ever actually registered previously that some extensions were putting object methods in their main libraries so they could add to them from Lua easily.

I added hs.getObjectMetatable() and converted all of the ones I've found so far, but things like the one in hs.hotkey.modal (an fnutils.each that was passing hotkey.enable as its fn) suggest to me that there will be other places that will need to be updated, if this is going to land.

I'm in two minds. I think this separation is the right thing to do, but I do have to admit that it makes writing extensions a little harder, by forcing more correctness on authors. I'm not sure yet if that's a bad thing or not.

Either way, I'm running this branch and will fix up everything I can find.

cmsj:
(One clear advantage is that it means we have a defined way of adding object methods from Lua - our answer to that wouldn't be "oh right, yeah you have to put them in the extension's main library, it's hacky, but it works" :)

cmsj:
@lowne I would be grateful if you could give this branch a spin too, see if any of your stuff is affected.

lowne:
@cmsj (first, I think you should cherry pick c49c730 into master right away)
After staring for a long while at registerLibraryWithObject and the ensuing adjustments in your commits I think I more or less understand what's going on, so here's what I think (with the caveat that I might have utterly misunderstood the code; if so all that follows is worthless); as always, I advocate for the Lua side of things :grinning: 
While I understand that from an ObjC perspective maybe a proper 'separation of concerns' is warranted, I disagree on that being the case for Lua; after all, so-called "methods" in Lua are mere syntactic sugar for functions, and that imo is a fundamental aspect worth preserving; in other words, as a Lua coder, the current way ("a module exports functions; some or all of these maybe take a 'special' first argument; we can call these methods, or not - they're still just pieces of functionality that reside in the module") seems perfectly natural, whereas it's the getmetatable rigamarole that feels hacky (if I saw this without any context for the first time, I'd say "oh, that's weird; I suppose it's necessary to accommodate things on the C side, which I know nothing about; oh well").
As an example, I wouldn't want to "hide" all the hs.func methods - I want the user to be able to use them on arbitrary (non-objectified) tables as well; that would still be the case, if at all possible (and it is), if hs.func was fully implemented in C.
In short: I'm all for making module registration more consistent via LuaSkin everywhere, but I don't agree with fencing off methods inside a metatable. If anything, wherever that might currently be the case, I'd fix that the other way around - have all methods on the module's table.

cmsj:
@Lowne I can't speak offhand for the pure Lua modules, but in the C ones, the vast majority of them are keeping their object methods in a metatable.

I absolutely don't want to force things like hs.func to do the same if it doesn't make sense, I'm really just tidying up what we have right now, with a few tweaks to what is where that could mostly be done without.

As to the wider question of whether all functions should be on the modules, and their objects should just inherit all of those, I'm somewhat torn. The OO purism angle says we should keep them separate. The pragmatism angle says we should keep them together.

What do folk like @habbie and @asmagill think here?

asmagill:
This got longer than I expected, so for those who look and go TL;DR, let me summarize here:  I favor the separation because 90% of the time a method has no use without it's object and separation makes it more obvious in what context a certain function has validity.  The new coder can't easily accidentally try out a function that expects a userdata as its first argument and get the enigmatic "Parameter #1 expected to be a userdata" error because they won't have access to it without first creating the userdata.  For the rare case where you do in fact know what you're doing and need to access it directly as a function, (a) this has always been possible but hacky, and (b) it's now easier with `hs.getMetaTable(...)`.


--- more explication for those who care to read it ---

 I lean toward the separation, personally in both worlds, but more strongly in the C one -- In the case of what we're choosing to call methods, that special first variable is (usually?) always a userdata -- and there is simply no other way to work with it except as an argument to a function... can't be added, looked at, compared, or nuttin (well, we can add metamethods and are starting to for many, but by default this is true).  The "function" has almost no purpose or use outside the context of an existing userdata created on the C side of things.

Another minor benefit of the separation that I like is that when I'm debugging or playing around in the console, if I have a userdata and I forget exactly what it's method names are, I can type 'hs.inspect(obj)` and get a list of methods which are applicable and not a list which contains the applicable PLUS everything else that is in the module including tables of constants, etc.  Of course the inline help system has progressed, so this isn't as useful as it used to be, but still... for the userdata's we haven't added a __tostring method to yet, this is also a way to figure out what it is if you're uncertain or have forgotten.

For the rare case that you might want to actually invoke the method as a function with an explicit "self" parameter (and I have done this twice, I think, in what I would consider "useful" as opposed to "playing around" code), the newly added hs.getMetaTable(...) function allows a clean way for a module to grab a reference to the functions without having to create a (possibly unnecessary) object first.

In the abstract, the separation of functions from methods makes it more clear what you're able to do with a variable which is treated as an object (this is usually a userdata, but I do have a couple of personal modules where I use a table containing a bunch of state information and treat the group as a single object)

Since the subject has been brought up, let me also add that I'm personally not in favor of defining lua functions in the format of "module:functionName(...)" rather than "module.functionName(self, ...)"... In my opinion, the later makes it more clear what is actually being passed into the function -- of course the case can be made that the former makes it more clear how you're expected to *use* the function, so I can see both sides of this... I just remember when I first started learning lua, I kept wondering where all these references to self came from and why it seemed to exist in some functions and not others...

lowne:
About how to type function definitions in Lua, I agree that it's useful to show the intended purpose, which is why I kept the `self` parameter explicit (even better, it's called `t` for table) in the hs.func methods; the Dash docset format doesn't really help to get the point across but I made the possibility explicit in the module description at the top.

I concede the general point on separation (not many useful things to do by calling a method function-like, if you don't have the userdata to begin with); the proposed ad-hoc workaround for module use will always feel weird, but I can live with it.

Now I'm about to go a bit off topic, but due to time constraints I couldn't add in my previous comment that in my even-more-impractical-than-usual version of an ideal world there would be no userdata to begin with, and no C-side metatables, no __gc headaches etc, as hs objects would be created Lua-side, and C libs would just be tables of 'dumb' functions (with minimal sanity checks) returning and getting passed back a single light userdata (or a list, for collections) pointing to the relevant ObjC-API object of interest (e.g. AXUIElementRef for hs.window). 

(Of course, coming back to the methods-vs-functions diatribe for a second, we could then still enforce proper separation, instead of putting module functions and object methods all together in one giant table, so that stuff like `hs.window.minimize(a_complex_expression_resulting_in_a_window)` wouldn't work (which I still think will be a minor loss, but I will manage without it)) 

The productivity benefits of moving up in the abstraction ladder (beside much less compilation, and radically simplified use of the Lua C API in the surviving C code) cannot be overstated imo; but I'm sure there's at least a dozen reasons why this isn't feasible. If I ever get a sudden urge to dip my toe into (what to me is the toxic hellstew of) ObjC, I might find out what these reasons are myself :) - and even if it turns out to be feasible, it'd still be a lot of work due to the sheer number of modules to convert.

asmagill:
I'll admit I've been dabbling a little bit at seeing what your "even-more-impractical-than-usual version of an ideal world" would take...  And the simple answer is a lot... but not, I think, as much as I once feared... It's still a long ways off, though.  I'll post more on that later in another thread when I'm ready to start laying the groundwork for it... the data conversion tools will be useful to module developers even without a true bridge.

cmsj:
Isn't that world really just lots of ffi?

asmagill:
I'm not familiar with ffi

Most of bridges I've looked at first of all don't support 5.3 and don't have plans to, and most require a mixing of C and lua code in the same file to be parsed out... the C code is either compiled at run time or requires the files be "precompiled" before using

I'm thinking more of something in which all of the code at runtime remains lua code, and actual NSObjects remain in C but have a "representation" in lua (usually as a table or an object with attached methods) so that lua developers don't have to know the underlying structure.

I may hit a wall where it just isn't worth it -- this is mostly just ideas at present, though I have made some progress with representing complex NSObjects as simpler ones in Lua (with some limitations at present) and with dynamically generating messages with selectors and parameters to be sent to these objects.

Long way to go, but I no longer lump it in the OMFG NO WAY category :-)  And I don't promise to open up the entire API either... but some of them might be nice and it would put more of the "logic" part of some modules back into Lua while the "implementation" part is hidden from the user.

cmsj:
FFI is a way of working directly with C data types and functions, from another language. It's not fun to work with!

Bridging dynamic selector stuff into the objc runtime might be nicer, but I think we strike a good balance already,much of the time

cmsj:
Since I've not come across any more issues with the conversion, and nobody has raised any, I'm going ot land this.

I also think we need to keep the more esoteric conversations about our general principles, on the mailing list :)

PR #488, opened 2015-08-13
[feedback?] hs.func

lowne:
So, continuing off from #455: I take the merging as a green light:
> If my proposal is accepted, I'll proceed with the rest of the module.

Which is what I did, but mid-way through working on the rest of the module, I realised there were too many weird things and assumptions in the code and therefore too many compromises if backward compatibility was to be assured (I can elaborate if anyone's interested).

So I propose to:
- revert #455 
- deprecate fnutils as a whole
- merge this :)

This module can be used just like fnutils (e.g. `t=hs.func.filter(t,fn) t=hs.func.map(t,fn2)`), or via more svelte method calls on "collection" objects: `hs.func(t):filter(fn):map(fn2)`; there's also a bunch of iterators, such as `for v in t:valuesByKeys() do ... end` (bringing in and completing the recent additions to fnutils); and sort-of-wrapper methods for `table.insert`, `.remove` and `.sort` for convenience.

### DIFF
--- a/extensions/func/init.lua
+++ b/extensions/func/init.lua
@@ -17,6 +17,7 @@
 local getmetatable,setmetatable=getmetatable,setmetatable
 local pairs,ipairs,next,type=pairs,ipairs,next,type
 local tpack,tunpack=table.pack,table.unpack
+local tostring=tostring
 
 local f={} -- module/class
 
@@ -36,12 +37,14 @@ local f={} -- module/class
 ---    that hs.func methods don't shadow your table object methods
 ---  * for the same reason, if you need to set the metatable to an already existing hs.func table, you can use
 ---    the `hs.func:setmetatable()` method, it will *insert* the new metatable at the top of the chain
-f._isFuncTable = true
+f._isFuncCollection = true
 local function new(t)
   if not t then t={} end
   local mt,nt=t,t
   while mt do
-    if mt._isFuncTable then return t end
+    if mt._isFuncCollection then return t end
+    -- not checking __index to see that it actually points to the metatable;
+    -- let's assume it does in almost every case; if it doesn't, issues will be filed
     nt,mt=mt,getmetatable(mt)
   end
   setmetatable(nt,{__index=f}) -- at the end of the metatable chain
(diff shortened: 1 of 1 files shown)


cmsj:
If we're deprecating fnutils, we'd need to convert the ~60 places in other extensions where fnutils is being used currently, to hs.func. Is that something you'd be interested in doing? :D

I'm not sure that I like the name "hs.func", it sounds a bit generic and non-descriptive (not that fnutils is great either). Maybe hs.tableutils? hs.functionalTables would be nice, but it's a bit long.

Also, how reasonable would it be to map the old fnutils API onto this? There are going to be a lot of configs out there which are using fnutils, and dropping it makes me nervous about annoying users.

lowne:
@cmsj sure, I'll eventually convert ALL the things (same goes for geometry); about the deprecation: fnutils should stick around as-is (even reverting #455) for a while *after* we put deprecation warnings in the console; after that period has elapsed, I'll replace fnutils with wrappers for func - there's a 1:1 mapping, but the functions do not behave *exactly* the same, there are all the corner cases where fnutils did unexpected things (which is why I gave up trying to keep backward compatibility in the first place); however for the *vast* majority of uses, users won't notice any difference. But the implementations still differ in significant ways when it comes to "mixed" tables, which is why I think we need to (eventually) "force" users off of fnutils. 
About the name: agreed `hs.func` isn't great; I wanted a short one to help when using the constructor (`hs.func(mytable)`) here and there in user configs without having to set it on a local, or without having to type too much.

lowne:
Todo:
[ ] decision on module name (need feedback)
[ ] unlike what I've written above, add aliasing fnutils->func right away into this PR (if deemed appropriate - need feedback)
[ ] after testing that stuff mostly doesn't break with change above, add another commit where fnutils calls within our modules are changed to the correct func implementations (if deemed appropriate for this PR as opposed to a separate one - need feedback)

lowne:
It just occurred to me that `hs.collection` can be a good name - the module has more to do with handling lists and sets than with functional programming after all.

PR #496, opened 2015-08-15
[WIP] Add luaskin reference support

### LINKED ISSUE #454: Make luaL_unref() less dangerous
I think it's fairly clear that we are playing with fire when it comes to luaL_unref(). I want to look at replacing it somehow in LuaSkin.

My thoughts so far are:

 * Instead of using LUA_REGISTRYINDEX for everything, we could have a ref table for each module, so they are unable to contaminate each other's references
 * We could replace luaL_unref() with a custom implementation that never removes objects, but simply replaces them with a sentinel object that has appropriate metamethods to trigger a log and a crash when a reference is used after being freed
 * We could replace luaL_ref()/luaL_unref() to use a single custom table, and only allocate monotonically increasing references, so we'd never re-use one

We could actually do all of the above, for maximum safety. Anyone have any thoughts on this?

cmsj:
Closes #454 

The idea here is that each time an extension is registered with its own objects, we create a separate table for it in LUA_REGISTRYINDEX and it should then use the new LuaSkin ref/unref/push wrappers to interact with its reference table.

This doesn't stop us from corrupting the reference tables by accidentally over-unreffing, but it does mean that the damage is isolated in each module, instead of removing/re-using arbitrary items in LUA_REGISTRYINDEX, as we have now, so it should make debugging these issues significantly easier.

Currently a WIP because all that exists right now is the base implementation in LuaSkin and a quick port of the hotkey module to use the new API.

### DIFF
--- a/Hammerspoon/MJLua.m
+++ b/Hammerspoon/MJLua.m
@@ -13,6 +13,7 @@
 
 static LuaSkin* MJLuaState;
 static int evalfn;
+int refTable;
 
 /// === hs ===
 ///
@@ -313,7 +314,7 @@ void MJLuaSetup(void) {
     MJLuaState = [LuaSkin shared];
     lua_State* L = MJLuaState.L;
 
-    [MJLuaState registerLibrary:corelib metaFunctions:nil];
+    refTable = [MJLuaState registerLibrary:corelib metaFunctions:nil];
     push_hammerAppInfo(L) ;
     lua_setfield(L, -2, "processInfo") ;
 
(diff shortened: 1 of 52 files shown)


cmsj:
(also for now this is based off the branch for #487, hence those commits showing up)

cmsj:
Ported to all the modules except hs.application/hs.window/hs.uielement (we really need to do something about their crazy setup) and hs.fs (because doing so would force me to accept that we're wildly diverging from LFS, not just adding to it).

(hs.notify has been ported, and for bonus points, ported to the LuaSkin library registration too)

cmsj:
I'm also going to land this, because I believe it's working, but I want to force some testing :)

PR #500, opened 2015-08-16
Add argument type checker to LuaSkin

cmsj:
This one I'm not so sure about - it may be a solution in search of a problem, or at least to a larger extent than the other LuaSkin additions of late :)

I like the way it works, it's reasonably lightweight, and I like the idea of frontloading all of the type checking as soon as you cross the Lua->C boundary, rather than maybe-hopefully having lua_check* at various points in the code that follows.

### DIFF
--- a/LuaSkin/LuaSkin/Skin.h
+++ b/LuaSkin/LuaSkin/Skin.h
@@ -11,6 +11,18 @@
 #import "lualib.h"
 #import "lua.h"
 
+// Define some bits for masking operations in the argument checker
+#define LS_TBREAK         1 << 0
+#define LS_TOPTIONAL      1 << 1
+#define LS_TNIL           1 << 2
+#define LS_TBOOLEAN       1 << 3
+#define LS_TNUMBER        1 << 4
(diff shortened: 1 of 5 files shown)


cmsj:
@asmagill do you think this is useful? I'm still undecided. I'm not completely happy with the API, but I can't think of a neater way to do it efficiently. I still like the idea of making it very explicit which argument types are acceptable, but I don't know if it's worth it. (sorry, I'm pinging you on everything today ;)

asmagill:
My fear with anything like this is that it's hard to come up with every possible argument combination that might be possible/make sense to the developer/user... I'd have to play with it and see.  As long as its not required (maybe encouraged, but not required, at least not until we see where it breaks down and what it might take to extend it), I see no problem with adding it as an option for those developers which choose to use it.

cmsj:
@asmagill I certainly wouldn't want it to be mandatory, at least not until we'd shown that it can cover all the bases.

Perhaps the thing to do is to land it and convert a few things over to using it, and see how it feels?

asmagill:
I'm fine with that.

PR #503, opened 2015-08-17
(do not) add hs.moses

lowne:


jhgg:
Why include minified version? Why not full version??

lowne:
Because it's smaller?

jhgg:
Does size matter??

jhgg:
Would rather be able to see stack traces through non minifed code.

lowne:
Fine. Obviously I had to remove all the comments.
(OT, for future reference: `make docs` fails in a very creative (and impossible to diagnose) way if a source file doesn't end in `\n`)

asmagill:
Just do a global replace of '---' to '--' or '-- -' before adding the hammerspoon comments.  The docs code only cares about comments that start with 3 consecutive dashes.

And yes, it's errors do suck :-)  At some point I hope to get around to that since our internal documentation system will support third party extensions adding their docs to the `help` command; until doing so is relatively painless for the developer and documented, however, it's kind of a useless feature :-)


And I would agree -- seeing the expanded original would help with debugging, especially since we're expecting a few hiccups if we move things away from fnutils.

asmagill:
Amusingly, luac likes the full version better than the min version:

~~~sh
$ luac -o 1 moses.lua 
$ luac -o 2 moses_min.lua
$ luac -o 3 -s moses.lua 
$ luac -o 4 -s moses_min.lua
$ ls
total 608
-rw-r--r--   1 amagill  staff  55073 Aug 17 13:17 1
-rw-r--r--   1 amagill  staff  55576 Aug 17 13:17 2
-rw-r--r--   1 amagill  staff  25668 Aug 17 13:18 3
-rw-r--r--   1 amagill  staff  25813 Aug 17 13:18 4
-rw-r--r--   1 amagill  staff  57506 Jul  8 02:24 moses.lua
-rw-r--r--   1 amagill  staff  17706 Jul  8 02:24 moses_min.lua
$
~~~

lowne:
Agreed - I originally went with the minified one just because I really didn't want to deal with the comments (and `make docs`) :)

lowne:
(kinda OT - btw I *love* `help.anything` from the console)

lowne:
Sigh... future reference #2: `moses(t)` does **not** "transparently" insert itself at the end of the metatable chain; it creates a whole new wrapper around `t` - therefore you must either

1. keep a local reference around if you can't chain everything (`local tmos=moses(t)`)
2. or unwrap the thing (`...:value()`) before passing it on

lowne:
`<rant>`
So after a couple days of porting everything to moses, I must say that I'd *much* rather keep `fnutils` and work around the original issue (#455) by looping manually with ipairs where necessary.
I find it *extremely* abstruse and unidiomatic (due perhaps to its JS inspiration); every single time I need the `k,v` pair, nope, no iterator, and conversely when I need just the value here I go with another `_,v` :cry: 
Sure it's powerful, and I even "took advantage" of `zip`:
```lua
return screens:map(function(_,s)return frame:intersect(s:fullFrame()).area end)
  :zip(screens:value()):reduce(function(state,v)return v[1]>state[1] and v or state end,{0,nil}):value()[2]
```
but that's just because I *had* to; compare to `hs.func`, no `_`, no `:value()`s, and no need for `zip`:
```lua
return screens:map(function(s)return s,frame:intersect(s:fullFrame()).area end)
  :reduce(function(maxs,maxa,s,a)if a>maxa then return s,a else return maxs,maxa end end,nil,0)
```
or how do I do `someValue=coll:find(predicateFn)`, other than the opprobrium `someValue=coll[moses.detect(coll,predicateFn)]`? And, again, why no in-order filtering? :sob: 
`</rant>`

So in the interest of performance, future readability, and sanity, I think I'll just unwrap and iterate manually everywhere; but I'd **very much** appreciate any code samples and/or argumentation meant to make me see the light, as there must be something wrong with me since I'm voicing the minority opinion here. (not being sarcastic, I really mean it; I like to take every chance to learn something new, and being dead wrong enables the highest form of learning :) )

asmagill:
While I admit to being impressed with the offerings described in moses, and definitely see weaknesses and inconsistencies within hs.fnutils, I have to admit, I've only used the bare minimum of moses so far...

I know we've talked about it as a replacement to fnutils, but what if we were to just include it as close to "as-is" as possible while continuing with hs.fnutils for our own shortcuts and helper functions... Is there too much overlap for this to make sense?  Or would we be better pulling (or replicating the effect of) what we like from moses and expanding fnutils and sticking with just the one module?

cmsj:
Well this took an unexpected turn!

@lowne I don't think your opinion is necessarily a minority one, and it's one I take very seriously

I'm largely ambivalent towards this because I don't really like any kind of functional programming. fnutils has some useful stuff in it, but I tend to look at long chains of maps/filters/zips/etc and think "I have zero interest in parsing that in my head" ;)

If porting our own code from fnutils to moses means making it longer, more complex and harder to reason about, then I see little reason to do it, which brings us back to the whole question of whether moses should be included, to which I have to re-state something I've said before, which is that I don't think Hammerspoon should be aiming to be a wholistic Lua development environment. Our modules are very heavily focussed towards exposing OS features. fnutils is one of the only exceptions to this, because it's been a very useful abstraction for common operations elsewhere.

If moses is not suitable for our own modules to use, then I revert my position to thinking we should not merge it. It's just going to be a maintenance burden (in terms of keeping it up to date, integrating its docs, etc) which is trivially solved for users by either luarocks or just dropping moses.lua into ~/.hammerspoon/

I'm going to point everyone at the topic in our IRC channel - we exist to make the runtime, funtime. If we're not having fun then we are making a mistake and need to correct it :)

lowne:
Agreed; I finished removing moses (previously converted from fnutils) from window, and the resulting code (besides being sometimes *much* faster, details on the upcoming PR) is **a lot** more parseable without requiring to look stuff up elsewhere, primarily because there are no "surprises" in a for loop with pairs/ipairs - it's well-defined and anyone can see what it does at a glance, and furthermore it's much easier to reason about wasteful computations/calls and bring those out of the loop as required. (no, memoization is *not* the solution; after all it's Lua we're talking about.)

So in short I suggest we remove all this stuff *from our code* (as I said, I'm on it already), for easier maintainability/refactoring/expansion down the line.

As for shipping stuff for users' benefit: I agree with @cmsj there's no point in shipping moses (or anything else) if it's not used internally, when for users it's just a matter of dropping whatever they want onto their .hammerspoon folder. 

As for fnutils, it's already there, but I still think we should avoid leading users into impossible-to-diagnose bugs (main culprit being silently out-of-order filtering, but there's a couple more there); this could be "fixed" by putting adequate warnings in the docs (there's no proper fixing fnutils with full backward compatibility as I didn't quite explain in #488 - just take my word for it), but then, in the user shoes, my immediate reaction would be "ok, careful this, can't do that, *why* don't you guys just fix it?" - so in the end I think it's better to deprecate and undocument it.

At this point I *might* argue again in favour of hs.func (or, better, hs.collection, since that's what it is in essence) as a proper (and more versatile to boot) fnutils replacement, but I'm now inclined to think that maybe it's better to just leave all this stuff out, as I bet the vast majority of users do not or (more importantly) should not care; for those who do, again, nothing's stopping them from dropping in moses/penlight/their own homemade collection library in their configs.

lowne:
I'll keep this open for a bit for further discussion, if any.

PR #509, opened 2015-08-20
WIP hs.notify updates

asmagill:
I'm putting this up for anyone who wants to take a look at it -- I'm still testing it, so won't be pushing it into core for a bit, but it's worked flawlessly for me for the last half day if anyone wants to check it out.

Changes:
* complete rewrite of custom NSUserNotificationCenterDelegate to utilize a single callback through LuaSkin for stability and standardization
* added support for customImage in notification (10.9 and later)
* un-deprecated show and related functions -- show is now a shorthand for simple immediate notifications, and register is used for registering callback tags (see below for reasoning)
* expanded documentation for formerly deprecated functions
* added support for scheduled notifications (to be displayed after time has passed)

The reasoning behind moving to a single callback and un-deprecating the register/unregister of tags instead of a single function per notification is two fold: simplicity and it provides a path for allowing notifications which are scheduled to be displayed when Hammerspoon is not running, or which are still in the notification center after a reload of your Hammerspoon configuration to "re-attach" to their callbacks.  For this to work, you do have to make sure your init.lua requires hs.notify (to install our delegate) and registers the tagged callback functions again with the same tag name, but it works.

A future update (I'm not ready to tackle it yet) will be to modify Hammerspoon's NSApplicationDelegate so that it can hold notifications which cause a launch of Hammerspoon to be eventually handed over to the appropriate delegate in hs.notify after the application has finished loading. Currently, if Hammerspoon is not running and you click on a notification, all it will do is launch Hammerspoon -- the first notification is not passed through because the NSUserNotificationDelegate doesn't exist yet.

### DIFF
--- a/extensions/notify/init.lua
+++ b/extensions/notify/init.lua
@@ -1,13 +1,68 @@
+
 --- === hs.notify ===
 ---
---- On-screen notifications using Notification Center
+--- This module allows you to create on screen notifications in the User Notification Center located at the right of the users screen.
+---
+--- Notifications can be sent immediately or scheduled for delivery at a later time, even if that scheduled time occurs when Hammerspoon is not currently running.  Currently, if you take action on a notification while Hammerspoon is not running, the callback function is not honored for the first notification clicked upon -- This is expected to be fixed in a future release.
+---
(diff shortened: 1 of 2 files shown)


cmsj:
The future NSApplicationDelegate work may also want to consider hs.urlevent, which might also appreciate being told about URL events that have caused Hammerspoon to launch :D

asmagill:
update: still looking good, so unless there are any objections, I'll probably push this sometime tomorrow...  want to stress it a little more.

FWIW with the additions in #512 to LuaSkin, its possible to "re-aquire" the actual (well, new, but equivalent) hs.notify objects for delivered (still in the notification center) and scheduled notifications after a reload or restart of Hammerspoon.  I'll add the necessary modifications (there are a couple) to this module when #512 lands.

PR #501, opened 2015-08-16
[READY] add hs.screen.find (needs #471)

lowne:
original message:
> hs.grid.show() is currently broken, merge this before next release

Whops, I've accidentally overwritten the wrong branch and lost the original commit... However it was an easy fix: https://github.com/Hammerspoon/hammerspoon/blob/master/extensions/screen/init.lua#L185 `screen`->`aScreen`

The *current* PR here does include the fix, but it has unmet dependencies (geometry #471), so it might not land for a while.

Issue #497, opened 2015-08-15
appfinder.windowFromWindowTitlePattern is very slow


STRML:
`appfinder.windowFromWindowTitlePattern` is very slow on latest `0.9.38`. It adds about 4-5 seconds to my config's init time.

Not sure why this is, or if it has something to do with my config. A dump of app:allWindows() lists about 28 items.

This is not a critical issue, but I figured I'd open it in case someone else was googling around. To work around it, I instead grab the application I expect the window title to be under, then cycle through the windows. This is much faster. For example:

```lua
function applyLayout(layout)
  return function()
    alert.show("Applying Layout.")
    for appName, place in pairs(layout) do
      -- Two types we allow: table, which is {appName, windowTitle}, or just the app itself
      if type(appName) == 'table' then
        local parentAppName = appName[1]
        local windowPattern = appName[2]
        local app = appfinder.appFromName(parentAppName)
        if app then
          for i, win in ipairs(app:allWindows()) do
            if string.match(win:title(), windowPattern) then
              applyPlace(win, place)
            end
          end
        end
      else
        local app = appfinder.appFromName(appName)
        if app then
          for i, win in ipairs(app:allWindows()) do
            applyPlace(win, place)
          end
        end
      end
    end
  end
end
```

cmsj:
I think this is a duplicate of #233, so I'm closing this issue.

lowne:
("crossposting" for completeness)
After having updated to 0.9.40 or later,  as soon as you (if you still) notice sluggishness or excessive delay (upon calling `windowHints`, `layout`, etc), please open the Console and run this:
```lua
hs.window._timed_allWindows()
```
(If you get no output, try again a bunch of times (up arrow+return) until you (hopefully) do)

Then please paste the output in issue #233.

Issue #277, opened 2015-04-27
focusWindowWest() very slow on OS X 10.9.5


posto:
The ```focusWindowWest()``` is very slow on my machine (focus east, north and west are also slow, for that matter).
It always takes about 10seconds to move the focus, which makes it ususable.

Can I somehow help to debug this issue?

Btw, ```hs.hints.windowHints``` is also very slow, might be somehow related.

cmsj:
@posto Is it also this slow after a reboot? I've seen instances where the OS X Accessibility API calls take a while to return, but only after the machine has been up for quite a while.

posto:
Just restarted. Still happens, takes 6 sec to switch :(

dktn:
Same here: focus and hints take 6 seconds.

cmsj:
In these situations, it's worth checking Activity Monitor for any processes that are listed as Not Responding.

The reason being, when we use the Accessibility API to go off and fetch window information, the replies actually come from the apps themselves, so if one of them has hung, or is just really busy, there will be a delay in processing the functions.

dktn:
It is not a case, none of the application is busy nor not responding. Processor idle 97-98%.
In contrast hs.application.launchOrFocus and all hs.grid operations work instantly.

cmsj:
I think this is a duplicate of #233, so I'm closing this issue.

lowne:
("crossposting" for completeness)
After having updated to 0.9.40 or later,  as soon as you (if you still) notice sluggishness or excessive delay (upon calling `windowHints`, `layout`, etc), please open the Console and run this:
```lua
hs.window._timed_allWindows()
```
(If you get no output, try again a bunch of times (up arrow+return) until you (hopefully) do)

Then please paste the output in issue #233.

Issue #233, opened 2015-02-26
WindowHints Performance


rphillips:
I'm running Yosemite 10.10.2 and hammerspoon's windowHints take about 5-6 seconds to display. Seems to be a bit slow.

``` lua
hs.hotkey.bind(hyper, "i", function()
    hs.hints.windowHints()
end)
```

cmsj:
Yeah, I've noticed this too. It seems to not happen immediately, but some time afterwards. If you move the mouse over the Hammerspoon menubar icon, you'll probably see a beachball.
I need to do some investigation, but my suspicion is that this means that the performance issue is coming from something in the OS X Accessibility API :(

atsepkov:
I'm seeing similar issue, occasionally. Sometimes hints open immediately (even with 20+ apps/windows), other times there is a lag of several seconds.

echiu64:
Same, i notice it's not just hammerspoon, but overall GUI responsiveness goes downhill. I've been rebooting Yosemite about 10X more often than my windoze machine. Disappointing...

rcoedo:
The same thing happens to me, `hs.hints.windowHints()` is completely unusable at the moment.

I also noticed that `hs.hints.windowHints()` somehow messes with the `hs.hotkey.modal` module. It's like the modal state is stuck after calling windowHints():

1. Make the hs.hints.windowHints() call
2. Switch to a window
3. Now if you use your modal key mappings without the prefix the functions are executed.

cmsj:
This should be addressed when #511 lands

lowne:
After having updated to 0.9.40 or later,  as soon as you (if you still) notice sluggishness or excessive delay (upon calling `windowHints`, `layout`, etc), please open the Console and run this:
```lua
hs.window._timed_allWindows()
```
(If you get no output, try again a bunch of times (up arrow+return) until you (hopefully) do)

Then please paste the output here.

ryfow:
```> hs.window._timed_allWindows()
took 6.00s for com.adobe.PDApp.AAMUpdatesNotifier
took 6.00s for org.pqrs.Karabiner-AXNotifier
table: 0x7ffca30b3280
```

lowne:
@ryfow thanks, added in #594

cmsj:
(fwiw, #594 is merged and will be in the next release)

asethwright:
Not sure if this is helpful but an old adobe verison appears to cause slowness. Once I kill the process the window hints work fine.

```
[...]

PR #511, opened 2015-08-20
[READY] improve window, application

### LINKED ISSUE #478: frame of fixed-size windows may be incorrect during animation
Since the asynchronous animation changes the window frame may be incorrect for fixed-size windows during an animation. This is due to changes to the way how `win:frame()` determines the frame of a window.

1. For example take a fixed-size window like the main window of Terminal.app.
2. Call `win:moveToUnit({x = 0.5, y = 0.5, w = 0.5, h = 0.5}):ensureIsInScreenBounds()` on the main window.
3. Depending on the font the Terminal window can only resize in rows and columns. Because of this the window frame after `moveToUnit` may be too large or too small and exceed the screen bounds.
4. The `ensureIsInScreenBounds` function does not work because it operates [on the frame from the `moveToUnit` animation because this animation is not finished](https://github.com/Hammerspoon/hammerspoon/blob/71e847ec217669105d16f64e49cd1139fe3fe1b0/extensions/window/init.lua#L110). The returned frame is not the actual frame of the window.

Workaround:
Let the `moveToUnit` animation finish before `ensureIsInScreenBounds`:
`win:moveToUnit({x = 0.5, y = 0.5, w = 0.5, h = 0.5}, 0):ensureIsInScreenBounds()`

lowne:
- added `hs.window.find()`, `hs.application.find()` (similar to #501, which is a prerequisite of this PR for `win:moveToScreen'dell'`) - `hs.appfinder` isn't needed anymore
- `hs.window.get()` and `hs.application.get()` are "strict" variants for `find`
- `hs.application:findWindow()` and `hs.application:getWindow()`
- added `hs.application.open()` (wraps the two launchOrFocus methods)
- `hs.application:title()->hs.application:name()` (users shouldn't have to suffer just because it derives from uielement)
- geometrified `hs.window` for cleaner code (#471, prerequisite), and to enable...
- `hs.window:move()` one-stop-shop for window movement; takes a point (moves by x,y), a rect (does setFrame) or a unit rect+screen (basically does :moveToScreen+:moveToUnit in one go) 
- unfnfied everything for correctness (there was a potential heisenbug in find_windows_in_direction), readability, maintainability and performance; in particular, **the occasional 3 to 5 seconds delay** when doing `hs.window.allWindows()` and derivatives (including stuff like #497) **seems now gone**; it used to happen quite often here, and I haven't seen it yet
- fixed some minor bugs, docs
- ensureIsInScreenBounds doesn't care about animations (fixes #478)
- also needs the windowfilter->window.filter PR (#513) ~~which apparently I haven't opened yet :/~~

### DIFF
--- a/extensions/application/init.lua
+++ b/extensions/application/init.lua
@@ -6,10 +6,9 @@ local uielement = hs.uielement  -- Make sure parent module loads
 local application = require "hs.application.internal"
 application.watcher = require "hs.application.watcher"
 local window = require "hs.window"
-local moses = require "hs.moses"
 
-local type,ipairs=type,ipairs
-local tunpack,tpack,tinsert=table.unpack,table.pack,table.insert
+local type,pairs,ipairs=type,pairs,ipairs
+local tunpack,tpack,tsort=table.unpack,table.pack,table.sort
 
 --- hs.application:visibleWindows() -> win[]
 --- Method
@@ -23,8 +22,8 @@ local tunpack,tpack,tinsert=table.unpack,table.pack,table.insert
 function application:visibleWindows()
   --  return moses.filter(self:allWindows(), window.isVisible)
   local r={}
-  if self:isHidden() then return r
-  else for _,w in ipairs(self:allWindows()) do if not w:isMinimized() then tinsert(r,w) end end end -- (barely) faster
+  if self:isHidden() then return r -- do not check :isHidden for every window
+  else for _,w in ipairs(self:allWindows()) do if not w:isMinimized() then r[#r+1]=w end end end
   return r
 end
 
@@ -88,10 +87,16 @@ function application.find(hint,exact)
   local r=application.applicationsForBundleID(hint)
   if #r>0 then return tunpack(r) end
   local apps=application.runningApplications()
-  r=moses.filter(apps,exact and function(_,a)return a:name()==hint end or function(_,a)return a:name():lower():find(hint:lower())end)
-  if #r>0 then return tunpack(moses.sort(r,function(a,b)return a:kind()>b:kind()end)) end -- gui apps first
-  r=moses.toArray(window.find(hint,exact))
-  if #r>0 then return tunpack(moses(r):map(function(_,w)return w:application()end):unique():value()) end
+
+  if exact then for _,a in ipairs(apps) do if a:name()==hint then r[#r+1]=a end end
+  else for _,a in ipairs(apps) do if a:name():lower():find(hint:lower()) then r[#r+1]=a end end end
+  tsort(r,function(a,b)return a:kind()>b:kind()end) -- gui apps first
+  if #r>0 then return tunpack(r) end
+
+  r=tpack(window.find(hint,exact))
+  local rs={} for _,w in ipairs(r) do rs[w:application()]=true end -- :toSet
+  for a in pairs(rs) do r[#r+1]=a end -- and back, no dupes
+  if #r>0 then return tunpack(r) end
 end
 
 
@@ -106,7 +111,7 @@ end
 ---    - the application's bundle ID as per `hs.application:bundleID()`
 ---
 --- Returns:
----  * the `hs.application` object of the launched or activated application; `nil` if not found
+---  * the `hs.application` object for the launched or activated application; `nil` if not found
 function application.open(app)
   if type(app)~='string' then error('app must be a string',2) end
   if application.launchOrFocus(app) then return application.find(app,true) end
--- a/extensions/application/init.lua
+++ b/extensions/application/init.lua
@@ -52,8 +52,27 @@ end
 --- Alias for `hs.application:title()`
 application.name=application.title
 
---- hs.application.find(hint[, exact]) -> hs.application object(s)
---- Function
+--- hs.application.get(hint) -> hs.application object
+--- Constructor
+--- Gets a running application
+---
+--- Parameters:
+---  * hint - search criterion for the desired application; it can be:
+---    - a pid number as per `hs.application:pid()`
+---    - a bundle ID string as per `hs.application:bundleID()`
+---    - an application name string as per `hs.application:name()`
+---
+--- Returns:
+---  * an hs.application object for a running application that matches the supplied search criterion, or `nil` if not found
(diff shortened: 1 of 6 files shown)


lowne:
I think there's a good chance that filtering non-gui apps in `window.allWindows()` (the "major speedup" commit here) will help wrt #277, #175 and the like.

cmsj:
@lowne What's the reasoning for the app:kind() test being >0? Shouldn't apps with a kind() of 0 be able to own windows?

lowne:
@cmsj wtf, yes... I explicitly remember thinking "kind must be >=0" (as in windowfilter) - it slipped through somehow. That was an important one to catch! Will fix.

lowne:
Fixed; and added `hs.window._timed_addWindows()` (to be run from the console) to help in diagnosing slowness problems such as in the previously mentioned issues.

cmsj:
@lowne also, it looks like all of the docstrings in hs.appfinder are being rewritten by this, to not be docstrings anymore? (i.e. --- replaced with --)

extensions/application/init.lua
```diff
@@ -19,7 +21,11 @@ application.watcher = require "hs.application.watcher"
 --- Returns:
 ---  * A table containing zero or more hs.window objects
 function application:visibleWindows()
-  return fnutils.filter(self:allWindows(), window.isVisible)
+  --  return moses.filter(self:allWindows(), window.isVisible)
```
cmsj: Could we remove the moses reference(s) please?

extensions/application/init.lua
```diff
@@ -48,4 +48,145 @@ function application:activate(_allWindows)
...
+---    - the full path to an application on disk (including the `.app` suffix)
+---    - the application's bundle ID as per `hs.application:bundleID()`
+---  * wait - (optional) the maximum number of seconds to wait for the app to get into "launching" state, if not already running; if omitted, defaults to 0;
+---           if the app takes longer than this to launch, this function will return `nil`, but the app will still launch
+---
+--- Returns:
+---  * the `hs.application` object for the launched or activated application; `nil` if not found
+---
+--- Notes:
+---  * the `wait` parameter will *halt the entire script* in order to return the application object "synchronously"; only use it if you
```
cmsj: Rather than "halt the entire script", maybe this should say "block all Hammerspoon activity", since timer.usleep() will block the entire application.

extensions/window/init.lua
```diff
@@ -8,14 +8,12 @@
 local uielement = hs.uielement  -- Make sure parent module loads
 local window = require "hs.window.internal"
 local application = require "hs.application.internal"
-local fnutils = require "hs.fnutils"
+--local moses = require "hs.moses"
```
cmsj: Ditto removing moses.

extensions/window/init.lua
```diff
@@ -25,30 +23,131 @@ local atan,cos,abs = math.atan,math.cos,math.abs
 --- hs.window.animationDuration = 3 -- if you have time on your hands
 window.animationDuration = 0.2
 
---- hs.window.allWindows() -> win[]
---- Function
+--- hs.window.allWindows() -> list of hs.window objects
+--- Constructor
```
cmsj: I don't think this is a constructor - they should typically be returning an object that has a metatable.

extensions/window/init.lua
```diff
@@ -25,30 +23,131 @@ local atan,cos,abs = math.atan,math.cos,math.abs
...
+    if app:kind()>=0 then local _=app:allWindows() r[app:name()]=timer.secondsSinceEpoch()-starttime end
+  end
+  for app,time in pairs(r) do
+    if time>0.1 then print(string.format('took %.2f s for %s',time,app)) end
+  end
+  return r
+end
+
+--- hs.window.visibleWindows() -> list of hs.window objects
+--- Constructor
```
cmsj: Ditto constructor

extensions/window/init.lua
```diff
@@ -25,30 +23,131 @@ local atan,cos,abs = math.atan,math.cos,math.abs
...
+  --  return fnutils.filter(window.allWindows(), window.isVisible) -- nope
+  local r={}
+  for _,app in ipairs(application.runningApplications()) do
+    if app:kind()>0 and not app:isHidden() then for _,w in ipairs(app:visibleWindows()) do r[#r+1]=w end end -- speedup by excluding hidden apps
+  end
+  return r
+end
+
+--- hs.window.orderedWindows() -> list of hs.window objects
+--- Constructor
```
cmsj: Ditto constructor

extensions/window/init.lua
```diff
@@ -25,30 +23,131 @@ local atan,cos,abs = math.atan,math.cos,math.abs
...
+---
+--- Notes:
+---  * see also `hs.window.find` and `hs.application:getWindow()`
+function window.get(hint)
+  return tpack(window.find(hint,true),nil)[1] -- just to be sure, discard extra results
+end
+window.windowForID=window.get
+
+--- hs.window.find(hint) -> hs.window object(s)
+--- Constructor
```
cmsj: Ditto constructor

extensions/window/init.lua
```diff
@@ -328,71 +393,66 @@ end
...
-  return lastscreen
+  if maxs then return maxs end
+  local mind=99999 -- if (impossibly) a window is totally out of bounds, get the closest screen
+  for _,s in ipairs(screens) do
+    local d=frame:vector(s:frame()).length
+    if d<mind then mind,maxs=d,s end
+  end
+  return maxs
+end
+--[[ -- moses version
```
cmsj: ditto moses

cmsj:
This should also help with #233, or at least allow us to apply similar fixes until that is solved.

lowne:
@cmsj yes, the idea being that application/window.find/get obviate the need for a separate module (but staying for now for backward compatibility) - I changed the appfinder functions to be thin wrappers.

cmsj:
@lowne I think I'd prefer to keep the docs around for a while, with a note at the top saying that the module is deprecated. It's pretty widely used, afaik.

lowne:
@cmsj done

PR #512, opened 2015-08-20
[WIP] NSObject->Lua conversion in LuaSkin

asmagill:
This is a work in progress and should not be considered for inclusion in core at this time.  I am creating this pull for comments and questions for the brave who want to check it out  😛

@cmsj - is there a reference guide for the commenting style you're using in LuaSkin?  I'd like to make mine match, but haven't really looked that closely at it yet.

Contains:
* NSObject to Lua Conversion (actual converters return an int - it should always be 1, but this allows these to be used as `return [skin push...]` in your own module functions.)

  * [x] `(int) [skin pushNSObject:obj]`and `(int) [skin pushNSObject:obj preserveBitsInNSNumber:bitFlag]`
    Takes an arbitrary NSObject and checks registered helpers, then the defaults for a class match and uses the appropriate method to convert the object into its Lua representation.  Without preserveBitsInNSNumber:YES, NSNumbers which are unsigned long long are treated specially (see below)
    
  * [x] *coded, but untested in LuaSkin* : `(int) [skin pushNSObject:obj withLocalHelpers:fnList ]` and `(int) [skin pushNSObject:obj withLocalHelpers:fnList preserveBitsInNSNumber:bitFlag]`
    Takes an arbitrary NSObject and checks a run-time supplied list of conversion functions, then the registered helpers, then the defaults for a class match and uses the appropriate method to convert the object into its Lua representation.  This method allows a module which knows its data might match another converter to override it for this specific call only if special needs apply.
    
  * [x] `(void) [skin registerPushNSHelper:helperFN forClass:className]`
    registers a c-function to handle conversion for NSObject types which are used within a module.  See the example in hs.application which allows NSRunningApplication objects to be converted to a hs.application userdata for use within Hammerspoon.
    
  * [x] `(void) [skin unregisterPushNSHelperForClass:className]`
    un-registers a c-function and removes it from the list of registered converters.

    * Individual Converters (in order)   (*not directly callable*)
      * [x] `(int) [skin pushNSNull:obj]`
    Converts [NSNull null] to lua's nil; mainly included for recursive conversion from NSArray or NSDictionary with null placeholders -- I don't expect this to be called directly unless you are very very weird.
      * [x] `(int) [skin pushNSNumber:obj]` and `(int) [skin pushNSNumber:obj preserveBits:bitsFlag]`
    Converts [NSNumber] to lua number or integer as appropriate. Without preserveBits:YES, unsigned long long is treated special.  See below.
      * [x] `(int) [skin pushNSString:obj]`
    Converts [NSString] to a UTF8 representation in a lua string.
      * [x] `(int) [skin pushNSData:obj]`
    Converts [NSData] to a byte-by-byte copy as a lua string (note that this probably won't print to the console properly, but can be manipulated as binary data)
      * [x] `(int) [skin pushNSDate:obj]`
    Converts [NSDate] to an integer useable with os.date() in lua.
      * [x] `(int) [skin pushNSArray:obj]`
    Converts [NSArray] into a lua table.  Recursively calls pushNSObject: on its contents.
      * [x] `(int) [skin pushNSSet:obj]`
    Converts [NSSet] into a lua table.  Recursively calls pushNSObject: on its contents.
      * [x] `(int) [skin pushNSDictionary:obj]`
    Converts [NSDictionary] into a lua table.  Recursively calls pushNSObject: on its keys and contents.
      * [x] `(int) [skin pushNSUnknown:obj]`
    Final check -- catches [NSObject] (which should be any NS type) and returns its string representation as %@ would in NSLog.

* Lua to NSObject Conversion
  * More limited in options because it is probably better/easier for the module to do what it needs rather than force it into a mold that would have to be managed anyway (e.g. tableFromIndex can return NSArray or NSDictionary, depending upon context, so you have to check on type of object returned, but using lua_next(...) doesn't care...  This is for cases where you really do want an NSObject: json, settings, nslog, etc.

  * [x] `(id) [skin toNSObjectFromIndex:idx]` - uses the following to return the appropriate object for whats on the stack at the specified index

  * Individual converters (*not directly callable - existing `lua_toX` methods are easier to use and understand*)  
    * [x] `(NSNumber*) [skin numberFromIndex:idx]` - handle Lua types number and integer
    * [x] `(id) [skin stringFromIndex:idx]` - If valid UTF8, will return NSString, otherwise returns NSData.
    * [x] `(id) [skin tableFromIndex:idx]` - if table contains only numeric keys, returns NSArray, otherwise returns NSDictionary.  Recursively calls toNSObject: on its keys and contents.
    * [x] `(NSNumber *) [skin booleanFromIndex:idx]` - returns an NSNumber wrapped BOOL -- necessary for boolean values to be put into NSArray and NSDictionary objects.
    * [x] <del>`(BOOL) [skin boolFromIndex:idx]` - returns BOOL (this version not used by  `toNSObjectFromIndex:`)</del>
    * [x] `(NSNull *) [skin nilFromIndex:idx]` - returns [NSNull null]... necessary for inclusion in NSArray and NSDictionary objects... like `pushNSNull:`, I expect this to be called directly very very seldom if ever.
    * [x] `(NSString *) [skin unknownFromIndex:idx]` returns the lua tostring representation to catch userdata, thread, etc.

* Known issues, limitations, or questions...

  * Should there be a facility to add converters to the lua->NSObject converters... perhaps so userdata types could be converted back to an NSObject, if appropriate?  Usually we store more than just an NSObject in the userdata, so it would be an imperfect conversion, but would it be useful?  Or would it be better to leave it as userdata and assume the module knows how to deal with it as such (like hs.drawing with hs.image or hs.application, hs.window, and hs.uielement...)

  * in registerPushNSHelper:forClass:, if a class converter exists, it is overwritten.  Should this raise an error?
  
  * in registerPushNSHelper:forClass:, converters are stored in a NSDictionary, so order is not guaranteed.  This means that an object that might match more than one class will be handled by whichever one is found first, and that might change from invocation to invocation.  If changed to an array, then order would be based upon which order modules are loaded into Lua, and even this can be non-deterministic.  Suggestions?  Wait and see if it bites us and deal with it then?
  
  * <del>Does not handle loops (e.g. a NSDictionary in which one of its objects is itself)... </del> nor is there currently a maximum depth for recursion. <del>Looking into options here because this does need to be addressed before including in core.</del>  At depth of 512, it crashes... but duplicate references to dictionaries, sets, and arrays now detected, so this is unlikely -- unless something really does have a depth of 512+... should we catch/prevent this or is it unlikely enough?

  * in pushNSNumber:, Lua uses a signed long long type; this means that an unsigned long long cannot be accurately represented.  For an unsigned long long, the current convertor checks to see if the value is less than 0x8000000000000000, and if so returns it as an int.  Otherwise it returns it as a number.  <del>This will be a problem if we're actually interested in doing bit manipulations on the value... Which behavior should be the default, and which should be an optional method which can be included via pushNSObject:withLocalHelpers:?</del>  *See `pushNSObject:preserveBitsInNSNumber:` above.*

Any other thoughts or suggestions or questions, go ahead and put in here as well.


asmagill:
Should individual converters (pushNSString, pushNSArray, etc.) be discouraged?  Since pushNSObject will handle any of these, should only pushNSObject: an its variants be advertised?  Also, registering your own handlers for NSArray, NSSet, or NSDictionary can bypass duplicate detection... should these be prevented in overrides?

cmsj:
@asmagill I'll loop back on this properly asap, but I just wanted to mention that the doc format is Xcode's thing. It's called HeaderDoc.

Very excited about this PR :D

asmagill:
The individual converters are now part of an extension to the LuaSkin class... I don't really see a reason for them to be directly callable -- when you know you have an NSString or NSNumber, etc. [skin pushNSObject:] works just as well as it does for NSArray and NSDictionary (and won't return nil when you are wrong, so data validation falls back to lua, where its easier to do anyways)  and when converting from Lua to NSObjects, there are so many special cases and variants that it seemed better to have a generic converter for those cases when you really do want mass conversion (hs.json, hs.settings, NSLog, etc.) but leave it to the modules to handle there data as is easiest for the situation.

Note that a reference to an NSObject is anything which is actually a descendant of NSObject -- things that are structures (like NSRect, NSPoint, etc.) are not covered.

I also added a method which will check the specified stack element and determine if it is valid UTF8 `(bool) [skin isValidUTF8AtIndex:idx]` which is used internally to determine if the stack element should be converted to NSString or NSData.  I'm also planning to look at MJConsole.m and hs.ipc and see if this might help with the issues there (#334).

asmagill:
Self reference going from lua to NSObject now has to be asked for explicitly (i.e. `[skin toNSObjectFromIndex:idx allowSelfReference:YES]`)

This is done because things like NSDefaults and NSLog won't reject loops in their NSArray's and NSDictionaries, but they will crash because of them.

I have not added this option to NSObject to lua conversion because it is my understanding that, while allowed, this is somewhat rare coming from the OS, and it's also a heck of a lot easier to deal with on the Lua side.

asmagill:
Ok, a somewhat simpler public interface -- I'll update the list later.  Also, you can check out an example of its use for converting lua tables into various things at https://github.com/asmagill/hammerspoon_text (make install will put it into ~/.hammerspoon/hs/_asm/text)

In this code, I've implemented converters for NSAttributedString, NSFont, NSParagraphStyle, NSColor and NSShadow.  The command `getTextWithStyle(hs.drawing.textobject)` will return a table that basically mirrors NSAttributedString and its attributes dictionary (for the whole text... not doing multiple dictionaries with distinct ranges).  You can modify the text in array position 1 or the table in array position 2 and then do `setTextWithStyle(hs.drawing.textobject, resultTable)` and it will apply the changes you've made.

By using the table converters I've put in here, any place that uses an NSAttributedString (currently only hs.drawing, but you get the idea) can use [skin pushNSObject:] and [skin tableAtIndex: toClass:] to get and set the object and not have to worry about coding their own interpreters.

Neither my skin changes or the example are well documented yet, it's still very definitely a work in progress, but I thought I'd push it in case anyone wants to play with it and make suggestions.

asmagill:
Note to self (to make sure it makes it into the final documentation): toNSObjectAtIndex: and pushNSObject: now both use lua_pushlstring even for NSString (rather than lua_pushstring) since valid UTF8 (and lua strings) can include embedded nulls.

asmagill:
@cmsj, @lowne -- question on style for tables as types:  should tableAtIndexToRect:, should tableAtIndexToPoint:, and should tableAtIndexToSize: cause an error if the correct parameters aren't found? e.g. should {} be the equivalent of {x=0,y=0,h=0,w=0}?

lowne:
@asmagill since it's (mostly) for screen coordinates, there's no meaningful default for those imo

asmagill:
Pros:
* simplified constructor when you know that you'll be calculating the values again, perhaps multiple times, before it's even made visible to the user
* more closely matches objective-c defaults for the creation of said structures under ARC
* they aren't really limited to screen coordinates/size (though this is their most common usage) and sometimes an origin point/size *is* what you want
* more consistent with almost all other table handling in our modules -- i.e. missing fields are given a default, not cause an error

Cons:
* validating (with useful error messages) is slightly longer to code, but then the whole point of sticking it into LuaSkin is to code it once
* introduces requirements, but doesn't address extra fields (e.g. is {x=1,y=1,h=5,w=5,a="boo!"} a rect? should it be?)
* hs.window is a notable divergence from the usual practice of being lax about requiring the fields -- it does require them, which means either it couldn't use the new methods if we're lax, or the method needs a flag indicating whether it should be lax or not.

lowne:
Ok, since hs.window and hs.screen uses are fully wrapped Lua-side in hs.geometry rects, C code can use the lax version without consequences; and for the same reason these functions must ignore any extraneous fields since we're passing around full objects (just like for color tables, which can be trivially extended for additional common functionality in a  Lua object wrapper)

cmsj:
@asmagill Some notes/questions as I read through the diff:

 * How come conversionSupport is being done as an extension to LuaSkin rather than just putting the methods into the main @interface in Skin.h?
 * I think isValidUTF8AtIndex should have a @note that it uses lua_tolstring(), which may modify the value in the stack (per Lua docs)

Otherwise, this is looking pretty awesome.

In terms of validation of the tables, I tend to think that it would be better at this stage to not do that.

asmagill:
As I understand it, both lua_tolstring and lua_tostring (IIRC, it's just lua_tolstring with nil as the length parameter) change the value on the stack *only if the value is a number*.  I'll look closer at the lua source later to verify this.  Still, the point about making a note of it is a good one, for both this and the toNSObjectAtIndex: methods.

As to the extension, I think this is now mostly methods that shouldn't be called from the outside -- for example the versions that carry around an NSDictionary of already visited objects so that loops and self-reference and multiple references to the same object can be detected.  I may not have the latest version currently in the pull, so I'll check that out and make sure everything is current when I get home later today.

cmsj:
@asmagill yeah I think you're right about the `tostring` functions, I just figured it was worth noting in the docs.

As for the extension (or "category" I think is the right ObjC term), that does indeed seem to be how people do private methods, so, rock on :)

asmagill:
To clean up my merge issues with #537 and #542, I blew away my original fork of Hammerspoon... I have a backup at home, should I find that I'm missing something that didn't make it into core in the past  (I doubt it, but you never know), so I'm working off a leaner, cleaner fork now.  I've recreated this pull there as well and will replace this with it once I'm comfortable I didn't miss anything in these changes.

Hopefully it can be landed sometime later this week or next -- I've been having to remind myself to *not* use some of these methods with webview and I'd like to stop having to do that :-)

asmagill:
@cmsj re the isValidUTF8AtIndex method... I added a check so it will return NO automatically if it's not a string or number (otherwise lual_tostring will return NULL and an empty string *is* valid UTF8, so...)

This makes it safe anywhere you were going to use lua_tostring or lua_tolstring anyways.

However, it occurs to me we could also just return NO if its a number as well... then (depending upon context) you might need an additional check to see why the NO.

toNSObjectAtIndex doesn't suffer from these issues because it checks the type before calling any other functions, but other code using this method...

Certain cases are going to require extra coding or consideration no matter which way we go, so... what's your preference?

asmagill:
Superseded by #549

PR #519, opened 2015-08-25
ipc and hs now deal in raw CFData, not UTF8 based CFString

asmagill:
fixes UTF8 data issues with hs cli tool and adds libedit (libreadline) library back into target requirements.


asmagill:
This should close #334 as well, now.

In console:

~~~lua
> utf8.charpattern
[∅-�-�][�-�]*
~~~

From the hs command line tool:

~~~sh
> utf8.charpattern
[-?-?][?-?]*
> hs.cleanUTF8forConsole(utf8.charpattern)
[∅-�-�][�-�]*
~~~

For reference, the actual value of utf8.charpattern is defined as `"[\0-\x7F\xC2-\xF4][\x80-\xBF]*"`.

The ? marks in the first line of the command line output is what Terminal does with byte-sequences it doesn't recognize... the output from `hs` is now treated as CFData all the way through, rather than converted to a CFString first.

asmagill:
Oh, and \x7F is valid (but not visible) UTF8... that threw me at first since it looked like something was missing in the output.

PR #522, opened 2015-08-26
WIP: Adding chaining for methods and __tostring for userdata to remaining modules

asmagill:
Adding chaining of methods to:
* [x] eventtap.event.post()
* [x] caffeinate
* [ ] sound

Updating docs for:
* [x] uielement
* [x] screen

Adding __tostring for userdata for:
* [x] application/watcher.m
* [x] audiodevice/internal.m
* [x] battery/watcher.m
* [x] caffeinate/watcher.m
* [x] eventtap/event.m
* [x] eventtap/internal.m
* [x] hints/internal.m
* [x] hotkey/internal.m
* [x] httpserver/internal.m
* [x] keycodes/internal.m
* [x] milight/internal.m
* [x] pathwatcher/internal.m
* [x] screen/internal.m
* [x] screen/watcher.m
* [x] sound/internal.m
* [x] spaces/watcher.m
* [ ] uielement/internal.m
* [x] usb/watcher.m
* [x] wifi/watcher.m

The sound methods are kind of a low priority... don't think anyone particularly cares if they chain, so I'll get to it for consistency at some point, but it shouldn't hold this up.

Uielement is going to take some thought... it looks like there are 3 different userdata types in use, but none are actually stored in the LUA_REGISTRY table?  It may be that __tostring methods don't make sense for it... or maybe belong on the enclosing tables... don't know, need to sleep first.

At any rate, if this meets with no objections, I'd like to add it into core in the next few days.

### DIFF
--- a/extensions/application/watcher.m
+++ b/extensions/application/watcher.m
@@ -286,6 +286,11 @@ static int app_watcher_gc(lua_State* L) {
     return 0;
 }
 
+static int userdata_tostring(lua_State* L) {
+    lua_pushstring(L, [[NSString stringWithFormat:@"%s: (%p)", USERDATA_TAG, lua_topointer(L, 1)] UTF8String]) ;
+    return 1 ;
+}
+
 static int meta_gc(lua_State* __unused L) {
     return 0;
 }
@@ -311,6 +316,7 @@ static void add_event_enum(lua_State* L) {
 static const luaL_Reg metaLib[] = {
     {"start",   app_watcher_start},
     {"stop",    app_watcher_stop},
+    {"__tostring", userdata_tostring},
     {"__gc",    app_watcher_gc},
     {NULL,      NULL}
 };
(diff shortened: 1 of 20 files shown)


asmagill:
Should I land this?  I personally like knowing what type of userdata I'm seeing when I print it to the console while testing things, but since it does mean adding another function to the C files, I don't know that its realistic to enforce it with all things that make it to core, unless someone like me gets around to adding it every so often... I'm willing, for now, if no one objects.  I'm also thinking I'm going to leave uielement alone for now... I need to start using it myself and see how everything interconnects before trying to modify it.

cmsj:
I haven't had a chance to review/test this, but I am strongly in favour of what it's doing, so I'm fine with it landing. 

I also think we need to rethink uielement. I also don't really understand what it's doing with its metatables  and have left it out of the various core changes I've been making :/

PR #527, opened 2015-08-28
Added release as a no-op for backwards compatibility

asmagill:
Added release as a no-op, added hs.notify:getFunctionTag() and hs.notify.warnAboutMissingFunctionTag as a flag to show/hide missing tag warnings during callback

### DIFF
--- a/extensions/notify/init.lua
+++ b/extensions/notify/init.lua
@@ -77,6 +77,12 @@ local emptyFunctionPlaceholder = "__emptyFunctionPlaceHolder"
 
 module.activationTypes = _makeConstantsTable(module.activationTypes)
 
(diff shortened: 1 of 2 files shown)


cmsj:
Interesting comment style there, moving them inside the function declaration. Was that intentional? I quite like it :)

asmagill:
Yeah, it allows me to collapse a function in BBedit and collapse the comments when I do so.  It's still not automatic to me, but if a file gets big enough, I start looking for ways to reduce scrolling around!

cmsj:
The only thing I'm not so keen on, is that it's not consistent. I think I can live with that though :)

asmagill:
Should I go ahead and land this?

cmsj:
+1

Issue #524, opened 2015-08-27
hs.drawing and window levels


asmagill:
I think we need to revisit the question of adding a way to explicitly set the windowLevel for hs.drawing objects...

With 10.11 allowing for Menubar hiding, I've been playing around with my bartender-ish status item bar and found that it was blocking the menu when it dropped down over it.  Putting some code in to change the level to match that of the menubar makes for a nice effect -- a mini bar of status/menu items managed by Hammerspoon in the upper right corner that persists across spaces and full screen apps, and then when the menubar appears, the menubar takes focus, as one would expect, rather than remain blocked by the images.

So, I'm now finding that our current bringToFront, sendToBack, orderAbove, and orderBelow are not sufficient... should I open it up to all levels or just a new one -- the menubar level?  Should we change bringToFront and sendToBack to just one: setLevel (the other two could be maintained as lua wrappers, if desired).

Thoughts?

cmsj:
Let's just open the flood gates and have an arbitrary setLevel (probably with some pre-defined levels?), otherwise we'll be dancing around this forever, adding more API to cope with different use cases.

asmagill:
I can make it take a string value that is has to match one of the predefined levels:

~~~
NSNormalWindowLevel      0
NSFloatingWindowLevel    3
NSSubmenuWindowLevel     3
NSTornOffMenuWindowLevel 3
NSMainMenuWindowLevel    24
NSStatusWindowLevel      25
NSModalPanelWindowLevel  8
NSPopUpMenuWindowLevel   101
NSScreenSaverWindowLevel 1000
NSDockWindowLevel        20
~~~

Probably without Dock (it's deprecated) and Submenu (since TornOffMenu is preferred)... not sure about Floating... it's the same numeric value as Submenu, but in the docs, it isn't listed as deprecated or less preferred...  Any others I should leave out?

I'll leave in orderAbove and orderUnder, since they work within a specific level as defined above, and the existing bringToFront and sendToBack will probably become wrappers to a call to the new method.

asmagill:
I'm having second thoughts about using a string and limiting it to just those keys -- they're the ones in Dash for NSWindow's setLevel:, but the CoreGraphics list, converted with CGWindowLevelForKey is:

~~~
kCGAssistiveTechHighWindowLevelKey 1500
kCGBackstopMenuLevelKey            -20
kCGBaseWindowLevelKey              -2147483648
kCGCursorWindowLevelKey            2147483630
[...]

PR #531, opened 2015-08-31
hs.drawing:setLevel() and hs.drawing.windowLevels

asmagill:
addresses #524


asmagill:
objections?  Or can I go ahead and merge this?

PR #536, opened 2015-09-04
[READY] setup.lua, hs.inspect, hs.window

lowne:
~~The "terminate event processing" mention in the docs assumes the strategy I described in #477~~

~~EDIT: whitespace troubles again :/ 
proper diff here: https://github.com/Hammerspoon/hammerspoon/pull/536/files?w=1~~

See comment at the bottom.


cmsj:
Can I have some detail about what's driving this? How come we're reverting the previous PR and adding hs.assert()?

lowne:
Yeah, sorry for the the utter lack of context...
- Reverting due to discussion in #477 - the short of it is, `hs.showError` is meant as the "after-the-fact" handler for uncaught errors; as such, we should (almost) never call it directly (as I was previously doing, which prompted me to #473); instead we should simply `error`/`lua_error`/`luaL_error` and let it propagate upstream, structured exception handling-style.
- The use case for `hs.assert` is to catch early (i.e. at the point of failure) potentially very-hard-to-diagnose bugs for new code (untested in the "real world") where we know that something *should* be true, but we can't rule out it isn't if the incoming data is "crazy" in unpredicted ways; typical examples are make sure we consumed all the incoming data, make sure the work data is "proper" after transformation/parsing of user data, etc. If we *do* get issues, the (hopefully included) paste should show what edge case wasn't accounted for.

asmagill:
Just to play devils advocate re inspect (since it's automatic inclusion of metatables sometimes bugs me too), did you know that you can do this with the stock version:

hs.inspect(hs.window,{process=function(item,path) if path[#path] == inspect.METATABLE then return nil else return item end end})

I have this and a version with {depth=1} set up as a couple of personal shortcuts myself...

lowne:
Hrm, all I have to say about inspect is, I don't know what language that is (jk, it's ruby), but it sure ain't Lua. ([this](https://github.com/pkulchenko/serpent/blob/master/src/serpent.lua) is Lua, for comparison)

lowne:
On a related note (about the `hs.` prefixes), I have `setmetatable(_G,{__index=hs})` at the top of my config - a lifesaver in the console, but I keep forgetting the prefix in code snippets here on gh too :)

lowne:
1. the few uses (by me) of the "fancier" (and wrong) `hs.showError` are long gone in the codebase, so zero harm in reverting
2. also no harm in adding `hs.assert` - currently never used (because it wasn't there), but it could be handy
3. while it is indeed possible to use a custom filter for `hs.inspect`, I expect that 99% of the time users don't want metatables; moreover a regular (and documented) flag to control inclusion/exclusion is certainly easier (also, __tostring for userdata, yay!)
4. global-level "setLogLevel" and "getLogLevel" will need further improvement, but for now it's better than nothing (there were 1 or 2 issues asking for them)
5. somewhat related to 4, treating (most of) setup.lua as a regular "extension" aids in faster iteration, but mainly it might be useful to a couple of power-users out there, and I see no drawbacks
6. with even more submodules coming to the window namespace, I'm less and less comfortable with loading them all no matter what; the lazy-loading solution is a wee bit inelegant as it relies on the assumption that `hs.window` inherits from `hs.uielement`, but I don't see that changing anytime soon (or ever), and if it *did* change a whole lot of code elsewhere would need amending anyway

cmsj:
This has now been merged with master, manually.

PR #537, opened 2015-09-05
[WIP] hs.webview - show web pages as hammerspoon objects

asmagill:
Creates webview objects which can display web pages from remote web sites or local files:

e.g.

~~~lua
> w = require("hs.webview")
> a = w.new({x=100,y=100,h=500,w=500},"http://www.google.com"):show()
> a:allowTextEntry(true)
hs.webview: (0x7fa8b48176e8)
> a:asHSDrawing():setAlpha(.8) -- treat it as an hs.drawing to access it's methods
hs.drawing: unknown type (0x7fa8b4857c18)
> a:url("file:///Users/amagill/Sites/index.html") -- change the url
hs.webview: (0x7fa8b48176e8)
> a:windowStyle({"closable","resizable"}) -- make it closable and resizable
hs.webview: (0x7fa8b48176e8)
> a:windowStyle("titled") -- a single string means to toggle the window style attribute
hs.webview: (0x7fa8b48176e8)

> b = w.new({x=100,y=100,h=500,w=500},"http://www.google.com"):allowTextEntry(true):windowStyle({"titled","resizable"}):show()

~~~

screen shot 2015-09-04 at 9 40 10 pm

I moved them a bit to show them in one picture.  There is also an asHSWindow() method to get access to things like minimize, maximize, close, etc. as well.


cmsj:
very nice :)

do these webviews share (e.g. cookie stores) with safari?

asmagill:
Not at present, but I do remember reading somewhere that it is possible...  I'm planning to look into it.  Plus it doesn't really give any errors at present -- if a URL doesn't resolve/returns an error/is malformed/etc.  it simply doesn't change the webview... I'd like something a little more definitive...  Hopefully I can add these options soon.  I figure in its current state, however, it's nice for things like  formatted help, status pages, etc...

asmagill:
redoing with WKWebView... seems it's considered more modern and better supported under OS X.

asmagill:
For the interested, it is ready as a basic web page display module.

My plans for the module:

Things I plan to add:
* [ ] the ability to pass in raw html and have it rendered without requiring the use of a temp file
* [ ] URL list of webview's history (i.e. the forward and backward list for the current view)

Things I want to figure out:
* [ ] how to fully clear a webview's cache of all cookies, data, history, everything, preferably during run as well as __gc.
* [ ] can something like private mode browsing be included?
* [ ] where do things go when you right click on them and save the link (it's not Downloads)
* [ ] can we share settings/cookies with Safari?  Most of what I'm finding is from pre-sandboxing and I'm not sure this can be done anymore, but it'd be nice to know for certain
* [ ] since we currently can't "open in another window", etc. can these be removed from the context menu?

Things I might add if there is interest or inspiration hits me
* [ ] WKUIDelegate method for creating new webviews (open in another window)
* [ ] WKUIDelegate methods for javascript alert, confirm, and input
* [ ] WKNavigationDelegate method for authentication challenge
* [ ] WKNavigationDelegate methods for callback for tracking load state (e.g. navigation started, completed, error, etc.) -- probably one callback with a flag set
* [ ] WKNavigationDelegate methods for accept/cancel navigation policy -- maybe for completeness, or if there is interest

asmagill:
Ok, that's weird... somehow it thinks I'm changing 34 files... I must have mixed branches up in a commit or something... I'll pull this and resubmit a cleaner version.

PR #538, opened 2015-09-07
Add window title pattern matching to hs.layout.apply().

wsmith323:


cmsj:
Thanks!

cmsj:
I had to revert this, see #546 for the details.

wsmith323:
Were you able to reproduce the described errors? I can't. I have even tried to make string.match() fail in the console in the way described in the ticket and I can't make it happen.

cmsj:
@wsmith323 it should be pretty easy to reproduce situations where the window title to search for, contains the special Lua pattern characters:

```
> string.match("test[123.txt", "test[123.txt")
[string "return string.match("test[123.txt", "test[123..."]:1: malformed pattern (missing ']')
```

Even if one happens to have a balanced set of special characters, it still doesn't behave how we would want:

```
> string.match("test[123].txt", "test[123].txt")
nil
```

I suspect that our only option here, if we want to support patterns, would be to have the string start with a special string that shouldn't ever occur in a window title (maybe something like `__HS_PATTERN:`, but I think that's getting a bit ugly :/

cmsj:
(I should add, this is my fault, I should have caught this before landing the patch)

wsmith323:
Ah. I didn't think about the unbalanced case. Still, the problem is not with the window title, it is with the pattern we are using to match it. This doesn't fail: string.match("test[123.txt", "test%[123.txt"). Here, the special character is escaped in the pattern.

The real problem is the expectations of the user that what they put in the apply() argument structure for window title will be evaluated with '==' instead of string.match, as historically has been the case. Those that have weird stuff in their window title arguments _may_ have to tweak those values to continue with the enhanced capability. However, I contend that it is most likely those users that have the use cases that this capability was intended to address.

Now, I understand and agree with your decision to revert my initial patch. It broke backwards compatibility. I also agree that adding some sort of marker to the window title arguments is ugly.

Perhaps we could add an optional boolean argument to apply() that changes the way window titles are compared. By default, they would be compared with '==' as usual.

Thoughts?

PR #540, opened 2015-09-08
[READY] fixes and improvements, preparation for hs.window.layout

lowne:
- prepare for `hs.window.layout` and `hs.window.tiling` (empty modules for now)
- add `hs.geometry.unitrect()` constructor
- add `"[X%,Y%,W%xH%]"` constructor notation for hs.geometry unitrects
- add `.table`, `.string` fields for hs.geometry objects, for serialization/logging respectively
- allow chaining `hs.geometry.setFIELDNAME()` calls
- remove __len metamethod for hs.geometry.objects (ambiguous notation)
- add `.x2y2` field for hs.geometry objects
- add `hs.geometry:floor()` method
- fixes for `hs.geometry:intersect()`, `:union()` and `:vector()`
- `hs.window:setFrame()` clips the rect to integer values (could this be problematic in some way?)
- fix `hs.screen:fromUnitRect()`
- `hs.window.filter(...)` is a shortcut for `hs.window.filter.new(...):getWindows()`
- tons of fixes and internal improvements for `hs.window.filter`

While all this is the result of ongoing work on hs.window.layout, it contains some fixes that shouldn't wait for it.


lowne:
As mentioned above, I math.floor the rect passed to window:setFrame to integer coordinates - could this cause problems? i.e. any reason to have actual *pixel* precision (with x=XXX.5) for window frames in HiDPI screens?

cmsj:
@lowne what's the motivation for floor()ing the rect there?

lowne:
@cmsj because window.tiling would set fractional frames causing a 're-tiling loop' in window.layout; and
- I assume there's no reason for fractional values
- I assume the AX layer ignores them anyway (not sure about this tho)
- hs.grid might also set fractional values
- I needed a fix for the loop
therefore I fixed the issue at its root. 

---

I just checked on the retina display and it seems that even though window/internal.m passes around floats, AX only deals in integer points anyway:

```
> w:_setFrame{x=5100,y=400,w=500,h=599.5}
hs.window: Hammerspoon Console (0x61800145eec8)

> w:_frame()
hs.geometry.rect(5100.0,400.0,500.0,600.0)
```

PR #542, opened 2015-09-09
WIP - hs.webview

asmagill:
Hopefully this will be cleaner -- I shouldn't have 34 files out of sync with master, I should only have 3!

At any rate, the current take now supports navigation and magnification gestures, a simple error page for failed navigation, a navigation callback with an option to override the error page with your own or even suppress it and use hs.notify or hs.showNotification instead, javascript alert, confirm, and prompt now have the appropriate delegate in place, and basic/digest auth is supported (sort of -- see below) so it can be used with hs.httpserver's pages even if a password is set.

The alerts used are modal to the window, not to Hammerspoon, so they are non-blocking (yay!), but at present the credential prompting shows two prompts for login -- one for username and one for password... should be a simple matter to combine them at some point, but I was tired and it works -- sort of... the caveat is that it doesn't store a successful credential in the memory cache, so the next request also prompts again... and again... etc.

I plan to add a callback to the navigationPolicy delegate methods and maybe to the credential query as well -- have a lua function provide the username and password for a site using basic or digest auth instead of a prompt.

I also need to implement the delegate method for creating subviews... for frames and "other window"... and maybe see if we can override the context menu so that it doesn't show things we either can't (yet) or won't do... right now they just do nothing with a very non-expressive notification of nothing...

### DIFF
--- a/Hammerspoon.xcodeproj/project.pbxproj
+++ b/Hammerspoon.xcodeproj/project.pbxproj
@@ -276,6 +276,10 @@
 		D02F95291A00221C00E28BB2 /* Hammerspoon_Tests.m in Sources */ = {isa = PBXBuildFile; fileRef = D02F95281A00221C00E28BB2 /* Hammerspoon_Tests.m */; };
 		D02F95341A003A2900E28BB2 /* init.lua in Resources */ = {isa = PBXBuildFile; fileRef = D02F95331A003A2900E28BB2 /* init.lua */; };
 		D077B5BF1A001B5300369B30 /* variables.m in Sources */ = {isa = PBXBuildFile; fileRef = D077B5BE1A001B5300369B30 /* variables.m */; };
+		D6065AC81BA0105D006E67AA /* Cocoa.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 9445CA0E19083251002568BB /* Cocoa.framework */; };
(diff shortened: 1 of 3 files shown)


asmagill:
Ok it's late and I'm missing something... what's the conflict?

asmagill:
I give up... I'll try again tomorrow, but any thoughts or suggestions would be helpful because I'm not seeing it right now...

cmsj:
It's Hammerspoon.xcodeproj/project.pbxproj in the Copy Extensions shell script, which changed in master in c0f6c443b33b4f1ae738654642a60bbf62dff09b.

cmsj:
FWIW, if you install the github command line tool, this sort of thing is fairly easy to track down:

`hub checkout https://github.com/Hammerspoon/hammerspoon/pull/542` (grabs this PR into a local branch called `asmagill-webview2`)
`git checkout master` (return to current master)
`git checkout -b lolmerge` (create a fresh branch of master we can mess up)
`git merge asmagill-webview2` (merge in the PR branch, which will fail due to conflicts, and tell you which files are in conflict).

Then edit the files that have conflicts and search for `<<<`. Then go cross-eyed trying to work out what the heck has changed. That last part I don't have handy hints for ;)

asmagill:
nope... merges fine on my end.  I'm gonna clear my local copy (well, move it somewhere else) and start with a whole brand-new fork and see how that fares

asmagill:
closing, see #544 instead

Issue #515, opened 2015-08-23
Extension request: Add lsqlite3 to the default lua packages installed


ketralnis:
This makes a lot of settings and tasks easier. An official-ish lua module to do this [already exists](http://lua.sqlite.org/index.cgi/index), and OS X ships with sqlite3 already

cmsj:
I think this is an interesting idea - the question is whether Hammerspoon should be shipping sqlite bindings, or if folk should use luarocks to install it.

We recently had the same question in regards to the Moses lua library, but I think there are two differences here:

 1. lsqlite3 is a C library, so users would need to have Xcode installed to be able to luarocks install it
 2. being able to access system databases is more of a Hammerspoon related feature than doing arbitrary functional programming is

So, I'm leaning towards saying yes, but I'd like some more opinions (@asmagill ?)

asmagill:
Originally I didn't have a strong opinion one way or the other on this... however, I think I'm going to need to delve into sqlite for webview if we want to give it complete access to it's cookie cache... the stuff I'm reading is confusing at times and very IOS-centric with OS X tacked on at the end (except when it's just plain outdated), but I believe the recommended current implementation with sandboxing under WKWebView is heavy on the JS injection and sqlite, so... I say go ahead.

asmagill:
We're going to either require a different module or include our own sqlite3 library -- trying to use the module with the stock libsqlite3.dylib gives the following error:

~~~lua
> l = require("lsqlite3")
error loading module 'lsqlite3' from file '/usr/local/lib/lua/5.3/lsqlite3.so':
	dlopen(/usr/local/lib/lua/5.3/lsqlite3.so, 6): Symbol not found: _sqlite3_enable_load_extension
  Referenced from: /usr/local/lib/lua/5.3/lsqlite3.so
  Expected in: flat namespace
 in /usr/local/lib/lua/5.3/lsqlite3.so
~~~

Forcing the use of the brew version with this:
~~~lua
> package.loadlib("/usr/local/opt/sqlite/lib/libsqlite3.dylib","*")
true

> l = require("lsqlite3")
~~~

works, however.

Trying the loadlib with the path to the system library in /usr/lib gives the same error as the first attempt, so it's definitely something that's not included in the stock version (Note, this is under 10.11 GM... I'm going to assume that the 10.11 release will be the same, and I never tried this with earlier versions, so I can't say if this is new or not.

bradparks:
[...]

PR #544, opened 2015-09-09
WIP - hs.webview

asmagill:
brand new fork, brand new branch...
No idea what got corrupted/changed in my previous fork, but whatever -- this merged (finally).

I'll pull the comments from the other aborted pulls in a bit and stick them in here... you can also try out:

~~~lua
f = function(...) print((inspect(table.pack(...)):gsub("\n"," "))) end
w = hs.webview
a = w.new({x=100,y=100,h=500,w=500},{developerExtrasEnabled=true}):windowStyle(1+2+4+8):allowTextEntry(true):url("https://www.google.com"):navigationCallback(f):allowGestures(true):show()
~~~


asmagill:
From #542 

At any rate, the current take now supports navigation and magnification gestures, a simple error page for failed navigation, a navigation callback with an option to override the error page with your own or even suppress it and use hs.notify or hs.showNotification instead, javascript alert, confirm, and prompt now have the appropriate delegate in place, and basic/digest auth is supported (sort of -- see below) so it can be used with hs.httpserver's pages even if a password is set.

The alerts used are modal to the window, not to Hammerspoon, so they are non-blocking (yay!), but at present the credential prompting shows two prompts for login -- one for username and one for password... should be a simple matter to combine them at some point, but I was tired and it works -- sort of... the caveat is that it doesn't store a successful credential in the memory cache, so the next request also prompts again... and again... etc.

I plan to add a callback to the navigationPolicy delegate methods and maybe to the credential query as well -- have a lua function provide the username and password for a site using basic or digest auth instead of a prompt.

I also need to implement the delegate method for creating subviews... for frames and "other window"... and maybe see if we can override the context menu so that it doesn't show things we either can't (yet) or won't do... right now they just do nothing with a very non-expressive notification of nothing...

A self signed certificate which you have accepted in Safari will work... if you haven't, the code for implementing NSErrorAttemptingRecovery isn't in place, so it will currently show an error... If I can figure out how to use the recovery object provided, I'll add it, but for now, visit the page with the self-signed cert in Safari first.

asmagill:
Just had an idea... when this lands, it might be nice to include the Hammerspoon.docset in the Resources of a release since we will now have a visual tool for viewing the docs even if you don't have Dash... just a thought...

cmsj:
@asmagill that's a nice idea!

asmagill:
I'm thinking this is about as far as I can take it within the WKWebView and WKUserContentController classes... There are still some limitations/oddities, most notably (also in comments at the top of internal.m):

~~~
// TODO: or ideas for updates
//   single dialog for credentials
//   downloads, save - where? intercept/redirect?
//   cookies and cache?
//   handle self-signed ssl
//   can we adjust context menu?
//   can we choose native viewer over plugin if plugins enabled (e.g. not use Adobe for PDF)?
~~~

But it is fairly powerful/functional as-is and many of the sites I've come across with suggestions concerning these involve using JavaScript injection which works now.  JS isn't my strong suit, so getting outside eyes on it would be helpful as well.

Should I go ahead and merge the pull request?

cmsj:
I'll try and have a play with it today, but from what I've seen of the code/discussion so far, I think it's going to be ok for landing :)

asmagill:
Example of Docset viewing with simple search function at https://gist.github.com/asmagill/633c8515e6ace3335d31

Does require lsqlite3, third-party sqlite3 installation (see #515), and an uncompressed copy of the DocSet, and is very simplistic, but it works :-)

screen shot 2015-09-19 at 5 36 34 pm
screen shot 2015-09-19 at 5 38 40 pm

jhgg:
Now what would be really cool is to have a menu item that pops up a webkit view. So we could start making some custom webkit powered menus.

asmagill:
Is there a webkit pop-up menu type?  you could definitely use the hs.drawing:onClick method to pop up a hs.webview with no window decorations (all of my screen shots show the titlebar, etc, but that's because I explicitly add them with has.webview:windowStyle), but the highlighting and menu selection, etc would probably require JavaScript injection...

If there is a class already in existence that handles the "menu" part of this, I'd be happy add it... the extent of my comfort with JS is expressed in the example gist I give above... I avoid it myself when possible because each browser implements it just a *leetle* differently...  of course here we only have to worry about WebKit/WebKit2, so that would be simpler...

asmagill:
@cmsj, barring objections, I'm going to merge this tonight... I'd like to get it in before one of us modifies project.pbxproj again and makes this require a re-pull or a manual merge.

cmsj:
@asmagill go for it :)

Issue #499, opened 2015-08-16
hs.window:setFrame() does not guard against absurd values


gibiansky:
Hammerspoon crashes and somehow causes the OS to crash on `setFrame`, when the arguments to `setFrame` are absurd. For example, something like this:

``` lua
local rect = {
  h = 1152000000.0,
  w = 2359296000.0,
  x = 4718592000.0,
  y = 33120000.0
}
hs.window.focusedWindow():setFrame(rect)
```

gibiansky:
I actually can't get it to crash the OS quite as drastically with my test code, but instead on this code Hammerspoon takes up 100% CPU and 50% of my 16 GB of memory.

cmsj:
@gibiansky Out of interest, how did you come across it? Did you just decide to try absurd values, or did some kind of bug end up with those values being chosen for you?

I can see the argument behind guarding ourselves against absurd values, but my question then is, how do you define absurd. The total size of all active displays? 2x that? 4x that?

gibiansky:
Yes, it was a bug in my scaling code. My issue actually resulted in an OS crash somehow, I think caused by this. Was very strange.

Does a setFrame for a size larger than the display actually work? I thought it just clipped it to the maximum size. If that's right, then Hammerspoon should just do the clipping earlier.

lowne:
It's not so straightforward... in fact, it would introduce quite a lot of overhead for every call, since HS itself doesn't keep state (enumerate screens, get frames, compute union, compute intersection).
For reference, this is how you could do it (in the near future, hopefully, as both geometry and moses aren't merged yet) script-side when `frame` comes from "dangerous" calculations upstream:

```lua
w:setFrame(hs.moses.reduce(hs.screen.allScreens(),function(s,v)return s:union(v:frame())end,hs.geometry('0,0 0x0')):intersect(frame))
```

lowne:
Should be addressed by using `:setFrameInScreenBounds()`, merged in #563

cmsj:
(I'm going to close this out because I don't think there is any further work to be done in Hammerspoon. Please re-open if you disagree!)

PR #545, opened 2015-09-09
[READY] small improvements for hs.window

lowne:
fix docs; add ensureInScreenBounds arg to :move(); alias :setFrameInScreenBounds() for :ensureIsInScreenBounds(); add optional frame arg to :setFrameInScreenBounds()

Handles cases like #499 (use `:setFrameInScreenBounds()` instead of `:setFrame()` to make sure things don't explode)

### DIFF
--- a/extensions/window/init.lua
+++ b/extensions/window/init.lua
@@ -158,10 +158,8 @@ end
 
 
 local animations, animTimer = {}
-
-
-local function animate()
-  --[[
+local DISTANT_FUTURE=315360000 -- 10 years (roughly)
+--[[
   local function quad(x,s,len)
     local l=max(0,min(2,(x-s)*2/len))
     if l<1 then return l*l/2
@@ -171,10 +169,11 @@ local function animate()
     end
   end
 --]]
-  local function quadOut(x,s,len)
-    local l=1-max(0,min(1,(x-s)/len))
-    return 1-l*l
-  end
+local function quadOut(x,s,len)
+  local l=1-max(0,min(1,(x-s)/len))
+  return 1-l*l
+end
+local function animate()
   local time = timer.secondsSinceEpoch()
   for id,anim in pairs(animations) do
     local r = quadOut(time,anim.time,anim.duration)
@@ -189,9 +188,10 @@ local function animate()
     end
     anim.window:_setFrame(f)
   end
-  if not next(animations) then animTimer:stop() end
+  if not next(animations) then animTimer:setNextTrigger(DISTANT_FUTURE) end
 end
-animTimer = timer.new(0.017, animate)
+animTimer = timer.new(0.017,animate)
+animTimer:start() --keep this split
 
 
 local function getAnimationFrame(win)
@@ -204,7 +204,7 @@ local function stopAnimation(win,snap,id)
   local anim = animations[id]
   if not anim then return end
   animations[id] = nil
-  if not next(animations) then animTimer:stop() end
+  if not next(animations) then animTimer:setNextTrigger(DISTANT_FUTURE) end
   if snap then win:_setFrame(anim.endFrame) end
 end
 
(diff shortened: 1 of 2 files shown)


extensions/window/init.lua
```diff
@@ -682,21 +681,28 @@ end
 ---    - an `hs.geometry` unit rect, or argument to construct one; will set the window frame to this rect relative to the desired screen;
 ---      if `screen` is nil or omitted, use the screen the window is currently on
 ---  * screen - (optional) An `hs.screen` object or argument for `hs.screen.find`; only valid if `rect` is a unit rect
+---  * ensureInScreenBounds - (optional) if `true`, use `setFrameInScreenBounds()` to ensure the resulting window frame is fully contained within
+---    the window's screen; `duration` is ignored (animation is not supported)
 ---  * duration - (optional) A number containing the number of seconds to animate the transition. Defaults to the value of `hs.window.animationDuration`
 ---
 --- Returns:
 ---  * The `hs.window` object
 function window:move(rect,toScreen,duration)
```
cmsj: How come this doesn't need another parameter in the function definition?

extensions/window/init.lua
```diff
@@ -682,21 +681,28 @@ end
...
+  local frame,inBounds
+  if type(toScreen)=='boolean' or (type(toScreen)=='number' and toScreen<20 and duration==nil) then duration=toScreen toScreen=nil end
```
cmsj: What's the purpose of the <20 here? Looks like it's a limit on the number of seconds for duration?

extensions/window/init.lua
lowne: Because it's either one or the other; if ensureInScreenBounds, duration is always 0

extensions/window/init.lua
lowne: As a side effect, yes; it's a crude way to disambiguate between a screen id and a duration.
- `w:move('[0,0,50,50]',5123423)` movesToUnitRect on that screen id, if found (otherwise, current screen)
- `w:move('[0,0,50,50]',10)` slooooowly movesToUnitRect on the window's current screen
- `w:move('[0,0,50,50]',634635,10)` slooooowly movesToUnitRect on that screen id

I'm ofc assuming that screen IDs can't be 0<x<20; that seems pretty unlikely

lowne: Perhaps it should be documented that this method can't handle durations>20s? Come to think of it, kiosks and similar might want to? (but even then, it seems just a bad idea)

cmsj: I'm going to firmly -1 this <20 thing. It says to me that we have gone too far down the road of accepting arguments in multiple formats. If we need an hs.screen object, the user should supply an hs.screen object. Trying to do clever dances to save them typing hs.screen.find(blah) is causing us more hassle than it's worth, IMO.

lowne: I had similar worries myself, so in this instance I agree; let's see if it's possible to find a middle-of-the-road way?

lowne:
With the latest commit, `:setFrameInScreenBounds()` behaves thusly:
- if animation is requested, but the window only allows discrete resize (Terminal), or can't be resized at all (e.g. dialogs), or is already shrunk to the minimum size, do not animate
- otherwise:
  - if the window is along a screen edge, momentarily move it off the edge (to prevent the OSX "stickiness" thing)
  - momentarily try resize it to the requested size, to see if it's viable
  - if the requested size is too small, change it to the smallest size
  - restore the window to its original frame; **these necessary "tests" will cause a brief "flicker" of the window** - is it acceptable, in exchange for animation support?
  - start the animation to the requested frame (adjusted within destination screen bounds)

- `:move()` is changed to allow both `inScreenBounds` and `duration`
- when calling `:move()` with a unit rect and a duration, `screen` isn't optional

Issue #548, opened 2015-09-12
LuaSkin protectedCallAndTraceback


asmagill:
I've noticed that whenever I use this, the stack pointer increases by the number of returned objects expected + 1... looking at the docs suggests that lua_pcall pops the function and the arguments, but does not pop the message handler -- I've not confirmed this in the lua source yet, but that's the way it reads to me.

Using lua_pcall in the same place with a msgh of 0 doesn't have this problem, supporting the theory.

We should either document this or remove it before returning... I'm leaning towards the removal myself.

cmsj:
+1 removal, I think it's super important that wherever possible, the LuaSkin functions don't leave unexpected things on the Lua stack - we can't necessarily predict what they will end up being used for, so they should do their utmost to be tidy!

Thanks for spotting this, I hadn't realised that the message handler doesn't get popped.

asmagill:
Agreed.  In looking at the examples of using lua_pcall in my books and other references, i notice that they always push it once but reference it's specific positive index number multiple times, so I suppose in retrospect, it should have been obvious to me that it wasn't getting popped, but I do wish the documentation were more clear on it.  At any rate, it's addressed now.

PR #549, opened 2015-09-12
Luaskin enhancements

asmagill:
##### Conversion from NSObjects into Lua objects
* `- (int)pushNSObject:(id)obj ;`
  converts an NSObject into the closest lua equivalent. Built-in convertors exist for:

  From               | To
  ------------------ | ---------
  NSString           | string
  NSData             | string
  NSNumber           | number or integer
  NSDictionary       | table
  NSArray            | array (note NSArray indexes start at 0, lua arrays at 1.  This preserves the array concept, so numerically the indexes shift by 1, but their position and use in loop structures does not.)
  NSSet              | array
  NSNull (nil, NULL) | nil
  NSObject           | string -- this is the fallback if no other match is found and converts the object into a string from the objects description method (what NSLog would output for %@)
  Other converters can be registered by modules and they will be checked before the above.

* `(int)pushNSObject:(id)obj preserveBitsInNSNumber:(BOOL)bitsFlag ;`
  Variant which optionally preserves bits rather than closest signed equivalent of numeric value for an NSNumber of type unsigned long long (the above sets bitsFlag to NO)

* `- (void)registerPushNSHelper:(pushNSHelperFunction)helperFN forClass:(char *)className ;`
  Register a special helper for a specific class.  The function should match this signature: `int function(lua_State *L, id obj)`. Registering the same className twice will overwrite the existing helper function -- specify nil as the function to remove the existing helper function.

* `- (int)pushNSRect:(NSRect)theRect ;`

* `- (int)pushNSPoint:(NSPoint)thePoint ;`

* `- (int)pushNSSize:(NSSize)theSize ;`
  NSRect (CGRect), NSPoint (CGPoint), and NSSize (CGSize) are actually structs, but common enough for their own converters.

##### Conversion from lua objects into NSObjects
* `- (id)toNSObjectAtIndex:(int)idx ;`
  Converts the item at the specified stack index from Lua into it's closest equivalent NSObject.  Built in converters exist for:

  From               | To
  ------------------ | ---------
  table  | NSDictionary or NSArray if all indexes are numeric and the number of items in the table == the largest key (index) in the table -- again, this preserves the array concept at the expense of sparse arrays which are challenging in Lua in their own right to iterate.
  number | NSNumber as float or integer
  string | NSString or NSData if the string cannot be fully expressed as valid UTF8
  nil    | NSNull
  other  | userdata, thread, function, etc. get convrted to NSString as their `tostring()` output

* `- (id)toNSObjectAtIndex:(int)idx allowSelfReference:(BOOL)allow ;`
  Variant of the above which optionally allows you to allow self-reference in tables being converted.  The above calls this with allowSelfReference equal to NO.

* `- (id)tableAtIndex:(int)idx toClass:(char *)className ;`
  Converts a table into an instance of the specified NSObject class.  Requires a helper to be registered first -- there are none pre-defined at present.  This has to be handled independant of the above converter because to lua, one table is like another and there is no way to tell if it should be considered as an NSDictionary or as the fields for an NSParagraphStyle (see https://github.com/asmagill/hammerspoon_text as an example of its use, which may one day be submitted for consideration as well)

* `- (void)registerTableHelper:(tableHelperFunction)helperFN forClass:(char *)className ;`
  Register a helper function for the above converter.  The function should match this signature: `id function(lua_State *L, int idx)`.  Registering the same className twice will overwrite the existing helper function -- specify nil as the function to remove the existing helper function.

* `- (NSRect)tableToRectAtIndex:(int)idx ;`

* `- (NSPoint)tableToPointAtIndex:(int)idx ;`

* `- (NSSize)tableToSizeAtIndex:(int)idx ;`
  Like the pushNSRect, Point, and Size methods above, these return the appropriate structures.  Missing fields are set to 0.0.

##### Other helpers

* `- (BOOL)isValidUTF8AtIndex:(int)idx ;`
  Returns YES or NO indicating if the item at the specified index can be converted to a valid UTF8 string.  Like `lua_tostring()` and `lua_tolstring()`, it will convert any number on the stack into a string, so don't use this in a loop being traversed with `lua_next()`

* `- (BOOL)requireModule:(char *)moduleName ;`
  Equivalent to lua's `require("moduleName") ;` -- Attempts to load the specified module, if not already loaded, and push its value (usually a table) onto the top of the stack.  If it fails, it will push a string containing the error instead.  Returns YES on success or NO on error.  This is useful for ensuring a module is loaded or for referencing a function in another module without relying on it existing within the global name space at a specific location.

Note that regular table conversions to and from NSDictionary, NSArray, and NSSet properly handle references to the same sub-tables (i.e. without duplication) and can handle self reference (i.e. a table which has itself as an element).  Self reference has to be requested for lua->NSObject conversion because this is not usually checked for in the Objective-C side of things -- NSLog, NSJSONSerialization, and NSUserDefaults all crash if you pass a self-referencing dictionary into them.    For NSObject->lua conversion, this is not quite as big of a concern because (a) it doesn't happen as often in values returned from the Mac OS APIs, and (b) lua crashes are less catastrophic to the Hammerspoon application.

### DIFF
--- a/LuaSkin/LuaSkin/Skin.h
+++ b/LuaSkin/LuaSkin/Skin.h
@@ -190,6 +190,8 @@ typedef struct tableHelpers {
  */
 - (void)checkArgs:(int)firstArg, ...;
 
+#pragma mark - Conversion from NSObjects into Lua objects
+
 /** Pushes an NSObject to the lua stack
 
  @note This method takes an NSObject and checks its class against registered classes and then against the built in defaults
@@ -244,6 +246,8 @@ typedef struct tableHelpers {
  @returns The number of items on the lua stack - this is always 1 but is returned to simplify its use in Hammerspoon modules.
  */
 - (int)pushNSSize:(NSSize)theSize ;
+
+#pragma mark - Conversion from lua objects into NSObjects
 
 /** Return an NSObject containing the best representation of the lua data structure at the specified index.
 
@@ -310,13 +314,24 @@ typedef struct tableHelpers {
  */
 - (NSSize)tableToSizeAtIndex:(int)idx ;
 
+#pragma mark - Other helpers
+
 /** Determines if the string in the lua stack is valid UTF8 or not.
 
  @note This method is used internally to determine if a string should be treated as an NSString or an NSData object.  It is included as a public method because it has uses outside of this as well.
+ @note This method uses lua_tolstring, which will convert a number on the stack to a string.  As described in the Lua documentation, this will causes problems if you're using lua_next with the same index location.
  @param idx - the index on lua stack which contains the string to check.
- @returns YES if the string can be treated as a valid UTF8 string of characters or NO if it cannot.
+ @returns YES if the string can be treated as a valid UTF8 string of characters or NO if it is not a string or if it contains invalid UTF8 byte sequences.
  */
 - (BOOL)isValidUTF8AtIndex:(int)idx ;
+
+/** Loads a module and places its return value (usually a table of functions) on the stack.
+
+ @note This method performs the equivalent of the lua command `require(...)` and places the return value (usually a table of functions) on the stack, or an error string on the stack if it was unable to load the specified module.
+ @param moduleName - the name of the module to load.
+ @returns YES if the module loaded successfully or NO if it does not.
+ */
+- (BOOL)requireModule:(char *)moduleName ;
 
 @end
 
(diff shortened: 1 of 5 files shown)


asmagill:
Replaces #512.  I've been using this with only minor tweaks for almost two weeks now with no problems, so barring anyone having issues with it, I think it's ready for inclusion. However, I'll wait for review and comments.

asmagill:
addressed #548

cmsj:
@asmagill looks great to me, +1 landing it :)

PR #550, opened 2015-09-13
[READY] hs.window.layout (needs #563)

lowne:
###### Removed
- removed `hs.window.filter:trackSpaces()`; it's now on a per-app-filter basis
- removed `windowShown` event; it's now `windowVisible`

###### Changed
- add event parameter to `:notify()` and `:subscribe()` callbacks
- allow passing {event1=fn1,...} maps to `:subscribe()` and `:unsubscribe()`
- add `windowVisible` (old `windowShown`) event
- add `windowNotVisible` (old `windowHidden`) event
- changed `windowHidden` event; it's only emitted when an app is hidden (cmd-h)

###### Added
- add `currentSpace` field to per-app filters, for tracking Spaces
- add `allowRegions` and `rejectRegions` fields to per-app filters
- add `allowScreens` and `rejectScreens` fields to per-app filters
- add convenience methods `:setCurrentSpace()`, `:setRegions()`, `:setScreens()`
- add `windowUnhidden`; emitted when an app is unhidden (cmd-h)
- add `windowInCurrentSpace`, `windowNotInCurrentSpace` events
- add `windowOnScreen`, `windowNotOnScreen` events
- add `:setSortOrder()` (for `:getWindows()` and `:notify()`) and relative constants (`focusedLast`, `focused`, `createdLast`, `created`)
- add optional sort order parameter to `:getWindows()`
- add __tostring metamethod

###### Fixed
- emit all relevant events on `windowCreated`
- disallow extraneous fields on `:setAppFilter()`
- `override` filter doesn't immediately return on allow (run app filters and `default` filter as well)
- improved performance of `:isWindowAllowed()`: use already-stored window properties for active instances instead of re-querying AX
- fixed `.copy()`
- preserve relative window ordering when (re)starting global watcher
- fixed passing maps to `:subscribe()`, `:unsubscribe()`

### hs.window.tiling
- new module with a single function `.tileWindows()` (unsurprisingly, it tiles/fits windows into a given rect)

### hs.window.layout
- new module

### DIFF
--- a/extensions/geometry/init.lua
+++ b/extensions/geometry/init.lua
@@ -55,7 +55,7 @@ local function gettype(t)
   if t._x and t._y then
     if t._w and t._h then
       for _,k in ipairs{'_x','_y','_w','_h'} do
-        if t[k]>1 or t[k]<0 then return 'rect' end
+        if t[k]>1.0000000000001 or t[k]<-0.0000000000001 then return 'rect' end
       end
       return 'unitrect'
     else return 'point' end
--- a/extensions/geometry/init.lua
+++ b/extensions/geometry/init.lua
@@ -129,7 +129,7 @@ local function maketable(t,x,y,w,h)
 end
 
 local function new(x,y,w,h)
-  if getmetatable(x)==geometry then return x end -- disable copy-on-new
+  if y==nil and getmetatable(x)==geometry then return x end -- disable copy-on-new
   local t=maketable({},x,y,w,h)
   for _,a in ipairs{x,y--[[,w,h--]]} do
     if a~=nil and (not t._x or not t._y or not t._w or not t._h) then t=maketable(t,parsearg(a)) end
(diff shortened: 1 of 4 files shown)


lowne:
> suggestions welcome for event names etc. to disambiguate between
1. a window that exists somewhere with a proper frame etc (i.e. not minimized and not hidden)
2. like 1, but the window is *currently* user-visible, i.e. somewhere on the screen(s), i.e. in the "current Space"
(in other words, case 1 includes windows in other Spaces)
I have windowVisible vs windowOnScreen but they seem just synonyms that would both apply to case 2
(windowVisible because it is inherited from hs.window:isVisible(), which is true for windows in any Space)

lowne:
"restore" and "guessLayout" still not implemented, but I think it's better to get it out there and get feedback for ux problems/bugs in the basic functionality now, before it gets even more complicated.

PR #551, opened 2015-09-14
Add custom window title comparison to hs.layout.apply().

wsmith323:


cmsj:
Thanks for re-visiting this. How about we go all out and make that optional argument be called something like `windowTitleComparator` and have it accept a function. Said function should accept two string arguments and return true if they match, false if not.

If the argument is omitted we supply our own function that just uses ==, otherwise we use the function the user supplied. This would let people supply `string.match` or any other thing of their choice.
(and we should suggest using `string.match` as an example)

wsmith323:
That is a great idea. I'll implement it.

wsmith323:
Done.

wsmith323:
Done.

wsmith323:
Done.

cmsj:
Thanks!

Issue #217, opened 2015-02-16
Resize up not working when window bottom near screen bottom


etienned:
I'm new to Hammerspoon and when doing the tutorial I found that when the bottom of a window is near the bottom of the screen I'm no longer able to resize the window vertically. For example if I hit `cmd-alt-ctrl-Left`, so the window is size to half screen on the left and then I press`cmd-alt-ctrl-Up` to trigger this function:

```lua
hs.hotkey.bind({"cmd", "alt", "ctrl"}, "Up", function()
  local win = hs.window.focusedWindow()
  local f = win:frame()
  local s = win:size()

  s.h = s.h - 10
      
  win:setSize(s)
end)
```

nothing happen. This function works correctly if the bottom of the window is at least at 18px of the screen border. I have the same problem if I use `win:setFrame(f)` instead of `win:setSize(s)`.

I also saw that if I change this function to resize 1px at the time (`s.h = s.h - 1`) the function stop working at around 26px from the screen bottom (instead of 18px). Also with resize set to 1px I'm not able to resize terminal windows.

I'm on Yosemite 10.10.1 with Hammerspoon 0.9.21.

cmsj:
I can reproduce this. I noticed that if I try to resize the window by hand, I have to drag a little way before it actually resizes. It seems like there's a stickiness to windows when they're at the bottom of the screen.

You could check if the bottom edge of the window is within that sticky zone at of the screen, and then adjust the amount you resize, to get it out of the sticky zone (optionally then resizing it back down to the point you actually wanted, if you don't mind the window looking like it's jumping around).

I have a horrible feeling that this is going to be very hard to work around in the code, because I'm not aware of that sticky area being at all optional :(

etienned:
You're right, there's stickiness to windows, but this stickiness is present on the menu side (top) and on the dock side (so it depends where's the dock). Honestly, I find this "feature" a bit a strange. I can't see the advantage of this stickiness that works only when sizing down/moving away from the side? It's not like a snap to edge.

Anyway, we'll have definitely to workaround this non-optional "feature". I found that the sticky zone is 24px large. So, if I use steps of 24px or greater, it works. When using the `win:setFrame(f)` method, the first move will jump, then after you will have the animation.

Concerning the Terminal.app case, I found that the minimum step size is in relation with the terminal font size. For example:
- font size 9pt -> 7px step
- font size 14pt -> 10px step
- font size 72pt -> 46px step

[...]

PR #563, opened 2015-09-24
[READY] fix screen edge stickiness, hs.geometry fixes, add window:sendToBack()

lowne:
### hs.geometry
- add `:fit(bounds)`
- ugly workaround for a precision error where `print(t.w) --> 1.0` but `t.w>1.0 --> true`; `.type()` will now correctly recognise these as unit rects
- fix `.string` for size objects
- fix `geometry.new(point,size) --> rect`

### hs.window
- add `:sendToBack()`
- improve `:setFrameInScreenBounds()`
- don't crash `.find()` if `:title()` is nil

### hs.application
- don't crash `.find()` if `:name()` is nil

### hs.screen
- don't crash `.find()` if `:name()` is nil

-----

As per my last comment in #217, I modified `setFrameInScreenBounds()` to also take into account (in addition to the likes of Terminal windows) edge stickiness. Same tradeoff as the previous incarnation - namely, if the window is potentially trouble, it will first jump around a little, regardless of whether it will then animate or not.

~~A case can be made to extend these safety precautions (minus the actual "ensure is in screen bounds part", of course) also to plain `:setFrame()`, however I haven't done so (yet); the question is **is the correct (or at least, expected) behaviour of setFrame in these edge cases worth the occasional annoying window dance?** Or shall we just inform the users and let them decide whether to use setFrameInScreenBounds() when dealing with these situations? (if so, I'll just add the relevant bits to the docstring and mark this PR as ready)~~ (added docs, see below)

(Also, following the 'fixes' theme, I snuck in some fixes for hs.geometry)

### DIFF
--- a/extensions/application/init.lua
+++ b/extensions/application/init.lua
@@ -112,7 +112,7 @@ function application.find(hint,exact)
   local apps=application.runningApplications()
 
   if exact then for _,a in ipairs(apps) do if a:name()==hint then r[#r+1]=a end end
-  else for _,a in ipairs(apps) do if a:name():lower():find(hint:lower()) then r[#r+1]=a end end end
+  else for _,a in ipairs(apps) do local aname=a:name() if aname and aname:lower():find(hint:lower()) then r[#r+1]=a end end end
   tsort(r,function(a,b)return a:kind()>b:kind()end) -- gui apps first
   if exact or #r>0 then return tunpack(r) end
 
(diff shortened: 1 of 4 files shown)


lowne:
So I decided to just mention the issues in the docstrings and let the user deal with it. :sunglasses: 
(and added window:sendToBack())

lowne:
Also added `geometry:fit()`

PR #567, opened 2015-09-26
non-activating panel instead of windows for drawing holders

asmagill:
Suggestion for hs.drawing which removes the bringing forward of the Hammerspoon console when a drawing object with a callback is clicked on.


cmsj:
lgtm, works in my testing :)

Issue #570, opened 2015-09-30
0.9.41 - > hs.mouse.getCurrentScreen() fails


bradparks:
Hey there! I just upgraded to Hammerspoon 0.9.41, and immediately noticed that the following method seems to be failing now

`hs.mouse.getCurrentScreen()`

It gives the following error when run via an extension, or run in the console:

```
> hs.mouse.getCurrentScreen()
-- Loading extension: mouse
...n.app/Contents/Resources/extensions/hs/geometry/init.lua:573: attempt to compare nil with number
```

I tried it on a system with both a single monitor (Macbook Air) and one with 2 additional external monitors and got the same result. Is there a better way to report a dump or bug?

cmsj:
damn!

cmsj:
(I mean, thanks, good catch. I'm just annoyed that we released with a silly bug :)

bradparks:
hey... no worries! I'm loving Hammerspoon so I'm very understanding.... thanks again! and I'm sure you've found it, but it looks to be at this line to me:

https://github.com/Hammerspoon/hammerspoon/blob/ff5d964dff657f75b5aeba524816b65bd455f315/extensions/geometry/init.lua#L573

with the `x2` and `y2` properties, i think....

cmsj:
@lowne looks like the parameters might be reversed here? Although I can't get it to return a value even if I do that.

lowne:
Whops. My bad, sorry everyone :/
#571 for the fix if urgent. I'll try to audit the rest of geometry to see if that commit (12ab412) had other unintended consequences.

lowne:
Unfortunately it appears there's no ready function for a temporary workaround 😭 - I'll add the possibility to pass arbitrary points and rects to screen.find (so that e.g. `hs.mouse.getCurrentScreen()` could be also done with `hs.screen.find(hs.mouse.getAbsolutePosition())`)

cmsj:
@lowne hmm, even with #571 applied, getCurrentScreen() still seems to fail on a nil/number comparison

bradparks:
@lowne - I'm finding the same as @cmsj, that it works in some situations, and not in others.... so this is still broken in some way at some times. I'm getting around it for the time being by using `hs.screen.primaryScreen()`....

lowne:
works fine here, can I have more details (and stack traces)?

bradparks:
[...]

PR #574, opened 2015-10-01
[READY] hs.screen and hs.window

### LINKED ISSUE #577: hs.gris.set glitchy on 0.9.41 (was ok on 0.9.40)
https://youtu.be/NnLkjDFm4-c

As can be seen in the video, my test config is fairly simple. When moving the window around, the window seem to resize more than once for a single hs.grid.set.

This occurs wether the dock is at the bottom or side. Same if it's set to auto-hide.

If I press mod+LEFT many times, the window still shrink a bit then grow to the proper position even though it's already in the right position in the first place.

This did not occur on 0.9.40

lowne:
- allow arbitrary rect in screen.find()
- add hs.window.setFrameCorrectness
- add hs.window:setFrameWithWorkarounds()
- fix crash when using "hide" in hs.window.layout

closes #577

### DIFF
--- a/extensions/application/init.lua
+++ b/extensions/application/init.lua
@@ -149,7 +149,7 @@ function application:getWindow(hint)
   return tpack(window.find(hint,true,self:allWindows()),nil)[1]
 end
 
---- hs.application.open(app[, wait]) -> hs.application object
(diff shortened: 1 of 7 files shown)


lowne:
Waiting on feedback on #570 before declaring this as ready.

Issue #579, opened 2015-10-02
Is it possible to move the hammerspoon configuration?


nfarrar:
I *really* hate clutting up $HOME with hundreds of files. I typically keep my application configurations in `$XDG_CONFIG_HOME` (`~/.config/<application>/`). Looked around and read through the docs, but didn't see a way to change this. Am I missing it? If it's not possible, any chance it might be in the future?

asmagill:
At present, no.

The location of MJConfigFile is hard coded in the source to "~/.hammerspoon/init.lua".

You could put the following at the top of ~/.hammerspoon/init.lua so that any *other* hammerspoon config files could be found in ~/.config/hammerspoon:

package.path = os.getenv("HOME").."/.config/hammerspoon/?.lua;"..os.getenv("HOME").."/.config/hammerspoon/?/init.lua;"..package.path
package.cpath =  os.getenv("HOME").."/.config/hammerspoon/?.so;"..package.cpath

@cmsj, while I've not considered this specifically for Hammerspoon before, I have noticed (and mostly like) a trend in many unix based things moving configuration files into ~/.config... do you object to checking a MJConfigFile entry in the application user defaults and defaulting to the current location if it's not present? Other than setting this var in variables.m, can you think of other places to check for impact?

asmagill:
@cmsj - re-reading this, it may not be clear -- I'm willing to do the work, and I know what needs to be done in the above specified file, just looking for pointers on other places it might affect (setting up the search paths in setup.lua I think is one... are there others that you can think of off the top of your head?)

nfarrar:
The reason for the widespread use of `~/.config/<application_name>` is because it's part of the [XDG Base Directory Specification](http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html). The `/.config/<application_name>` path is the *default* path used for user configuration files. It makes it *very* easy to version control all your dotfiles and sync them between machines.

cmsj:
@asmagill I think search paths should be it. I'm generally in favour of doing this, the question for me is purely about how we present the setting to users.

I think something we need to be very clear about is that the config location must be a directory, and we will unconditionally load `init.lua` from that directory. I think it would be a big mistake to let people point at arbitrary files, simply because then the search path stuff gets increasingly weird, and might interact with other files unintentionally.

So, I'd be ok with it hiding as an advanced option in our `Info.plist`, but I'd also be ok with there being a button in the Prefs window to select a directory to load the config from.

[...]

PR #582, opened 2015-10-03
allow user setable init file location

asmagill:
as suggested in #579, you can set MJConfigFile from the command line with:

~~~
defaults write org.hammerspoon.Hammerspoon MJConfigFile "~/.config/hammerspoon/init.lua"
~~~

@cmsj, I think I've left your test stuff ok by only doing the check for the user default if your XCTest in MJAppDelegate.m is false, but you should look at it to make sure.


asmagill:
Been running this with my files in ~/.config/hammerspoon/ for 4 days now and have seen no adverse affects.  Objections to merging?

molleweide:
worked perfectly!thank you very very much guys!

PR #585, opened 2015-10-05
add hs.grid.getCell to get the geometry rect of a grid cell

jfchevrette:
Implement hs.grid.getCell which returns the hs.geometry rect for the cell

### DIFF
--- a/extensions/grid/init.lua
+++ b/extensions/grid/init.lua
@@ -243,10 +243,10 @@ function grid.getCell(cell, scr)
   cell.w=max(1,min(cell.w,screengrid.w-cell.x)) cell.h=max(1,min(cell.h,screengrid.h-cell.y))
   local cellw, cellh = screenrect.w/screengrid.w, screenrect.h/screengrid.h
   local newframe = {
-    x = (cell.x * cellw) + screenrect.x + margins.w,
-    y = (cell.y * cellh) + screenrect.y + margins.h,
-    w = cell.w * cellw - (margins.w * 2),
-    h = cell.h * cellh - (margins.h * 2),
+    x = (cell.x * cellw) + screenrect.x,
+    y = (cell.y * cellh) + screenrect.y,
+    w = cell.w * cellw,
+    h = cell.h * cellh,
   }
   return newframe
 end
--- a/extensions/grid/init.lua
+++ b/extensions/grid/init.lua
@@ -222,6 +222,35 @@ function grid.get(win)
   }
 end
 
+--- hs.grid.getCell(cell, screen) -> hs.geometry
+--- Function
+--- Gets the `hs.geometry` rect for a cell on a particular screen
+---
+--- Parameters:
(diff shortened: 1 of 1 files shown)


cmsj:
Thanks. I think the docstring could do with some clarification - it's talking about "the window", but if I understand correctly, this isn't typically being used in conjunction with a particular window, it's just getting the rect for an arbitrary cell on a screen, right?

jfchevrette:
You are right. I updated the docstring in various places to remove refenrences to windows and also change the screen parameter to remove the (optional) tag.

extensions/grid/init.lua
```diff
@@ -222,6 +222,35 @@ function grid.get(win)
...
+---
+--- Parameters:
+---  * cell - a cell object, i.e. an `hs.geometry` rect or argument to construct one
+---  * screen - an `hs.screen` object or argument to `hs.screen.find()` where the cell is located
+---
+--- Returns:
+---  * the `hs.geometry` rect for a cell on a particular screen
+function grid.getCell(cell, scr)
+  scr=screen.find(scr)
+  if not scr then log.e('screen cannot be nil') return grid end
```
lowne: better `return end`

extensions/grid/init.lua
```diff
@@ -222,6 +222,35 @@ function grid.get(win)
...
+  cell=geom.new(cell)
+  local screenrect = scr:frame()
+  local screengrid = getGrid(scr)
+  -- sanitize, because why not
+  cell.x=max(0,min(cell.x,screengrid.w-1)) cell.y=max(0,min(cell.y,screengrid.h-1))
+  cell.w=max(1,min(cell.w,screengrid.w-cell.x)) cell.h=max(1,min(cell.h,screengrid.h-cell.y))
+  local cellw, cellh = screenrect.w/screengrid.w, screenrect.h/screengrid.h
+  local newframe = {
+    x = (cell.x * cellw) + screenrect.x + margins.w,
```
lowne: should probably remove margins here, as they are meaningful when talking about window frames but I don't they should apply to the "abstract" grid cell

extensions/grid/init.lua
```diff
@@ -222,6 +222,35 @@ function grid.get(win)
...
+--- hs.grid.getCell(cell, screen) -> hs.geometry
+--- Function
+--- Gets the `hs.geometry` rect for a cell on a particular screen
+---
+--- Parameters:
+---  * cell - a cell object, i.e. an `hs.geometry` rect or argument to construct one
+---  * screen - an `hs.screen` object or argument to `hs.screen.find()` where the cell is located
+---
+--- Returns:
+---  * the `hs.geometry` rect for a cell on a particular screen
```
lowne: ..., or nil if the screen isn't found

jfchevrette:
made the changes as suggested by @lowne.

This PR had gotten off my mind, sorry for the delay!

asmagill:
Is this still relevant/wanted?

cmsj:
@asmagill I think it is

Issue #576, opened 2015-10-01
hs.window.filter and AX roles for windows spawned by HS


lowne:
The premise: `hs.window.filter` should completely disregard container windows for `hs.drawing` and `hs.alert` objects; since at any given moment there could be tons of drawings (and, to a lesser extent, alerts) being created/destroyed (e.g. `hs.grid.show()`), this is for performance reasons, and to avoid cluttering the log.

This used to be the case until 0.9.40 - drawing windows, just like alerts, had a `:subrole()` of `AXUnknown` so it was easy to filter them out.

With recent changes (I suppose having to do with activating vs not activating HS) it seems that drawings now are `AXSystemDialog`, at least for rects (but I haven't checked all different the ways to create drawings). To complicate things further, `hs.webview` windows (which we do _not_ want to filter) can have varying subroles; what I have found so far is `AXUnknown` for `windowStyle(0)`, `AXDialog` when titled, and `AXFloatingWindow` for "utility".

So:
1. I assume these AX subroles are somehow determined by the OS, but just to be sure, is there no way to assign them directly?
2. If not, is there a way to have drawings behave as desired, but still being `AXUnknown`? Do alerts (which still are AXUnknown) exhibit the same problem? If not, can whatever method is used for them be also used for drawings?
3. If not, is it possible to find a stable way so that drawing containers are _always_, say, `AXSystemDialog`, and webviews are _never_ that?
4. Any other way to disambiguate drawings/alerts/webviews/HS Console/HS Preferences?

(To be clear, if we can get back to AXUnknown for everything, I'm ok with "sacrificing" title-less AXUnknown webviews since they should be "background", or at best transient, windows - in that case I'll just mention in the docs for windowfilter to use titled webviews if the user wants to be able to address them)

(ping @asmagill)

asmagill:
The AX roles and subroles are, as far as I can determine, set by the system and not directly changeable by us.  There are some patterns as shown in pull #566, though I can try and get a more formal list this weekend if you like.  I'll also dig a little deeper into NSPanel and see if we can identify them another way.

lowne:
[...]

PR #591, opened 2015-10-08
added some image functions

asmagill:
cleaned up some compiler warnings in hs.drawing and added some methods:

hs.drawing:rotateImage - rotate image counterclockwise around its center
hs.drawing:imageScaling - adjust how image is (or is not) scaled within its view
hs.drawing:imageAnimates - enable/disable animation of animated GIF
hs.drawing:imageFrame - set image frame type
hs.drawing:imageAlignment - how image is aligned with view
hs.drawing:clickCallbackActivating - set whether or not clicking on an image with a clickCallback will make Hammerspoon active.  Setting this to true changes the AX subrole which impacts the default window filters.  Defaults to false. See #576.

### DIFF
--- a/extensions/drawing/internal.m
+++ b/extensions/drawing/internal.m
@@ -75,7 +75,7 @@ NSMutableArray *drawingWindows;
         return nil;
     }
 
-    self = [super initWithContentRect:contentRect styleMask:NSBorderlessWindowMask | NSNonactivatingPanelMask
+    self = [super initWithContentRect:contentRect styleMask:NSBorderlessWindowMask
                                                     backing:NSBackingStoreBuffered defer:YES];
     if (self) {
         [self setDelegate:self];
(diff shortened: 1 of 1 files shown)


lowne:
@asmagill so if I understand correctly when I call :setClickCallback on a drawing its role becomes AXSystemDialog, and if I *also* call :clickCallbackActivating(true) then it reverts to AXUnknown - correct?

asmagill:
If clickCallbackActivating is not called for a drawing object, or is called with a value of true, then the subrole will be "AXUnknown".  If it is called with false, it will be "AXSystemDialog".

Whether or not the callback is actually set has no affect on this... I considered renaming the method because actually setting a callback is not necessary but as you describe in a previous comment, there may be rare cases where a non clickable interface element should still be identified separately from other drawings. I decided not to rename it because, for the most part, a user will only care about this flag when the drawing is clickable (i.e. has a function assigned) and will likely ignore it for all other cases.

lowne:
got it, thanks.

PR #593, opened 2015-10-08
Fix getColorWithDefaultFromStack. Closes #592

### LINKED ISSUE #592: hs.drawing:setFillGradient() doesn't work
Always seems to error with `attempt to index a number`.

I think I've tracked this down to `getColorWithDefaultFromStack()` in `hammerspoon.h`

cmsj:


cmsj:
@asmagill I'd appreciate it if you could take a quick look at this. I think the -1 was only working by accident for other methods (e.g. `setFillColor()`) because the colour happened to be the final argument, so would be the top of the stack.

cmsj:
(err, for context, I came across this by noticing that `setFillGradient()` always produces an error about trying to index a number, and its final argument is a number)

asmagill:
Your fix is correct; the -1 in the field queries was most likely a leftover from the original source of the function back when it was used in a single place and not in multiple places.

cmsj:
Thanks :)

PR #602, opened 2015-10-12
hs.drawing.color

asmagill:
Adding more color options, moving all color and image decoding/encoding to LuaSkin

prereq for full NSAttributedString support and ANSITerminal color support in hs.drawing.text (well, the color stuff is at any rate; the image stuff is for consistency)

### DIFF
--- a/Hammerspoon.xcodeproj/project.pbxproj
+++ b/Hammerspoon.xcodeproj/project.pbxproj
@@ -283,6 +283,11 @@
 		D64125651BA07D09000FB5F8 /* WebKit.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 9473A75C1959F9AB0080C329 /* WebKit.framework */; };
 		D64125661BA07D20000FB5F8 /* internal.m in Sources */ = {isa = PBXBuildFile; fileRef = D64125591BA07CDE000FB5F8 /* internal.m */; };
 		D643B73D1B8BFE7700B372EB /* libedit.dylib in Frameworks */ = {isa = PBXBuildFile; fileRef = D643B73C1B8BFE7700B372EB /* libedit.dylib */; };
+		D6461BDC1BCC32EC00E9A9B1 /* Fabric.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 4F0F97561ADE9F8600F2A199 /* Fabric.framework */; };
+		D6461BDD1BCC32EC00E9A9B1 /* Carbon.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 4FBE2D131B72635400407788 /* Carbon.framework */; };
+		D6461BDE1BCC32EC00E9A9B1 /* Cocoa.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 9445CA0E19083251002568BB /* Cocoa.framework */; };
+		D6461BDF1BCC32EC00E9A9B1 /* LuaSkin.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 4F9850E81B722CBF00F7E909 /* LuaSkin.framework */; };
+		D6461BE51BCC331D00E9A9B1 /* internal.m in Sources */ = {isa = PBXBuildFile; fileRef = D6461BD51BCC32B900E9A9B1 /* internal.m */; };
 		D64B03371BA954290006C468 /* webview.h in Headers */ = {isa = PBXBuildFile; fileRef = D64B03361BA9541E0006C468 /* webview.h */; };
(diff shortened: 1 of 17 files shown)


cmsj:
lgtm :)

asmagill:
Hmm... just noticed the merge conflicts... grr... oh well, will worry about it tomorrow -- I almost have `hs.drawing.styledtext` ready to add, so I'll include it before resolving the conflict so it all can be included together.

The new module already supports conversion form HTML and RTF to a styled text string (OS X provides this internally), and provides the framework where an ANSI Terminal sequence conversion tool could be written in Lua (I may do so at some point, but will probably release this first)

Questions:

hs.drawing.styledtext replaces/exceeds the following hs.drawing methods and I want to know how you think its best to proceed with them:

`hs.drawing.fontNames`, `hs.drawing.fontNamesWithTraits`, and `hs.drawing.fontTraits`
* these are basically moved to hs.drawing.styledtext and support a few more options but are fully backwards compatible... should I just remove them from hs.drawing completely or should they be a "shortcut" to the new versions in hs.drawing.styledtext?

`hs.drawing.getTextDrawingSize`
* the new version at `hs.drawing.styledtext.getTextDrawingSize` supports a slightly different format for the style table -- most notably the font and size attributes now make up a font subtable, and the paragraph style elements are in their own subtable as well -- this more closely matches how they are stored internally so it made the conversion tools more flexible.  Should this remain to support the old style table format or should we just remove the old style completely since the new one offers quite a bit more?  And, in either case, does it remain a shortcut to the newer version or should its path change as well?

`hs.drawing:setTextStyle`
* like hs.drawing.getTextDrawingSize, this uses an older format for the style array...

I guess it comes down to should I break backwards compatibility or force the two to live side by side?  Internally the old style would be converted to the new one, and certainly the more advanced features would only be supported by the new functions, but is there enough use of the old style functions to justify not breaking an outdated format?

cmsj:
@asmagill It's tough. I hate backwards compatibility stuff littered all over the place, but I also hate breaking working configs. I'm also increasingly conscious of the crash/download data suggesting to me that many users are going a long time between upgrades, so they won't see release note warnings about deprecations.

I wonder if we could have a generic mechanism that presents a nice error about the use of deprecated stuff, and then assign any old variable/function/etc to that mechanism, so referencing the deprecated thing, will halt the app and let the user know what they used, and what they should be using?

Bit complex, but every other option involves either endless complexity increases for us (which will cause regressions when we change something and forgot about its deprecated alias in another module), or confusing "why did my config start complaining about things being nil?" pain for users. Unless anyone else has bright ideas?

cmsj:
so let's not get sidetracked on the blue sky discussion about deprecation methods, now that we have a separate issue to discuss that in.

@asmagill how much work is backwards compatibility here? If it's trivial, let's do it, if not, let's rip the band-aid off and remove the old bits.

asmagill:
Been focusing on adding metamethods and replicating the lua string library so that you can treat styedtext objects as strings in most places (concatenation, substring manipulation, etc.) to make writing conversion tools in Lua (which is what I intend for the ANSI sequences) easier... should have something later this weekend for review...

As much as I would like to "clean up" hs.drawing to keep things better focused and with less sprawl, it's pretty minor to put a wrapper around methods like hs.drawing.setTextStyle so they can feed directly into hs.styledtext (I'm thinking its got enough power/potential in its own right with the HTML and RTF -- and maybe a few more file types -- conversions to be its own module now and not tied specifically to drawing).  It would be a one way kind of thing... if you manipulate the style with hs.styledtext, you can't apply it with the old methods, but the new ones can certainly inherit from the old.

Its just... inelegant 😁

lowne:
+1 to `hs.drawing.styledtext` to `hs.text`, that can feed into `hs.drawing` (and perhaps `hs.webview`?)

asmagill:
This pull will be closed in favor of one which contains these changes, #611, and hs.styledtext hopefully this weekend.  Adding ANSI color sequence support should then be a simple matter of lua string pattern matching then, and will follow shortly after.

I prefer the name hs.styledtext to hs.text because the latter sounds, to me, like it provides support for manipulating text in many ways, while this addition is specifically targeted at what NSAttributedString offers... however, I'm not wedded to it... I'm guessing @lowne favors hs.text... does anyone else have a preference/thought?

lowne:
@asmagill I'm fine with `hs.styledtext` - I prefer shorter names where possible but you have a point.

asmagill:
@lowne, the new pull will allow converting hs.styledtext into HTML which can then be fed into hs.webview... that seems cleaner to me than modifying webview to accept NSAttributedString.

hs.drawing is different in that it already deals with an object (NSTextView) which can take NSString or NSAttributedString, so it makes sense to ensure it can accept our (Hammerspoon's) representation of either (for almost all purposes, a Lua string is effectively equivalent to NSString except when it contains arbitrary binary (non UTF8) data).

asmagill:
closing this because it will never be used... #620 contains this

PR #607, opened 2015-10-15
Multiple additions to the Spotify module

zzamboni:
- Renamed `hs.spotify.play()` to `hs.spotify.playpause()` (because that is
  its behavior) and added a new `hs.spotify.play()` that only plays instead
  of toggling.
- Added new functions `getPlayerState()`, `isRunning()`, `isPlaying()`.
- Added new constants `hs.spotify.state_paused`, `hs.spotify.state_stopped`
  and `hs.spotify.state_playing` to use for checking the return values from
  `getPlayerState()`.

### DIFF
--- a/extensions/spotify/init.lua
+++ b/extensions/spotify/init.lua
@@ -7,6 +7,21 @@ local spotify = {}
 local alert = require "hs.alert"
 local as = require "hs.applescript"
 
+--- hs.spotify.state_paused
+--- Constant
+--- A string that, when returned by hs.spotify.getPlayerState() indicates Spotify is paused
+spotify.state_paused = "'kPSp'"
+
+--- hs.spotify.state_playing
+--- Constant
+--- A string that, when returned by hs.spotify.getPlayerState() indicates Spotify is currently playing a track
+spotify.state_playing = "'kPSP'"
+
+--- hs.spotify.state_stopped
+--- Constant
+--- A string that, when returned by hs.spotify.getPlayerState() indicates Spotify is stopped (initial state after application startup)
+spotify.state_stopped = "'kPSS'"
+
 -- Internal function to pass a command to Applescript.
 local function tell(cmd)
   local _cmd = 'tell application "Spotify" to ' .. cmd
@@ -18,7 +33,7 @@ local function tell(cmd)
   end
 end
 
---- hs.spotify.play()
+--- hs.spotify.playpause()
 --- Function
 --- Toggles play/pause of current spotify track
 ---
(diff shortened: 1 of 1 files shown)


extensions/spotify/init.lua
```diff
@@ -7,6 +7,21 @@ local spotify = {}
 local alert = require "hs.alert"
 local as = require "hs.applescript"
 
+--- hs.spotify.state_paused
+--- Constant
+--- A string that, when returned by hs.spotify.getPlayerState() indicates Spotify is paused
```
cmsj: Constants don't really need to specify their type, so this could just be "Returned by `hs.spotify.getPlayerState()` to indicate Spotify is paused" (note the use of backticks around the hs.spotify.getPlayerState() - we should really be doing that every time we refer to part of the API)

extensions/spotify/init.lua
```diff
@@ -125,4 +153,54 @@ function spotify.getCurrentTrack()
     return tell('name of the current track')
 end
 
+--- hs.spotify.getPlayerState()
```
cmsj: I think this might be better as `hs.spotify.getPlaybackState()` in case we want to add other state related things in the future. `Player` seems a bit too generic.

extensions/spotify/init.lua
```diff
@@ -125,4 +153,54 @@ function spotify.getCurrentTrack()
...
+--- Function
+--- Gets the current state of Spotify
+---
+--- Parameters:
+---  * None
+---
+--- Returns:
+---  * A string containing one of hs.spotify.state_stopped, hs.spotify.state_paused, hs.spotify.state_playing
```
cmsj: I tend to prefer lists to be done with Markdown, so:

 * A string containing one of the following constants:
  * `hs.spotify.state_stopped`
  * `hs.spotify.state_paused`
  * `hs.spotify.state_playing`

extensions/spotify/init.lua
```diff
@@ -27,10 +42,23 @@ end
...
 --- Returns:
 ---  * None
-function spotify.play()
+function spotify.playpause()
   tell('playpause')
 end
 
+--- hs.spotify.play()
+--- Function
+--- Play current spotify track
```
cmsj: Small detail, but the single line description should be more in the form of "Plays the current Spotify track"

extensions/spotify/init.lua
```diff
@@ -125,4 +153,54 @@ function spotify.getCurrentTrack()
...
+--- Function
+--- Returns whether Spotify is currently open. Most other functions in hs.spotify will automatically start the application, so this function can be used to guard against that.
+---
+--- Parameters:
+---  * None
+---
+--- Returns:
+---  * A boolean value indicating whether the Spotify application is running.
+function spotify.isRunning()
+   return hs.application.get("Spotify") ~= nil
```
cmsj: By convention (and not properly documented anywhere, so this one is absolutely my fault) we don't call other modules with absolute names - it relies on the automatic module loading, which users may have disabled. Instead, add a `local application = require "hs.application"` at the top of the file, and use `application.get(blah)`

zzamboni: Thanks, I was wondering about this, and also about calling functions within the same module (e.g. the call to `hs.spotify.isRunning()` from `isPlaying()`. Is the way I did it correct?

PR #611, opened 2015-10-17
Added X11 web colors

zzamboni:
Added the full list of X11 standard web colors from the CSS3
specification, as listed at https://en.wikipedia.org/wiki/Web_colors#X11_color_names.

The color names were made all lowercase for consistency.


zzamboni:
I just noticed the existence of #602, which also makes other extensive changes. If that is merged, the new colors in this PR could be added to the new `drawing/color/init.lua` file.

asmagill:
Ideally, colors should be kept in collections (groups) once #602 lands, rather than defined loose as hs.drawing.color.-colorname-.  Do you mind if I put them in as hs.drawing.color.x11.-colorname- (i.e. a collection named "x11")?  Or do you prefer a different collection name?

zzamboni:
@asmagill Sounds good to me! Having them in collections actually makes a lot of sense, since there are so many color sets out there. Thanks!

zzamboni:
Closing in favor of #620

PR #614, opened 2015-10-20
Tweaks to the layout module:

ravasthi:
* Default to the main screen if the passed-in screen isn't found; useful for layouts activated using the screen watcher, meaning that screens in layouts may not be in the current screen configuration.
* Correct the print statement for "Unable to find display: "; the concat will only work with strings and numbers, and if a table or function is passed in there, the print craps out and throws a misleading error message.

### DIFF
--- a/extensions/layout/init.lua
+++ b/extensions/layout/init.lua
@@ -159,13 +159,18 @@ function layout.apply(layout, windowTitleComparator)
                 display = _row[3]()
             elseif fnutils.contains(screen.allScreens(), _row[3]) then
                 display = _row[3]
+            else
+                -- Default to the main screen if the passed-in screen isn't found; useful for
+                -- layouts activated using the screen watcher, meaning that screens in layouts may
+                -- not be in the current screen configuration.
+                display = screen.allScreens()[1]
             end
         else
             display = screen.allScreens()[1]
         end
 
         if not display then
-            print("Unable to find display: " .. _row[3])
+            print("Unable to find display: ", _row[3])
         else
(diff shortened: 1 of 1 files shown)


extensions/layout/init.lua
```diff
@@ -159,13 +159,18 @@ function layout.apply(layout, windowTitleComparator)
                 display = _row[3]()
             elseif fnutils.contains(screen.allScreens(), _row[3]) then
                 display = _row[3]
+            else
+                -- Default to the main screen if the passed-in screen isn't found; useful for
+                -- layouts activated using the screen watcher, meaning that screens in layouts may
+                -- not be in the current screen configuration.
+                display = screen.allScreens()[1]
```
cmsj: Could this use `screen.primaryScreen()` please? Right now it does the same thing as your code, but that may not always be the case.

ravasthi: Good point. Sure, I'll do that ASAP.

extensions/layout/init.lua
```diff
@@ -159,13 +159,18 @@ function layout.apply(layout, windowTitleComparator)
...
             end
         else
             display = screen.allScreens()[1]
```
ravasthi: By that reasoning, would it also be good to change this to `screen.primaryScreen()`?

cmsj: Good point, thanks!

ravasthi:
@cmsj I've updated the PR based on your comments; please review. Thanks!

Issue #597, opened 2015-10-09
Can hs.drawing.text make use of ANSI color escapes?


chdiza:
If HS is to ever really replace GeekTool for me, it will have to be able to display the output of shell scripts (or ruby scripts, etc.) like GeekTool does.  GeekTool will honor color escapes that are emitted by the shell script, e.g.,
```sh
yellow="\033[33m"
none="\033[0m"
echo -e "${yellow}This output${none} is colored"
```
If that script is in a "shell geeklet", there will be yellow text followed by text of whatever text color selection one makes in GeekTool's color picker.

I can't seem to get `hs.drawing.text` to honor these color escapes.  HS prints  `[33mThis output...`

If it makes any difference, I seem to have to use `hs.execute` to first get the shell script's output, then `hs.drawing.text` to put it on the screen.

asmagill:
The problem is that as far as I have seen so far, there is no direct support for ansi sequence to NSAttributedString conversion... and at present, I'm not sure if NSAttributedString with multiple regions (which would be necessary if using that class) would break hs.drawing or not... it would certainly make the text methods more complex.

I'd be curious to know what container class GeekTool is using for its text output... I chose NSAttributedString because its a subclass of NSString and required few changes to the core of hs.drawing... by accepting the limitation of only manipulating the object's entire region and not allow subregions.

Is GeekTool opensource?  If so, it's not out of the question that someone could see how they're handling this and create a new module around that model and not have to worry as much about breaking hs.drawing then...

While I've used Geektool some, I've had issues with it -- probably my fault, as I know people that love and swear by it -- so migrating/replicating things like it's color flexibility has been a low priority for me since I can get most of what I used to use it for already.


I guess the first thing to find out is if we can get at its source... that would go a long way towards determining how much effort a true replacement would take.

cmsj:
@asmagill GeekTool has a long and complex history with being open source, but it has previously had source releases and there are forks still hanging around:

https://github.com/balthamos/geektool-3/blob/ac91b2d03c4f6002b007695f5b0ce73514eb291f/NerdTool/classes/ANSIEscapeHelper.m

chdiza:
Unfortunately I'm unable to help code, but I am more than happy to test whatever anyone comes up with.

HS has an opportunity to gain some converts.  GeekTool has been decaying slowly and slightly with each new OS X release.  But it's seriously buggy now on ElCap, bordering on unusable unless one's geeklets are already created and adjusted they way one wants.

[...]

Issue #494, opened 2015-08-15
getColorFromStack - what it does and what it should do


asmagill:
Partially to disentangle differing ideas in thread #477, partially because I am uncertain what is "best" myself, and partially so we can decide on common practices to provide a consistent experience for the users of Hammerspoon, I bring up this helper function which I added to the hammerspoon.h file a while back for discussion...

This function is used in a few different places (mostly in hs.drawing but a couple other places as well, I think and at least one I will probably push soon) and it does one thing:  Takes a table on the lua_stack and checks for keys which define a color and returns an NSColor object which can be used elsewhere in the C code.

There are actually two versions of this function: One which takes a fallback NSColor to use if the lua parameter is nil or not present, and one which uses [NSColor blackColor] in this case (see step 2 below).  Currently I don't believe anything actually uses the version of this function which allows for a different fallback color.

The expected table will look like this, where each # is a number between 0.0 and 1.0 inclusive:

``` lua
color = { red = #, blue = #, green = #, alpha = # }
```

What the function currently does and why:
1. if the item at the stack location which is supposed to contain this table is not a table and is also not nil or none (i.e. the absence of a passed parameter) then it generates an error via luaL_error and returns up the function call chain (i.e. execution of this function ends, but the lua function which called the function which uses this helper can either handle the error or pass it back up the chain as well).   _-- This differs from before my changes in that before it would return nil but not an error._
2. If the item at the stack location is nil or none (a non-existant parameter -- the difference between `function()` and `function(nil)`) -- then the fallback color is returned.
3. Since we have a table, a "place holder" NSColor is created with red, blue, and green = 0, and alpha = 1.  _-- this differs from before where the initial place holder had alpha = 0 as well._
4. if red, blue, green, and/or alpha are found in the table and its value is a number, then the placeholder's key of the same name is changed to match the number assigned to the tables key.
5. if a key name other than those in 4 is found or if the value found for one of the above keys is not a number, it is silently ignored.
6. the resulting placeholder NSColor is returned and used by the functions caller in whatever way it wanted the NSColor for.

Now, where this differs from before:
[...]

PR #620, opened 2015-10-28
WIP: hs.styledtext

asmagill:
I'm not quite ready for it to go live yet, as I'm tracking down an intermittent crash and working up some documentation for it, but if you're interested and don't mind experimental...

This incorporates the X11 colors from #611, the enhanced hs.drawing.color of #602 (and the changes to hs.image, etc. to use LuaSkin rather than hammerspoon.h for creating image objects), and provides an ANSI color converter as requested in #597.


asmagill:
With these additions, anyplace that uses LuaSkin to get a color (I've updated menubar, drawing, and image here, and I know I need to update my third party console module... are there others?) can accept a table of { red=, blue=, green=, alpha= } or { hue=, saturation=, brightness=, alpha= } or { list=, name= } (for OS X's system color lists), but the list and name fields only work for the system color lists, not for the collections we add to `hs.drawing.color`.  Currently a lua side defined color has to be retrieved via `hs.drawing.color.colorFor("list").name` (this latter also works for the system lists)...

Should the { list=, name= } approach support the collections as well?  My leaning is yes, but it's not quite as simple, since they're defined in Lua and not available via a simple Objective-C method directly...

Thoughts?  Should I add this before adding the pull?  Or is documenting the difference sufficient?

cmsj:
@asmagill being able to access the collections that way sounds nice, but I don't think we need to go crazy to add that

lowne:
@asmagill there's always the possibility of a thin lua wrapper for these conveniences - see also my request in #494

asmagill:
the problem with just accepting an array of numbers is what are they?  RGB? HSB?

What is 1 number? 2?  Since we now allow lists with names, should 2 strings be allowed?  1 and default to the "Apple" list for lookups?

I forgot to mention above that { white=, alpha=} is also allowed... basically the generic's in the NSColor class.  I go back and forth about adding CMYK; since it's in NSColor as a device specific color space, I've avoided it for now, but from past experience, I know that someday I'll probably want it... especially if someone ever decides to tackle a module that can print.

But that would add a possible confusion for 4 numbers as well.

My general feeling is that the keys should be required, even if it does make it a little longer to type since its clear what is intended, but... I'll go with the consensus, so vote!  Vote early, vote often!

And as for the lookups into the collections, the latest pull adds it... I was mainly dreading dealing with possible loops, but I think I addressed that cleanly enough.

asmagill:
I think this is ready to land, pending any last minute requests/changes...

For testing, I primarily used the file at https://gist.github.com/asmagill/3f7f08e5934c4327db3e saved to my desktop as `test.sh`.  Then:

~~~lua
> c = hs.styledtext.ansi(hs.execute("~/Desktop/test.sh"), {font={name="Menlo",size=10}, color={list="Apple",name="White"}, backgroundColor = {alpha=.5}})
> a = hs.webview.new({x = 100, y = 200, h = 70, w = 330}):html(c:convert()):show()
> b = hs.drawing.text({x = 100, y = 100, h = 70, w = 330}, c):show()
~~~

Rendered:

screen shot 2015-10-29 at 10 07 15 pm

asmagill:
In addition, I am in the process of writing a couple of documents which I will add to the Wiki... a more "how do I do..." approach to modifying style with this module, rather than just the built in documentation, and a developers guide/example of how to use `hs.styledtext` in your C module without having to re-invent everything... this later may also serve as an explanation of how the LuaSkin converters work, depending upon how generic it is... or maybe that will be a third document; however, these shouldn't hold up inclusion of this module in the app.

lowne:
Keys 1..4 should mean RGB[A]. Even if the documentation didn't spell it out (and why shouldn't it), I don't see how anybody would think that a color-from-array constructor would default to HSB or CMYK or any other colour space over RGB - without precluding the possibility of allowing said color spaces with named keys.

lowne:
Alternatively, something like this will do:
`function RGBAColor(c) return color{red=c[1],green=c[2],blue=c[3],alpha=c[4]} end`
and I suppose it'll just be used everywhere, even without the dict/list polymorphism (because, why would people want to use the dict anyway)

asmagill:
I'm still not seeing the benefit of brevity in this case - it limits thinking to one color model when others may make more sense programmatically, depending upon what your use case is.

However, if you want to add it, or if there is enough demand, it would only require modifying the `table_toNSColorHelper` function in extensions/drawing/color/internal.m and possibly `NSColor_tolua`, if you want to change what is returned when an Objective-C NSColor object is queried.

For now, I'm going to apply this and it can always be updated later.

zzamboni:
@asmagill I compiled Hammerspoon with these changes, and your example works fine. Thanks!

Unfortunately, some of my old configurations were causing Hammerspoon to crash. I eventually tracked it to my use of `setTextStyle()`. In particular, this statement causes the crash: `text:setTextStyle({ ["alignment"] = "center", ["size"]=16.0})`.

I assume the arguments of `setTextStyle()` have changed, but it would be nice if old-style calls continued to work, or at least didn't cause a crash.

asmagill:
They were supposed to (be silently accepted and not crash), and my test set worked...  apparently I missed something!  I'll take a close look at it today.

Thanks!

asmagill:
@zzamboni , this is fixed in #625. I have a couple of more things I'm tracking down and then I'll add that pull to master.

zzamboni:
@asmagill Thanks!

PR #624, opened 2015-10-30
Added hs.menubar:frame()

### LINKED ISSUE #609: Add hs.menubar:frame()
It would be useful to know where on the screen the menubar items are. http://stackoverflow.com/questions/5413784/how-to-get-frame-for-nsstatusitem/10375784#10375784 may help, but we'll need to check that the proposed solutions work on 10.8.

ocurr:
Closes #609. One thing that I am unsure of is that the frame still exists if no icon or title is set, I have currently set it up to just return nil if this occurs, but I want to know if this is functionality that others would want.


asmagill:
my guess is that there should still be a frame -- I accidentally removed both the title and icon from a menu item once and if i recall correctly, I was still able to click on it if I positioned the mouse in the "perfect" place... however, it should be easy enough to test by calling :_frame instead of :frame since you don't suppress the unwrapped version.

ocurr:
Whoops, my bad, I wasn't clear in my first post. I have tested it and a frame does exist even if no icon or title is set. What I want to know is if others want the ability to access the frame even if no title or icon is set.

asmagill:
do you recall how wide it is? i can imagine using it to position the mouse and using eventtap to send a mouseclick ... might be a way to have an unobtrusive menu "hidden" most of the time, though i admit this is a bit contrived...

ocurr:
It ended up being about half the width of the hammerspoon menubar icon.

ocurr:
Since nobody else has responded, and I can't come up with any reasons not too, I went ahead and changed the behavior to return a frame whether or not the user has set an icon or title.

PR #625, opened 2015-11-01
styledtext fixes

asmagill:
* work-around for ANSI constructor when string contains UTF8 chars
* invalid font no longer crashes Hammerspoon

### DIFF
--- a/extensions/drawing/internal.m
+++ b/extensions/drawing/internal.m
@@ -1996,7 +1996,7 @@ static int default_textAttributes(lua_State *L) {
 /// Parameters:
 ///  * styledTextObject - an object created with the hs.styledtext module or its table representation (see `hs.styledtext`).
 ///
-///  The following simplified format is supported for backwards compatibility, but it is recommended that you use the `hs.styledtext` module instead.
+///  The following simplified style format is supported for use with `hs.drawing:setText` and `hs.drawing.setTextStyle`.
 ///
 ///  * theText   - the text which is to be displayed.
 ///  * textStyle - a table containing one or more of the following keys to set for the text of the drawing object (if textStyle is nil or missing, the `hs.drawing` defaults are used):
@@ -2025,7 +2025,7 @@ static int default_textAttributes(lua_State *L) {
 ///  * The size returned is an approximation and may return a width that is off by about 4 points.  Use the returned size as a minimum starting point. Sometimes using the "clip" or "truncateMiddle" lineBreak modes or "justified" alignment will fit, but its safest to add in your own buffer if you have the space in your layout.
 ///  * Multi-line text (separated by a newline or return) is supported.  The height will be for the multiple lines and the width returned will be for the longest line.
 static int drawing_getTextDrawingSize(lua_State *L) {
-    [[LuaSkin shared] checkArgs:LS_TANY, LS_TTABLE | LS_TOPTIONAL, LS_TBREAK] ;
+    [[LuaSkin shared] checkArgs:LS_TANY, LS_TTABLE | LS_TNIL | LS_TOPTIONAL, LS_TBREAK] ;
 
     NSSize theSize ;
     switch(lua_type(L, 1)) {
--- a/extensions/drawing/internal.m
+++ b/extensions/drawing/internal.m
@@ -2076,6 +2076,23 @@ static int userdata_tostring(lua_State* L) {
     return 1 ;
 }
 
+static int drawing_wantsLayer(lua_State *L) {
+    [[LuaSkin shared] checkArgs:LS_TUSERDATA, "hs.drawing",
+                                LS_TBOOLEAN | LS_TOPTIONAL,
+                                LS_TBREAK] ;
+
+    drawing_t       *drawingObject = get_item_arg(L, 1);
+    HSDrawingWindow *drawingWindow = (__bridge HSDrawingWindow *)drawingObject->window;
+    HSDrawingView   *drawingView = (HSDrawingView *)drawingWindow.contentView;
+
+    if (lua_type(L, 2) != LUA_TNONE) {
+        [drawingView setWantsLayer:(BOOL)lua_toboolean(L, 2)];
+        lua_pushvalue(L, 1) ;
+    } else
+        lua_pushboolean(L, (BOOL)[drawingView wantsLayer]) ;
+
+    return 1;
+}
 
 // Lua metadata
 
@@ -2092,6 +2109,7 @@ static const luaL_Reg drawinglib[] = {
 };
 
 static const luaL_Reg drawing_metalib[] = {
+    {"wantsLayer",          drawing_wantsLayer},
     {"setStroke", drawing_setStroke},
     {"setStrokeWidth", drawing_setStrokeWidth},
     {"setStrokeColor", drawing_setStrokeColor},
--- a/extensions/drawing/internal.m
+++ b/extensions/drawing/internal.m
@@ -764,6 +764,7 @@ static int drawing_setTextStyle(lua_State *L) {
         }
 
         NSMutableParagraphStyle *style = [[attributes objectForKey:NSParagraphStyleAttributeName] mutableCopy] ;
+        if (!style) style = [[NSParagraphStyle defaultParagraphStyle] mutableCopy] ;
 
 // NOTE: If we ever do deprecate setTextFont, setTextSize, and setTextColor, or if we want to expand to allow
 // multiple styles in an attributed string, move font and color into attribute dictionary -- I left them as is
--- a/extensions/drawing/internal.m
+++ b/extensions/drawing/internal.m
@@ -573,14 +573,50 @@ static int drawing_newImage(lua_State *L) {
 /// Sets the text of a drawing object
 ///
 /// Parameters:
-///  * message - A string containing the text to display.  May also be any of the types supported by `hs.styledtext`.  See `hs.styledtext` for more details.
+///  * message - A string containing the text to display
 ///
 /// Returns:
 ///  * The drawing object
 ///
 /// Notes:
 ///  * This method should only be used on text drawing objects
+///  * If the text of the drawing object is emptied (i.e. "") then style changes may be lost.  Use a placeholder such as a space (" ") or hide the object if style changes need to be saved but the text should disappear for a while.
 static int drawing_setText(lua_State *L) {
+    drawing_t *drawingObject = get_item_arg(L, 1);
+    HSDrawingWindow *drawingWindow = (__bridge HSDrawingWindow *)drawingObject->window;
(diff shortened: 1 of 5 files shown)


lowne:
@asmagill I have a bunch of `drawing.text` objects (that stay in memory), all with the same style ("global" font, alignment etc - no string subregions), so they're instantiated once with an immediate `:setTextStyle(...)`; the actual text changes dynamically via `:setText`.
If `:setTextStyle()` is deprecated, what's the "right" way of achieving that? From the docs it seems I'd have to re-create a new `styledtext` every time, is there a better (and less wasteful) way?

asmagill:
That should still partially work.  The setStyledText method is kind of a hybrid... for font, size, and color, it's really more of a wrapper for setTextFont, setTextSize, and setTextColor in that it applies the settings to the textfield and not to the string... The alignment/linebreak are set to the entire string as a simplification of what hs.styledtext does; this can't be applied to the textfield as it doesn't have a paragraph style attribute, only an attributed string does.

I'll run some tests, but I suspect that the alignment and linebreak settings set by this method are overwritten when you next call setText, but that the color and font settings remain in effect; since you're not setting a color or font with the string, they fall back to the textfield's settings.

I'll run some tests, but I believe the correct hs.styledtext way to do it is:

~~~lua
drawing:setText(drawing:getText():replaceSubstring(newString))
~~~

As long as newString is a lua string (not an hs.styledtext userdata or table form), this will replace the text without replacing any style attributes associated with the text, and replaceSubstring defaults to the entire string if you don't specify indices.  The getText() method is required so replaceSubstring knows what the current style is (line break and alignment, if all you're using is the setTextStyle method.)

lowne:
@asmagill if that's the case, and if I understand correctly that a styledtext object is always used for drawings behind the scenes, I propose to:
1. rename `styledtext:asString()` to `styledtext:getString()` (because it's not a transparent conversion, it's lossy)
2. add `styledtext:setString(s)` as sugar for `styledtext:replaceSubstring(s))`
3. rename `drawing:getText()` to `drawing:getStyledText()`
4. add `drawing:getString()` (`drawing:getStyledText():getString()`)
5. rename `drawing:setText()` to `drawing:setStyledText()`
6. add `drawing:setString(s)` (`drawing:getStyledText():setString(s)`)
7. for (long-term) backward compatibility, keep around `drawing:getText()` and `drawing:setText()` as aliases for the *string* methods
8. un-deprecate `drawing:setTextStyle(styleTable)`, and have it as sugar for `drawing:getStyledText():setStyle(styleTable,nil,nil,true)`; remove the translation code and the documentation for the old table format and point to the new format. `setTextStyle` is a fairly recent addition and I don't think we need to worry about compatibility

lowne:
Also: I haven't tested it nor looked at the implementation (and I don't know how NSAttributedString and its subregions work), but we should make sure that we can instantiate a styledtext with an empty string while retaining all the attributes for subsequent calls to `replaceSubstring`; relatedly, if my styledtext *does* have multiple regions with different styles, whatever strategy is used by `replaceSubstring(newString,nil,nil)` to decide which attributes will be given to the new string (I imagine whatever attributes were applied to the first or last character in the previous contents) should be documented.

asmagill:
1. makes sense... I was thinking of it as mirroring the asTable method, but you're right... it is lossy in a way that asTable isn't.
2. setString *instead of* replaceSubstring... then it more closely follows the pattern of setStyle (which can also work on sub-string regions with optional indices)
3. & 5. makes sense
4. prior to styledtext, there was no way to get the text from an already existent drawing object... style manipulation benefits from getting the current value, but do we need a shorthand to get just the string value when you can still get it via chaining to styledtext's getString?
6. & 7.  It would probably be better to leave it as setText, but have it do the getStyledText():setString() chain if the value is a string or number (as opposed to a userdata or table-form styled text)... I need to mull a little to make sure this doesn't break anything in the conversion functions, but I don't think it will.
8. Since half of what setTextStyle does is unique to hs.drawing (its defaults) and not styledText itself, I want to ponder this some more... 

As to your followup post, unfortunately an empty string drops any attributes you set to it... it needs at least one character to anchor the attributes to.

lowne:
Agreed on 2.

About `drawing.get/setString` + `drawing.get/setStyledText` vs `get/setText`, I also generally prefer (all else being equal) one overloaded method that deals with different argument types, but in this case I fear there are enough fundamental differences when dealing with strings vs styledtext (e.g. see this very thread) to warrant a clear-cut distinction; or in more general terms, if it's not possible to have consistently overloaded methods across the whole API, or if doing so would be excessively cognitively burdensome for the API user (e.g. having to deal with `hs.styledtext` at all if all I care about is quick and simple text labels). But *maybe* we can get away with it in this case...

About 8: another possibility (that would enforce clear separation of concerns) is to build up drawing's constructors:
- `drawing.rect(frame,attributes)` where `attributes` is `{fillColor=...,strokeWidth=...}` etc
- `drawing.[othershapes](frame,attributes)` ...
- `drawing.text(frame,attributes,styledText)`
- sugar: `drawing.text(frame,attributes,nakedString)`
- sugar: `drawing.text(frame.attributes,nakedString,styledTextAttributes)` -> `drawing.text(frame,attributes,styledtext.new(nakedString,styledTextAttributes)`
where the last two sugars reflect a similar adjustment you made to `getTextDrawingSize`.
With this, and provided `setString/setStyledText` or a unified `setText`, `setTextStyle` can go away entirely; if I want to change styles after the fact I'll have to use `styledtext` (which is appropriate, for such cases).

The `attributes` for the "pure" drawing object in the constructors is provided for completeness/symmetry/efficiency, and incidentally it would help with the "white rectangle syndrome" as described in #629 (which I'm similarly experiencing in my drawing-heavy modules).

About "9" (empty string): this imo is an implementation detail and should be semantically abstracted from API users, by e.g. substituting an empty string in the constructor, `setString` etc. with ZERO WIDTH SPACE `U+200B`.

asmagill:
Re 9, possible... to keep the string functions and metamethods consistent, it will mean that either there is no equivalent of the empty string for hs.styledtext or, if we silently convert it, no way to represent a single `U+200B` in a hs.styledtext object, but I don't think that will be too much of a problem...

lowne:
Well, if we really want *full* transparency, we can always have a separate `isEmpty` flag (when true, pass `"\u200b"` downstream to NSAttributedString, but the normal `""` upstream); then the user can `styledtext.new('\u200b') with abandon (`isEmpty`<-false, and `:getString()` retains it) :)

cmsj:
How are we looking on this one? I feel like we are due a release, so once hs.task has stabilised, I'll be aiming to get 0.9.43 out.

asmagill:
I should have (hopefully) final changes up tomorrow... still working out the implications of having a non-empty empty string and seeing if the gains are worth it.

asmagill:
* documented hs.drawing:wantsLayer
* hs.drawing:setText and hs.drawing:setTextStyle are back to their original form
* hs.drawing:setStyledText and hs.drawing:getStyledText are the ones which use `hs.styledtext`
* some other minor changes
* I elected to not include a "placeholder" for an empty string because (1) in hs.styledtext, the whole point is that you can have multiple styles; an empty string means there is nothing to have a style, therefor their is no style.  and (2) hs.drawing.text and hs.drawing:setTextStyle already document this gotcha (but still support changing font and color via setTextStyle), and it's not been made an issue before.

asmagill:
I think this is ready, unless anyone has objections/thoughts?

asmagill:
stupid auto fill making my commit notes make no sense...

lowne:
+1

cmsj:
I'm wondering about the webview_gc change, but I haven't dived into the context of it - just want to make sure it's not going to leak webview objects.

Other than that, no objections :)

asmagill:
@cmsj, I was seeing some occasional crashes during Lua's final cleanup when killing and restarting the Lua state... it was about the space lua had allocated being changed after it had already been released.  This seemed to help, so I'm not sure myself... I suppose one of these days I may have to dig into the guts of Lua's garbage collection, but I don't look forward to it...

@chdiza, other than wantsLayer not really helping you, any objections?  I'll keep digging and see if we can make things sharper, but in the meantime, I don't want this holding things up.

cmsj:
@asmagill huh, fair enough :)

chdiza:
@asmagill  No objections.  Not even about `wantsLayer`, since I'm happy to put my text geeklets on an opaque background.

chdiza:
I hadn't realized that this PR was updated a few days ago.  I think I'd been testing the version of it prior to the update.

Now that the update is there and has been merged, I'm seeing strange RAM behavior even though my init file and geeklets are the same (other than having to change `setText` to `setStyledText`).  My analog of "refreshing GeekTool" is to `nil`-out my hs.timer objects and hs.task objects, `:delete()` my hs.draw objects, and then recreate them.  It appears that after this PR was merged, the RAM they were using is never getting thrown pack into the pond.  Every "refreshing" operation adds 15MB of RAM to HS's total, where as with the older version of this PR in play, only about 2MB would be added.  (I don't see why that should be added either, but it's much better than 15MB.)

asmagill:
Hmmm... can't see why that would change...  since you had wantsLayer, I think I know which version you were running...  I've munged together a regression but with the same function names as the current master.  Download this file and expand it in your hammerspoon home directory, then fully restart hammerspoon and let me know if it makes a difference.

https://cumulus.dysfunctionalmonkey.info:8192/public.php?service=files&t=bea98216bee8efe17c7f4c7606a96576

chdiza:
I'll do this, but in the meantime I'd already changed my init quite a bit in attempts to figure out which "geeklet" were contributing.  I don't think I can recreate the way the init was when I had an early version of this PR installed.

Turns out, trying to isolate which geeklets are contributing reveals that `hs.drawing.line` objects, despite seeming like the simplest drawing objects, eat a TON of RAM.  I had three of them, and simply replacing them with hs.drawing.rectangle objects took about 10-12 MB of RAM out of play, and it also stopped the "gain 15MB of RAM every refresh" phenonomenon and replaced it with what had been there before (i.e., "gain 2-5 MB of RAM every refresh").

asmagill:
TBH I'm hoping the regression doesn't help -- it's back when I tried to overload setText.  I'd be real surprised if it helped at all, but from your description, I'm pretty sure it's what you were running before you mentioned the spike in memory.

chdiza:
Well it didn't help, but then again I was unable to restore my init file to the way it was before the PR was merged.

Thanks for throwing that together, though, I appreciate it :)

It seems that the main problem is likely to be `hs.drawing.line` anyway.  I'll continue to investigate, but so far, loading only my 3 `line` geeklets and no others produces higher RAM usage than if I load all the others plus 3 rectangles instead of 3 lines.  Then there is the matter of the persistent (permanently un-GC'ed) 5-7 MB on every "refresh" (i.e., destruction & re-creation), despite my (seemingly) delete()ing or nilling-out everything involved.

PR #627, opened 2015-11-02
allow hs.pasteboard to get image contents 

### LINKED ISSUE #626: feature request: support images in hs.pasteboard
I don't know about other people, but I use ⌘⌃⇧4 all the time to take partial screenshots to the clipboard, frequently to upload to imgur.

In Chrome on https://imgur.com, you can then ⌘V with image data in your pasteboard and hit go. Very easy. Unfortunately it doesn't work in Safari, which I prefer these days for general browsing. Why not just set up a hotkey to upload it to imgur and return the url to the pasteboard? Seems relatively easy and something I definitely want. 

Right now hs.pasteboard is hardcoded to use `NSPasteboardTypeString`. I think we should also support `NSPasteboardTypePNG` since I can think of a number of useful utilities to make with it as an 'image-heavy' user. There are a bunch of [`NSPasteboardTypes`](https://developer.apple.com/library/mac/documentation/Cocoa/Reference/ApplicationKit/Classes/NSPasteboard_Class/#//apple_ref/doc/constant_group/Types_for_Standard_Data_OS_X_v10.6_and_later_) but I imagine text/images cover >99% of use cases. I am willing to implement this myself but I would like some advice on the preferred way to get the data to Lua land. Probably best just to have something explicit like `hs.pasteboard:getImageContents` and leave `getContents` alone for now.

I had started to ask the best way to get an `NSImage` to Lua space, but found an example in `hs.image.imageFromPath` using `[[LuaSkin shared] pushNSObject:]`. I was then going to ask how to 'cast' the Lua-loaded `NSObject` to `hs.image` (I'm new at all this low level language interfacing business), but decided to just build and try it and it automagically worked and returned a valid `hs.image`, I suppose because of `[[LuaSkin shared] registerPushNSHelper:NSImage_tolua forClass:"NSImage"]` in `hs.image`. Great!  It looks like `userdata_gc` in `hs.image` properly releases the NSImage.

It seems I finished it and will create a pull request if there are no objections.

```objc
static int pasteboard_getImageContents(lua_State* L) {
    NSImage *image = [[NSImage alloc] initWithData:[lua_to_pasteboard(L, 1) dataForType:NSPasteboardTypePNG]];
    
    if (image && image.valid) {
        [[LuaSkin shared] pushNSObject:image];
    } else {
        return luaL_error(L, "No valid image data in pasteboard");
    }
    
    return 1;
}

//...

static const luaL_Reg pasteboardLib[] = {
//...
    {"getImageContents", pasteboard_getImageContents},
//...
};
```

heptal:
Adds a getImageContents() method to hs.pasteboard, which returns an `hs.image` object if available.
Resolves #626

Also, I find it annoying to navigate the docs without a link back to the index, so I added one:

### DIFF
--- a/extensions/pasteboard/init.lua
+++ b/extensions/pasteboard/init.lua
@@ -5,6 +5,7 @@
 --- This module is based partially on code from the previous incarnation of Mjolnir by [Steven Degutis](https://github.com/sdegutis/).
 
 local module = require("hs.pasteboard.internal")
+require("hs.image")
 
 -- private variables and methods -----------------------------------------
 
--- a/extensions/pasteboard/internal.m
+++ b/extensions/pasteboard/internal.m
@@ -24,6 +24,27 @@ static int pasteboard_getContents(lua_State* L) {
     return 1;
 }
 
+/// hs.pasteboard.getImageContents([name]) -> hs.image object or nil
+/// Function
+/// Gets the first image of the pasteboard
+///
+/// Parameters:
+///  * name - An optional string containing the name of the pasteboard. Defaults to the system pasteboard
+///
+/// Returns:
+///  * An `hs.image` object from the first pasteboard image, or nil if an error occurred
+static int pasteboard_getImageContents(lua_State* L) {
+    NSImage *image = [[NSImage alloc] initWithData:[lua_to_pasteboard(L, 1) dataForType:NSPasteboardTypePNG]];
+
+    if (image && image.valid) {
+        [[LuaSkin shared] pushNSObject:image];
+    } else {
+        return luaL_error(L, "No valid image data in pasteboard");
+    }
+
+    return 1;
+}
+
 /// hs.pasteboard.setContents(contents[, name]) -> boolean
 /// Function
 /// Sets the contents of the pasteboard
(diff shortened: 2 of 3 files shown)


zzamboni:
This is very nice. Any chance for a `setImageContents()`? ;)

heptal:
Sure. Took only a few minutes to implement using the image representation techniques used in `hs.image`. Will create a pull request later. 

```objc
static int pasteboard_setImageContents(lua_State* L) {
    NSImage*  theImage = [[LuaSkin shared] luaObjectAtIndex:1 toClass:"NSImage"] ;
    NSPasteboard* thePasteboard = lua_to_pasteboard(L, 2);

    NSData *tiffRep = [theImage TIFFRepresentation];
    if (!tiffRep)  return luaL_error(L, "Can't create internal image representation");

    NSBitmapImageRep *rep = [NSBitmapImageRep imageRepWithData:tiffRep];
    if (!rep)  return luaL_error(L, "Can't wrap internal image representation");

    NSData* pngImageData = [rep representationUsingType:NSPNGFileType properties:@{}];

    [thePasteboard clearContents];
    BOOL result = [thePasteboard setData:pngImageData forType:NSPasteboardTypePNG];

    lua_pushboolean(L, result);
    return 1;
}
```

zzamboni:
Works great, thanks!

Issue #644, opened 2015-11-19
hs.task and some modifications


asmagill:
I want to make some mods to hs.task, but before doing so, I want to make sure I'm not changing things that we're put in place for a reason...

First, I want to implement an `isRunning` method -- I know that the NSTask isRunning method is deprecated, but we can mimic it with terminationStatus and wrapping it in a @try/@catch block.  To distinguish between an exception because the task hasn't started yet and a task that is currently running, I'll need to add a flag to the userdata structure, if that's acceptable.

Second, can I change start, pause, and resume so that they return the hs.task object on success and false on failure?  You could still do `if (task:start()) then...` to test for success, but is there a reason/need for it to specifically return true?  The current implementation prevents doing something like `task = hs.task.new(...):start()`

Third, when I first conceived of a task module (I never finished it, hence your module, @cmsj), I wrote in code which duplicated the environment variables of Hammerspoon and assigned them so that the developer could use setEnvironment/getEnvironment to see and change the tasks environment vars... I know that if you pass nil (as is currently done in this implementation) that it inherits Hammerspoon's, but I don't remember if setting a few environment variables, but not others, with setEnvironment is additive (i.e. inherit what isn't defined) or not... duplicating them up front takes away ambiguity plus lets the user check to make sure they're what is wanted.  Can/Should I add this with the other proposed changes?

chdiza:
I like the `:isRunning()` idea.  Should allow me to do some debugging wrt seemingly stalled tasks.

I have no view on the other two questions.

cmsj:
@asmagill I think the third item here is still open for business? I think the duplication idea makes sense.

asmagill:
Check out what's in master... I updated the environment functions at that time. It works as I thought I remembered it (tested with the following shell script):

~~~sh
#! /bin/sh
printenv
~~~

If you don't do anything with `hs.task:environment` or `hs.task:setEnvironment`, you get what Hammerspoon was launched with.  If you set an environment, but don't first get it with `hs.task:environment`, then only what you set is present.  If you get the environment, modify it, and then reset it (all before starting the task, of course), then the modifications are included.

cmsj:
perfect :)

cmsj:
@asmagill This is all taken care of now, I think?

asmagill:
Yeah, this was merged with #646, so I'll close it.

PR #646, opened 2015-11-20
added some methods to hs.task

asmagill:
Proposed additions as described in #644, plus documentation for environment and setEnvironment

I followed the existing convention in the module of returning false and printing to the console on error, rather than use luaL_error.  I can change this if desired for consistency with (most) other modules.

### DIFF
--- a/extensions/task/internal.m
+++ b/extensions/task/internal.m
@@ -173,7 +173,7 @@ static int task_getWorkingDirectory(lua_State *L) {
 ///  * The hs.task object, or false if the working directory was not set (usually because the task is already running or has completed)
 ///
 /// Notes:
-///  * You can only set the working directory if the path has not already been started.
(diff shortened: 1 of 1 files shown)


asmagill:
thoughts? objections?

extensions/task/internal.m
```diff
@@ -138,6 +140,61 @@ static int task_new(lua_State *L) {
...
+/// Sets the working directory for the task.
+///
+/// Parameters:
+///  * path - a string containing the path you wish to be the working directory for the task.
+///
+/// Returns:
+///  * The hs.task object, or false if the working directory was not set (usually because the task is already running or has completed)
+///
+/// Notes:
+///  * You can only set the working directory if the path has not already been started.
```
cmsj: s/path/task/ :)

cmsj:
that one word correction aside, lgtm!

Issue #613, opened 2015-10-19
hs.clearConsole or tabs for console output by "id"


bradparks:
Hey! I find the console super busy with messages from lots of things...

I'd love to be able to:
- programmatically and in the console window itself clear the console output
- have the console break into tabs optionally by "id", so if I did

```
local log = hs.logger.new("test")
log.showInTab()
```

then in the console window, there'd be a separate to for any outputs by log messages by id "test".

Maybe another simple way to get this would be to have a watcher for log events, so optionally an app could catch their own logs msgs, and do whatever they wanted with them?

asmagill:
A certain amount of this already exists in hs.logger, but not all modules have been changed to take advantage of it yet.  I personally find the output from hs.hotkey a little overwhelming because it defaults to logging every registered key and when they are activated/deactivated, etc.  You can adjust this for hs.hotkey by putting something like this into your `init.lua` file:

      require("hs.hotkey").setLogLevel("warning")

where "warning" can be "nothing", "error", "warning", "info", "debug" or "verbose".

What other modules are you seeing a lot of output from?  Some of them may also be adjustable via hs.logger using a similar method, but not all of them have been converted yet... you can always take a look at making the updates yourself and submitting them for inclusion, or submit a specific list and we can try and address them as time and availability permits.


As to clearing the console, I've got a set of functions I've added in a third party module that can be found at https://github.com/asmagill/hammerspoon_console.  Specifically the `hs._asm.clearConsole()` function, but some of the others may be of interest as well. Currently I don't have a pre-compiled version available for download, because it relied on functions which weren't in the core application when I last worked on it, but I think it should be fine now... I'll see about getting a precompiled version up later today or tomorrow, or you can download it and compile it yourself.

--
A-Ron

bradparks:
I just tried it and that custom console you have in your repo and it seems to work great! That plus the idea of turning off the hotkey logging will totally do all I want... thanks a bunch!

asmagill:
@bradparks, can this be closed?

cmsj:
Should we pull `clearConsole()` into the core? It might be interesting to people.

bradparks:
I'm fine with it being closed anytime. I think having clearConsole in core would be a great idea, imho....

asmagill:
simple enough to add... anything else in the module worth adding to the hs space? i'm biased so i leave it to the group to decide

heptal:
+1

[...]

PR #653, opened 2015-11-24
adding hs.console module

asmagill:
Some mods to my earlier module, but repackaged for inclusion, pending objections/corrections/suggestions.

Partially address #613

Includes replacement print which support hs.styledtext and a few other additions/modifications.

### DIFF
--- a/extensions/console/internal.m
+++ b/extensions/console/internal.m
@@ -8,18 +8,18 @@
 
 @interface MJConsoleWindowController : NSWindowController
 
-+ (instancetype) singleton;
-- (void) setup;
++ (instancetype)singleton;
+- (void)setup;
 
 @end
 
 @interface MJConsoleWindowController ()
 
-@property NSMutableArray* history;
+@property NSMutableArray *history;
 @property NSInteger historyIndex;
-@property IBOutlet NSTextView* outputView;
-@property (weak) IBOutlet NSTextField* inputField;
-@property NSMutableArray* preshownStdouts;
+@property IBOutlet NSTextView *outputView;
+@property (weak) IBOutlet NSTextField *inputField;
+@property NSMutableArray *preshownStdouts;
 
 @end
 
@@ -27,11 +27,11 @@
 
 typedef struct _drawing_t {
     void *window;
-    BOOL skipClose ;
+    BOOL skipClose;
 } drawing_t;
 
-int refTable = LUA_NOREF ;
-// int hsDrawingRef = LUA_NOREF ;
+int refTable = LUA_NOREF;
+// int hsDrawingRef = LUA_NOREF;
 
 /// hs.console.asHSDrawing() -> hs.drawing object
 /// Function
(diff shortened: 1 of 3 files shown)


extensions/console/internal.m
```diff
@@ -0,0 +1,382 @@
...
+///  * You can specify the console content as a string or as an `hs.styledtext` object in either userdata or table format.
+static int console_setConsole(lua_State *L) {
+    [[LuaSkin shared] checkArgs:LS_TANY | LS_TOPTIONAL, LS_TBREAK] ;
+    MJConsoleWindowController *console = [MJConsoleWindowController singleton] ;
+
+    if (lua_gettop(L) == 0)
+        [[console.outputView textStorage] performSelectorOnMainThread:@selector(setAttributedString:)
+                                           withObject:[[NSMutableAttributedString alloc] init]
+                                        waitUntilDone:YES];
+    else {
```
cmsj: I'm generally not a fan of braceless conditional blocks, but mixing braced and braceless seems like a fairly bad idea to me :)

extensions/console/internal.m
```diff
@@ -0,0 +1,382 @@
...
+/// Parameters:
+///  * None
+///
+/// Returns:
+///  * an array containing the history of commands entered into the Hammerspoon console.
+static int console_getHistory(__unused lua_State *L) {
+    [[LuaSkin shared] checkArgs:LS_TBREAK] ;
+    MJConsoleWindowController *console = [MJConsoleWindowController singleton] ;
+
+    [[LuaSkin shared] pushNSObject:[console history]] ;
```
cmsj: It'll be marginally more efficient to just grab a local LuaSkin pointer, rather than calling through to `shared` multiple times in a function.

extensions/console/internal.m
```diff
@@ -0,0 +1,382 @@
...
+///
+/// Returns:
+/// * the current color setting as a table
+///
+/// Notes:
+///  * See the `hs.drawing.color` entry in the Dash documentation, or type `help.hs.drawing.color` in the Hammerspoon console to get more information on how to specify a color.
+static int console_inputBackgroundColor(lua_State *L) {
+    NSTextField *input = [MJConsoleWindowController singleton].inputField ;
+
+    if (lua_type(L, 1) != LUA_TNONE) {
```
cmsj: Is there a reason to not be using `LuaSkin checkArgs` for the various functions here that have arguments?

cmsj:
Few minor objections, otherwise lgtm :)

asmagill:
A good compiler should take care of the multiple [[LuaSkin shared]...] thing, so it really shouldn't affect efficiency much if at all, and I don't know why I do it... it seems more like the alloc/init thing for NSObjects, and in *most* functions I write, it ends up being used only once or twice.

As to the brace/braceless thing, I'd have to look at the specific instance, but likely it comes from copy/pasting/merging things from different sources... I generally try to keep it the same for a given if/then/else block.  I can address these later tonight.

cmsj:
I'm not sure if the compiler can optimise that, it would need to identify the same parameter to objc_msgSend and understand it was returning a static variable every time. Maybe I'm underestimating clang/llvm :)

Either way, it's not performance critical code.

asmagill:
You may be right... I was thinking about it in terms of the value returned and not the fact that the whole message-passing thing is actually reworked by a pre-parser before actual compilation occurs into a series of obj_msgSend functions...  I would hope that a good compiler would optimize it some, but in the end, it couldn't guarantee that the value didn't change without sending at least one message to check each time...

PR #657, opened 2015-11-26
Slightly increase LuaSkin test coverage

cmsj:


cmsj:
So this is an interesting one. Previously, LuaSkin would `abort()` if it detected code execution on the non-main thread. That is problematic for automated testing, but since this is a pretty crucial safety feature, I feel like it ought to be tested.

We have an Info.plist setting to die on unhandled exceptions, so I've switched `[LuaSkin shared]` over to throwing an exception if it detects non-main thread Lua work, however, this does mean that if someone modifies the plist, we won't die.

I think that's an acceptable trade-off, because almost everyone (including the developers, I hope!) isn't going to be disabling that plist setting ∴ we'll catch these issues just as before. The one part I'm not 100% on yet, is whether Crashlytics will catch this.

Anyone have thoughts in either direction about this?

(Separately, our test coverage for LuaSkin has fallen from almost-100%, to 17%, which is something we badly need to address)

asmagill:
I don't see this as a problem... if you're changing the plist, then you're already running something we don't test against and either know what you're doing or deserve what you get :-)

Seriously, we can't protect against everything; just give appropriate defaults and warnings if you choose to play differently.

Issue #668, opened 2015-12-08
webview - hammerspoon protocol not recognized


trzeciakp:
Hi,
I've created webview with hammerspoon link and handler in hammerspoon. It works ok in chrome, but in webview I got the error:

    An Error code: -1002 in NSURLErrorDomain occurred during navigation:
    Description: unsupported URL

Is there a way to use hammerspoon links in webview?

cmsj:
Could you post the configure you're using?

trzeciakp:
What do you mean? Webview configure?
    local rect = hs.geometry.rect({1, 1, 200, 800})
    local web = hs.webview.new(rect)
    web:windowStyle({'closable', 'titled'})
    web:url("http://localhost:9000")
    web:allowTextEntry(true)
    web:closeOnEscape(true)
    web:show()

and in the html:
    
    window.location.href = 'hammerspoon://break-report?breakType=' + element.innerHTML

cmsj:
sorry, yes, I meant to type "config", but even that may have been unclear. Thanks for the pasting that, it should be enough to figure out what's going on here :)

cmsj:
This may be one for @asmagill

trzeciakp:
I've prepared jsfiddle to ease you testing:
http://jsfiddle.net/trzeciakp/8t1535o5/light/

Unfortunately it does not work in web view so below is corresponding html file.

here is hammerspoon code:

local linkToHtml = PASTE_LINK_HERE
    hs.hotkey.bind({"cmd", "alt", "ctrl"}, "W", function()
        showButtonsWebView()
    end)
    
    function showButtonsWebView()
        local rect = hs.geometry.rect({1, 1, 800, 200})
        local web = hs.webview.new(rect)
        web:windowStyle({'closable', 'titled'})
        web:url(linkToHtml)
        web:allowTextEntry(true)
        web:closeOnEscape(true)
        web:show()
    end
    
    hs.urlevent.bind('break-report', breakReportCallback)
    function breakReportCallback(eventName, params)
      hs.alert.show(params.breakType)
    end

and here html code


     <html>
       <head>
         <meta http-equiv="content-type" content="text/html; charset=UTF-8">
           <title>Hammerspoon link example</title>
           <link rel="stylesheet" type="text/css" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.6/css/bootstrap.min.css">
[...]

PR #671, opened 2015-12-08
[READY] webview hands off unrecognized schemes to OS

asmagill:
Addresses #668 

This update allows you to specify which URL schemes are to be handled by the WKWebView at the heart of this module.  If the scheme isn't in the list, it is handed off to the OS to handle.

This is probably the more user-friendly approach, but a more secure approach would have been to switch directions: list the schemes we're willing to hand out to the OS; however I didn't think of that until just now.

Preferences on whether it should be a list of schemes to handle (and thus hand off what we don't) or whether it should be a list of what we hand off (and thus attempt to handle what isn't listed, causing an Unrecognized protocol error page when WKWebView doesn't really handle it)?

### DIFF
--- a/extensions/webview/init.lua
+++ b/extensions/webview/init.lua
@@ -21,9 +21,6 @@ local module       = require("hs.webview.internal")
 module.usercontent = require("hs.webview.usercontent") ;
 local http         = require("hs.http")
 
--- assign the default schemes handled natively by the WKWebView.  Anything else is handed off to the OS.
-module.nativeURLSchemes({"http", "https", "about", "file"})
-
 -- private variables and methods -----------------------------------------
 
 local _kMetaTable = {}
--- a/extensions/webview/init.lua
+++ b/extensions/webview/init.lua
@@ -21,6 +21,9 @@ local module       = require("hs.webview.internal")
 module.usercontent = require("hs.webview.usercontent") ;
 local http         = require("hs.http")
 
+-- assign the default schemes handled natively by the WKWebView.  Anything else is handed off to the OS.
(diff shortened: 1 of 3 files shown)


cmsj:
Which way around does Safari choose? I suggest we mirror that :)

asmagill:
No idea...

The problem as I understand it is that WKWebView doesn't automatically check registered Application or protocol URLs in the same way that UIWebView or WebView did (supposedly -- I've never used them myself).  Nor have I been able to find a list of "officially recognized" protocolos for the WKebView class, generated or otherwise.

It comes down to do we want to field "why isn't X scheme type handled in the webview itself" or "why do I get unrecognized URL type" questions?

asmagill:
I should add that if we pass off what we don't recognize, the OS might pop up a dialog about not recognizing the URL scheme like this:

screen shot 2015-12-08 at 6 22 41 pm

If we attempt to do it ourself and it fails, we get the error page that led to #688.

I suppose I might be able to patch into the didFailNavigation delegate instead of the didAcceptNavigation one... the delegate method doesn't receive the same arguments, but I think we can still get the URL scheme out of one of them.  That might even remove the need for the new function all together, but the module would no longer ever handle the Unrecognized protocol error, since that would be specifically passed out... must mull and test...

asmagill:
Never mind all that above... moved the hand off to didFailProvisionalNavigation when the error type is NSURLErrorUnsupportedURL; that way we don't have to keep track of recognized versus unrecognized... the OS handles anything that WKWebView doesn't handle first.

asmagill:
Note you can still reject any URL you like with the navigationCallback -- that happens before the didFailProvisionalNavigation message occurs.

PR #677, opened 2015-12-11
Reordered log message before the msg is deleted

bchapman:
Fixed ordering of log message to be before the msg has been deleted.

Otherwise, got this error when deleting a hotkey:
```
...oon.app/Contents/Resources/extensions/hs/hotkey/init.lua:123: attempt to concatenate a nil value (field 'msg')
stack traceback:
	...oon.app/Contents/Resources/extensions/hs/hotkey/init.lua:123: in method 'delete'
```

Code had the msg attribute trying to be printed, after it had been deleted by:
```for k in pairs(self) do self[k]=nil end --gc```


cmsj:
Good catch, thanks!

Issue #676, opened 2015-12-10
hs.wifi


asmagill:
Is there interest in being able to turn the wifi power on or off (still no bluetooth without using private API's, however, so its not a true "Airplane Mode" by itself) or doing a background (i.e. non-blocking) scan with full details?

junkblocker:
Yes, +1

oschrenk:
Would be nice. Then I can replace, this

```
  os.execute("/usr/sbin/networksetup -setairportpower " .. wifiInterface .. " off")
```

cmsj:
This should be closable now :)

PR #685, opened 2015-12-12
[READY] Wifi

asmagill:
Additions include:
* backgroundScan(fn) - scans for available networks without blocking.  Returns full details (channel, PHYmode, rssi, noise, etc.)  Invokes callback with results.
* interfaceDetails([interface]) - full details about the interface including above plus known networks, all supported channels & modes, etc.
* interfaces() - list of all recognized wifi devices on the machine
* setPower(state, [interface]) - turns wifi on or off
* disassociate([interface]) - disconnect from current network

All functions take an optional interface argument and should work on systems which have multiple airport devices... I've not been able to test this myself, but the code matches the API and *should*.  Feedback would be appreciated.

The background scan is done by doing the scan in a new thread.  All lua activity occurs on the main thread, however.  @cmsj, I'd appreciate a once over to make sure I haven't missed anything, but I'm pretty sure I've covered my bases.

Because I'm a fan of examples, I've thrown together a simple rssi/noise graph here: https://github.com/asmagill/hammerspoon-config/blob/master/utils/_keys/wifiSNR.lua  which looks like this when it's been running for a while:
screen shot 2015-12-11 at 11 28 43 pm

### DIFF
--- a/extensions/wifi/init.lua
+++ b/extensions/wifi/init.lua
@@ -1,9 +1,12 @@
 --- === hs.wifi ===
 ---
 --- Inspect WiFi networks
---- NOTE: This extension assumes that you have one and only one WiFi interface on your Mac. If you have more than one, behaviour is undefined.
 
 local wifi = require "hs.wifi.internal"
 wifi.watcher = require "hs.wifi.watcher"
+local log    = require("hs.logger").new("hs.wifi","warning")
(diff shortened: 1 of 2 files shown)


asmagill:
addresses #676

cmsj:
lgtm :)

Issue #682, opened 2015-12-11
Allow communicating with hs.task via stdin


yoavg:
I'd like to be able to construct a string which I could then send to an external process via stdin.

Currently, hs.task allows to define commandline arguments to be sent, but not stdin value (or is there a way to do it which I missed?)

asmagill:
At present, I don't believe its possible.  What kind of interface are you thinking of?  Something where you can provide a string with the input when creating the task, or would there be an additional method which could be called to "feed" the stdin of a running task?

I've not actually tried a task yet that wanted input yet... what does it do?  Does it stall? or continue as if the input was empty?

yoavg:
I was thinking of providing the input when creating the task, or perhaps changing the start() method to take the string as a parameter.

My current use-case for this is a dynamic menu with fuzzy-find functionality, i.e. feeding text into a selector such as https://github.com/sdegutis/choose and acting on the result.

cmsj:
@asmagill I've just pushed up some code for doing stdin with both streaming and non-streaming tasks, based on what you had in #686. I don't seem to have as much Exception protection as you, and I'm not sure if that's necessary. Do you have some gnarly tests you could hurl at my code?

asmagill:
At least one of the exception tests was around code which was failing to exit gracefully before I put it on the main thread.  I have never been able to actually trigger an exception using the writeData: message, even though its documentation reads "This method raises an exception if the file descriptor is closed or is not valid".  Even when it was failing before, it wasn't with an exception -- it was with a crash.  Using `dispatch_sync(dispatch_get_main_queue(), ^{` took away the crashes, but I still never saw an exception.

So, at a guess, I don't think it will matter. I'll try some tests tonight or tomorrow, though, and let you know.

cmsj:
Ok thanks :)

asmagill:
Ok, against what's currently merged, for the following script `t.sh`:

~~~bash
#! /bin/bash
sleep 15
echo "begin"
a=$(cat -)
echo $a
echo "end"
~~~

(the sleep is in there so I can do things like restart Hammerspoon, test resetting the input while it's already running, etc.)

Here are some of the things I've tested:

~~~lua
> n = hs.task.new("./t.sh", function(...) print(inspect(table.pack(...))) end)
> n:setInput("test1")
hs.task: ./t.sh  (0x60800014dd48)
> n:start()
hs.task: ./t.sh  (0x60800014dd48)
[...]

PR #686, opened 2015-12-12
[READY] setInput for hs.task

asmagill:
Added a method to hs.task which allows setting the stdin to a string (which can be arbitrary binary data -- i.e. not proper UTF8) as the STDIN for an `hs.task`.

See #682

### DIFF
--- a/extensions/task/internal.m
+++ b/extensions/task/internal.m
@@ -107,19 +107,16 @@ void create_task(task_userdata_t *userData) {
 ///
 /// Returns:
 ///  * An `hs.task` object
-///
-/// Notes:
-///  * Raises an error if the launchPath does not refer to an executable file.
 static int task_new(lua_State *L) {
     // Check our arguments
     LuaSkin *skin = [LuaSkin shared];
     [skin checkArgs:LS_TSTRING, LS_TFUNCTION|LS_TNIL, LS_TTABLE|LS_TOPTIONAL, LS_TBREAK];
 
     NSString *thePath = [skin toNSObjectAtIndex:1] ;
-    if (![[NSFileManager defaultManager] isExecutableFileAtPath:[thePath stringByExpandingTildeInPath]]) {
-        return luaL_error(L, [[NSString stringWithFormat:@"%s.new: %@ is not executable", USERDATA_TAG,
-                thePath] UTF8String]) ;
-    }
+//     if (![[NSFileManager defaultManager] isExecutableFileAtPath:[thePath stringByExpandingTildeInPath]]) {
+//         return luaL_error(L, [[NSString stringWithFormat:@"%s.new: %@ is not executable", USERDATA_TAG,
+//                 thePath] UTF8String]) ;
+//     }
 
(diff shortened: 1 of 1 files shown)


asmagill:
hold off... can crash under some circumstances... working on it

asmagill:
Ok the situation is this:  In order to ensure that the stdin is available when the task requires it, we use `[NSFileHandler writeabilityHandler:]`; however this dispatches its block immediately when the task is launched *even if the launch of the task fails*.  This is most likely to happen if you put the wrong path to an executable into the task, or forget the path altogether (e.g. "ls" instead of "/bin/ls").  The writability handler attempts to write the data out to a dead pipe.  Well, maybe not dead, since the docs say that should throw an exception which we should in theory catch (but for some reason aren't), but not connected at any rate.  The write blocks until the pipe is closed (garbage collection) and then Hammerspoon crashes... no exception is thrown, no "Hammerspoon unexpectedly crashed" alert, just... dead from a Broken Pipe signal (13).  And I can't figure out how to catch it/detect it/stop it in a safe survivable way (I suppose a signal handler could do it, but modifying Hammerspoon to allow clean re-entry from a signal handler is not a trivial task).

The only thing I can think of is to take advantage of the fact that 90%+ of the time, the task launch will fail because the path isn't correct -- this latest update checks that before creating the task, so this specific cause should never happen again.  However, I cannot guarantee that this is the *only* reason the launch might fail, just by far the most common.

So... what to do?

Being able to pass stdin is very useful (I'm liking sdegutis's choose utility a lot, now that I've played with it), but this isn't as wrapped/protected as I'd like.  

OTOH, unless you're using stdin, nothing has changed, except that you know sooner if your launch path is wrong.

I'm not sure if there is a way to use some sort of notification/background-thread write through the pipe to NSTask... all of the examples I've seen of this so far are writing to files/sockets and not NSTask, but... I'll keep looking.

In the mean time, do we add this or hold off?  I can always put it up with my other modules and offer it as a downloadable option to those who want it (and the risk), and leave it out of core.

Thoughts?

asmagill:
Ok, ignore previous post.  I don't know why I didn't try the same approach used for the stdout and stderr pipe's and make sure they were handled only on the main thread.  Putting the same dispatch wrapper around the writing and subsequent close allows this to fail gracefully like the rest of the module.

I believe this is ready for core now.

cmsj:
I'd like to delay this PR a little - I have a patch in flight which allows for tasks to run indefinitely, with a callback whenever they produce output. I'd like to make sure that this PR makes sense with it, and can work sanely for feeding input over time to long-running tasks.

asmagill:
@cmsj have you tested these additions against your recent changes to `hs.task`?  I do see that the branch has conflicts -- are you planning to address it, or should I look into it after the LuaSkin refactor?

cmsj:
@asmagill I have briefly and I need to make some changes. Unfortunately I ran out of wakefulness last night, but it's on my list :)

cmsj:
(see comments in #682 around this)

PR #688, opened 2015-12-12
[READY] add hs.redshift

lowne:
~~Needs #536~~ 

~~It's necessary to add the module to the build script~~


lowne:
If/when anyone tries out this module: I find colour temperatures between about 1400 and 2000 to be a bit too greenish, is it just me? 
I used to have a custom function for the gamma calculation based on entirely subjective/biased/depending on my specific calibration observations, but I thought that it'd be better to use the more scientific approach in redshift (whence the current colour ramp)... and the excess green I see might just be a result of my specific monitor calibration.
Thoughts?

extensions/screen/internal.m
```diff
@@ -447,14 +447,14 @@ static int screen_gammaSet(lua_State* L) {
     lua_getfield(L, 2, "blue");
     whitePoint[2] = lua_tonumber(L, -1);
     lua_pop(L, 1);
-
+    /*
```
cmsj: So these two conditional overrides serve as crude protection against the gamma values being set such that nothing can be seen on the screen anymore. Rather than just disable them, perhaps we can figure out a way to keep the ability to do an inverted mode, without losing the "oh damn I can't see my display anymore" protection? (initial, untested thought, we could check if the white/black points are equal)

lowne:
@cmsj those checks need to go, because all they do is prevent colour inversion (ironically resulting in the "can't see anything" scenario when redshift does try to do colour inversion). As you say, the way to guard against "can't see anything" is to make sure that whitepoint!=blackpoint, but that's currently not implemented so we're not losing anything - and if that hasn't been necessary until now, then it's still unnecessary because hs.redshift will never cause it.

But if you think we do need it, then I'll add a proper guard to hs.screen.

lowne:
(@cmsj to be clear I think the necessary - and not totally insignificant - overhead on every call isn't worth removing a bucket of danger out of the the "ways to shoot yourself in the foot with HS" ocean)

asmagill:
I've not played with this yet, so I don't know how easy or hard it is to shoot yourself in the foot, or get out of it once you do, but...

My general rule of thumb is if quitting or reloading Hammerspoon will clear whatever is "wrong" (both of which can easily be attached to a hotkey) then I don't generally worry too much, as long as it doesn't crash. But that's me...

lowne:
@asmagill that's the case here; to be honest, there's still a possibility of explicitly having (pseudocode) `setGamma(0.5,0.5)` at the top of one's `init.lua` - then you'd just have a brief moment of visibility at every reload; but "normal" variable-powered "accidents" would be cleared.

PR #696, opened 2015-12-17
WIP hs.doc update

asmagill:
I've noticed some trying to create relative links in the Dash documentation by using `[description](#function)`... I like the idea, it works well in Dash and in HTML versions of the docs for creating links within a module to the related item.

However it looks bad in the text version of the help  (see `help.hs.caffeinate.get`).

This is a fix for that... any substring on a single line that contains 
~~~
[something1](#something2)
~~~
is replaced with
~~~
`something1`
~~~

I need to make sure it doesn't break other things and that it won't cause problems with other links, so for now this is WIP, but I encourage you to try this if you're using links in internal documentation in your own WIP modules.

### DIFF
--- a/extensions/doc/init.lua
+++ b/extensions/doc/init.lua
@@ -144,6 +144,13 @@ local validateJSONFile = function(jsonFile)
     end
 end
 
+local fixLinks = function(text)
+    -- replace internal link references which work well in html and dash with something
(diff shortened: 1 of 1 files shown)


asmagill:
Any objections to merging this?

cmsj:
(retrospectively, no objections :)

PR #701, opened 2015-12-21
Add support for hs.task tasks to run indefinitely

cmsj:
So this one is a little hairy. It adds a second (but optional) callback to hs.task.new() which will be called whenever the OS decrees that there is new data to be read on either the stdout or stderr pipes, and the callback returns true to keep the background reading going, or false to stop it (although that applies per-channel, which is unfortunate, but seemingly unavoidable).

Where things get really hairy though, is that for this callback to be properly useful, it needs to be passed the hs.task object that fired it, but Lua has no API for pushing an existing userdata object onto the stack, so I've had to borrow some code from the Internet that is effectively a clone of lua_newuserdata(). I don't know if this should go in or not, and seek opinions (on subjects such as whether pushUserData is a reasonable addition to LuaSkin,, or whether indefinite task support is still useful without the task object argument, or on any other subject).

Paging Dr @asmagill :)

### DIFF
--- a/LuaSkin/LuaSkin/Skin.h
+++ b/LuaSkin/LuaSkin/Skin.h
@@ -14,6 +14,8 @@
 
 
 #import <Foundation/Foundation.h>
+#import "lobject.h"
+#import "lapi.h"
 #import "lauxlib.h"
 #import "lualib.h"
 #import "lua.h"
@@ -445,6 +447,14 @@ typedef struct luaObjectHelpers {
  @returns YES if the module loaded successfully or NO if it does not
  */
 - (BOOL)requireModule:(char *)moduleName ;
+
+/*!
+ @abstract Pushes an existing Lua userdata object onto the Lua stack
+
+ @important This is a terrible hack and could easily break with future Lua versions. Its use is discouraged
+ @param userData - a pointer to memory that was allocated by lua_newuserdata()
+ */
+- (void)pushUserData:(void *)userData;
 
 @end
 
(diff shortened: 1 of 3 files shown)


asmagill:
I'll take a look at it tonight, but from what you're describing you need to do to pass the same userdata around, you might take a look at what I do in `hs.speech.listener`... 

In a nutshell, I subclass the object, add a udref property to it that is set to LUA_NOREF in init, then use a helper function to pushNSObject: for the subclassed type and use pushNSObject in both the constructor and whenever needing to pass it from a delegate, etc.... in the helper function, if the udref property = LUA_NOREF, then I create a userdata, save it in the registry and generate a ref for it and save it in the objects udref property... otherwise, the property is used to push the object onto the stack and return without creating a new one.  The downside to this approach (which is why I don't use include the save to registry for all userdata objects in my code) is that you have to explicitly delete the object for it to be garbage collected because the saved reference in the registry keeps the reference count above 0, even if the user has set all lua-side variables for it to nil.

`hs.speech` (the synth portion of the module) uses something a little different... it's has its own reference counter so that __gc can be called on all of the "copies" of the "new" userdata for the object, but it only does the actual __gc work for the object when the reference count drops to 0.  The down side to this approach is that you can't use the userdata as a key in a table on the lua side because its address does change for each "copy" of the userdata, but you can usually work around this if the __eq metamethod exists.

At any rate, if either of these sound useful, check them out; I'll look at the actual code later tonight and see if I can think of anything else that might help.

cmsj:
@asmagill I like the selfRef plan, I have switched to using that, but I'd appreciate if you'd check over what I did :)

cmsj:
I'm going to land this - it's doing fine in all the tests I've thrown at it so far.

Issue #699, opened 2015-12-21
hs.logger question about history


asmagill:
How hard would it be to add a history to hs.logger so that, even if my current level is suppressing some of the messages, I can go to the console and review the last 50 (or whatever number)?

lowne:
@asmagill not very, and something that I have wanted to do for a while, but it needs appropriate hooks to the console texture (plus a "filter" text input, I suppose) (plus #536 to begin with).

lowne:
To be precise: it doesn't "need" the hooks, but I guess that'd be the logical next step (for ease of use) after implementing something like `hs.logger.getHistory(filter,entries)`.
Should I start with the internal implementation?

cmsj:
@lowne I'm not sure I understand what you need here from the Console? Wouldn't this just be returning a table of the last N strings that were sent to hs.logger? (implying that hs.logger would be keeping a fixed-length buffer)

lowne:
Yes; let's do things the other way around. I'll get to it as soon as I'm done with the current PRs.

lowne:
To clarify: I initially thought of allowing manipulation of the console (changing log level on the fly (and retroactively), filtering by id,..) from `hs.logger`'s Lua code; but I'll just implement `hs.logger.getHistory()` which will return a list of `{timestamp,id,level,message}` entries.

asmagill:
Not to get all kefedokhles and side-tracked again, but take your pick:

1. a format, as defined for use in `string.format` can reference 0 parameters, and ..., as defined in the lua manual refers to 0 or more arguments

2. It's one less thing on the traceback stack.

3. It doesn't matter, it works.

4. If you don't like it, write/modify it yourself... the delegate code will be very simple and found in MJLua.m... I'll let you know when the main code is ready and it should be only about 3 lines of code that will need to be modified to set the callback function

My main reason was 2, not for performance reasons, but because it seems more elegant to avoid increasing it solely for a functional syntax (why do you even have a version that takes no parameters when nothing about string.format or ...  has an issue with 0?) that I don't believe is really needed anyways.

asmagill:
I should also probably add

5. when I first started attempting the integration, I hadn't decided for certain if I would be passing multiple args or not... I'm still not 100% that it would never be useful, but that's probably a case-by-case thing... 90% of the time, and probably 100% of the time coming from the LuaSkin log methods, I probably won't pass additional args.

lowne:
[...]

PR #704, opened 2015-12-24
[READY] add hs.logger history stuff

lowne:


asmagill:
I've been thinking about this while working on adding logging methods to LuaSkin... would there be objections to my creating a "global" logging instance (maybe with the tag "LuaSkin" or "hammerspoon" that could capture items we log through these methods wth LuaSkin? Using this from the C-side isn't that difficult -- I've done it with hs.speech, if anyone wants to review the steps.  I'd make it part of the delegate, so the actual integration with this module would probably be in MJLua.m, rather than LuaSkin, but thats an implementation detail.

lowne:
No objections, other than to id it 'HS-core' (LuaSkin is meaningless to users, and hammerspoon too generic (and long)); I should also point out (just in case) that if the proposed implementation uses shortcuts to "require" and use `hs.logger` before the user's `init.lua` is loaded (in MJLua or _coresetup), any logging there will be lost because I'm defaulting the history to disabled; so if that's desirable for some reason, either history must be enabled by default, or some other workaround is necessary.
Lastly, after quickly looking at `hs.speech`, I don't understand why you're using the 'f' method variants since it seems that all string formatting is done in-situ via `stringWithFormat` (which is appropriate).

asmagill:
Unless you've changed something recently, when I last looked at the source for the module (admittedly a while back), the non-f versions were wrappers that called the f versions... thus calling them directly is one less function call in the path.

lowne:
Correct, but they expect a format argument: `log.i(s)=>log.f('%s',s)` - which I can't find in `hs.speech` (I'm probably just misreading line 38). Speaking of performance, Lua's overhead seems to be minimal anyway: if I call `logger.printHistory(...)` right after a config reload, it takes north of 3 seconds, if I store the string on another table and take out the `print`, it takes 0.05s - afaict the bottleneck must be in OjbC's logmessage.

asmagill:
(moving the conversation from #699 back to where it belongs)

Your argument re the '%' in the text is a good one... I'll switch things over (I think at atm the only core module affected is hs.speech, though I'll probably switch it to the LS method, once its in place)

As a side note, it is a little confusing having both, especially as I see it (rightly or wrongly) as an analog to print, and lua doesn't have a printf, you're expected to use print(string.format(...))... I go back and forth as to whether I consider that a "Good Thing"™, so not a comment/suggestion as much as an observation.

I'll review the docs more closely later and perhaps make some suggestions to make the distinctions more clear, if they're not in the present incarnation -- I admit, its been a while since I've looked closely at hs.logger's details.

And if I came across as defensive, consider it lack of sleep, too many holiday visitors, etc... I didn't take it badly, I just thought I was being clever removing an (as I saw it) unnecessary step.  But I had completely forgotten about a raw % in the text.

asmagill:
(incidentally, the C-API does have a `lua_pushfstring`, so it is curious that it doesn't have a `printf`)

lowne:
I find `print(string.format(...))` (and having to liberally sprinkle `string.format`s around in general) quite annoying, so despite my Lua-purity advocacy I wouldn't object to exposing a global `printf`; similarly, I exposed the 'f' variants in logger. (Typing aversion for frequently used stuff is also the reason the methods are called `d`/`df` etc., instead of `debug`/`debugf`)
The docs are admittedly a bit sparse (to be charitable) so improvements are welcome.

cmsj:
+1 a global printf(), I hate doing string.format or "foo"..bar

lowne:
Question: should the log history survive reloads (which requires persisting it)?

asmagill:
You can use something like

~~~lua
local tmp = hs.crash.crashLogToNSLog()
hs.crashLogToNSLog(true)
hs.crash.crashLog("whatever")
hs.crashLogToNSLog(tmp)
~~~

to put things into the system logs available via the Console application, so I don't see this as a "must have" option... we'd either need to add aging or a size limit to keep it from getting too large anyway, so unless someone has a strong opinion, I wouldn't bother.

lowne:
The history table is already size-limited to whatever is passed to `logger.historySize()`, with a hard ceiling of 10k entries (which might need increasing), so we could just dump/retrieve the table as is.

asmagill:
A thought re `printf`: might it be more appropriate to make it `hs.printf`?  Making it printf "pollutes" the lua global namespace, which we're generally trying to avoid.  It also makes it obvious that the command isn't part of Lua, but rather a Hammerspoon specific addition.

It's simple enough in a module or even a users `init.lua` to add `printf = hs.printf` for simplicity when typing (tbh, I probably will, if the name change occurs, much like I do with `hs.inspect`).  

No strong feelings one way or the other, but did feel I should point out that it is a divergence from our normal practice.

cmsj:
I think hs.printf is probably best, with a Notes suggestion to alias it into the global namespace

cmsj:
Looks like Travis failed transiently - it was in the setFrame() UI test. Perhaps they are not as reliable as I hoped :(

In theory, closing the PR and re-opening it should trigger a rebuild.

lowne:
@asmagill yep
@TravisPlusTheTestWriter(s) very regression such awesome 10/10 would test again
Although watching the build log go, I'm always reminded of [this](http://downlode.org/Etext/nine_billion_names_of_god.html), which I read decades ago - skimming through it right now I immediately found an apt quote to justify my feeling, as it describes the experience with utter accuracy:
> For weeks now the Mark V had been churning out acres of sheets covered with gibberish

(I'm merely following in asmagill's footsteps with the scifi references, so blame him 😀)

Issue #698, opened 2015-12-20
LuaSkin and throwing a lua error


asmagill:
Please someone tell me I'm wrong, but I think we're going to have to rethink/refactor a large chunk of LuaSkin.  I know it's a little long, but please bear with me...

Lua's error functions (any of `luaL_check*`, `luaL_error`, `luaL_argerror`, and `lua_error`) do not return.  In the context of the console or when running any lua code in the Lua state, this isn't a problem, as the whole environment is written around knowing this.

But we also use LuaSkin in Objective-C delegates, Objective-C blocks, etc.

To give an example of what I am fearing this means, imagine the delegate for webview... the user clicks on a link, and for whatever reason it fails... this triggers a call to the WKWebView delegate for the `webView:didFailNavigation:withError:` message, which, if defined, calls a lua callback function.  Since I wrapped that in `protectedCallAndTraceback:nresults:` an error in the callback doesn't matter, but what does matter is that the delegate method looks at the results returned by that callback, and if it is a string, assumes it's HTML and displays it as the error page... 

Actually, this is a bad example because I don't do what I'm about to describe, but it was the first delegate I could think of in enough detail without pulling up the source code... 

If I were to verify that this was a string with `luaL_checktype(L, -1, LUA_TSTRING)` or if for example, I was expecting a rect table instead of a string and used `[skin tableToRectAtIndex:]` and the table was malformed, `luaL_check*` functions in LuaSkin would, if I understand how this works correctly, _never return to the delegate method_ so whatever else the delegate has to do to cleanup after itself wouldn't happen...  I don't know if this will lead to a crash, necessarily, but I fear it could in some cases.

Sometimes in our C code, we call a lua c-function, which currently LuaSkin can't wrap, so if I don't use `lua_pcall` for these (and we don't in a couple of places I came across recently) then an error thrown by `[skin checkArgs:]` could also cause this...

Basically, if I'm understanding the problem correctly, _any_ use of a lua error function in LuaSkin is potentially vulnerable to causing this kind of situation.

Please tell me I'm missing something or mis-understanding something.

Modifying the `tableTo*` methods and `toNSObjectAtIndex:` and `luaObjectAtIndex:toClass:` methods to silently return zeroed structures or NSNull is one option to fix those, but I like that they throw errors when used within the console or in the evaluation/execution of lua files.

Another option is to add a flag to every LuaSkin method that indicates whether terminating errors can be thrown or not.

[...]

PR #705, opened 2015-12-24
[WIP] Lua skin error refactor

asmagill:
* still need to make core logging methods "delegatable"
* document new LuaSkin methods
* test the &%$#$&%^$ out of this (though in theory, it *should* crash *less* in odd places)

but so far it looks promising.

Not sure if we want to leave tracebackWithTag: in LuaSkin... for now I plan to use it to test some of the questions concerning stack boundary checking in the push methods, but it may be useful enough to leave in...

### DIFF
--- a/Hammerspoon/MJLua.h
+++ b/Hammerspoon/MJLua.h
@@ -1,5 +1,14 @@
 #import <LuaSkin/LuaSkin.h>
 
+@interface MJLuaLogger : NSObject <LuaSkinDelegate> {
+    lua_State *_L;
+}
+
+@property (atomic, readonly) lua_State *L;
+
+- (instancetype)initWithLua:(lua_State *)L  ;
+@end
+
 void MJLuaAlloc(void);
 void MJLuaInit(void);
 void MJLuaDeinit(void);
--- a/Hammerspoon/MJLua.m
+++ b/Hammerspoon/MJLua.m
@@ -14,6 +14,9 @@
 static LuaSkin* MJLuaState;
 static MJLuaLogger* MJLuaLogDelegate;
 static int evalfn;
+
+static lua_CFunction oldPanicFunction ;
+
 int refTable;
 
 static void(^loghandler)(NSString* str);
@@ -385,6 +388,14 @@ void MJLuaReplace(void) {
 
 # pragma mark - Lua environment lifecycle, low level
 
+static int MJLuaAtPanic(lua_State *L) {
+    CLS_NSLOG(@"LUA_AT_PANIC: %s", lua_tostring(L, -1)) ;
+    if (oldPanicFunction)
+        return oldPanicFunction(L) ;
+    else
+        return 0 ;
+}
+
 // Create a Lua environment with LuaSkin
 void MJLuaAlloc(void) {
     LuaSkin *skin = [LuaSkin shared];
@@ -392,6 +403,7 @@ void MJLuaAlloc(void) {
         [skin createLuaState];
     }
     MJLuaState = skin;
+    oldPanicFunction = lua_atpanic([skin L], &MJLuaAtPanic) ;
 }
 
 // Configure a Lua environment that has already been created by LuaSkin
--- a/Hammerspoon/MJLua.m
+++ b/Hammerspoon/MJLua.m
@@ -42,7 +42,7 @@ void MJLuaSetupLogHandler(void(^blk)(NSString* str)) {
         NSArray *stateLabels = @[ @"OK", @"YIELD", @"ERRRUN", @"ERRSYNTAX", @"ERRMEM", @"ERRGCMM", @"ERRERR" ] ;
         CLS_NSLOG(@"logForLuaSkin: error, state %@: %s", [stateLabels objectAtIndex:(NSUInteger)errState],
                                                          luaL_tolstring(_L, -1, NULL)) ;
-        lua_pop(_L, 1) ;
+        lua_pop(_L, 2) ; // lua_pcall result + converted version from luaL_tolstring
     }
 }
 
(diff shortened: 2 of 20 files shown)


asmagill:
addresses #698

lowne:
No issues so far for me (but I haven't been looking for them, just running my config). I suppose that my ancient "terminate event processing on error" proposal (or at least the easy way to implement it, in LuaSkin itself) is definitively ruled out now because of the need to keep LuaSkin "pure", correct?

asmagill:
Not sure I remember the behavior of your proposal... do you recall which issue it is described in or care to describe it again?  The main goal of this pull/wip is to ensure that any activity which invokes lua from "the outside" completes properly and isn't left in a half-completed state because of a non-returning lua error.  If they should be wrapped and a flag set so an appropriate error condition or state is acted upon once the Hammerspoon application loop is back in control of the Lua state, it may be possible.

lowne:
It's at the bottom of #477: no more callbacks (optionally) once an error is encountered, while preserving the Lua state for live diagnosis. Yes, it'd require wrapping. I don't know the details of what the behaviour is with this PR; perhaps just a(nother) C-to-Lua (global) hook to notify Lua that an error occurred within a callback would be sufficient.

asmagill:
Still needs documenting, and testing, especially with contrived errors to make sure behavior is more predictable as expected.

All error methods in LuaSkin are wrappers to a single delegate method, defined in MJLua.m, which calls `hs.handleLogMessage`... This is defined in `hs._coresetup` so we can easily expand/change/replace it without requiring a full recompile.  Currently all levels print to the console with all levels except verbose (5) also logging via `hs.crash.crashLog`, and level error (1) invokes `hs.showError` instead of printing (since it also prints to the console).  I figure this makes it easy to wrap with hs.logger later if desired, but for testing I want to see everything.

I'm experimenting with a wrapper in Lua which also attempts to attach file and function information via luaL_where, but I'm not sold on it yet... it may end up being more reliable (if more verbose) to just tack debug.traceback onto certain levels, like warn and error, but if it works, it's imo a little cleaner.

The toNSObject and pushNSObject methods now use the options as described in #698 with one exception... if self-reference is not allowed (by default it isn't) when converting a lua table into an NSArray or NSDictionary, it doesn't return nil for the entire conversion; it returns [NSNull NULL] for that object... what this means is that this Lua code:

~~~lua
a = {}
a[1] = a
a[2] = 2
~~~

will convert to this obj-C object:

~~~objc
@[ [NSNull NULL], @(2) ]
~~~

I don't know what the implications of this wrt to things like `hs.settings`, etc. are yet, but it was easier to code... I think I'm going to have to bite the bullet and do as described in #698, however... but I'm getting a little glassy eyed right now and need a break away for a bit... and the rest of it needs testing as well, and it compiles as is right now and at least loads my config without problem, so that's a start! :-)

asmagill:
@lowne, re "terminate event processing"...

My gut tells me that there are just too many different ways Lua callbacks get called to provide safe way to blanket terminate them all, and adding the ability for exceptions (like hotkeys) just makes it harder... 

And I'm still not sold that its a good idea in general... I certainly understand it when debugging certain things, but in general if 90% of my setup works, I want it to keep working, even if something else isn't right now (for example, a couple of my background NSTasks don't handle it well when I have no network access... I know I need to fix them due to the crap that keeps popping up in my console, but since everything else keeps working, it's not a priority at the moment since it *all* works *most* of the time.)

As to it being "easier" to do in LuaSkin, there is nothing that requires a module to use LuaSkin except convention (and maybe us "requiring" it in potential additions to core, but since LuaSkin itself is also a moving target, and not all modules are using it yet, "how much use" becomes a question too)... while the intent of LuaSkin is to make things easier and faster for the developer, we had successful modules before it was around that had callbacks, so even if we did stick code in there, it would be an incomplete solution at best.

The `hs.handleLogMessages` at least gives you a hook to set a flag when a certain error level occurs, but I think the only way to safely implement the event stoppage would be with a wrapper in each of your callbacks... plus that gives you complete control over your exceptions.

If you do go this route, I'd like to hear your experiences... as I said above, my gut thinks it a bad idea in general except maybe for specific situations, but... I'm willing to be proven wrong.

And the purity question... yeah, if there are plans/designs/ideas about LuaSkin being used outside of Hammerspoon, then purity does become important... @cmsj would have to chime in on that.  It's also good practice in general if a framework can stand on its own (as much as possible) because someone else down the road can always think of a use that we haven't.

asmagill:
~~~lua
> a = {}
> a[1] = a
> a[2] = 2
> inspect(a)
<1>{ <table 1>, 2 }
> hs.settings.set("test",a)
*** WARN:    [string "return hs.settings.set("test",a)"]:1: :lua table cannot contain self-references
[string "return hs.settings.set("test",a)"]:1: NSInvalidArgumentException: Attempt to insert non-property list object (
    "<null>",
    2
) for key test
> 
~~~

At least settings doesn't crash with an exception... that was my primary fear.

asmagill:
Discussion question: 

By default (this will be adjustable via options in the module code to the appropriate LuaSkin methods) should unknown types be an error or silently ignored? (for Lua->NSObject, this would be userdata which doesn't have a converter registered, for NSObject->Lua this would be any NSObject not caught by a registered convertor or the default types)

Current implementation (i.e. what is in the master branch) for Lua->NSObject uses luaL_tolstring to call the `__tostring` method and returns an NSString; NSObject->Lua uses the equivalent of `[obj debugDescription]` to return a lua-string, and these would still be available as options, just not the default when no options are provided.

cmsj:
@asmagill I think if we don't have a converter for an object, we should probably error out

asmagill:
@cmsj, because these methods can be called outside of code which can be guaranteed to be wrapped with lua_pcall, the best we can do is return nil and hope the receiver is checking for that (it really should, anyways, since nil is always a possible return value for anything expecting a type of `id`.)  So, that's the path I've taken as default.  There are options which will return the description instead and which will attempt to actually use a [NSNull null] placeholder, but you have to specify them.

So far, the number of places where the previous default behavior was preventing an exception from being thrown has been pretty low (except when crafting bad data on purpose, of course 😁) and as hoped, the logs generated have been specific enough I could easily find where to fix it right where the problem was, rather then hip deep in lua internals or lua_destroy called who knows how many instructions later, so I'm happy with the results so far.

cmsj:
@asmagill excellent, good point about the pcall stuff

PR #710, opened 2016-01-02
More event types for hs.caffeinate.watcher

fanf2:
These patches allow you to watch fast user switching and screensaver events. My motivation is to be able to automatically delete keys from the ssh-agent when the screen is locked, or when switching users, or when the machine sleeps.

The new events are:

 * sessionDidResignActive
 * sessionDidBecomeActive
 * screensaverDidStart
 * screensaverWillStop
 * screensaverDidStop
 * screensDidLock
 * screensDidUnlock

The first two are related to fast user switching.

### DIFF
--- a/extensions/caffeinate/watcher.m
+++ b/extensions/caffeinate/watcher.m
@@ -79,23 +79,23 @@ typedef enum _event_t {
     }
 }
 
-- (void)applicationDidWake:(NSNotification*)notification {
+- (void)caffeinateDidWake:(NSNotification*)notification {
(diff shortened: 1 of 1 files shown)


asmagill:
+1, haven't had a chance to test yet, but I've been thinking about wanting to know when the screen saver starts/stops but hadn't looked into what it would take.  Look forward to checking this out later this week!

cmsj:
Thanks, sorry it took longer than expected to merge

PR #713, opened 2016-01-06
Add support for mapping userdata tags to NSObject subclasses, allowin…

cmsj:
…g toNSObjectAtIndex to converted supported userdata objects in tables to NSObjects

This is definitely one for the eyes of @asmagill :)

The idea here is that I want to be able to bridge a Lua table which contains hs.image objects, into an NSDictionary without walking it and converting each thing individually, so I've taught the toNSObjectAtIndex parts of LuaSkin to be able to inspect userdata and a pre-supplied store of USERDATA_TAG to NSObject class mappings.

This depends on 3d63eb65351431675a10469c59947399113ac49f for storing USERDATA_TAG in the userdata objects, which is already in master because it didn't seem as controversial as this :)

### DIFF
--- a/LuaSkin/LuaSkin/Skin.h
+++ b/LuaSkin/LuaSkin/Skin.h
@@ -393,11 +393,21 @@ typedef struct luaObjectHelpers {
 /*!
  @abstract Register a luaObjectAtIndex:toClass: conversion helper function for the specified class
 
- @important This method registers a converter functions for use with the luaObjectAtIndex:toClass: method for converting lua tables into NSObjects
+ @important This method registers a converter function for use with the luaObjectAtIndex:toClass: method for converting lua tables into NSObjects
  @param helperFN - a function of the type 'id (*luaObjectHelperFunction) (lua_State *L, int idx)'
  @param className - a string containing the class name of the NSObject type this function can convert
  */
 - (void)registerLuaObjectHelper:(luaObjectHelperFunction)helperFN forClass:(char *)className ;
+
+/*!
+ @abstract Register a luaObjectAtIndex:toClass: conversion helper function for the specified class and record a mapping between a userdata type and the class
+
+ @important This method registers a converter function for use with the luaObjectAtIndex:toClass: method for converting lua tables into NSObjects. It builds on registerLuaObjectHelper:forClass by also storing a mapping between the NSObject class and Lua userdata type so userdata objects of this type can be automatically converted
+ @param helperFN - a function of the type 'id (*luaObjectHelperFunction) (lua_State *L, int idx)'
+ @param className - a string containing the class name of the NSObject type this function can convert
+ @param userdataTag - a string containing the Lua userdata type that can be converted to an NSObject
+ */
+- (void)registerLuaObjectHelper:(luaObjectHelperFunction)helperFN forClass:(char *)className withUserdataMapping:(char *)userdataTag;
 
 /*!
  @abstract Convert a lua geometry object (table with x,y,h, and w keys) into an NSRect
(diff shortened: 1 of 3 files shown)


cmsj:
(See particularly the FIXME. I'm not sure where the right balance is here - I want this to be safe, but I also want to have the minimum possible impact on the Lua stack)

asmagill:
Looks good, I'll try and add it to my work on #705 / #698 later tonight and give it a go to make sure it doesn't break anything, but it doesn't look like it should.

The internal documentation is a little misleading referring to converting lua tables, but then it's also misleading for the method you patterned it off of... I plan to completely re-write these documentation entries when I add docs to my additions in the mentioned WIP, so don't worry about it.

As to your FIXME line, I think it looks fine... you don't use lua API functions which might throw a lua error, and you do clean up after yourself -- unfortunately a lot of the data conversion methods can and do occur outside of a protected call, so the best we can do is make sure we use the absolute index, stay away from functions which might end in an error, and clean up after ourselves if we do anything which adds to the stack.

It's still possible to introduce problems in the actual helper functions, but that's always been possible and we can't wrap them with lua_pcall, since they don't follow the lua_CFunction template, so I've been reviewing those as well and will continue to do so...

cmsj:
@asmagill thanks :)
(fwiw, the specific use case for this is passing a table of data to the Chooser module, which includes an optional hs.image object for each row - pretty much just like the Alfred UI. I'm still making progress on hs.chooser, but it's slow going)

(also I haven't forgotten about the hs.task input stuff, that's next on my list after hs.chooser)

asmagill:
Sorry, didn't get a chance to look at yet, but I like the idea... one of the weaknesses of converting Lua objects into NSObjects when they are in a table has always been that non-basic types weren't easily converted because we can't easily tell (without adding a label like you do for userdata metatables registered through LuaSkin) if a table or userdata is a specific class or structure or just a regular ol' NSArray or NSDictionary...  hence the `luaObjectAtIndex:toClass:` method, but as you demonstrate, it doesn't work on elements within a table without writing your own loop.

Of course, this only works with userdata objects and only ones registered through LuaSkin, but that's still worlds better than before.

I've gone back and forth over this in my head, but since it's the mirror of your update for tables, what do you think about the idea of checking for a label in tables so we can auto-convert things with `toNSObjectAtIndex:` like a table such as `{ red = 1, blue = 1, __type="hs.drawing.color" }` and convert it into an NSColor automatically as well?  It puts the onus on the lua side code to make sure to include these when manually crafting such a table (we can easily add the type label in the functions which return such tables from NSObjects, so "machine generated" tables will already be labelled)... ultimately I've not bothered pursuing the idea because you'd still end up having to include code when writing a module function which checks for the `__type` field in the table and either return an error or call `luaObjectAtIndex:toClass:` anyways if its missing.  But since the door is open, what are your thoughts?

I'll see how this merges with the updated #705 tomorrow (I added documentation this evening and I need to go through it again and check for typos, etc. when I'm fresh, but I think it's almost ready for prime time) and see if there are any conflicts.

cmsj:
@asmagill that's an interesting idea. I worry a little that it would mean that if people aren't using our helper functions to create the tables, behaviours inside modules/LuaSkin would start to subtly change. I'd say let's keep the ideas separate the ideas for now - userdata seems like an easy win, we can register all (well, window/application/uielement are still not using LuaSkin because they are weird) userdata objects automatically with a `__type` and then slowly opt-in to auto-conversion as we find we need it.

Adding type hinting to tables can be explored separately to see how much of a win it is. I almost wonder though, if it might really be better to be switching the semi-opaque table structures into userdata objects.

One other thing I'm wondering about, is whether `__type` is too generic and liable to clash with something else. This is entirely hidden from users, so maybe we should be more verbose? `__HSuserdataTypeHint` or something?

PR #714, opened 2016-01-07
[WIP] Lua skin error refactor plus713

asmagill:
Minor merge conflicts between the two (in comments and internal functional prototypes), plus changed register support functions to return a BOOL indicating whether or not the registration of the helper function succeeded or not.

I put this up for people to test with now, if desired.  This may (or may not) replace #705 and #713 if there are no significant changes to either and they are both deemed "merge-worthy" at roughly the same time.


LuaSkin/LuaSkin/Skin.m
```diff
@@ -558,11 +621,50 @@ - (BOOL)isValidUTF8AtIndex:(int)idx {
...
+
+    return [[NSString alloc] initWithData:dest encoding:NSUTF8StringEncoding] ;
+}
+
 - (BOOL)requireModule:(char *)moduleName {
     lua_getglobal(_L, "require"); lua_pushstring(_L, moduleName) ;
     return [self protectedCallAndTraceback:1 nresults:1] ;
 }
 
+// FIXME: Is this actually used anywhere or a leftover from a removed NSTask hack?
```
cmsj: It's not used anywhere, I'm fine with removing it

cmsj:
This is looking great! I'm going to rebase my hs.chooser work on top of it, so I can implicitly test it by running it for a few days

cmsj:
I've not run into any issues so far :)
 @Asmagill - what do you still want to do on this before it lands?

asmagill:
I'll look over it one last time tonight, but I think it's basically ready...

asmagill:
The main thing left is to determine what the default behavior is for each log level and update `_coresetup/init.lua` accordingly.  The levels match that of hs.logger, so we could create a "general" or "Skin" instance of hs.logger and pass them along to it.

What levels should also be logged with `hs.crash.crashLog` and/or `hs.showError` as well?

FYI: I will be without reliable internet access for the next 3 or 4 days, so if you want this merged sooner, you or @lowne should go ahead and make the changes... unless you've uncovered anything recently that I have missed, I'm comfortable enough with the Objective-C side to go ahead and merge it once the logging details are worked out.

cmsj:
FWIW, I think that however we handle the logging, it probably shouldn't go through Lua, at least for error logging - we should be avoiding touching the Lua stack in error situations, imo. I'm a bit short on time at the moment, but if nobody else gets to it first, I don't mind doing something here.

asmagill:
@cmsj, I'm not sure I completely agree... if we leave lua completely out of the equation, then why bother having levels?  The only thing we can log completely without Lua interaction is via NSLog or crashlog.  

If we're going to end up calling hs.showerror or print_to_console (i.e. mimic what was in hammerspoon.h) then I'd rather have a helper that handles it so it can be customized or to change the level of output when debugging versus running in production.  Of course we can (and probably should) still log via crashlog behind the scenes, either for all levels or for just some.

cmsj:
@asmagill I'm happy to go with that for now, but longer term I think we should have error logs going outside of Lua and being handled by ObjC, simply to make them more reliable in the face of Lua weirdness, however, at this stage I'm really just interested in getting this all landed and a release cut :)

PR #703, opened 2015-12-23
[WIP] menubar, styledtext, image additions

asmagill:
Adding support in hs.menubar to use hs.styledtext and/or hs.image objects as menu items.

hs.image can create an image copy at a specific size (scaled or absolute) for use in hs.menu, to save screen/window shots at specific size, etc. (places where a "bounds" isn't set like it is with hs.drawing)

1. Needs more documentation,
2. testing on Retina (I make some assumptions about the best image size for onStateImage, offStateImage, and mixedStateImage that I need to test later this week when I have access to a machine with a Retina display),
3. and possibly extending hs.pasteboard as well to handle additional types (which will tie in closely with hs.styledtext and hs.image, hence the possible combination with this update)

PR #717, opened 2016-01-08
Add "Run in Hammerspoon" to OS X Services menu

asmagill:
Adds a service which allows you to select text in any Services aware application and tap Cmd-Shift-H (by default) to execute it in Hammerspoon and return the results.

You may need to log out and log back in to get the service to show up when you select text... you could also try `/System/Library/CoreServices/pbs -flush` in the terminal.

I wasn't able to find a way to enable the service automatically... I also had to go into System Preferences->Keyboard->Shortcuts->Services and enable "Run in Hammerspoon" in the Text section.

Since adding a service menu item seems to require editing the application info.plist, it seemed best to go ahead and stick it in MJLua.m/MJLua.h.

Would it be useful/interesting to also have a service which can call a user-defined callback function?  As far as I can tell, each service registered has to be described in an info.plist file somewhere, either in the application or in a bundle installed into Library/Services, so it would be just one callback or have to decide which callback function to invoke based upon something in the selection itself (some of the text, file extension if we add support for files, etc.)

(if I'm wrong about the plist requirement, please chime in, because I'd love to be able to dynamically add them via code in a users init.lua file... I just can't find a way to do it with the searching I've done.)

### DIFF
--- a/Hammerspoon/Hammerspoon-Info.plist
+++ b/Hammerspoon/Hammerspoon-Info.plist
@@ -86,6 +86,33 @@
 	<string>MainMenu</string>
 	<key>NSPrincipalClass</key>
 	<string>NSApplication</string>
+	<key>NSServices</key>
+	<array>
+		<dict>
+			<key>NSKeyEquivalent</key>
+			<dict>
+				<key>default</key>
+				<string>H</string>
+			</dict>
+			<key>NSMenuItem</key>
+			<dict>
+				<key>default</key>
+				<string>Run in Hammerspoon</string>
+			</dict>
+			<key>NSMessage</key>
+			<string>hammerspoonAsService</string>
+			<key>NSPortName</key>
+			<string>${EXECUTABLE_NAME}</string>
+			<key>NSSendTypes</key>
+			<array>
+				<string>public.utf8-plain-text</string>
+			</array>
+			<key>NSReturnTypes</key>
+			<array>
+				<string>public.utf8-plain-text</string>
+			</array>
+		</dict>
+	</array>
 	<key>SUEnableAutomaticChecks</key>
 	<true/>
 	<key>SUFeedURL</key>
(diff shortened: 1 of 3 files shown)


cmsj:
So this is interesting. I'm not sure I'd ever want to select text and run it in Hammerspoon, but it'd be a fun feature to have. I think I'd also want a service that can just apply to anything, which connects to a user supplied callback. It's a real shame that we can't register the services dynamically though, since the callback might not be connected.

asmagill:
As I've been thinking about it, I *think* that we might be able to write a module which can take a single compiled bundle (not sure exactly what the compiled portion of the bundle will entail yet) but dynamically create the info.plist with a unique NSUserData string which can be used to lookup a callback function... 

Creating a new service would require duplicating the bundle (with the modified plist) in ~/Library/Services and calling NSUpdateDynamicServices()... and cleaning them out when Hammerspoon isn't running or a service is "deleted"... not sure about that yet either.  Maybe use the recently added termination callback feature...

The problem is that the name of the service, the types it accepts/returns (if any), the key binding, whether or not it accepts files (well, technically an NSURL referring to the file)... all of this is defined in an info.plist file, not in function calls we can perform as needed... at least I haven't found any yet.

We'd still need to register the service provider, probably still in MJLua, since an application can only have one, so if a callback for a specific tag is no longer registered, the service simply does nothing.

Executing Hammerspoon code is, I admit, not the most useful thing I can think of, but remember it's executing it as if you typed it in to the console... anything you can do there, can be done in any place you can type text now.  I can see it being useful when writing documentation and not quite remembering if I have the syntax correct... or executing a predefined function which has side effects I want without having to pull up the console... but yeah, not a "killer feature" :-)

At any rate it was a nice diversion from LuaSkin for a short while, and I'll see if I can automate creating service bundles so we can make it more useful probably next week.

asmagill:
I'm going to close this one until I have a chance to determine what it will take to allow dynamically added services with user defined callbacks.  This may also benefit from furthering #703 with the pasteboard additions I have in mind, since NSPasteboard is used by services to pass information back and forth.

PR #725, opened 2016-01-14
LuaSkin updates missed in recent merge

asmagill:
Includes documentation corrections and registration return code additions missed in #720 multi-merge


cmsj:
Did I miss these when merging? It was not fun to land the mega PR made up of so many others, I can believe I missed something :/

asmagill:
Yes, these were missing when I grabbed master last night... minor things... since we've removed throwing a a lua_error for most LuaSkin things, the registration functions now return a boolean indicating success or failure and a few documentation tweaks.  I think these were maybe the last things I added.

Otherwise, looks good!

cmsj:
Apologies. It would have been much safer to merge the individual PRs in order, instead of trying to be clever and land the last one while squashing it into fewer themed commits :/

Issue #737, opened 2016-01-19
improve hs.pasteboard image handling


heptal:
Right now 'getImageContents' only works for PNG data, which is fine for screenshots directly to clipboard. It seems copying in other contexts loads it onto the pasteboard as "public.tiff" as shown by `hs.inspect(hs.pasteboard.contentTypes())`. TIFF is apparently being used as a generic container for 'image data that is not PNG or not PNG loaded via screenshot-to-pasteboard or a few other techniques such as hs.pasteboard.setImageContents', or, 'pretty much all image data with a few exceptions'. 

See [How do I put a GIF onto an NSPasteboard?](http://www.cocoabuilder.com/archive/cocoa/151166-how-do-put-gif-onto-an-nspasteboard.html)
>The usual way is to use an NSFilesPromisePboardType, which was a longer
discussion in this list a long time ago.
Or you can write the gif data to the filesystem and use the
NSFilenamesPboardType.
Or you cheat and write the gif data to the NSTIFFPboardType

It seems Apple has taken the 'cheating' method instead of creating pasteboard types for all image formats. Today it is called `NSPasteboardTypeTIFF`. 

I intend on tackling this one unless @asmagill has done any work on supporting more pasteboard types as I saw him mention awhile back.

asmagill:
I started playing with NSPasteboard's `readObjectsForClasses:options:` and `writeObjects:`, which allow directly getting and putting Cocoa framework classes which support the NSPasteboardWriting and NSPasteboardReading protocols (NSImage does) onto the pasteboard without a whole lot of hassle, but I haven't gotten to NSImage yet (hoping to tonight or tomorrow)...  Adding support for promised types does not look trivial if we want to make it extendable and not just for a few types we hand-select, but I may be missing something.

I know we can load a GIF image from a file into hs.image, which is at heart just an NSImage object, so I'm hopeful that using `writeObjects:` may make this easier, but (if I understand correctly, which is a big if) this only works if the receiver can take an NSImage and apply it in the way it wants... for testing purposes, do you have a specific use case I can try out (an application you are getting from or sending to)?

I'll push what I have tonight if you want to try it out yourself or extend it... so far, I'm just adding what I'm referring to as "experimental" functions/methods to hs.pasteboard, so it shouldn't break anything you already have in place.

heptal:
GIF Information gets lost going to or from the pasteboard

[...]

PR #744, opened 2016-01-20
[wip] hs.pasteboard additions

asmagill:
This is being added at this time for review and possible testing by @heptal for #737... new methods do not include Hammerspoon documentation strings yet.

@heptal, check out `readImage` and `writeObjects` and see if the generic NSImage support is sufficient for your GIF needs... if not, can you provide more details on the applications you are using or your specific needs?  I've been able to successfully load a GIF image with `hs.image` and paste it into Word and TextEdit, and retrieve a GIF images copied from Preview.

Note that while this module takes advantage of additions provided by #740 and #741, they are not required for compiling -- you just won't be able to get or write `hs.styledtext` or `hs.sound` objects until they are also added -- they are not required just to test the `hs.image` support.

* `deletePasteboard(pasteboard)` - verifies that a system pasteboard is not named

Potential additions:

* `allContentTypes([pasteboard])` - returns an array of arrays; each element array contains the content types for each item on the pasteboard.  When only one item is on the pasteboard, is equivalent to `{ hs.pasteboard.contentTypes() }`
* `uniquePasteboard` - returns the name of pasteboard guaranteed to be unique with respect to other pasteboards on the computer.

Additional additions -- note that these types take advantage of the fact that `writeObjects` and `readObjects` work with classes that support the `NSPasteboardReading` and `NSPasteboardWriting` protocols without requiring additional code or use of promised data.  If other types exist that support these, they can be easily added.  Other classes or data types may require additional support to include and need to be evaluated on a case-by-case basis:

* `typesAvailable([pasteboard])` - returns an array with one or more of the following keys set to true, if an element of that type is available on the pasteboard: `string`, `styledText`, `sound`, `image`, `URL`, or `color`
* `readString([pasteboard], [all])` - returns the first (or all in an array, if the parameter `true` is provided) element on the pasteboard which can be expressed as a string (NSString class)
* `readStyledText([pasteboard], [all])` - returns the first (or all in an array, if the parameter `true` is provided) element on the pasteboard which can be expressed as an `hs.styledtext` object (NSAttributedString class) -- requires #741.
* `readSound([pasteboard], [all])` - returns the first (or all in an array, if the parameter `true` is provided) element on the pasteboard which can be expressed as an `hs.sound` object (NSSound class) -- requires #740.
* `readImage([pasteboard], [all])` - returns the first (or all in an array, if the parameter `true` is provided) element on the pasteboard which can be expressed as an `hs.image` object (NSImage class)
* `readURL([pasteboard], [all])` - returns the first (or all in an array, if the parameter `true` is provided) element on the pasteboard of type NSURL, as a string (NSURL class)
* `readColor([pasteboard], [all])` - returns the first (or all in an array, if the parameter `true` is provided) element on the pasteboard of typeNSColor as an `hs.drawing.color` table (NSColor class)
* `writeObjects([pasteboard], object)` - writes the object or objects onto the specified pasteboard.  `object` can be a string, hs.styledtext object, hs.image object, hs.sound object, { url="..." }, or `hs.dwaring.color` table (e.g. { red=,green=,blue=,alpha= }) or an array (i.e. numerically indexed table) of any combination of the previously identified object types.  Note that most applications only look at the first entry on the pasteboard, though some will look for the first of a type they can use.  Generally passing multiple types at once is most useful for inter-application communication where both the sender and receiver know that they are getting a group of items.


asmagill:
added documentation and swapped parameter order for `writeObjects` to be more consistent with module format.

asmagill:
Before tweaking this and merging, the new functions replicate: 
`hs.pasteboard.getContents` -- mimicked by `readString`
`hs.pasteboard.getImageContents` -- mimicked by `readImage`
`hs.pasteboard.setContents` -- mimicked by `writeObjects(string)`
`hs.pasteboard.setImageContents` -- mimicked by `writeObjects(hs.image)`

My current thought is to leave things as they are, since the old functions are in use and well tested, at least for what they do (ok, so the image ones are newer, but still not as new as these additions).  After the new methods get a more complete workout, we can make a decision about deprecating the old and/or replacing them with wrappers to the new functions.

Other thoughts or suggestions?

cmsj:
+1 for caution wrt replacing the existing functions

heptal:
FWIW, after testing, I now consider get/setImageContents deprecated (and I wrote them) in favor of readImage and writeObjects.

For example, if I copy a non-PNG image in OSX, `getImageContents` will not work because it fetches data from the pasteboard with `dataForType:NSPasteboardTypePNG`. Even if it is a PNG copied from Safari, Messages, and other sources, I'll get "No valid image data in pasteboard" because OSX will almost always represent it on the pasteboard as a TIFF with the exception of a few cases like screenshot-to-pasteboard (which was my initial use-case). When I realized the current implementation was so limited, it prompted me to extend it, but these additions more robustly support `NSImage`. I would move to remove the existing functions or make them aliases of these new ones.

heptal:
I will note that it is kind of cool how each object on the pasteboard can have multiple representations. When you copy an image from Safari, `readImage` will return `hs.image` and `readURL` will return its associated URL (the URL within enclosing anchor tag if present, otherwise actual image URL).

Now, exploring this has presenting some interesting behavior when there are multiple data types on the image. Right-click->copy image in Safari to get such an image on the pasteboard.

```lua
--after copying my avatar (64x64 px) here in Safari

hs.inspect(hs.pasteboard.contentTypes())
--output
{ 
"public.tiff", 
"dyn.ah62d4rv4gu8zs3pcnzme2641rf4guzdmsv0gn64uqm10c6xenv61a3k", --Safari web history (found with mdfind)
"dyn.ah62d4rv4gu8yc6durvwwaznwmuuha2pxsvw0e55bsmwca7d3sbwu", --??
"public.url", 
"public.url-name", 
"public.utf8-plain-text", 
"com.apple.flat-rtfd" --rich text with attachments (RTFD) data
}

i=hs.pasteboard.readImage()
--hs.image:  (0x7ff9ca6c6e88)
u=hs.pasteboard.readURL()
--https://github.com/heptal
t=hs.pasteboard.readString()
--https://github.com/heptal
st=hs.pasteboard.readStyledText()
--hs.styledtext: ￼ (0x7ff9ca645d08)

d=hs.drawing.image(hs.geometry.rect(100, 100, 100, 100), i):show() 
d:delete()
--displays image as expected, scaled to 100x100 px 

d=hs.drawing.text(hs.geometry.rect(100, 100, 100, 100), u):show() 
d:delete()
--displays string as expected

d=hs.drawing.text(hs.geometry.rect(100, 100, 100, 100), t):show() 
d:delete()
--displays string as expected

d=hs.drawing.text(hs.geometry.rect(100, 100, 100, 100), st):show()
d:delete()
--expected URL or alt-text of some kind
--displays image unscaled as original 64x64 px without text
```
"com.apple.flat-rtfd" aka "rich text with attachments (RTFD)" made me try `readStyledText`.
The image was obviously (to me at least) unexpected. However it turns out this is totally normal, as `NSAttributedString` can have `NSTextAttachment`s and there is `+ (NSAttributedString *)attributedStringWithAttachment:(NSTextAttachment *)attachment` as seen [here](http://stackoverflow.com/questions/11204929/saving-nsattributedstring-with-nstextattachment-into-file-how-to), apparently invoked by `readObjectsForClasses`. Just thought that was interesting.

heptal:
I will also suggest that `readImage` check `NSImage`'s `valid` property upon reading from pasteboard and error out if false instead of returning `nil`.

asmagill:
The get/setImage functions are pretty new, only in the last two Hammerspoon releases I think, so it shouldn't break things (or at least not much) if removed.  I'd be much more leery of removing the get/setContents since they have been around so long and we know that there are people using them.

On the other hand, basic text and images are probably going to make up 90%+ of the use of hs.pasteboard, so wrapping them as special cases makes a certain kind of sense, too... unless there are objections, I'll probably go that route.

Re checking the valid property for readImage... was this on a pasteboard that claimed it had an image, but the image data was corrupt/invalid?  I can definitely see a reason for an error in that case; if the pasteboard has no image, however, I think nil is probably ok since nil is an accurate representation of the fact that there are no images on the pasteboard.

And your example above has given me something to mull over... the current writeObjects will take multiple data types, but adds them as separate entries in the pasteboard, not as differing representations of the same entry... I wonder how hard it would be to make them for the same entry (or more accurately, to support both options) to more closely match your example from Safari... will have to ponder and test before merging this...

heptal:
Yea, after all I suppose I agree that nil is fine if there is no actual image. It was just present in `getImageContents`. I also agree that `get/setContents` should stay as it's been there so long and the default case is text, but `get/setImageContents` should be removed as it is relatively new and now redundant. I suppose almost all images that are actually on the pasteboard would be valid since OSX seems to thoroughly alter image representation internally before putting it there.

I would encourage a sooner merge rather than later, as I found everything to work just fine (and I need fixes from both latest master and this merge).

heptal:
As far as writeObjects supporting 'multiphasic' objects as in Safari-copied images, perhaps the same API could be kept, but an array object containing a table like `{ myImageobject, "my associated string" }` could create the 'multiphasic' pasteboard entry. I don't think it is necessary for this merge, though.

Information on these custom data types with multiple representations is found [here](https://developer.apple.com/library/mac/documentation/Cocoa/Conceptual/PasteboardGuide106/Articles/pbCustom.html) and [here - NSPasteboardItemDataProvider](https://developer.apple.com/library/mac/documentation/Cocoa/Reference/NSPasteboardItemDataProvider_Protocol/index.html#//apple_ref/occ/intf/NSPasteboardItemDataProvider).

asmagill:
Hmm... yeah, dealing with a data provider and promised data... I thought I'd looked into multiple representations of the same type and decided to hold off on it, and now I remember why... I think if we're going to go that route, keeping the same interface might be a little challenging since it's needs a persistent object to act as the provider until the pasteboard owner changes again... screams new data type (i.e. userdata) to me, and I didn't want to go this route until a need arose.

Ok, I have a few things to clean up in this round (mostly in the docs) ( #740, #741, #742, #743, and this one) and will merge later today or tomorrow unless anyone sees any reason to hold off... @cmsj?

PR #747, opened 2016-01-21
Luaskin userdata struct abstraction

cmsj:
This adds userdata abstraction to LuaSkin by always having the module's userdata pointer be stored in a struct. This way we have a guaranteed separation of the management of memory allocation between Lua and CoreFoundation/Cocoa/ObjC/etc.

As a proof of what this looks like in use, this PR converts hs.chooser, hs.screen and hs.timer. Respectively, they were using their own struct, a pointer-to-NSObject-pointer and a struct containing a CFTypeRef.

In the case of hs.chooser and hs.screen, the NSObjects are given to LuaSkin, in the case of hs.timer, a new NSObject is created to encapsulate both the CFTypeRef and the associated properties that were previously in the userdata struct.

While this isn't a completely huge win, I think that simplifying the pointer-to-NSObject-pointer case alone justifies this, and over time I would expect all of the modules to a) migrate to this abstraction, b) use local NSObjects to encapsulate all their userdata state. For modules that already have an NSObject class relating to userdata, I'll migrate any associated properties into the class. For those that don't, I'll replace their existing structs with a class.

@asmagill I know you're a little unsure of this, but I hope it seems at least unobjectionable from the POV of simplifying another area of module<->Lua interaction :)

### DIFF
--- a/LuaSkin/LuaSkin/Skin.h
+++ b/LuaSkin/LuaSkin/Skin.h
@@ -363,7 +363,7 @@ typedef struct _userdata_t {
  @important You must have already registered an object with LuaSkin with the same userdata type that you pass to this method
  @important After calling this method, the userdata object will be at the top of the Lua stack
  @param type A const char * with the name of the userdata type. This is used to fetch the appropriate metatable to apply to the object
- @param obj A pointer to some data structure you wish to store in the userdata object
+ @param obj A pointer to some data structure you wish to store in the userdata object. Note that if this pointer is to an object tracked by ARC (e.g. something derived from NSObject), you should pass it in with a (__bridge_retained void*) cast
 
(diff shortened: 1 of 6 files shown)


cmsj:
I'm going to suggest pausing this for a while. It's starting to seriously concern me how many regressions we're getting in each release, so I think all refactoring efforts should be put on hold until we can figure out a better testing story

cmsj:
When I'm done with writing tests for hs.audiodevice, I'll move on to timer, screen and chooser, so we can have reasonable confidence about landing this PR, and I'll also think about how to add an implicit self reference, for @asmagill :)

cmsj:
I'm going to close this, I don't think it's useful anymore in its current form. More thought on this subject is required, but I still feel like we need to standardise our userdata handling.

Issue #752, opened 2016-01-22
More hs.image systemImageNames values


cmsj:
There seem to be quite a few undocumented image template names:

http://hetima.github.io/fucking_nsimage_syntax/

We should include them :)

asmagill:
I looked for such a list when first writing the module but didn't find a good one... I'll take a look later and see about including them... my main concern would be if any are OS version specific and whether that might cause a problem... I think since they're compiled in as strings, the concern is more with the compiling machine version than the running machine version (if you request an image name that doesn't exist, it either returns nil or an empty image -- forget which, but I know it doesn't crash).

PR #757, opened 2016-01-25
[WIP] add image names from #752

asmagill:
I'd like to test these to make sure that they are current... no reason to include things we know aren't in the latest OS X, plus it gives me an excuse to play with hs.chooser  😁

And, should they all be in one list or broken out into the arrays like in this pull?


cmsj:
I think it might be useful to have examples for iconForFileType - eg I don't know what an encoded hfs type is ;)

asmagill:
The text is mostly copied from the docs... It thought it was the 4 character sequence used to identify file types when HFS was first introduced (and resource forks were used)... but in testing, it doesn't seem to work with the few I remember... a quick google shows people using `NSHFSTypeCodeFromFileType` to generate the code from a more traditional file type, so maybe it has to be encoded in a special way... 

I'd be just as happy removing the text from the doc strings unless you think it should be explored further.

cmsj:
It sounds very legacy, so I think t can go. File extensions and UTIs seems plenty sufficient

cmsj:
@asmagill OOI, what's still WIP on this?

asmagill:
nothing, just forgot about it :-)  go ahead and land it or I will later today.

Issue #759, opened 2016-01-25
Character encoding display issue in console


heptal:
```lua
*** ERROR: hs.http callback error: The data couldn‚Äôt be read because it isn‚Äôt in the correct format.
stack traceback:
	[C]: in function 'hs.json.internal.decode'
```
Could be caused by improper NSLog formatting?
http://stackoverflow.com/questions/10940423/my-apostrophe-turns-into-Äô

Root cause from another SO answer:
>You have declared your file to be encoded with UTF-8, so your byte-string literal is in UTF-8. The curly apostrophe is U+2019. In UTF-8, this is encoded as three bytes, \xE2\x80\x99. Those three bytes are written to your output file. Then, when you examine the output file, it is interpreted as something other than UTF-8, and you see the three incorrect characters instead.

>In Mac OS Roman, those three bytes display as ‚Äô.

>Your file is a correct UTF-8 file, but you are viewing it incorrectly.

asmagill:
I'd need to see what the actual text is that the json decoder is choking on.  The issue with the ‚Äô characters is unrelated -- that's because the error message is being generated as UTF8 by `[NSJSONSerialization JSONObjectWithData:data options:NSJSONReadingAllowFragments error:&error];` in the `hs.json` module, but the code which `hs.http` uses to display it:


~~~lua
    if (![skin protectedCallAndTraceback:3 nresults:0]) {
        const char *errorMsg = lua_tostring(L, -1);
        [skin logError:[NSString stringWithFormat:@"hs.http callback error: %s", errorMsg]];
    }
~~~

treats the error message as a C-string (not inherently Unicode aware) rather then as an NSString (where encoding is assumed UTF8 unless otherwise specified).

heptal:
Here is the code that generated the error. The error occurred on the `local response = hs.json.decode(body)` line.
```lua
hs.hotkey.bind(hyper, "u", function() 
    local image = hs.pasteboard.readImage()

    if image then 
        image:saveToFile("/tmp/tmp.png")
        local base64 = hs.execute("base64 -i /tmp/tmp.png")
        local b64 = string.gsub(base64, "\n", "")
        b64 = hs.http.encodeForQuery(b64)

        local headers = {["Authorization"] = "Client-ID abc123"}

        local url = "https://api.imgur.com/3/upload.json"

        local payload = "type='base64'&image="..b64

        hs.http.asyncPost(url, payload, headers, function(status, body, returnedheaders)
            print(status)
[...]

PR #762, opened 2016-01-26
reset accumulated data upon new response in accordance with docs

heptal:
Occasionally, data corruption as described in #759 can occur if more than one `didReceiveResponse` message is received by `NSURLConnectionDelegate` (a retry if you will) and the accumulated `receivedData` buffer is not cleared because the metadata provided by the response (content-length, etc) is describing invalidated data. Clearing the buffer upon receipt of a new response should keep the `receivedData` buffer and the response metadata in sync and prevent data corruption.

From [Apple](https://developer.apple.com/library/mac/documentation/Foundation/Reference/NSURLConnectionDataDelegate_protocol/index.html#//apple_ref/occ/intfm/NSURLConnectionDataDelegate/connection:didReceiveResponse:)
>In rare cases, for example in the case of an HTTP load where the content type of the load data is multipart/x-mixed-replace, the delegate will receive more than one connection:didReceiveResponse: message. When this happens, discard (or process) all data previously delivered by connection:didReceiveData:, and prepare to handle the next part (which could potentially have a different MIME type).


heptal:
this one got dirtied with merge and unsynced with my fork...recreating

PR #763, opened 2016-01-27
Autocomplete: Show nothing if nothing to show

wvierber:
This leaves following error behind when you try nonsense.


cmsj:
Good catch, thanks!

PR #773, opened 2016-02-01
add hs.network submodules

asmagill:
Adds:

hs.network.reachability - check internet and network availability, includes callback when status changes

hs.network.host - ip-to-hostname and hostname-to-ip resolution in both blocking and non-blocking (callback) styles

hs.network.configuration - get network information from the Dynamic Store and monitor keys in store for changes


cmsj:
I think this is changing more than it should, in the .xcodeproj. The rest of the code looks fine though, so I'll land this manually.

asmagill:
@cmsj, are you still planning to come back to this, or should I try re-integrating it?  I don't know why it munged .xcodeproj so much, I don't think I did anything different, but I'm willing to try again and watch more closely if you like.

cmsj:
I looked at the diff more closely and it actually wasn't as bad as I thought, so I have landed it as-is :)

PR #787, opened 2016-02-09
Grid, add pure move; Closes #716

### LINKED ISSUE #716: hs.grid.show for movement only?
Is there a key or flag I'm missing to the hs.grid.show interface which allows just moving a window (without resize) so it's upper left corner matches the letter space typed?

If not, can this be considered a feature request?  It would actually probably be best if there were a key (maybe return?) as the second keypress which indicates that this should be considered a move rather than a move-and-resize... this way I can choose which I want on a case-by-case basis rather than have to map the function to multiple keys, depending upon which I want to accomplish for a given invocation.

wvierber:


wvierber:
Hmm, should I do some kind of `log.f('pure move to... )` when moving the window?

PR #791, opened 2016-02-12
[READY] hs.socket

heptal:
Adds `hs.socket` for communicating with arbitrary protocols using [`CocoaAsyncSocket`](https://github.com/robbiehanson/CocoaAsyncSocket).

TCP sockets are available under `hs.socket`
UDP sockets are available under `hs.socket.udp`

### DIFF
--- a/extensions/socket/internal.m
+++ b/extensions/socket/internal.m
@@ -5,6 +5,7 @@
 // Definitions
 @interface HSAsyncSocket : GCDAsyncSocket
 @property int readCallback;
+@property int writeCallback;
 @property int connectCallback;
 @property NSTimeInterval timeout;
 @property NSMutableArray* connectedSockets;
@@ -14,26 +15,13 @@
 static const char *USERDATA_TAG = "hs.socket";
 #define getUserData(L, idx) (__bridge HSAsyncSocket *)((asyncSocketUserData *)lua_touserdata(L, idx))->asyncSocket;
 
-// Callback on data reads
-static void readCallback(HSAsyncSocket *asyncSocket, NSData *data, long tag) {
-    LuaSkin *skin = [LuaSkin shared];
-    NSString *utf8Data = [[NSString alloc] initWithData:data encoding:NSUTF8StringEncoding];
-
-    [skin pushLuaRef:refTable ref:asyncSocket.readCallback];
-    [skin pushNSObject: utf8Data];
-    [skin pushNSObject: @(tag)];
-
-    if (![skin protectedCallAndTraceback:2 nresults:0]) {
-        const char *errorMsg = lua_tostring(skin.L, -1);
-        [skin logError:[NSString stringWithFormat:@"hs.socket read callback error: %s", errorMsg]];
-    }
-}
 
 // Delegate implementation
 @implementation HSAsyncSocket
 
 - (id)init {
     self.readCallback = LUA_NOREF;
+    self.writeCallback = LUA_NOREF;
     self.connectCallback = LUA_NOREF;
     self.timeout = -1;
     self.connectedSockets = [[NSMutableArray alloc] init];
@@ -48,9 +36,10 @@ static void readCallback(HSAsyncSocket *asyncSocket, NSData *data, long tag) {
     if (sock.connectCallback != LUA_NOREF) {
         [skin pushLuaRef:refTable ref:sock.connectCallback];
         sock.connectCallback = [skin luaUnref:refTable ref:sock.connectCallback];
+
         if (![skin protectedCallAndTraceback:0 nresults:0]) {
             const char *errorMsg = lua_tostring(skin.L, -1);
-            [skin logError:[NSString stringWithFormat:@"hs.socket connect callback error: %s", errorMsg]];
+            [skin logError:[NSString stringWithFormat:@"%s connect callback error: %s", USERDATA_TAG, errorMsg]];
         }
     }
 }
@@ -86,13 +75,34 @@ static void readCallback(HSAsyncSocket *asyncSocket, NSData *data, long tag) {
 }
 
 - (void)socket:(HSAsyncSocket *)sock didWriteDataWithTag:(long)tag {
-    [[LuaSkin shared] logInfo:@"Data written to socket"];
+    LuaSkin *skin = [LuaSkin shared];
+    [skin logInfo:@"Data written to socket"];
+
+    if (self.writeCallback != LUA_NOREF) {
+        [skin pushLuaRef:refTable ref:self.writeCallback];
+        [skin pushNSObject: @(tag)];
+        self.writeCallback = [skin luaUnref:refTable ref:self.writeCallback];
+
+        if (![skin protectedCallAndTraceback:1 nresults:0]) {
+            const char *errorMsg = lua_tostring(skin.L, -1);
+            [skin logError:[NSString stringWithFormat:@"%s write callback error: %s", USERDATA_TAG, errorMsg]];
+        }
+    }
 }
 
 - (void)socket:(HSAsyncSocket *)sock didReadData:(NSData *)data withTag:(long)tag  {
-    [[LuaSkin shared] logInfo:@"Data read from socket"];
+    LuaSkin *skin = [LuaSkin shared];
+    [skin logInfo:@"Data read from socket"];
+    NSString *utf8Data = [[NSString alloc] initWithData:data encoding:NSUTF8StringEncoding];
 
-    readCallback(self, data, tag);
+    [skin pushLuaRef:refTable ref:self.readCallback];
+    [skin pushNSObject: utf8Data];
+    [skin pushNSObject: @(tag)];
+
+    if (![skin protectedCallAndTraceback:2 nresults:0]) {
+        const char *errorMsg = lua_tostring(skin.L, -1);
+        [skin logError:[NSString stringWithFormat:@"%s read callback error: %s", USERDATA_TAG, errorMsg]];
+    }
 }
 
 - (void)socket:(HSAsyncSocket *)sock didReceiveTrust:(SecTrustRef)trust completionHandler:(void (^)(BOOL))completionHandler {
@@ -199,7 +209,7 @@ static int socket_parseAddress(lua_State *L) {
 /// Parameters:
 ///  * host - A string containing the hostname or IP address
 ///  * port - A port number [1-65535]
-///  * fn - An optional callback function to execute after establishing the connection. Takes no parameters
+///  * fn - An optional single-use callback function to execute after establishing the connection. Takes no parameters
 ///
 /// Returns:
 ///  * The [`hs.socket`](#new) object
@@ -346,6 +356,7 @@ static int socket_read(lua_State *L) {
 /// Parameters:
 ///  * message - A string containing data to be sent on the socket
 ///  * tag - An optional integer to assist with labeling writes
+///  * fn - An optional single-use callback function to execute after writing data to the socket. Takes the tag parameter
 ///
 /// Returns:
 ///  * The [`hs.socket`](#new) object
@@ -355,13 +366,17 @@ static int socket_read(lua_State *L) {
 ///
 static int socket_write(lua_State *L) {
     LuaSkin *skin = [LuaSkin shared];
-    [skin checkArgs:LS_TUSERDATA, USERDATA_TAG, LS_TSTRING, LS_TNUMBER|LS_TOPTIONAL, LS_TBREAK];
+    [skin checkArgs:LS_TUSERDATA, USERDATA_TAG, LS_TSTRING, LS_TNUMBER|LS_TNIL|LS_TOPTIONAL, LS_TFUNCTION|LS_TOPTIONAL, LS_TBREAK];
 
     HSAsyncSocket* asyncSocket = getUserData(L, 1);
     NSString *message = [skin toNSObjectAtIndex:2];
 
     long tag = -1;
     if (lua_type(L, 3) == LUA_TNUMBER) tag = lua_tointeger(L, 3);
+    if (lua_type(L, 4) == LUA_TFUNCTION) {
+        lua_pushvalue(L, 4);
+        asyncSocket.writeCallback = [skin luaRef:refTable];
+    }
 
     if (asyncSocket.userData != SERVER) {
         [asyncSocket writeData:[message dataUsingEncoding:NSUTF8StringEncoding] withTimeout:asyncSocket.timeout tag:tag];
--- a/extensions/socket/internal.m
+++ b/extensions/socket/internal.m
@@ -133,12 +133,12 @@ static void listenSocket(HSAsyncSocket *asyncSocket, NSNumber *port) {
 /// Creates an asynchronous TCP socket object for reading and writing. A callback is required to read data from the socket
 ///
 /// Parameters:
-///  * host - A optional string containing the hostname or IP address. If `nil`, a listening socket is created (same as `hs.socket.server`)
+///  * host - A optional string containing the hostname or IP address. If `nil`, a listening socket is created (same as [`hs.socket.server`](#server))
 ///  * port - A port number [1024-65535]. Ports [1-1023] are privileged
-///  * fn - An optional callback function to process data on reads. Can also be set with the `setCallback` method
+///  * fn - An optional callback function to process data on reads. Can also be set with the [`setCallback`](#setCallback) method
 ///
 /// Returns:
-///  * An `hs.socket` object
+///  * An [`hs.socket`](#new) object
 ///
 static int socket_new(lua_State *L) {
     LuaSkin *skin = [LuaSkin shared];
@@ -172,16 +172,16 @@ static int socket_new(lua_State *L) {
     return 1;
 }
 
-/// hs.socket:connect() -> self
+/// hs.socket:connect(host, port) -> self
 /// Method
-/// Connects an unconnected `hs.socket` instance
+/// Connects an unconnected [`hs.socket`](#new) instance
 ///
 /// Parameters:
 ///  * host - A string containing the hostname or IP address
 ///  * port - A port number [1024-65535]. Ports [1-1023] are privileged
 ///
 /// Returns:
-///  * The `hs.socket` object
+///  * The [`hs.socket`](#new) object
 ///
 static int socket_connect(lua_State *L) {
     LuaSkin *skin = [LuaSkin shared];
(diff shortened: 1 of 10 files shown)


extensions/socket/internal.m
```diff
@@ -0,0 +1,362 @@
+#import <LuaSkin/LuaSkin.h>
+#import <CocoaAsyncSocket/GCDAsyncSocket.h>
+
+static const char *USERDATA_TAG = "hs.socket";
+
+int refTable;
```
cmsj: (This isn't documented anywhere yet, or even widespread convention, but) this should be static too

extensions/socket/internal.m
```diff
@@ -0,0 +1,362 @@
...
+@property int callback;
+@end
+
+
+static void callback(HSAsyncSocket *asyncSocket, NSData *data) {
+    LuaSkin *skin = [LuaSkin shared];
+    NSString *utf8Data = [[NSString alloc] initWithData:data encoding:NSUTF8StringEncoding];
+
+    if (!asyncSocket.callback) {
+        lua_getglobal(skin.L, "print");
```
cmsj: Is this more of a debugging choice? Normally we'd throw an error at not having a callback defined when we're getting system callback data

extensions/socket/internal.m
```diff
@@ -0,0 +1,362 @@
...
+    } else {
+        [skin pushLuaRef:refTable ref:asyncSocket.callback];
+    }
+
+    lua_pushstring(skin.L, [utf8Data UTF8String]);
```
cmsj: Note that you can do [skin pushNSObject] here, it understands what to do with NSString :)

extensions/socket/internal.m
```diff
@@ -0,0 +1,362 @@
...
+///  * host - A string containing the hostname or IP address
+///  * port - A port number [1024-65535]. Ports [1-1023] are privileged
+///  * fn - An optional callback function accepting a single parameter to process data. Can be set with `:setCallback`
+///
+/// Returns:
+///  * An `hs.socket` object
+///
+static int socket_new(lua_State *L) {
+    LuaSkin *skin = [LuaSkin shared];
+    [skin checkArgs:LS_TSTRING, LS_TNUMBER, LS_TFUNCTION|LS_TNIL|LS_TOPTIONAL, LS_TBREAK];
```
cmsj: The LS_TNIL here should only be needed if you're expecting it, which in this function probably makes no sense.

extensions/socket/internal.m
```diff
@@ -0,0 +1,362 @@
...
+- (void)socket:(HSAsyncSocket *)sock didReadData:(NSData *)data withTag:(long)tag  {
+    [[LuaSkin shared] logInfo:@"Data read from socket"];
+
+    callback(self, data);
+}
+
+@end
+
+
+/// hs.socket.new(host, port) -> hs.socket object
```
cmsj: /// hs.socket.new(host, port[, fn]) -> hs.socket object

extensions/socket/internal.m
cmsj: (I know I just added it in hs.timer, if that's what you're referencing, but that is because the various Lua wrappers for hs.timer.new implicitly pass a nil. It doesn't hurt to have it here, and may make life easier for wrappers later, but it's up to you)

extensions/socket/internal.m
```diff
@@ -0,0 +1,362 @@
...
+        lua_pushvalue(L, 3);
+        asyncSocket.callback = [skin luaRef:refTable];
+    }
+
+    NSError *err;
+    if (![asyncSocket connectToHost:theHost onPort:[thePort unsignedShortValue] error:&err]) {
+        [skin logError:[NSString stringWithFormat:@"Unable to connect: %@", err]];
+    }
+
+    [skin pushNSObject:asyncSocket];
```
cmsj: I'm not a huge fan of this pattern, it means that every operation on a userdata object creates a new one. I think I'd rather leave this as a lua_newuserdata() call, and have the equivalent calls in the methods do lua_pushvalue(L, 1).

(The userdata abstraction PR will let us refactor this stuff later)

extensions/socket/internal.m
```diff
@@ -0,0 +1,362 @@
...
+
+    if (!asyncSocket.callback) {
+        [skin logWarn:@"No callback! Defaulting to print()"];
+    }
+
+    switch (lua_type(L, 2)) {
+        case LUA_TNUMBER: {
+            NSNumber *bytesToRead = [skin toNSObjectAtIndex:2];
+            NSUInteger bytes = [bytesToRead unsignedIntegerValue];
+            [asyncSocket readDataToLength:bytes withTimeout:-1 tag:-1];
```
cmsj: I wonder if we might want to make the timeout a property on the socket object, and expose a method to read/set it?

extensions/socket/internal.m
cmsj: I see from reading further in that the callback is optional. I'm not sure that's particularly necessary, except maybe during development?

extensions/socket/internal.m
cmsj: I also think we should expose the tags throughout, as optional parameters to the methods, and as a parameter to the callback. It seems like a super handy way to implement socket protocols.

extensions/socket/internal.m
```diff
@@ -0,0 +1,362 @@
...
+///
+/// Returns:
+///  * The `hs.socket` object
+///
+static int socket_setCallback(lua_State *L) {
+    LuaSkin *skin = [LuaSkin shared];
+    [skin checkArgs:LS_TUSERDATA, USERDATA_TAG, LS_TFUNCTION|LS_TNIL|LS_TOPTIONAL, LS_TBREAK];
+
+    HSAsyncSocket*  asyncSocket = [skin luaObjectAtIndex:1 toClass:"HSAsyncSocket"];
+
```
cmsj: Before doing anything with the LUA_TFUNCTION, this should do asyncSocket.callback = [skin luaUnref:refTable asyncSocket.callback] - it's a no-op if the callback is LUA_NOREF, but if a value has been stored, it is released from the reference table before we add another one

extensions/socket/internal.m
```diff
@@ -0,0 +1,362 @@
...
+        const char *errorMsg = lua_tostring(skin.L, -1);
+        [skin logError:[NSString stringWithFormat:@"hs.socket callback error: %s", errorMsg]];
+    }
+}
+
+
+@implementation HSAsyncSocket
+
+- (id)init {
+    return [super initWithDelegate:self delegateQueue:dispatch_get_main_queue()];
```
cmsj: the callback property should be initialised to LUA_NOREF

extensions/socket/internal.m
```diff
@@ -0,0 +1,362 @@
...
+    LuaSkin *skin = [LuaSkin shared];
+    [skin checkArgs:LS_TUSERDATA, USERDATA_TAG, LS_TBREAK];
+
+    HSAsyncSocket*  asyncSocket = [skin luaObjectAtIndex:1 toClass:"HSAsyncSocket"];
+    [asyncSocket disconnect];
+
+    [skin pushNSObject:asyncSocket];
+    return 1;
+}
+
```
cmsj: I'd suggest we have an hs.socket:connected() here, even though the same information is in :info()

extensions/socket/internal.m
cmsj: (I'd suggest copying the very simple struct userdata object pattern that I used in hs.chooser, IMO it's the easiest one to reason about)

extensions/socket/internal.m
heptal: I will admit I got tired of putting in callbacks in my iterations, and that I also newly discovered the ability to seamlessly patch in a function from global and thought it was cool. In my opinion there are valid uses for firing stuff without needing to read or process anything, but yea, in the case of reading, restoring the usual error might be better to more appropriately notify the user they're 'doing it wrong'.

extensions/socket/internal.m
heptal: Agreed. Intended to but just forgot.

extensions/socket/internal.m
```diff
@@ -0,0 +1,362 @@
...
+    void *ptr = luaL_testudata(L, idx, USERDATA_TAG);
+    if (ptr) {
+        return (__bridge HSAsyncSocket *)*((void **)ptr);
+    } else {
+        return nil;
+    }
+}
+
+static int socket_objectGC(lua_State *L) {
+    HSAsyncSocket *asyncSocket = [[LuaSkin shared] luaObjectAtIndex:1 toClass:"HSAsyncSocket"];
```
cmsj: Note that, however the userdata stuff is handled, there needs to be a (__bridge_transfer HSAsyncSocket *) cast somewhere to balance out the __bridge_retained. In this current iteration (and probably most others), it would be here in the __gc function.

extensions/socket/internal.m
heptal: Yea, I did think about having a timeout property on the socket object as this gets fleshed out. I thought the tagging stuff was pretty neat but haven't thought much about it yet. Good idea on having them as a callback parameter.

extensions/socket/internal.m
```diff
@@ -0,0 +1,657 @@
...
+- (id)init {
+    self.readCallback = LUA_NOREF;
+    self.writeCallback = LUA_NOREF;
+    self.connectCallback = LUA_NOREF;
+    self.timeout = -1;
+    self.connectedSockets = [[NSMutableArray alloc] init];
+    return [super initWithDelegate:self delegateQueue:dispatch_get_main_queue()];
+}
+
+- (void)socket:(HSAsyncSocket *)sock didConnectToHost:(NSString *)host port:(UInt16)port {
```
cmsj: Is this guaranteed to be called on the main thread by GCDASyncSocket? I am guessing it must be, otherwise you would have had failures from LuaSkin when it noticed it was being accessed from a non-main thread, but it seems like everything else here is expecting to be happening on random threads.

extensions/socket/internal.m
```diff
@@ -0,0 +1,657 @@
...
+    [skin pushNSObject: @(tag)];
+
+    if (![skin protectedCallAndTraceback:2 nresults:0]) {
+        const char *errorMsg = lua_tostring(skin.L, -1);
+        [skin logError:[NSString stringWithFormat:@"%s read callback error: %s", USERDATA_TAG, errorMsg]];
+    }
+}
+
+- (void)socket:(HSAsyncSocket *)sock didReceiveTrust:(SecTrustRef)trust completionHandler:(void (^)(BOOL))completionHandler {
+    // Allow TLS handshake without trust evaluation for self-signed certificates
```
cmsj: I think this needs to be documented, and potentially made into a variable that can be tuned. It has significant security implications.

extensions/socket/internal.m
```diff
@@ -0,0 +1,657 @@
...
+/// Parameters:
+///  * delimiter - Either a number of bytes to read, or a string delimiter such as `\n` or `\r\n`. Data is read up to and including the delimiter
+///  * tag - An optional integer to assist with labeling reads that is passed to the read callback
+///
+/// Returns:
+///  * The [`hs.socket`](#new) object or `nil` if no callback error
+///
+/// Notes:
+///  * If called on a listening socket with multiple connections, data is read from each of them
+///
```
cmsj: FWIW these trailing /// lines are not necessary :)

extensions/socket/internal.m
```diff
@@ -0,0 +1,657 @@
...
+
+        default:
+            break;
+    }
+
+    lua_pushvalue(L, 1);
+    return 1;
+}
+
+/// hs.socket:write(message[, tag]) -> self
```
cmsj: This is missing `, fn` in the optional parameters

extensions/socket/internal.m
```diff
@@ -0,0 +1,657 @@
...
+/// trusted 3rd party certificate agency (e.g. verisign) and that the certificate is not expired.
+/// However it will not verify the name on the certificate unless you
+/// give it a name to verify against via the kCFStreamSSLPeerName key.
+/// The security implications of this are important to understand.
+/// Imagine you are attempting to create a secure connection to MySecureServer.com,
+/// but your socket gets directed to MaliciousServer.com because of a hacked DNS server.
+/// If you simply use the default settings, and MaliciousServer.com has a valid certificate,
+/// the default settings will not detect any problems since the certificate is valid.
+/// To properly secure your connection in this particular scenario you
+/// should set the kCFStreamSSLPeerName property to "MySecureServer.com".
```
cmsj: `kCFStreamSSLPeerName` probably should be `peerName` and the rest of the sentence might need slight rewording to support that

extensions/socket/internal.m
```diff
@@ -0,0 +1,657 @@
...
+    [skin checkArgs:LS_TUSERDATA, USERDATA_TAG, LS_TBOOLEAN|LS_TOPTIONAL, LS_TSTRING|LS_TOPTIONAL, LS_TBREAK];
+
+    HSAsyncSocket* asyncSocket = getUserData(L, 1);
+    NSDictionary *tlsSettings = nil;
+
+    if (lua_type(L, 2) == LUA_TBOOLEAN && lua_toboolean(L, 2) == false) {
+        tlsSettings = @{@"GCDAsyncSocketManuallyEvaluateTrust": @YES};
+    }
+
+    if (lua_type(L, 3) == LUA_TSTRING) {
```
cmsj: I'm not sure this is right - the docstrings above suggest that both arguments are optional, so `L,2` might also be `LUA_TSTRING` if the user is omitting the `verify` argument. This would also need the `checkArgs` signature to change to `LS_TBOOLEAN|LS_TSTRING|LS_TOPTIONAL`.

extensions/socket/internal.m
```diff
@@ -0,0 +1,657 @@
...
+
+
+static int userdata_gc(lua_State *L) {
+    asyncSocketUserData *userData = lua_touserdata(L, 1);
+    HSAsyncSocket* asyncSocket = (__bridge_transfer HSAsyncSocket *)userData->asyncSocket;
+    userData->asyncSocket = nil;
+
+    [asyncSocket disconnect];
+    [asyncSocket setDelegate:nil delegateQueue:NULL];
+    asyncSocket.readCallback = [[LuaSkin shared] luaUnref:refTable ref:asyncSocket.readCallback];
```
cmsj: I think it would be a good idea fo also `luaUnref` the `writeCallback` and `connectCallback` since it's possible that `__gc` might be called while a socket is live and waiting for data/connect/whatever.

They're all initialised to `LUA_NOREF`, and it's a safe no-op to call `luaUnref` on a variable that's `LUA_NOREF` :)

extensions/socket/internal.m
```diff
@@ -0,0 +1,657 @@
...
+
+    {"__tostring", userdata_tostring},
+    {"__gc", userdata_gc},
+
+    {NULL, NULL} // This must end with an empty struct
+};
+
+int luaopen_hs_socket_internal(lua_State *L __unused) {
+    LuaSkin *skin = [LuaSkin shared];
+    refTable = [skin registerLibraryWithObject:USERDATA_TAG
```
cmsj: FWIW, in my testing/refactoring work, I'm moving things to making separate calls to `registerLibrary` and `registerObject`. It's not a big deal at all, but I think it's slightly neater to separate them (there was a time when libraries and their objects shared a metatable, and having a single method for both made more sense). Eventually I'd like to remove `registerLibraryWithObject`.

cmsj:
@heptal this is looking really good. I commented with a few little bits and bobs, but on the whole I am super pleased with this PR! Thanks very much :)

extensions/socket/internal.m
heptal: Yes. You can init with `[super initWithDelegate:self delegateQueue:NULL]` and GCDAsyncSocket will create its own serial queues, or specify your own serial queue like:
```objc
- (id)init {
    dispatch_queue_t socketQueue = dispatch_queue_create("socketQueue", NULL);
...
    return [super initWithDelegate:self delegateQueue:socketQueue];
}
```
Then, in each of the delegate methods that operate on the UI/Lua in any way, you must wrap those operations like so:
```objc
        @autoreleasepool {
            [skin logInfo:@"Data written to socket"];
            // do some lua callback or something
        }
    });
```
Failing to `dispatch_async` to the main thread will indeed crash as you say. I initially chose `dispatch_get_main_queue` as the `delegateQueue` because it 'just worked', but it does seem a bit more performant if I use separate queues and call `dispatch_async` to the main thread as shown above, presumably because GCDAsyncSocket avoids doing all of its internal operations on the main thread, waiting until the results are in. I have an uncommitted implementation of this and wanted your opinion before really putting it in.

extensions/socket/internal.m
heptal: Added a comment there to clarify that this delegate method is only called when startTLS is invoked with` GCDAsyncSocketManuallyEvaluateTrust == YES`

extensions/socket/internal.m
heptal: Is it alright if I feel like it looks clearer?

extensions/socket/internal.m
heptal: Oops. Fixed.

extensions/socket/internal.m
heptal: Done.

extensions/socket/internal.m
heptal: Agreed. Changed.

extensions/socket/internal.m
heptal: Oversight. Fixed.

extensions/socket/internal.m
heptal: How feasible/difficult would it be to have LuaSkin's log* methods automatically call `dispatch_async` to the main thread for log operations performed on background threads, with everything else being subjected to the `if (![NSThread isMainThread])` check performed in `+ (id)shared `? Seems to me like that wouldn't be terribly unsafe, but might be too much work for not a lot of benefit.

In any case I've converted it to use the background queue with `dispatch_async` to the main thread where needed and it passes all tests.

asmagill: The most reliable way to do this is to add class methods for the logX: methods... anything that relies on [LuaSkin shared] will be rejected before the method is ever entered. (e.g. `[LuaSkin logWarn:...]` rather than/in addition to `[[LuaSkin shared] logWarn:...]`

It's not a bad idea, maybe even include a prefix indicating when it's from another thread; @cmsj objections to my adding this?  I'd also like to see `hs.logger` used in the Lua side of the output so verbose and debug can be turned on and off as desired and can add that at the same time.

heptal:
I think this is ready. It exposes pretty much everything in CocoaAsyncSocket.

`hs.socket` at 99% coverage
`hs.socket.udp` at 96% coverage
Uncovered bits are errors that I don't know how to induce (and presumably are quite rare).

PR #795, opened 2016-02-16
Add optional duration parameter to hs.alert.closeAll.

drn:
I have a few bindings that display alerts that I regularly run in quick succession (for example changing system volume). If I run them within the 0.15 default fade out duration, the alerts end up stacking up and change displayed screen position. With a 0 duration and bypassing the fade, the alerts are always displayed in the same position.


cmsj:
Thanks!

PR #802, opened 2016-02-24
Allow floats when setting audio device volume.

drn:
Hello! I ran into the issue of not being able to set audio device volume with floats when replicating the OS X volume indicator in my bindings. OS X uses a 6.25% volume per step (16 steps) in their indicator, and I wasn't able to replicate it with only integers.

### DIFF
--- a/extensions/audiodevice/internal.m
+++ b/extensions/audiodevice/internal.m
@@ -536,7 +536,7 @@ static int audiodevice_volume(lua_State* L) {
     };
 
     if (AudioObjectHasProperty(deviceId, &propertyAddress) && (AudioObjectGetPropertyData(deviceId, &propertyAddress, 0, NULL, &volumeSize, &volume) == noErr)) {
-        lua_pushinteger(L, (int)(volume * 100.0));
+        lua_pushnumber(L, volume * 100.0);
     } else {
         lua_pushnil(L);
     }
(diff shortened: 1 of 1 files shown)


cmsj:
Good catch, thanks! We should also change `audiodevice_volume()` - it's currently fetching a Float32, but then casting it to `int` for `lua_pushinteger()` (where it should just be handing the float directly to `lua_pushnumber()`).

I'm happy to make that change if you want?

drn:
Cool, made the change you mentioned and tested them

cmsj:
Thanks very much!

cmsj:
@wvierber this stuff is all a bit loose. The official Apple SDK docs say:

```
kAudioHardwareServiceDeviceProperty_VirtualMasterVolume
A Float32 value that represents the value of the volume control.

The range for this property’s value is 0.0 (silence) through 1.0 (full level)
```

which is fairly vague and makes no declarations about the number of discrete volume values allowed, which is likely because different audio devices will support different levels of precision in their volume specification.

I'll expand the Hammerspoon docs a little, but the basics are that you may not get exactly the volume value you asked for, because it will be translated into a value that your specific hardware can handle. The values we supply to the OS, and read back from it, are ranged 0-100, so they are just the underlying kAudioHardwareServiceDeviceProperty_VirtualMasterVolume values multiplied by 100.

It may be possible to query more information about a specific audio device and properly smooth out the values across 0-100, but I'm not aware of such a facility.

Does that all answer your question?

wvierber:
Yeah, thanks @cmsj. I'm glad it isn't caused by the way Hammerspoon handels the numbers.
Getting more info about should be interesting...

PR #803, opened 2016-02-24
suppress excessive GNU statement expression warnings

heptal:
~~Added `-Wno-gnu` to `Other Warning Flags` to both `Hammerspoon` Project and `Hammerspoon Tests` Target build settings, because adding it to `Hammerspoon` Project's `Other Warning Flags` alone doesn't cover `Hammerspoon Tests` for some reason. Adding it to `Other C Flags` in `Hammerspoon` Project's build settings covers both, but `Other Warning Flags` seems like the right place over `Other C Flags`.~~

### DIFF
--- a/Hammerspoon Tests/HSTestCase.h
+++ b/Hammerspoon Tests/HSTestCase.h
@@ -8,6 +8,7 @@
 
 #import <XCTest/XCTest.h>
 #import "MJLua.h"
+#pragma clang diagnostic ignored "-Wgnu-statement-expression"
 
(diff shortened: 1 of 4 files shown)


heptal:
Added #pragma clang diagnostic ignored "-Wgnu-statement-expression" to HSTestCase.h (the RUN_LUA_TEST macro was generating the excessive warnings).

cmsj:
Thanks for this. I had a global fix for this queued up already, so I've landed that as part of c04ad20.

PR #804, opened 2016-02-25
add imageFromMediaFile constructor to hs.image

heptal:
This `hs.image` constructor attempts to obtain album artwork from the given audio file or album directory, or generate a thumbnail for a video file.

The function first determines the containing directory, if it is not already a directory
It checks if any of the following common, somewhat standard filenames for album art are present:
* cover.jpg
* front.jpg
* art.jpg
* album.jpg
* folder.jpg

If one of the common filenames is found, it is returned as an `hs.image` object
If none are found, it attempts to extract image metadata from the file. This works for .mp3/.m4a files
If embedded image metadata is found, it is return as an `hs.image` object, otherwise `nil`
This allows for obtaining artwork associated with file formats such as .flac/.ogg and it is much faster than extracting metadata from every file

### DIFF
--- a/extensions/image/internal.m
+++ b/extensions/image/internal.m
@@ -779,100 +779,6 @@ static int imageFromApp(lua_State *L) {
     return 1;
 }
 
-/// hs.image.imageFromMediaFile(file) -> object
-/// Constructor
-/// Creates an `hs.image` object from a video file or the album artwork of an audio file or directory
-///
-/// Parameters:
-///  * file - A string containing the path to an audio or video file or an album directory
-///
-/// Returns:
-///  * An `hs.image` object or `nil`, if no album art was found
-///
-/// Notes:
-///  * For audio files, this function first determines the containing directory (if not already a directory)
-///  * It checks if any of the following common filenames for album art are present:
-///   * cover.jpg
-///   * front.jpg
-///   * art.jpg
-///   * album.jpg
-///   * folder.jpg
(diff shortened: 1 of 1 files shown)


cmsj:
Does the AVFoundation thing also return valid artwork for, say, an iTunes movie file? If so, I'd suggest renaming this `imageFromMediaFile`.

heptal:
It does not. I had it named that originally but it returns nil for any .mp4/.m4v/.mov I've tried, including iTunes downloads

extensions/image/internal.m
```diff
@@ -778,6 +779,100 @@ static int imageFromApp(lua_State *L) {
...
 
+/// hs.image.imageFromMediaFile(file) -> object
+/// Constructor
+/// Creates an `hs.image` object from a video file or the album artwork of an audio file or directory
+///
+/// Parameters:
+///  * file - A string containing the path to an audio or video file or an album directory
+///
+/// Returns:
+///  * An `hs.image` object or `nil`, if no album art was found
```
cmsj: "album art was found" -> "image could be created"

(although I find myself wondering whether we should return a nil here, or get the OS to give us the filetype image as a fallback)

extensions/image/internal.m
```diff
@@ -778,6 +779,100 @@ static int imageFromApp(lua_State *L) {
...
+///
+/// Notes:
```
cmsj: I think the notes section should also cover how video files are handled.

PR #806, opened 2016-02-25
hs.osascript with NSAppleEventDescriptor parsing

heptal:
Unifies `hs.applescript` and `hs.javascript` with `OSAScript`'s  `-initWithSource:language:` which is much simpler/concise.

Parses the `NSAppleEventDescriptor` of arbitrary complexity into an NSDict/NSArray/NSString/primitive/whatever it may be and returns it as a table/string/number/boolean/whatever, which greatly simplifies the Lua-side parsing to just creating error tables when needed.

The `applescript` and `javascript` targets can be removed with this. Those top-level extensions could just have trivial `init.lua`s referencing this module to keep API compatibility, although the API is slightly changed now that a fully parsed object is being returned in addition. The question of API breakage needs discussion.


asmagill:
The itunes, spotify and messaging modules (are there others?) should be checked against using this directly... if they work (or can be easily made to), then I don't see a technical reason to keep applescript/javascript around, though the question of how/when to break backwards compatibility should still be had.

I personally don't think these are "high volume" modules, and as has been pointed out in another thread, we are still pre 1.0, so with some documentation/notification in the release notes, I wouldn't have a problem with just dropping applescript/javascript after testing this a little more.

heptal:
Is it possible to create 'top-level' modules from within this module, or are the folder groups with their `init.lua`s required?

asmagill:
Not if we want to keep module autoloading working for hs.applescript and hs.javascript -- the autoloading code works by identifying the directories of modules included within the Hammerspoon application, so no directory means no reference in the autoload metamethods.

I suppose the autoload code could have some explicit exceptions added, but it's just as simple to change them to lua-only modules which "wrap" this module... I'll defer to @cmsj on this since I don't use them (or rely on the autoload functionality much) that often myself.

heptal:
I added all the Java/AppleScript tests for this module ripped out the `applescript` and `javascript` modules except for simple `init.lua`s that reference this one. API compatibility is the same as far as the itunes, spotify and messaging modules go. Whether or not we go even further and drop those `init.lua`s to reduce clutter is another question.

PR #807, opened 2016-02-25
fix bugs in vimperator hint mode

xofyarg:
1. Remove unnessary key binding when enter hint. they are already done
    in setupModal. There will be a performance issue after several
    rebindings.
2. Vimperator mode's hint chars should be fixed to use A-Z.
3. Fails(also in default mode) if hintChar does not start with A.

### DIFF
--- a/extensions/hints/init.lua
+++ b/extensions/hints/init.lua
@@ -14,6 +14,9 @@ local modal_hotkey = hotkey.modal
 --- The default is the letters A-Z.
 hints.hintChars = {"A","B","C","D","E","F","G","H","I","J","K","L","M","N","O","P","Q","R","S","T","U","V","W","X","Y","Z"}
 
+-- vimperator mode requires to use full set of alphabet to represent applications.
+hints.hintCharsVimperator = {"A","B","C","D","E","F","G","H","I","J","K","L","M","N","O","P","Q","R","S","T","U","V","W","X","Y","Z"}
+
 --- hs.hints.style
 --- Variable
 --- If this is set to "vimperator", every window hint starts with the first character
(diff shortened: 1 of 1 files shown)


cmsj:
Thanks for this! One issue I notice is that `hintChars` is intended to be exposed as a user-controllable variable, so we ideally shouldn't be smashing it with the vimperator equivalent.

xofyarg:
Did you mean people may change hint style without reload this module, in which way, hintChars will lose track of default/user one?

cmsj:
yes, http://www.hammerspoon.org/docs/hs.hints.html#hintChars says that people are allowed to change the contents of `hintsChars` and the line in your patch:

```
+    hints.hintChars = hintCharsVimperator
```

means that their changes will be discarded. I think it might make more sense to expose `hintsCharsVimperator` to users as well, and have the function use a local copy of either `hints.hintChars` or `hints.hintCharsVimperator` so the user changes are preserved.

xofyarg:
Using local copy of these variables in functions sounds good. 

The question is what the behavior you want it to be if users change hints.hintCharsVimperator to a subset of alphabet?
e.g. the variable set by user is `hints.hintCharsVimperator = {"A", "B", "C"}`, they cannot select windows like "Google Chrome", "Safari", etc. (`#2` in my first comment)

xofyarg:
>  I think it might make more sense to expose hintsCharsVimperator to users as well, and have the function use a local copy of either hints.hintChars or hints.hintCharsVimperator so the user changes are preserved.

PR updated.

cmsj:
Thanks for doing this, sorry it got so delayed!

Issue #810, opened 2016-02-27
Clean up XCode Project


heptal:
It seems the time to streamline the xcodeproj is well-overdue. It turns out a few simple tasks can clean it up tremendously, removing many thousands of lines from the .pbxproj, which should make just about everything easier and more robust going forward. Creating this issue to track general progress and for further suggestions
### Clean up frameworks

Every target has a random selection of frameworks linked against in build settings. They can have some or all or none of Cocoa, Foundation, LuaSkin, etc. It is no longer [necessary](http://clang.llvm.org/docs/Modules.html#semantic-import) as long as they are modules (as frameworks are) and in the right place. System frameworks load automatically, Podfile-managed frameworks load automatically, and the LuaSkin framework does as well with correct build settings.
- [x] **Move Fabric/Sparkle to Podfile and update others** Just a few minor version updates. Nothing breaks here
- [x] **Trash Pods project, `pod install`, use generated .xcconfig** Regenerates everything freshly
- [x] **Change Crashlytics/Fabric `#include`s to `@import`s**
- [x] **Entirely remove Frameworks group** This automatically removes all references from the project, including all the `Link Binary With Libraries` build phases throughout the project. The `Embed Pods Frameworks` and `Copy LuaSkin.framework` build phases on the Hammerspoon target take care of everything. _ALL_ contents of PBXFrameworksBuildPhase and their corresponding references in PBXNativeTarget can be removed. Very satisfying!
- [ ] **Remove instructions in contributing docs for linking frameworks**  

In my local repo all tests pass with the above. Can PR
### Migrate to using .xcconfig files #786

I have played with [BuildSettingExtractor](https://github.com/dempseyatgithub/BuildSettingExtractor) and explored the output a bit. These files can `#include` one another. When I checked for dupes, it turned out there are just a few unique variants of build settings for all the targets. We can have a just a few .xcconfig files that cover just about everything: a `regular` module build file, a `strict` one (`regular` with -Weverything & -Werror added for modules have have undergone refactoring/testing). Each target's .xcconfig would then just have single line importing one of the common ones, and possibly a line or 2 for module-specific build settings (like OS version requirements). Exploring some scripts to do this reliably.

heptal:
The way `.xcconfig` files work is by cascading from right to left, from OSX defaults, to project build settings, to `.xcconfig` files, and finally in-XCode build settings.
![](http://i.imgur.com/Un7RQi0.png)

[...]

PR #811, opened 2016-02-27
Streamline Frameworks, Pods, and Build Settings

heptal:
See #810 

Passes all tests! 😁 (Except for that audio one due to float)

Edit: Having some trouble with LuaSkin on a pristine build in Travis. It works perfectly on fresh checkouts on my local machine.


cmsj:
So, I think this is an excellent idea, but as with #747 I have serious concerns about landing it while we have no test suite. It's become fairly normal to spend a couple of releases tidying up regressions from a previous large change, because we can't detect the things we've broken very easily.

How about we keep this as WIP until we've gotten the test suite more fleshed out? (which I know is my responsibility, and I am moving slowly on, I am suffering for time at the moment)

cmsj:
(and I realise that by not landing it immediately, we'd be pretty much forced into doing the work all over again)

heptal:
It's alright, I've had to scriptify the process to even do it in the first place

heptal:
Closing this one for now. Separating into one for moving crashlytics to podfile and will do another one for xconfig stuff later.

Issue #809, opened 2016-02-26
proposed additions to LuaSkin


asmagill:
I may have mentioned some of these in the past, but I want to put these forth and see if/how they may fit or affect plans others may have for LuaSkin/Hammerspoon...

* `checkArgs:`
  * add `LS_TINTEGER`.  This would remove a followup check that often occurs after verifying something is a number and make the error message more consistent since they would be coming from the same code source.  Should this be separate from `LS_TNUMBER` or something bitwise-or'ed like `LS_TOPTIONAL`?
  * there needs to be a way to specify `LS_TUSERDATA` without requiring the followup of the userdata type... this is required both so that `LS_TUSERDATA` can be paired with `LS_TOPTIONAL` and so you can specify that you want a userdata, but don't care the type (or will check the type later, because it could be more than one type in that argument position).
  * remove `LS_TNONE`. It serves no purpose as there is no way to specify a place holder that can't have something in it's place... It makes sense in `LUA_Txxx` type equivalents, since `nil` and `none` are identical to some of the `lua_toXXX` functions, and the don't require a terminating `LS_TBREAK`, but we do in `checkArgs:`
  * which leads to `LS_TBREAK`... should it be required?  Currently (or at least I haven't figured it out) there is no way to say "... and whatever else follows, as much as you want, following the initial requirements specified".  Or should we add something like `LS_TVARIABLE` as an alternative/or'ed  to/with `LS_TBREAK`?

* `pushNSObject:`
  * I finally have a reasonable prototype for an `NSValue` converter (well, I will later today/tomorrow) that can handle the known types it can hold, as well as provide the contents and some information about size and obCType when it doesn't hold something well known like NSRect, NSSize, etc.  I plan to put this in so rects, etc. stored in an NSValue because they are in an NSArray or NSDictionary can be properly pushed into Lua as their Hammerspoon equivalent table formats.
  * which poses the question... do we mirror this in the `toNSObjectAtIndex:` method by allowing it to check tables for a `__type` field which can be set to known types (primarily `color`, `rect`, `point`, and `size`, but there may be others that could benefit from this, like `styledtext`, etc.)

I'll probably think of more later (I always seem to!), but that will do for what is on my mind at the moment...

Agreements?  Disagreements?  Is something pending and I should wait?

asmagill:
... so I don't forget it:

[...]

PR #812, opened 2016-02-27
[WIP] LuaSkin additions

asmagill:
Per discussion in #809 

Adds LS_TINTEGER as optional addition to LS_TNUMBER
Adds LS_TSOFTBREAK as alternative to LS_TBREAK.  A soft break does not check to see if their are additional arguments on the stack

Adds NSValue support to pushNSObject: methods

A pushNSObject helper function can now return -1 (actually, any negative number, in case we eventually want to include different types of error) if it does not handle the conversion for pushNSObject: methods.  This allows for a helper function to allow another helper function to handle the conversion, or for the object to fall through to the default converters.  Useful if a class and a sub-class of that class should be handled differently or by different modules/functions.


LuaSkin/LuaSkin/Skin.m
```diff
@@ -294,6 +338,15 @@ - (void)checkArgs:(int)firstArg, ... {
                 break;
             case LUA_TNUMBER:
                 lsType = LS_TNUMBER;
+
+                if ((spec & lsType) && (spec & LS_TINTEGER)) {
+                    int isInteger ;
+                    lua_tointegerx(_L, idx, &isInteger) ;
+                    if (!isInteger) {
```
cmsj: I think this might break down a little if someone were to specify LS_TNUMBER | LS_TINTEGER. I'm not necessarily advocating that that is a sensible thing to do, it was just a thought that occurred to me. I think given the choice I'd rather document that that is not a sensible thing to do, than make this code more complex.

LuaSkin/LuaSkin/Skin.h
```diff
@@ -43,10 +47,14 @@
...
 /*! @define LS_TUSERDATA maps to LUA_TUSERDATA */
 #define LS_TUSERDATA      1 << 8
-/*! @define LS_TNONE maps to LUA_TNONE */
+/*! @define LS_TNONE maps to LUA_TNONE.  Deprecated, as this serves no real use in checkArgs except to provide parity with Lua's LUA_TNONE, which is handled by optional argument tagging or as an argument count error. */
 #define LS_TNONE          1 << 9
 /*! @define LS_TANY indicates that any Lua variable type is accepted */
 #define LS_TANY           1 << 10
+/*! @define LS_TINTEGER Can be OR'd with LS_TNUMBER to specify an integer argument.  This option is ignored if paired with other types. */
+#define LS_TINTEGER       1 << 11
+/*! @define LS_TSOFTBREAK Similar to LS_TBREAK, in that it signals the end of the argument list @link //apple_ref/occ/instm/LuaSkin/checkArgs: checkArgs @/link should validate; however, the argument count is not verified, so any additional arguments beyond the validated list are the responsibility of the module function to check and use or ignore. */
```
cmsj: How would you feel about making the name of LS_TSOFTBREAK more explicit? Something like LS_TVARARG?

asmagill: NP, will do so in the next update

LuaSkin/LuaSkin/Skin.m
asmagill: Not sure I follow... "pass a float"?  Do you mean in the arguments on the Lua stack?  That's the whole point here -- to detect whether or not they've put in something that isn't an integer (or immediately convertible into and integer) or not...

Do you have an example of what you have in mind?

cmsj: So I'm thinking that `static int module_foo()` calls `checkArgs` with `LS_TNUMBER | LS_TINTEGER` and from Lua, someone calls `hs.module.foo(9.1)`. I think this change will see that `LS_TINTEGER` is specified, look at the `9.1` and set `isInteger` to false, which causes us to raise a `luaL_error()`. I haven't tested whether this is true, it was just a thought that occurred to me as I was reading the PR.

LuaSkin/LuaSkin/Skin.h
cmsj: Thanks. It should probably also be explicitly documented that this only works as the final argument (and afaict from the code, it shouldn't matter if `LS_TBREAK` is specified afterwards, so we probably don't need to explicitly say anything about whether or not `LS_TBREAK` should follow it)

LuaSkin/LuaSkin/Skin.m
asmagill: Well, since it's set in the labels passed to `checkArgs:`, that's kind of the point... checkArgs throws lua errors when you pass in the wrong arguments.

However, I can see the argument that since we haven't done this in the past and because Lua does have a default behavior (it returns 0 for `lua_tointeger`) that doesn't throw an error (unless you explicitly use `luaL_checkinteger`), perhaps using `logError:` might be better.  It would be different than other argument errors, but perhaps kinder to pre-existing code. OTOH, if the module truly does want a proper integer, it will still have to use it's own integer check to know if 0 is because 0 was specified or because the number wasn't a proper integer.

I'll switch if you think it's a better approach.  Let me know.

cmsj: @asmagill I am doing a very bad job of explaining myself. In my example above, `9.1` is a valid argument, because the `checkArgs` spec was `LS_TNUMBER | LS_TINTEGER`, which would mean that both floats and integers are acceptable, but afaics the code will reject the floats.

asmagill: Ah... then I did a bad job in my description/documentation... I went with LS_TINTEGER being a modifier of LS_TNUMBER and not a stand-alone type... it's like LS_TOPTIONAL and means nothing by itself.  My bad for not being clearer...  Do you think it should be stand-alone?  My thinking was that if LS_TINTEGER truly was stand alone, then LS_TNUMBER | LS_TINTEGER has no meaning, since LS_TNUMBER would be sufficient without the alternative valid type... it seems more like the optional tag to me in that it modifies something but isn't a full definition itself.

cmsj: Oh! Sorry, I completely missed that in the docstring :(

cmsj:
I'm afraid that I've broken the cleanliness of your merge here. I'm done changing LuaSkin's code for a little while.

One other thing I'd say is that this should probably be the last time we land new code in LuaSkin without associated tests. We already have slipped significantly on code coverage - LuaSkin is currently standing at 47% coverage, which we need to improve soon.

asmagill:
I can work out the corrections...

I don't have a problem working up tests, but can you provide a good sample/walkthrough of what you're looking for?

cmsj:
I think we probably need to do some work on a more featureful test setup for LuaSkin, because right now it doesn't have a standardised way of loading an `init.lua` when the tests are running.

I'd say don't wait on that for this PR, I'll get to work on beefing up the test harness, and adding some tests, then it should be more obvious how to proceed with adding other tests, if you're interested in doing that :)

asmagill:
Almost ready to update the pull, but I noticed after the merge, the portion which generates the docs for LuaSkin gave this (which I've not seen before):

~~~
.../hammerspoon/LuaSkin/LuaSkin/Skin.h:273:Unknown field type @code in HeaderDoc comment.
.../hammerspoon/LuaSkin/LuaSkin/Skin.h:273:Unknown field type @endcode in HeaderDoc comment.
.../hammerspoon/LuaSkin/LuaSkin/Skin.h:297:Unknown field type @remark in HeaderDoc comment.
.../hammerspoon/LuaSkin/LuaSkin/Skin.h:315:Unknown field type @remark in HeaderDoc comment.
.../hammerspoon/LuaSkin/LuaSkin/Skin.h:326:Unknown field type @remark in HeaderDoc comment.
.../hammerspoon/LuaSkin/LuaSkin/Skin.h:339:Unknown field type @remark in HeaderDoc comment.
.../hammerspoon/LuaSkin/LuaSkin/Skin.h:374:Unknown field type @remark in HeaderDoc comment.
~~~

Does the Makefile need updating as well to adjust the arguments to `headerdoc2html` (or do I need a new version/addition to it?)

asmagill:
`LS_TVARARG` is now an option to be OR'd with `LS_TBREAK` instead of an alternative to it.

cmsj:
@asmagill. *%$😍🔥🔫ing HeaderDoc. I was getting errors from Xcode so I switched some tags around, now the CLI generator is erroring?!

I'll see what I can do, thanks for mentioning it :)

cmsj:
Also +1 to LS_TVARARG being an OR with TBREAK :)

asmagill:
@cmsj, I have no recollection as to why I put the dictionaries which hold info about the helper functions in LuaSkin.m as global variables in the file instead of as properties or in an extension to the class...

https://github.com/asmagill/hammerspoon/blob/LuaSkin/LuaSkin/LuaSkin/Skin.m#L85-L89

Since I'm already working on this, do you know of a reason to or not to change this?  A part of me wants to because it seems to make more sense and is more "correct" for an Objective-C application and another part of me says "don't fix what ain't broke"...

cmsj:
@asmagill So the best answer to that is to add tests for all the helper stuff and then move them, thus having confidence that nothing broke :D

but yes, I do think they should move away from being global variables. I'd actually been meaning to point that out, since it may have important relevance to your threading fork.

fwiw, I'm working on the test harness for LuaSkin at the moment, and attempting to somewhat unify the HS/LS harnesses, without making LS depend on HS for anything. It's going well and I'd hope I have something landable today/tomorrow.

asmagill:
@cmsj, and while I'm rooting around in LuaSkin, per your comment regarding looking for __type in a userdata at https://github.com/asmagill/hammerspoon/blob/LuaSkin/LuaSkin/LuaSkin/Skin.m#L1056...

Upon further thought, the reason this works for us is because all of our userdatas so far have an `__index` metatable.  The first one that doesn't will fail with that all too descriptive error "tried to index a userdata" or some-such.

I'm thinking the proper thing to do is something along the lines of:
~~~objc
...
        case LUA_TUSERDATA: // Note: ...
             lua_pushcfunction(self.L, helper-function) ;
             lua_pushvalue(self.L, realIndex) ;
             if ((lua_pcall(self.L, 1, 1, 0) == LUA_OK) && (lua_type(L, -1) == LUA_TSTRING) {
                 userdataTag = (char *)lua_tostring(self.L, -1);
             }
             lua_pop(L, 1) ; // if it errors, the error is on the stack... if it doesn't, our result is
...
~~~

where `helper-function` is defined somewhere as

~~~objc
static int helper-function(lua_State *L) {
    if (lua_getfield(self.L, 1, "__type") != LUA_TSTRING) {
        lua_pop(L, 1) ;
        lua_pushnil(L) ; // only a string is allowed to return, everything else is nil
    }
    return 1;
}
~~~

To keep this HS independent, `helper-function` (with a better name of course) needs to go into `Skin.m` somewhere before the class is implemented.

Do you agree with my thinking?  I'll add this as well, if you see no objections.

asmagill:
ok, technically the pop in helper-function isn't needed, and self.L should be just L... it's late, so I'll go before I break something and review again later 😴

cmsj:
@asmagill are we expecting to have userdatas without that in their metatable? I rather like living in a world where everything has a __type, __tostring, etc :)

asmagill:
Anything in core and anything that registers through LuaSkin should never not have the field... but third-party or stuff inherited from outside?  As an example, I don't recall if hs.fs uses a userdata for anything, but outside of a few additions and the watcher, it's not our code and doesn't use LuaSkin...

as the note (you?) left in the file says, it could potentially be dangerous outside of a protected call and it finally hit me as to why and how to fix it...

cmsj:
yeah I agree we should try and avoid the dangerous pcall, I'd just like it if we could keep userdata objects in a common format (fwiw I don't think hs.fs has any that we inherited from lfs)

asmagill:
@cmsj, I think the lua_pcall (I don't use the LuaSkin `protectedCallAndTraceback:nresults:` because I don't care about the traceback, so no need for the overhead) is actually safer then the unprotected lua_getfield... getfield can cause a lua error to occur, pcall doesn't.  Granted all of our userdata types should be safe without this (and have been), but this protects those that we may miss in the future or that are in third-party modules.

This method converts lua objects on the stack into NSObjects... it is called either from within the c code of a module function or immediately after a lua function callback to convert results into usable Objective-C data... in either case, I think it's a safe bet that the Lua stack can stand up to a lua_pcall at the points where this code will be invoked.

Unless you can think of a case that I'm missing right now...

At any rate, a potential pull with both of the latest discussion items in it that I will be testing out today can be found at https://github.com/asmagill/hammerspoon/tree/LuaSkin2 if you want to take a look.  It builds on this one, so if it passes muster, I'll probably merge them into this one, rather than create a second pull.

asmagill:
(FWIW, the LuaSkin2 branch does seem to work quite well with my not-simple Hammerspoon setup and greatly simplifies the LuaSkinThread overrides in hs._asm.luathread... obviously I will test it more, but it looks promising)

asmagill:
@cmsj, If using `@code/@endcode` and `@remark` really do make things better in XCode or another tool you're using, the only work around I've found so far with the html generation from the makefile would be to do something like:
~~~
sed -e 's/@code/<pre>@textblock/' -e 's/@endcode/@\/textblock<\/pre>/' -e 's/@remark/@attributeblock Notes/' LuaSkin/LuaSkin/Skin.h > test.h
~~~
and then run `headerdoc2html` on that file...

As far as I can determine, Apple has stopped using `@code/@endcode` and `@remark` in their own files as well... only `/System/Library/Frameworks/GSS.framework/Versions/A/Headers/gssapi_protos.h` seems to still have `@code/@endcode` in it's header file.

asmagill:
of course... bc master changed... another conflict... brb

asmagill:
I'm ready to merge this, once Travis verifies I haven't done anything to break anything.

I'll open a new pull when I get around to looking closer at expanding userdata options to checkArgs:

@cmsj, let me know if you're in the middle of something and I should hold off...

cmsj:
All I'm doing at the moment is writing tests for LuaSkin, no changes outstanding for the framework itself :)

PR #814, opened 2016-02-29
enable strict building for LuaSkin

heptal:
-Wno-nullable-to-nonnull-conversion can be fixed with a few (String * _Nonnull)s

Some of the omitted errors would really ugly it up if the code was changed, though.


cmsj:
I think this should be covered by 11b574bff94ca8451cd18c2f15df27692ea1eb19

PR #817, opened 2016-03-02
hs.fs tests and refactoring

heptal:
hs.fs coverage to 96%
hs.fs.volume coverage to 97%
-Weverything -Werror
Removed win32 code and other such unused `#define`s
Reindented


extensions/fs/test_fs.lua
```diff
@@ -0,0 +1,331 @@
+-- hs.fs tests
+
+homeDir = os.getenv("HOME").."/"
```
cmsj: I'm a little nervous about depending on the environment for a path, and then appending "/" to it, it has all the makings of a "rm -r /" bug.

How about we just hard code to using `/tmp/hs_fs_test_dir/`?

cmsj:
Looking good, just had one comment :)

heptal:
Fixed

cmsj:
Thanks :)

(I tweaked it ever-so-slightly to just use `/tmp/` - I'm not sure if we're guaranteed that `/private/tmp`, which is after-all the same actual folder - will remain where it is)

PR #822, opened 2016-03-06
Fix hs.notify.show() subTitle display

heptal:
The `hs.notify.show()` convenience constructor was using the wrong key ('subtitle' instead of 'subTitle')
Improves function-tag handling in `hs.notify`
Reindents lua code

Also makes some error logging more consistent in `hs.osascript` and `hs.socket`


cmsj:
Thanks for fixing `hs.notify.show()`. FWIW, it would be easier for me if PRs like this could be split up more, particularly when whitespace is involved :)

heptal:
Noted :)

PR #824, opened 2016-03-06
add ability to set identity image for notification

heptal:
In addition to `hs.notify:contentImage`:

Adds `hs.notify:setIdImage`:

Requires @asmagill's review and approval because this uses a private API method:

http://stackoverflow.com/questions/19797841/nsusernotification-mavericks-custom-images

### DIFF
--- a/extensions/notify/init.lua
+++ b/extensions/notify/init.lua
@@ -319,13 +319,13 @@ hs.getObjectMetatable("hs.notify").contentImage = function(...)
   end
 end
 
---- hs.notify:setIdImage(image[, hasBorder]) -> notificationObject
+--- hs.notify:setIdImage(image[, withBorder]) -> notificationObject
 --- Method
 --- Set a notification's identification image (replace the Hammerspoon icon with a custom image)
 ---
 --- Parameters:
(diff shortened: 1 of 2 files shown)


cmsj:
As I said on IRC, I'm minded to accept this because it's a fairly simple addition to an API we're already using, and it's wrapped in a `respondsToSelector`, but I want a +1 from @asmagill :)

asmagill:
+1

This is actually something I had in the back of my mind to look into at some point... I came across https://github.com/indragiek/NSUserNotificationPrivate when I was last working on this module, but never got around to doing anything with it.

heptal:
Oh, nice. I'll add the `_identityImageHasBorder` property as well.

cmsj:
@heptal I'm not sure what "whether or not the image has a border" means - is this an argument that says "please draw a border", or "please don't draw a border because the image already has one"? I think the docs should be worded in a way that makes that question irrelevant. Apart from that, this is good to land :)

heptal:
It specifies whether or not OS X should draw a border - the image in my comment has such a border drawn (which is the system default, but imo shouldn't be the user-exposed default because most images seem to look better without it).

PR #825, opened 2016-03-09
[WIP] luaTable mapping to NSObject support

asmagill:
* Adds support for adding `__luaSkinType = "something"` to allow `[skin toNSObjectAtIndex:#]` to convert tables as well as userdata directly into objects

* Adds support for NSValue as a __luaSkinType'd table, returning any of the structures supported by NSValue wrapped in NSValue.  e.g. for a table:

~~~lua
bounds = { __luaSkinType="NSRect", x = 1.0, y = 2.0, h = 100.0, w = 200.0 }
~~~

You can get at the NSRect in two ways:

~~~objc
NSRect boundsRect = [skin tableToRectAtIndex:1] ;
// or
NSRect boundsRect = [[skin toNSObjectAtIndex:1] rectValue] ;
~~~

The benefit being that `[skin toNSObjectAtIndex:1] ` can be a key/value in an array or dictionary (so easier to pass around in batches), whereas `[[skin toNSObjectAtIndex:1] rectValue]` and `[skin tableToRectAtIndex:1]` can't.

* Also allows unrecognized NSValue structures/pointers to be passed in and out as `{ __luaSkinType = "NSValue", data = "data-as-bytes-in-string", objCType = "@encode compatible string"}`; probably most useful for realizing we need to add another convertor of some sort.

* Added tableMapping for hs.drawing.color tables (so `[skin toNSObjectAtIndex:1]` can return NSColor without specific code in place that knows its a color before hand) and tableMapping for table form of hs.styledtext (NSAttributedString).

* Needs more testing, possibly adding more table forms to other convertors already in palce.

* Not sure what to do with `hs.geometry` yet...  There needs to be some logic in place to determine which __luaSkinType label gets added to the table when passed into obj-c land, but I haven't looked at the module yet to determine if this will be difficult to pin down specifically, since lua can be fluid about rect/point/size, but obj-c cannot.

* @cmsj - I changed all `(char *)` arguments/parameters in LuaSkin to `(const char *)`, since that's what they are 99% of the time -- either a return value from a `lua_Xstring` function or an actual c-string literal.  Is that going to be a problem for anything you're planning or working on?  I can change back, but I got tired of having to cast in various places...


cmsj:
@asmagill I have nothing brewing that would be affected by making all the chars, const :)

cmsj:
We also need a testing story here :D

LuaSkin tests are at about 70% code coverage now, but there are big gaps in the NSValue and NSNumber conversions, because I wasn't sure exactly what some of those things are, so I was hoping I could enlist your help fleshing those parts out, in return for getting you up to speed on how it all works? :)

asmagill:
I've been thinking about how to best test them myself... some of the NSValue conversions aren't really necessary at present, but I wanted to include them to provide a complete coverage of the class and because we don't know if or when they might become important and/or necessary... the CLLocationCoordinate2D and MKCoordinateSpan structures, for example, would be useful with some extensions to `hs.location` I've been considering...

I'll take a look at the LuaSkinTests code and see if I understand it well enough to add to it...

asmagill:
@cmsj, Ok, for adding tests... I'm thinking I add them to LuaSkinTests.m in the LuaSkinTests folder... do I then have to add them to a list somewhere?  And to run the tests... I assume this has to be done within XCode?  Or is their something in the scripts folder I should be looking at?  And to your comment in the file about NSValue not having a valueWithCMTime method... the proper imports should be in LuaSkin.h now... have you tried it recently, or maybe they need to be added to the LuaSkinTests.m file as well?  I thought having them in the framework's header would be enough, but I'm not positive.

heptal:
The easiest way is to open Xcode and click the diamond (which means tests):


Clicking any of the tests under LuaSkinTests will open LuaSkinTests.m and bring you to that function. Simply adding a new test function here and saving will automatically populate the test list in the sidebar with your new test. You can then hit the 'play' or the diamond next to any function to run that test individually:


Sometimes you have to click away and back again before it will run the test because Xcode. I'm sure you can do this from your preferred editor and some xcodebuild invocation but it's easier just to do it from Xcode imo.

asmagill:
Yeah, I figured doing it in XCode would be easiest/necessary... I generally don't use XCode directly when I don't have to because I do most of my work on my personal machine, which is a MacBook Air with only 4GB of memory... (fingers crossed that may change in a couple of months).  At least the first few times, I'll follow your example in XCode... once I'm more comfortable with it, we'll see.

cmsj:
'xcodebuild test' will run the tests, but it will run all of them, which takes ages (just over 5 mins on a quad core 4GHz i7). I'm not sure if there's a way to specify a subset of tests to run.

cmsj:
@asmagill to answer your first question, you do add them to LuaSkinTests.m, but you don't need to add them to a list, XCTest will run any method in that class whose name starts with `test`.

As for running the tests: `xcodebuild -workspace Hammerspoon.xcworkspace/ -scheme LuaSkin test` will just run the tests for LuaSkin, because we have a separate scheme for that. It's going to be more of an issue for you when you're working on an extension, but we can cross that bridge when we come to it :)

Re the missing method on NSValue, I did try that very recently, and the imports were there. It did just occur to me that I didn't explicitly link the frameworks, but I thought the `@import` would do that for me.

asmagill:
@cmsj: first, are the additions I've made to the testPushNSObject function in line with the kind of things you want?

Second, got a weird one that must have something to do with differences in the test environment and Hammerspoon as an actual application (even when running attached to XCode)...  The SCNVector3 and SCNVector4 tests are commented out because within the test environment, the encoding type is coming back as NSRect, so the table returned is wrong.  However, when the following function is in a module loaded into a running Hammerspoon, it outputs what you'd expect (i.e it's correct):

~~~objc
static int testNSValueEncodings(lua_State *L) {
    LuaSkin *skin = LST_getLuaSkin(); //[LuaSkin shared] ;
    NSValue *theRect  = [NSValue valueWithRect:NSMakeRect(0,1,2,3)] ;
    NSValue *thePoint = [NSValue valueWithPoint:NSMakePoint(4,5)] ;
    NSValue *theSize  = [NSValue valueWithSize:NSMakeSize(6,7)] ;
    NSValue *theRange = [NSValue valueWithRange:NSMakeRange(8,9)] ;
    NSValue *vector3  = [NSValue valueWithSCNVector3:SCNVector3Make(1,2,3)] ;
    NSValue *vector4  = [NSValue valueWithSCNVector4:SCNVector4Make(4,3,2,1)] ;

    lua_newtable(L) ;
    lua_newtable(L) ;
    [skin pushNSObject:theRect withOptions:LS_NSDescribeUnknownTypes] ;  lua_setfield(L, -2, "rect") ;
    [skin pushNSObject:thePoint withOptions:LS_NSDescribeUnknownTypes] ; lua_setfield(L, -2, "point") ;
    [skin pushNSObject:theSize withOptions:LS_NSDescribeUnknownTypes] ;  lua_setfield(L, -2, "size") ;
    [skin pushNSObject:theRange withOptions:LS_NSDescribeUnknownTypes] ; lua_setfield(L, -2, "range") ;
    [skin pushNSObject:vector3 withOptions:LS_NSDescribeUnknownTypes] ;  lua_setfield(L, -2, "vector3") ;
    [skin pushNSObject:vector4 withOptions:LS_NSDescribeUnknownTypes] ;  lua_setfield(L, -2, "vector4") ;

    lua_setfield(L, -2, "raw") ;
    lua_newtable(L) ;
    lua_pushstring(L, [theRect objCType]) ;  lua_setfield(L, -2, "rect") ;
    lua_pushstring(L, [thePoint objCType]) ; lua_setfield(L, -2, "point") ;
    lua_pushstring(L, [theSize objCType]) ;  lua_setfield(L, -2, "size") ;
    lua_pushstring(L, [theRange objCType]) ; lua_setfield(L, -2, "range") ;
    lua_pushstring(L, [vector3 objCType]) ;  lua_setfield(L, -2, "vector3") ;
    lua_pushstring(L, @encode(SCNVector3)) ; lua_setfield(L, -2, "vector3e") ;
    lua_pushstring(L, [vector4 objCType]) ;  lua_setfield(L, -2, "vector4") ;
    lua_pushstring(L, @encode(SCNVector4)) ; lua_setfield(L, -2, "vector4e") ;
    lua_setfield(L, -2, "objCType") ;
    return 1 ;
}
~~~

The output is:

~~~lua
> inspect(_xtras.testNSValue())
{
  objCType = {
    point = "{CGPoint=dd}",
    range = "{_NSRange=QQ}",
    rect = "{CGRect={CGPoint=dd}{CGSize=dd}}",
    size = "{CGSize=dd}",
    vector3 = "{_SCNVector3=ddd}",
    vector3e = "{SCNVector3=ddd}",
    vector4 = "{_SCNVector4=dddd}",
    vector4e = "{SCNVector4=dddd}"
  },
  raw = {
    point = {
      __luaSkinType = "NSPoint",
      x = 4.0,
      y = 5.0
    },
    range = {
      __luaSkinType = "NSRange",
      length = 9,
      location = 8
    },
    rect = {
      __luaSkinType = "NSRect",
      h = 3.0,
      w = 2.0,
      x = 0.0,
      y = 1.0
    },
    size = {
      __luaSkinType = "NSSize",
      h = 7.0,
      w = 6.0
    },
    vector3 = {
      __luaSkinType = "SCNVector3",
      x = 1.0,
      y = 2.0,
      z = 3.0
    },
    vector4 = {
      __luaSkinType = "SCNVector4",
      w = 1.0,
      x = 4.0,
      y = 3.0,
      z = 2.0
    }
  }
}
~~~

Any ideas?

heptal:
I would bet it's an Apple bug. I had a lot of issues with newer frameworks like SceneKit within the last year (filed bugs w/ Apple and they acknowledged the issue, tho never got conclusive followup), though I don't understand how such a simple addition to NSValue can be broken.

asmagill:
@heptal, understood, though what's weird to me about the whole thing is that it works properly *outside* of the test environment...

@cmsj, trying to run the tests from the command line nets me this:

~~~sh
$ xcodebuild -workspace Hammerspoon.xcworkspace/ -scheme LuaSkin test
xcodebuild: error: Scheme LuaSkin is not currently configured for the test action.
~~~

some googling found that this can sometimes occur when a scheme name exists twice in the workspace, so:

~~~sh
$ xcodebuild -workspace Hammerspoon.xcworkspace/ -list
Information about workspace "Hammerspoon":
    Schemes:
        Hammerspoon
        Release
        alert
        applescript
        application
        applicationwatcher
        audiodevice
        base64
        battery
        batterywatcher
        brightness
        caffeinate
        caffeinatewatcher
        crash
        dockicon
        drawing
        eventtap
        eventtapevent
        fs
        hash
        hints
        host
        hotkey
        http
        httpserver
        image
        ipc
        hs
        json
        keycodes
        location
        menubar
        milight
        mouse
        notify
        pasteboard
        pathwatcher
        screen
        screenwatcher
        settings
        sound
        spaceswatcher
        timer
        uielement
        urlevent
        usb
        usbwatcher
        wifi
        wifiwatcher
        window
        Pods-Hammerspoon
        Pods-Hammerspoon-ASCIImage
        Pods-Hammerspoon-CocoaAsyncSocket
        Pods-Hammerspoon-CocoaHTTPServer
        Pods-Hammerspoon-CocoaLumberjack
        LuaSkin
        lua
        alert copy
        webview copy
        drawing copy
        audiodevicewatcher
        drawing copy 2
        alert copy 2
        applescript copy
        task
        alert copy 4
        battery copy
        batterywatcher copy
        fsvolume
        chooser
        ASCIImage
        CocoaAsyncSocket
        CocoaHTTPServer
        CocoaLumberjack
        alert copy 6
        networkconfiguration copy
        networkreachability copy
        LuaSkin
        socket
        socketudp
        alert copy 8
~~~

Sure enough, LuaSkin is listed twice... and some other things that I suspect could be removed/renamed/whatever as well.  I'm not nearly proficient enough with Xcode to attempt a cleanup, but I thought I'd document it in case someone else wants to do so.

Being able to run individual tests in Xcode isn't too bad; doing a full compile/full test run with the limited memory is a pain, but for single shot testing while adding them, it's not a problem.

asmagill:
Arrrggghhh... now SCNVector3/4 is failing outside of the test environment and the only change I've made is to upgrade Xcode...

Ok, the plan now, unless anyone has objections, is to reduce auto NSValue support down to just NSRect, NSPoint, NSSize, NSRange, and NSValue with objCType and data fields... while I would like to support all of the "native" formats, apparently even that's not quite as "supported" as the docs suggest.

The LuaSkin additions I've made do allow creating custom helper functions that are called when a LuaTable has a specific label, so the other formats can be added as helper functions if and when needed.

asmagill:
Unless anyone has any objections, I think this is ready to go as well.  I will add additional unit tests for more things over time, but I think that should be a separate effort from the functionality this adds.

asmagill:
@cmsj, I'd like your review of the changes made to compile in XCode 7.3... especially for `hs.fs` and `hs.osascript`...

Mostly the fixes included ensuring proper casts to remove implicit numeric conversions, but those two apparently use methods which aren't available in all OS versions the project is supposedly built for, and 7.3 doesn't take kindly to that.

It may be that the project target OS needs to be updated, or some warnings-as-errors relaxed, but as that means changing the XCode project files, rather than source, I'm leery of touching it (though if you decide to go that route, check the workspaces list I mention above... as good a time as any to clean it up at the same time)

At any rate, for the immediate future, assuming Travis doesn't have a problem with this, I'd say this is ready to land since I've added nothing that should affect compiling in 7.2, pending your review of course.

cmsj:
@asmagill so I think the casting stuff is good, it makes us more explicit about what we are telling the compiler to do. The partial availability one I think is potentially more tricky - the compiler warnings (so, errors in our case) are telling us that we're using features that were added >10.8. Disabling the warnings isn't necessarily wrong, but we're effectively saying "it's fine if this crashes on 10.8/10.9/whatever".

I continue to not be particularly in favour of supporting old releases, because I don't ever use/test them, so we could easily be super broken there in all sorts of ways and not know.

tl;dr, I'm fine with the changes you have for now, and I definitely don't want to make any of them be global build options, because I think we should be casting things correctly, and I think the partial availability warnings/errors are helpful to keep us from accidentally breaking older OSes :)

I'm also +1 for the rest of this PR landing. Thanks!

asmagill:
We should wrap things that are potentially going to cause problems with earlier versions, and I agree that  suspending the warnings for the newer code is a band-aid, but we have things that have been added/removed in every OS X version... just dropping 10.8 and even 10.9 won't fully remove the problem... NSAppleEventDescriptor didn't support doubleValue until 10.11, and some additions I'd like to eventually put in webview were just added recently as well...

Are you aware of a tool which would check the code against a range of OS X versions so we can see where we are missing appropriate checks?  Or do we have to just try it with a different target selected and generate such a list manually?

At any rate, that's beyond the scope of this pull, so I'm going to land it.

PR #826, opened 2016-03-10
add lua-stdlib

heptal:
### hs.stdlib

This module provides functionality from lua-stdlib
Documentation for [`std`](https://github.com/lua-stdlib/lua-stdlib): http://lua-stdlib.github.io/lua-stdlib/
Documentation for [`std.functional`](https://github.com/lua-stdlib/functional): http://lua-stdlib.github.io/functional/
Documentation for [`std.prototype`](https://github.com/lua-stdlib/prototype): http://lua-stdlib.github.io/prototype/
Documentation for [`std.strict`](https://github.com/lua-stdlib/strict): http://lua-stdlib.github.io/strict/
#### Usage:

``` lua
std = require 'hs.stdlib'
table = std.table

k = table.keys({ a=1, b=true, c="hello" })
hs.inspect(k)
 { "b", "c", "a" }

f = std.functional
m = f.map(f.lambda '=_1*_1', {1, 2, 3, 4})
hs.inspect(m)
 { 1, 4, 9, 16 }
```

Not referencing lib contents in `project.pbxproj` as it seems unnecessary and cluttering.

### DIFF
--- a/Hammerspoon.xcodeproj/project.pbxproj
+++ b/Hammerspoon.xcodeproj/project.pbxproj
@@ -704,6 +704,7 @@
 /* Begin PBXFileReference section */
 		19B5A52CAF8760B10C86C2D0 /* Pods-Hammerspoon.release.xcconfig */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = text.xcconfig; name = "Pods-Hammerspoon.release.xcconfig"; path = "Pods/Target Support Files/Pods-Hammerspoon/Pods-Hammerspoon.release.xcconfig"; sourceTree = "<group>"; };
 		1A11170BCF0BA210FA17BF37 /* Pods-Hammerspoon.debug.xcconfig */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = text.xcconfig; name = "Pods-Hammerspoon.debug.xcconfig"; path = "Pods/Target Support Files/Pods-Hammerspoon/Pods-Hammerspoon.debug.xcconfig"; sourceTree = "<group>"; };
+		4C049C241C9106B900AC6FC8 /* init.lua */ = {isa = PBXFileReference; lastKnownFileType = text; name = init.lua; path = extensions/stdlib/init.lua; sourceTree = "<group>"; };
 		4C3D55A91C7F05D00023B633 /* init.lua */ = {isa = PBXFileReference; lastKnownFileType = text; path = init.lua; sourceTree = "<group>"; };
 		4C3D55AA1C7F05D00023B633 /* internal.m */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.objc; path = internal.m; sourceTree = "<group>"; };
 		4C3D55AD1C7F05D00023B633 /* test_javascript.lua */ = {isa = PBXFileReference; lastKnownFileType = text; path = test_javascript.lua; sourceTree = "<group>"; };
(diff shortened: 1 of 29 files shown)


cmsj:
@asmagill any thoughts on this? I'm torn, I like stdlib, but I'm not sure about adding one more thing to our plate for future migration/deprecation (hs.fnutils)

asmagill:
**TL;DR** - basically I like what this offers, but accepting/rejecting it should be with a specific plan forward that addresses at least point 1 below.

I do think fnutils should probably be deprecated -- this is the second (third?) time something more powerful and more widely used has been proposed which, with minor modifications to our existing code, could replace 95%+ of fnutils... the one or two others could always be added to `hs` if they're truly necessary...

My thoughts fall in slightly different directions...

1. We would be basically replicating something that *can* be included through other means already -- LuaRocks or a local copy from the git repo... we do it with hs.inspect, and modified lfs (hs.fs), so there is precedent, but what is/where is the line?  Maybe including LuaRocks support (I haven't look at it in a long time, so I don't know if this is even feasible) so that people aren't required to have Homebrew installed to add third party things like this would be better than evaluating these individually each time it comes up...

2. Documentation... A lot of third party stuff uses LDoc, and I was hoping to take a look at that this weekend and see how hard it would be to add support for it in hs.doc...  I've already added some methods to hs.doc (don't recall if I've added them to core yet or if its only local to my dev machine, but none-the-less it's in progress) which can generate the necessary XML from both obj-c files and lua files and then either be piped into the local parser for `hs.help` or into an HTML filter for Dash, a web page, etc.  (I use it to create a markdown skeleton for the README.md files in my personal repos, for example) I plan to see what it would take to add LuaDoc (and maybe later HeaderDoc as well) so that all Documentation generation can be done with one central toolset.

cmsj:
We probably have almost all of the pieces required for a luarocks client hosted inside Hammerspoon that downloads libraries into ~/.hammerspoon/ but I'm not sure how useful it is. I find myself never installing things with luarocks, so I'm not able to judge how worth it is to support it.

To specifically address point 1, I don't know where the line is. The judgement was pretty easy for inspect/lfs, but I don't really have an opinion on stdlib. I like it, but fnutils isn't particularly bothering me.

asmagill:
I'm going to take a crack at modifying the core module usage of fnutils to see what it might take if we did decide to move to this; however, whether we truly do remove `hs.fnutils` or not, the following needs to be added to the top-level `init.lua` file of `hs.stdlib` before the first `require` if we're going to include this:

~~~lua
local stdlibPath = package.searchpath("hs.stdlib", package.path):match("^(.*)/init.lua$")
package.path = package.path..";"..stdlibPath.."/lib/?.lua;"..stdlibPath.."/lib/?/init.lua"
~~~

jasonm23:
Any movement on this, IMO luarocks support would basically relieve hammerspoon of many maintenance costs, and put more power in the hands of users.

:+1:

cmsj:
We already support luarocks. Just use it to install things and Hammerspoon can require() them :)

cmsj:
I keep seeing this PR open and it's annoying me. Not the PR, just that we haven't figured out what to do about it.

How about this? I'm willing to include stdlib if the following conditions are met:
 * as close to zero modifications as possible vs upstream, so we can keep it updated easily. Ideally zero modifications.
 * If modifications to stdlib are required, the PR branch should contain two commits - the first being the inclusion of stock upstream stdlib, the second being the necessary modifications (so it's really easy later to refer back to the changes that were made)
 * hs.fnutils is also rewritten to, as much as is possible, just be a wrapper on top of stdlib.

In theory it would also be nice if our usage of hs.fnutils was migrated to using stdlib, but I think that's an unreasonably high bar at this stage.

Basically, my aim here is to make it easy for us to maintain stdlib in the future, and to make sure that we're actually using it.

Any objections? Any volunteers to get this/another PR in line with the above?

cmsj:
Seems like there's no interest in doing this, or at least no available time (which is fine, we're all busy :) so I'm going to close this out. If someone wants to resurrect it later, please feel free :)

PR #836, opened 2016-03-14
Completed description info for hs.keycodes extension

Lisergishnu:
When browsing the API docs I noticed there was no mention on the main document about keyboard layouts. After checking the code I noticed the hs.keycodes extension has the exact functionality I was looking for but the module info didn’t mention it at all. This small patch adds that info so it appears on the docs.


cmsj:
Thanks, good catch!

Lisergishnu:
No problem! :ok_hand:

Issue #831, opened 2016-03-13
MenuIcon:setColor?


kevinastone:
Could we set the color of the menubar icon?  Is that feasible?

(as a bonus ask, can we set/manipulate the hammerspoon menubar icon?)

Seems like a great way to flag or notify the user.

asmagill:
It actually should be pretty easy to modify the title method to support hs.styledtext like the menu items do now... not really certain why I missed it when last updating the module.  It may take a few days, but I'll add it to my list.

As for the Hammerspoon menubar icon, currently no, it can't be modified.  What I've done is hide the menu icon from the application preferences panel and created my own "Hammerspoon" menu using the same icon... if you did something similar, you could easily change the icon based upon timers, etc.

asmagill:
addressed with #838

Issue #821, opened 2016-03-04
hs.menubar question


asmagill:
Since this involves using unpublished selectors in NSStatusBar and NSStatusItem, I figured it more appropriate to ask here rather than the Google group...

Is there interest in being able to insert `hs.menubar` items with different priorities to give some control over positioning?

I'm using a modified version of code found at https://github.com/shergin/NSStatusBar-MISSINGOrder, and it seems to work well under 10.11 -- I've been able to place and move menubar items before and after the Notification Center icon, the Spotlight Icon, the set of System icons and the set of "other" (i.e. application) icons in the system status bar... but it does make use of NSInvocation to invoke unpublished selectors in the classes.

cmsj:
from looking at the repo you linked to, this seems reasonably safe to do. Perhaps we could play doubly safe by only invoking their codepaths at all, if the user specifies order arguments?

asmagill:
I'll take a look at my modified module... I think I set it up as a separate method... certainly none of the existing methods/constructors need to be touched.

I'm also going to address #831 when I get to this sometime in the next few days. (hence the note to myself here :-P )

asmagill:
addressed with #838

Issue #832, opened 2016-03-13
addition to hs.application:getMenuItems


asmagill:
Any objections to adding the key kAXMenuItemCmdGlyphAttribute to the attributes returned when getting an applications menus?  I need to find a more formal source to add a proper lookup for the numeric result it returns then http://superuser.com/questions/415213/mac-app-to-compile-and-reference-keyboard-shortcuts, but it would allow programs like @dharmapoudel's at https://github.com/dharmapoudel/hammerspoon-config to include menu items that have a visible glyph rather than a command character (think arrow keys, delete, etc.)?

The update would be to the _getMenuStructure function in hs.application's internal.m file at around line 802... I don't believe it should affect anything else since it's just an addition to the query dictionary.

dharmapoudel:
That would be an awesome addition. The cheatsheet copycat is pretty limited to AXMenuItemCmdChar . Also is there any way (or workarounds) to detect long press of any keys?

asmagill:
closed with #838

asmagill:
@dharmapoudel, to detect long keypresses, you need to delve into `hs.eventtap` and `hs.timer`.  I recently updated a modified version of your cheatsheet code at https://github.com/asmagill/hammerspoon-config/blob/master/utils/_keys/cheatsheet.lua which watches for a 3 second press of just the command key that seems to work well... I just added the command key detection, so consider it still in "beta testing", but you're welcome to take a look at and use any portion of it that appeals to you.

(My version does require the glyphs which were recently added to the development version of Hammerspoon, but the eventtap stuff should work right now with a stock Hammerspoon release if that's what you're running.)

dharmapoudel:
Awesome. Thanks!

PR #838, opened 2016-03-17
[WIP] add attributedTitle support and priority positioning to hs.menubar

asmagill:
This pull adds:
* styledText support to the `setTitle` and `title` methods
* `hs.menubar.newWithPriority` to create a menubar item with a positioning priority other than the default
* `hs.menubar:priority` to get/set an existing menubar item's positioning priority

Addresses #831 and #821

While I believe it is ready for integration (I have no further expected additions to this module at present), I want to run a few more tests, but I figured I'd put it out there for anyone who's been following the referenced issues and wants to work with it now.

### DIFF
--- a/extensions/application/init.lua
+++ b/extensions/application/init.lua
@@ -185,6 +185,156 @@ function application.open(app,wait,waitForWindow)
   return r
 end
 
+--- hs.application.menuGlyphs
+--- Variable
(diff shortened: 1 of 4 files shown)


asmagill:
Adding menu glyphs per #832

Assuming Travis like this, and no one has any objections, I think this is ready to go.

PR #847, opened 2016-03-26
add LS_TTYPEDTABLE to checkArgs:

asmagill:
Forgot comment...

adding LS_TTYPEDTABLE to check args to further support tables with __luaSkinType field.

E.g.

~~~objc
[skin checkArgs:LS_TTYPEDTABLE, "NSPoint", LS_TBREAK]
~~~
will verify that the table at the first index position includes the __luaSkinType field with a value of "NSPoint", as is used by LuaSkin to determine if it can automatically convert a table into an Objective-C type of some sort.

Needs some additions to the test suite.

@cmsj syntacticly, is this better as a separate type (as defined here) or as an optional addition to LS_TTABLE (for example `LS_TTABLE | LS_TTYPED` instead of `LS_TTYPEDTABLE`)?

### DIFF
--- a/LuaSkin/LuaSkin/Skin.m
+++ b/LuaSkin/LuaSkin/Skin.m
@@ -1,4 +1,4 @@
- //
+//
 //  Skin.m
 //  LuaSkin
 //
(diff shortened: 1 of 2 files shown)


LuaSkin/LuaSkin/Skin.m
```diff
@@ -335,7 +335,24 @@ - (void)checkArgs:(int)firstArg, ... {
                 lsType = LS_TFUNCTION;
                 break;
             case LUA_TTABLE:
-                lsType = LS_TTABLE;
+                if (spec & LS_TTYPEDTABLE) {
+                    lsType = LS_TTYPEDTABLE;
+                    char *expectedTableTag = va_arg(args, char*);
+// NOTE: Make sure LS_TTYPEMASK in Skin.h is updated as options are added. Since va_args can't check type for us, we have to do this to make sure we get an address rather than an integer cast as an address... it's a bit of a hack for bit-position based flags since at some point we enter the realm of a valid setting being a valid address, but its the best I've been able to come up with so far...
+                    if (((size_t)expectedTableTag & ~LS_TTYPEMASK) == 0)
```
cmsj: @asmagill I think I'm fine with crashing to motivate the developer, that's always been my approach to `LS_TUSERDATA` :)

asmagill:
Reverted hack to check for missing strings in `checkArgs:` parameters.

cmsj:
lgtm, the only thing it needs is documenting the new type in Skin.h

asmagill:
I also want to add some tests to the suite (ideally, I'd like to chip away at this each time I submit something that has a test suite... we'll see how that goes!), so give me a day or two unless you're pushing for a release before then.

cmsj:
np

Issue #851, opened 2016-03-30
hs.httpserver request


asmagill:
How hard would it be to pass the html request headers to the callback function in addition to the method and path? Specifically, I'm thinking of the requesters IP address, so it can be checked against an access list, but the other headers could be useful as well.  I'm thinking maybe a table of key-value pairs as the third argument so existing callback functions still work as expected, they just ignore the third argument.

I don't know anything about CocoaHTTPServer, but can find some time to dig later this week; I just figured it might be a simple addition to someone already familiar with the Pod.

cmsj:
I think this can close, the related PR was merged alread.

PR #852, opened 2016-03-30
add requestheaders to hs.httpserver callback

asmagill:
Addresses #851.

Not quite as fully detailed as I would have liked, but then again, this is a minimalist web server... I can get what I need by checking the host key:

label | value
----|----
DNT | 1
Accept-Language | en-us
User-Agent | Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_4) AppleWebKit/601.5.17 (KHTML, like Gecko) Version/9.1 Safari/601.5.17
Host | localhost:7734
Connection | keep-alive
Accept-Encoding | gzip, deflate
Cache-Control | max-age=0
Accept | text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8

### DIFF
--- a/extensions/fs/internal.m
+++ b/extensions/fs/internal.m
@@ -910,6 +910,43 @@ static int hs_fileuti(lua_State *L) {
     return 1 ;
 }
 
+/// hs.fs.fileUTIalternate(fileUTI, type) -> string
+/// Function
+/// Returns the fileUTI's equivalent form in an alternate type specification format.
+///
+/// Parameters:
+///  * a string containing a file UTI, such as one returned by `hs.fs.fileUTI`.
+///  * a string specifying the alternate format for the UTI.  This string may be one of the following:
+//     * `extension`  - as a file extension, commonly used for platform independant file sharing when file metadata can't be guaranteed to be cross-platform compatible.  Generally considered unreliable when other file type identification methods are available.
+///    * `mime`       - as a mime-type, commonly used by Internet applications like web browsers and email applications.
+///    * `pasteboard` - as an NSPasteboard type (see `hs.pasteboard`).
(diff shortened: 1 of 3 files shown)


asmagill:
When the POST or PUT size exceeds the max size, it just serves a blank page... I want to figure out how to get it to put up a proper error page.  I also want to work out a set of functions for parsing the data for the most common POST encodings; once these are in place, any objections to adding this to core?

cmsj:
@asmagill no objections, lgtm

asmagill:
I'm working on a more "complete" web server module than just hs.httpserver provides (some CGI support, access list, etc.) and may have a couple more tweaks to httpserver to offer, so unless you're itching for a formal release soon, give me another couple of days...

cmsj:
I do need to do a release, but I'm fine with this feature missing it if you are?

asmagill:
I'm in no rush, so use your judgement.

FWIW, the "compliant" web server addition is very incomplete at the moment, and documentation is lacking -- working on that now -- but it can serve static pages off of a documentRoot.

If you want to take a look:  https://github.com/asmagill/hammerspoon_asm/tree/master/hsminweb... I go back and forth as to whether or not I think it should maybe eventually be added to core, since it's probably something I will be tweaking for quite a while and Hammerspoon will never be a replacement for a true dedicated web server...

The most likely remaining additions to `hs.httpserver` will possibly be some additional "server generated" request/response header fields (like "X-Client-Ip"), possibly a way to handle partial requests (byte ranges would allow an almost complete WebDAV implementation to be handled in Lua), and partial (or at least delayed) responses so that CGI isn't blocking.  Nothing that anyone has asked for yet, so this shouldn't be a release blocker.

PR #860, opened 2016-04-14
resize console input field for history items

heptal:
When cycling through history items, the input field would not expand, making it tedious to edit. This fixes that.

It now uses `setSelectedRange` instead of `moveToEndOfDocument` as any of the NSResponder methods will 'scroll' the content upwards out of view, leaving the final line of Lua code on the first line of an empty text box which had just been resized to fit it. Even moving left or right a single character via the `NSResponder` methods will cause this. Could be a limitation of `NSTextField` not implementing the necessary scroll methods, being intended for single-line use?

### DIFF
--- a/Hammerspoon/HSGrowingTextField.m
+++ b/Hammerspoon/HSGrowingTextField.m
@@ -50,7 +50,7 @@
         }
         
         if (intrinsicSize.height > 100) {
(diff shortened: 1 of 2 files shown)


cmsj:
thanks :)

PR #866, opened 2016-04-15
allow hs.grid to use user-supplied frame (per screen)

### LINKED ISSUE #792: Feature Request: Add an option to hs.grid for utilizing screen:fullFrame()
Currently, `hs.grid` hard-codes the use of `screen:frame()`; I'd love for that to be configurable so `screen:fullFrame()` could be used for its internal calculations instead.

My use case is that I auto-hide my dock and menu bar to maximize screen real estate...however, even with an auto-hiding dock, OS X reserves some pixels unnecessarily on whichever edge your dock is located (Googling reveals a lot historical complaints about this behavior). Always having that gap on the edge of the screen is not ideal, and utilizing `screen:fullFrame()` would fix the issue.

Let me know if you need any additional details, or a screenshot demonstrating the current behavior.

lowne:
Resolves #792, but it's more general. Someone on IRC (iirc, maybe it's an issue here somewhere) was asking to move the whole grid down, leaving a strip at the top for a custom bar; users can now pass an arbitrary frame to `hs.grid.setGrid()` (well, not fully arbitrary, it'll be `:intersect()`ed to the screen's `:fullFrame()` to keep the grid from wandering off)

### DIFF
--- a/extensions/grid/init.lua
+++ b/extensions/grid/init.lua
@@ -39,12 +39,13 @@ local setmetatable,rawget,rawset=setmetatable,rawget,rawset
 
 
 local gridSizes = {[true]=geom'3x3'} -- user-defined grid sizes for each screen or geometry, default ([true]) is 3x3
+local gridFrames= {} -- user-defined grid frames; always defaults to the screen:frame()
 local margins = geom'5x5'
 
 local grid = {setLogLevel=log.setLogLevel,getLogLevel=log.getLogLevel} -- module
 
 
---- hs.grid.setGrid(grid,screen) -> hs.grid
+--- hs.grid.setGrid(grid,screen,frame) -> hs.grid
 --- Function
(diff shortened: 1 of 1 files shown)


elasticdog:
I think there might be a bug here with the intersection code...using the suggested fix for my original issue in #792, an even larger buffer of pixels is put on the left where my hidden dock is rather than completely eliminating it. This buffer of pixels also appears now on the right side of the screen as well, which leads me to believe the calculation of the intersection is adding some space from the menu bar on the top to the horizontal sides.

I'll see if I can step through the changes to figure out where this happens. If you need some screenshots for a better explanation, let me know.

lowne:
@elasticdog it might also be a rounding issue, so yes, please provide screenshot, the exact code you're using - including the grid size -  and the rects for your screen's `:frame()` and `:fullFrame()`

PR #876, opened 2016-04-23
Hammerspoon Documentation Web Server

asmagill:
Still need to add `hs.httpserver.hsminweb`, but this gets the markdown support added.  Expect updates in the next few days.

@cmsj, this uses the C code portion of the Ruby gem `github-markdown`, and I think I've attributed that well enough in the docs, but do we have a common place or format for listing third party licenses?  (I checked, it's MIT, so I just need to know if I need to add it somewhere in addition to the source for `markdown.m`)


heptal:
I've actually been working on related stuff. I'm working on a code snippet gallery that fetches gists via the api and displays them using [marked](https://github.com/chjj/marked) and/or [highlight.js](https://github.com/isagalaev/highlight.js) with great success. 

I had also already been thinking about an internal doc display system just like this along similar lines.

The app bundle already has random scripts in the Resources folder including `gencomments`, `genhtml`, `librelease.sh`, etc (probably needs an issue). Perhaps it could also include (jsified) templates to convert docs.json to markdown, and use marked/highlight.hs to render them in a webview.

asmagill:
I tried various external markdown scripts for the web server, and they all introduced a very noticeable drop in speed in terms of rendering the web pages (I will admit to not trying node.js's stuff, but since it would still have been started via `os.execute` or `hs.execute`, I suspect it would have been similar)... the purpose here is to provide the same level of markdown support that we do for the Dash docset and do it fast and do it fully self contained -- no additional software on the users part will be necessary; if you have the Hammerspoon application (not source), you have what this documentation rendering solution needs.

In terms of rendering the documentation, that's pretty much done now with the second commit.  What's still lacking are a few things to bring the web server portion to where I want it in terms of standards compliance and adding a default hs.webview mini browser to the example in the hsdocs folder.

asmagill:
Speaking of the scripts in the application bundle... do we really need/want them in the released Application?

* gencomments, genjson, and gensql can be done in Hammerspoon itself, if needed, and they really aren't useful without the rest of the source code
* genhtml will be doable with this merge, once I finish it -- and in fact isn't even usable as is without the erb files
* librelease.sh, release.sh, and copy_extensions_to_bundle.sh aren't really needed either, once the App is made

With this merge, we could, in theory, incorporate third-party documentation at run time (something I've been wanting to add for a while), so the gen* scripts will likely end up in `hs.doc` in some form or another, if needed; but I can't come up with another reason these scripts are currently in the released bundle... am I missing something?

asmagill:
Not sure what's up with CocoaDev... but I have been merging back against the moving master, so maybe when I'm done, I should create a new merge and dump this one?  Still, since it will merge, I'll leave this until I'm finished in case people want to test it as I tweak it.

cmsj:
meh, don't worry about CodeCov if it's getting confused. If the Travis tests pass, it's all good.

cmsj:
@asmagill good point about the build/release scripts being in the app bundle. I've removed them :)

cmsj:
@asmagill as for adding attributions, Credits.rtf would be the place to do that :)

(This is looking great btw :)

asmagill:
Functionally ready, but I hope to add a sample browser with tie in to help function, and add some docs this weekend for lua template file usage this weekend.

Issue #880, opened 2016-04-24
Add LS_NSConversionOptions option for preserving nils?


cmsj:
@asmagill what do you think about adding a conversion option that preserves a Lua nil into a simple ObjC nil, instead of NSNull?

asmagill:
If you've a use case for it, go ahead... It's been awhile, but I think my reasoning for [NSNull null] being the default has to do with the fact that container classes (NSDictionary, NSArray, etc.) can't use nil as a key or value, but they can use [NSNull null]; however, when you know you're not working with such a container, nil can be easier to work with.

cmsj:
Take a look at hs.distributednotifications - there are a couple of points where I'm processing optional Lua args that could be a single toNSObject instead of a few lines of conditionals around lua_isnoneornil()

asmagill:
It wouldn't be too hard to add level detection in toNSObjectAtIndex: so that if it hasn't recursed yet (i.e. descended into a table), it returns a regular `nil` and if it has, then it returns `[NSNull null]`... I don't think that would break anything we currently have in place.  Maybe add options for *not* doing the detection... it would keep the code in the module simpler (you wouldn't have to remember to add an option when you're just getting your arguments, only when you know you want to do something "abnormal").

Which would you prefer?

cmsj:
Ooh, I like the idea of having it be magically done when we're not inside a collection type. I'm ever-so-slightly nervous about changing the behaviour, however, I think it's very unlikely that there are any cases where a non-collection conversion to NSNull is useful, so I think having it convert to nil by default, if it hasn't recursed yet, is fine. I'm not even sure it's necessary to have an option to override that, but I would not be opposed to adding it such an option.

cmsj:
@asmagill this can be closed now, right?

asmagill:
I'd forgotten about it and nothing has blown up, so I'm closing it.

Issue #891, opened 2016-04-30
Question about hs.task


asmagill:
@cmsj, in looking at `hs.task`, I noticed that the `selfRef` is assigned when a task is created (`new`), but never unreferenced *except* in the `__gc` function... this means that an `hs.task` object will *never* get collected, since it's entry in the registry prevents it being collected (except during a reload, of course).

Looking at the registry with `debug.getregistry()` confirms this... If you have `hs.tasks` which are repeated (like the "geeklets" that spawned `hs.styledtext`, etc.), you can find it's luaRef table with something like:

~~~lua
f = debug.getregistry()
a = {} ; for i,v in pairs(f) do if type(v) == "table" and #v > 0 then print(i, #v) ; table.insert(a, i) end end
~~~

then look for one with a semi large number (in my case, this time it was 1061), so...

~~~lua
> #f[1061]
182

-- the time it's taken me to write this passes --

> #f[1061]
350
~~~

Is there a reason you're keeping the selfRef?  If so, can we make sure it gets removed as soon as the task terminates (cleanly or otherwise)?  Or will this cause other problems?

cmsj:
I'm out on my phone, so I can't be sure, but that sounds like a bug on my part

chdiza:
I hope that the cause of the "fake GeekTool" memory leak has been found :)

A couple of weeks ago I left my office Mac running and awake all weekend, with HS running and my "geeklets" all running, but otherwise I didn't touch the machine at all (e.g., no ssh-ing in from afar, etc.).  When first fired up on Friday afternoon, HS consumed about 24MB of RAM.  On the ensuing Monday, HS was consuming 2.4 Gigabytes of RAM.

asmagill:
Inserting the following at line 158 of hs.task's internal.m file seems to be working for me:

~~~
userData->selfRef = [skin luaUnref:refTable ref:userData->selfRef];
~~~

I haven't made a pull or edited master yet because I don't use the features which use `streamCallback` or `inputCallback` myself so I don't know how to test whether or not they are affected... I don't think they should be, since this line is in the `terminationHandler`, but I leave the final determination of whether this is sufficient/correct to Chris.

chdiza:
@asmagill I inserted that line, and it also *seems* to have made a big difference.

[...]

PR #892, opened 2016-05-04
uielement and timer tweaks

asmagill:
* uielement now uses refTable like other modules instead of LUA_REGISTRYINDEX
* timer clears metatable on __gc and logs info that may help in debugging when it fails

I'm hoping these can help in tracking down some of the recurring crashes we're seeing that are related to timerCallback... or at least give a better picture of which timer so we can figure out where to focus...

@cmsj, please take a look and if you have no objections, I'd like to get this in master so those that use developer builds can start chiming in... 

And if you can think of more checks to add or stuff to log, please let me know... this is vexing me in its inconsistency!

EDIT: also modified toNSObjectAtIndex: so that it returns nil rather than [NSNull null], if the function has not yet recursed, per #880.

### DIFF
--- a/Hammerspoon/MJLua.m
+++ b/Hammerspoon/MJLua.m
@@ -283,6 +283,32 @@ static int checkForUpdates(lua_State *L) {
     return 0 ;
 }
 
+/// hs.canCheckForUpdates() -> boolean
+/// Function
+/// Returns a boolean indicating whether or not the Sparkle framework is available to check for Hammerspoon updates.
+///
+/// Parameters:
+///  * None
+///
+/// Returns:
+///  * a boolean indicating whether or not the Sparkle framework is available to check for Hammerspoon updates
+///
+/// Notes:
+///  * The Sparkle framework is included in all regular releases of Hammerspoon but not included if you are running a non-release or locally compiled version of Hammerspoon, so this function can be used as a simple test to determine whether or not you are running a formal release Hammerspoon or not.
+static int canCheckForUpdates(lua_State *L) {
+    LuaSkin *skin = [LuaSkin shared];
+    BOOL canUpdate = NO ;
+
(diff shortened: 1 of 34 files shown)


asmagill:
switched to logBreadcrumb so hopefully we get some info even during a crash

heptal:
Though not merged, I do have some tests for `hs.uielement` in #870. Hopefully we can eventually overhaul the uielement/application/window interdependencies, and perhaps incorporate your useful-looking axuielement code in a centralized way instead of being spread over several modules.

asmagill:
Note to module writers: we really need to be carful to clear the stack of *all* possible returned values, including error messages, when using `[LuaSkin protectedCallAndTraceback:(int)nargs nresults:(int)nresults]`

asmagill:
This is turning into a collection of minor bug fixes... @cmsj, if you approve the fix I mentioned in #891 for hs.task, I'll add that here as well... I just want a second opinion regarding whether or not it will affect the module's ability to stream output as it arrives or input as it is required.

asmagill:
Assuming Travis likes this (I don't know what drugs CodeCov is on), I would appreciate any feed back... it's a lot of little minor fixes, mainly to keep the stack clean around protectCallAndTraceback, since most of the time these are invoked by callbacks or timers, so the LuaState isn't known at the time and we really should be keeping things as clean as possible since we can't be positive that Lua will clean up after us like it does when invoking a C-function from the Lua interpreter itself.

I'll be checking out hs.webview as I get back to finishing the internal documentation updates, but any additional feedback would be appreciated, especially if anyone uses hs.task with streaming output or input.

I've also addressed #891 and #893.

asmagill:
Assuming Travis likes this and no one objects, I plan on landing this pull this weekend.

PR #896, opened 2016-05-10
add websocket capability to hs.httpserver

heptal:
Adds websocket capability to `hs.httpserver`
```lua
server = hs.httpserver.new();
server:setCallback(function() return "hello", 200, {} end);
server:setPort(8888);
server:websocket("/ws", function(msg) return msg.."!!!!!" end)
server:start();
```
Browser console:
```js
var ws = new WebSocket("ws://localhost:8888/ws") //if https: wss://localhost:8888/ws
ws.onmessage = (e => console.log(e.data))
ws.send('asdo')
> asdo!!!!!
```
Can also send messages directly:
`hs.timer.doEvery(1,function() server:send(tostring(math.random())) end)`
Browser console:
```
0.39438292663544
0.78309922339395
0.79844003310427
0.91164735751227
...
```

### DIFF
--- a/extensions/httpserver/internal.m
+++ b/extensions/httpserver/internal.m
@@ -3,6 +3,7 @@
 #import "CocoaHTTPServer/HTTPMessage.h"
 #import "CocoaHTTPServer/HTTPConnection.h"
 #import "CocoaHTTPServer/HTTPDataResponse.h"
+#import "CocoaHTTPServer/WebSocket.h"
 #import "CocoaAsyncSocket/GCDAsyncSocket.h"
 #import "CocoaLumberjack/CocoaLumberjack.h"
 #import "MYAnonymousIdentity.h"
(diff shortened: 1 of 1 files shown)


extensions/httpserver/internal.m
```diff
@@ -51,6 +59,48 @@ - (id)init {
...
+        [skin pushLuaRef:refTable ref:self.callback];
+        lua_pushstring(skin.L, [msg UTF8String]);
+
+        if (![skin protectedCallAndTraceback:1 nresults:1]) {
+            const char *errorMsg = lua_tostring(skin.L, -1);
+            [skin logError:[NSString stringWithFormat:@"hs.httpserver:websocket callback error: %s", errorMsg]];
+        } else {
+            response = [skin toNSObjectAtIndex:-1];
+        }
+    };
```
asmagill: You really should clean up the stack with `lua_pop(skin.L, 1)` here -- callbacks don't occur as a result of the lua interpreter invoking the function, so we can't count on Lua cleaning up for us.  A single instance before `};` will catch both your result or an error message from an error in the callback function.

asmagill:
For my edification, how do websockets differ from traditional http server requests?  Just curious... other than the line note re stack cleanup, lgtm.

extensions/httpserver/internal.m
heptal: Thanks. I've also got websocket clients in hs.http working, just needs to be cleaned up.

heptal:
Certain realtime APIs are only available over WebSocket, for example the [Slack API](https://api.slack.com/rtm). One can then react to or print activity in realtime.

heptal:
They are essentially full-duplex channels over TCP that, once negotiated, eliminate header overhead and the need for polling between client and server. HTTP is only used for the handshake although it generally shares the same ports (ws:// on 80 and wss:// on 443). Commonly used for things like chat, games, and streaming events such as tweets.

PR #920, opened 2016-06-17
[WIP] webview additions

asmagill:
* adds toolbar support (hs.webview.toolbar)
* adds 10.11 additions to WKWebView
  * private browsing
  * cache purging for webview (hs.webview.datastore)
  * better OS X version checking
  * remove asHSDrawing -- too easy to cause crashes; added relevant methods to webview directly
* Similar fix for hs.console wrt asHSDrawing

Toolbar documentation and methods will likely undergo some modifications yet -- it is now possible to create new items after the toolbar has been created, and the way groups are handled has been completely redone.

Probably some more fixes as well, hence the WIP.

### DIFF
--- a/extensions/webview/init.lua
+++ b/extensions/webview/init.lua
@@ -56,8 +56,9 @@ local USERDATA_TAG = "hs.webview"
 
 local osVersion = require"hs.host".operatingSystemVersion()
 if (osVersion["major"] == 10 and osVersion["minor"] < 10) then
-  hs.luaSkinLog.wf("%s is only available on OS X 10.10 or later", USERDATA_TAG)
-  return nil
+    hs.luaSkinLog.wf("%s is only available on OS X 10.10 or later", USERDATA_TAG)
+    -- nil gets interpreted as "nothing" and thus "true" by require...
+    return false
 end
 
 local module       = require(USERDATA_TAG .. ".internal")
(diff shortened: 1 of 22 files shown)


asmagill:
screen shot 2016-06-18 at 5 05 55 am

Still need to clean up the toolbar docs, add a few methods, add a few more items that 10.11 supports in WKWebView (custom useragent is the main one still missing), but it's coming along.

asmagill:
probably ready for inclusion; will give it another once over this weekend.

extensions/webview/init.lua
```diff
@@ -17,14 +17,45 @@
 ---
 --- This module provides support for injecting custom JavaScript user content into your webviews and for JavaScript to post messages back to Hammerspoon.
 
-local module       = require("hs.webview.internal")
-module.usercontent = require("hs.webview.usercontent") ;
+local USERDATA_TAG = "hs.webview"
+
+local osVersion = require"hs.host".operatingSystemVersion()
```
chdiza: Is there a missing space (or parenthesis) between 'require' and '"hs.host"' here?

cmsj: @chdiza no, it's one of Lua's (imo) more ugly pieces of syntax, but it's valid :)

chdiza: Ah, I see.  It's a bit out of place compared to other nearby occurences of "require".

asmagill: to me it scans better than including parenthesis or adding a space when you're really only going for one function or table from the module... and as no one's published style requirements for HS... but I'll change them if anyone cares enough

cmsj: fwiw, I don't care, we have never been good at consistency, and I don't think this is the PR to start :)

chdiza: I don't care enough either :)

I only pointed it out in case it was either a mistake or an unwanted style inconsistency.  Turns out, it's neither.

asmagill:
unless there are objections, I'm going to land this.

cmsj:
no objections :)

Issue #103, opened 2014-10-27
make help(hs.ipc) work


Habbie:
As compared to `help('hs.ipc')`

cmsj:
@asmagill I do like the `help.hs.foo` syntax we have, but I think it's non-obvious, where `help()` is something that people probably expect to work. At some point I will get around to either using @Habbie 's suggestion of autopopulating the magic variables, or just add it to all of the modules by hand.

cmsj:
I'm going to close this out now- realistically I don't think we're ever going to get around to the metadata that would be necessary to make this work. It would be very intrusive and cumbersome.

asmagill:
With command preprocessing with `hs._consoleInputPreparser` this has been possible for some time... see #921

PR #921, opened 2016-06-19
Console tweaks

asmagill:
* adds console input pre-parser function `hs. _consoleInputPreparser`, addresses #103 

Adding the following to your `init.lua` file

~~~lua
hs._consoleInputPreparser = function(s)
    local helpWhat = s:match("^help%s*%((.*)%)%s*$")
    if helpWhat then
        if helpWhat == "" then
            s = "help"
        else
            s = "help." .. helpWhat
        end
    end
    return s
end
~~~

will allow typing `help(hs.foo)` as an alternative to `help.hs.foo`.  The pre-parser is invoked on the raw input typed into the console input field and if it is not a function (or nil) or if the function generates an error, an error is logged to the console, and the input is passed on through `runstring` un-modified.

* errors that occur during Hammerspoon initialization (i.e. processing setup.lua) before reaching user `init.lua` now pop up alert

This primarily impacts module developers that are making adjustments to core modules which may be used by `setup.lua` before loading the users `init.lua` file (which is already wrapped with `xpcall`).  Instead of getting a non-responsive console with just the welcome message when, for example, `hs.doc` generates a run-time error, it causes an alert with the error text to be shown.

screen shot 2016-06-19 at 1 14 28 am

### DIFF
--- a/Hammerspoon/MJLua.m
+++ b/Hammerspoon/MJLua.m
@@ -475,12 +475,21 @@ void MJLuaInit(void) {
     lua_pushboolean(L, [[NSFileManager defaultManager] fileExistsAtPath: MJConfigFileFullPath()]);
     lua_pushboolean(L, [[NSUserDefaults standardUserDefaults] boolForKey:HSAutoLoadExtensions]);
 
-    lua_pcall(L, 7, 2, 0);
-
-    evalfn = [MJLuaState luaRef:refTable];
-    completionsForWordFn = [MJLuaState luaRef:refTable];
-    MJLuaLogDelegate = [[MJLuaLogger alloc] initWithLua:L] ;
-    if (MJLuaLogDelegate) [MJLuaState setDelegate:MJLuaLogDelegate] ;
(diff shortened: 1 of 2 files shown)


asmagill:
Updated `help` parser, if we add this to a FAQ or such:

~~~lua
-- make sure to chain these, if there might be multiple pre-parsers added
local previousParser = hs._consoleInputPreparser
hs._consoleInputPreparser = function(s)
    if previousParser then s = previousParser(s) end

    if s:match("^help") and not s:match("^help%.") then
        -- allow help(what) without quotes
        local helpParen = s:match("^help%s*%(([^\"]*)%)%s*$")
        if helpParen then
            if helpParen == "" then
                s = "help"
            else
                s = "help." .. helpParen
            end
        end

        -- allow `help what`
        if s:match("^help%s+[^\"]") then s = s:gsub("^help%s+", "help.") end
    end

    return s
end
~~~

The default supports:
* `help.hs`
* `help "hs"`
* `help("hs")`

This adds:
* `help hs`
* `help(hs)`

PR #942, opened 2016-07-23
Hints max title size

mwean:
This comes from #286. I just rebased it against master and removed the title substitution stuff because I couldn't get it working.


cmsj:
@mwean what's the `split()` function for? Am I being blind - it doesn't seem to be used?

mwean:
@cmsj I actually don't know... I started with #286 and removed the title substitution. It's very likely that it's not used.

mwean:
@cmsj ok, removed that function

PR #952, opened 2016-08-09
URL escape the file URLs we create ourselves. Fixes #756

### LINKED ISSUE #726: SCNetworkReachability
been playing with SCNetworkReachability to have a callback when a vpn network or even the internet itself becomes available/unavailable regardless of the interface it is on... should this go into `hs.host`, or are we approaching the point where we might have enough network related tidbits floating around to justify a separate module as `hs.network` or `hs.networking`?

### LINKED ISSUE #756: hs.urlevent problem with special characters
When trying to open a local file

```
Sun Jan 24 00:35:00 2016: "Jan 24 00:33:01 HOSTNAME Hammerspoon[82365]: ERROR: Unable to parse 'file:///Users/USER/Documents/foo/bar/Dog's Tail/Dog's Tail.HTML' as a URL"
```

junkblocker:


cmsj:
@junkblocker I removed the "Fixes", because I'm pretty sure this is unrelated to #726

junkblocker:
Thanks, @cmsj !

Issue #930, opened 2016-07-03
Feature Request: Indefinite Uptime for hs.alert


cwre:
I'd love to be able to have this up indefinitely. My usage is to notify I'm in a prefix mode. Currently I have it up for 100 seconds, but leaving it up until I have exited the prefix mode would be awesome.

Thanks!

asmagill:
By prefix mode, are you talking about modal hotkeys?  Under that assumption, your question intrigued me... we actually no longer need `hs.alert` to be an objective-c module at all, so I whipped up a potential lua/hs.drawing only version that I will be testing and possibly replacing the core one with down the road.

I've also added the ability to this test version to set an alert to display indefinitely and to remove it when requested.

If you want to test it as well, setting it up initially is a little tricky, since we want to make sure that the stock hs.alert is never loaded, so... do the following:

* Copy the file at https://github.com/asmagill/hammerspoon-config/blob/master/_scratch/alerts.lua to somewhere in your Hammerspoon configuration directory (I'll assume you save it as `~/.hammerspoon/newalerts.lua` for the rest of this writeup)
* add the following to the *top* of your Hammerspoon init.lua file before **ANY** use of the `hs.alert` module or possibly including any file which may include it (you can see mine, with a slightly different path of course, at https://github.com/asmagill/hammerspoon-config/blob/master/init.lua#L21):
~~~lua
local R, M = pcall(require,"newalerts")
if R then
   print()
   print("**** Replacing internal hs.alert with experimental module.")
   print()
   hs.alert = M
   package.loaded["hs.alert"] = M   -- make sure require("hs.alert") returns us
   package.loaded["hs/alert"] = M   -- make sure require("hs/alert") returns us
else
   print()
   print("**** Error with experimental hs.alert: "..tostring(M))
   print()
end
~~~
* Now, you use it normally, but if you save the return value for `hs.alert(..., true)` or `hs.alert.show(..., true)`, you can use that return value as an argument to `hs.alert.closeSpecific(...)` when you want to close it.  The `true` argument to the `hs.alert.show` command is optional -- you can still specify a duration, and using `hs.alert.closeSpecific` will close it early instead of waiting for the normal timeout.  You can see an example where I'm using it at https://github.com/asmagill/hammerspoon-config/blob/master/utils/_keys/applications.lua.

[...]

PR #964, opened 2016-08-25
add indefinite duration option to hs.alert

asmagill:
Also moves alert to an entirely lua based module using hs.drawing and hs.timer instead of obj-c code


asmagill:
address #930

asmagill:
closes #930

asmagill:
oops... didn't merge it :-P

PR #966, opened 2016-08-26
network additions

asmagill:
Some additions to the docs and hs.network module itself

addresses additional comments to #726

### DIFF
--- a/extensions/network/configurationinternal.m
+++ b/extensions/network/configurationinternal.m
@@ -1,7 +1,7 @@
-#import <Cocoa/Cocoa.h>
-#import <LuaSkin/LuaSkin.h>
-#import <SystemConfiguration/SystemConfiguration.h>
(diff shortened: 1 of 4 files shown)


cmsj:
lgtm. I might encourage you to use commit messages that trigger github to auto-close issues though :D

asmagill:
"commit messages"?  You mean mention the issue number in the final message where you then hit "Close and commit"?  Or something else I'm not familiar with?

I generally mention the issue number somewhere in the pull conversation, especially when it's a pull that may stay in the queue for a while, so that those most interested will know about it and can try it out if they want before it makes it into a formal commit, but if there is a better or more "correct" way, I'll switch to it.

I hadn't really paid that much attention as to when/where the issue had to be referenced to get an auto-close of issues, so clarify or point me to a link that does, and I'll start doing it.

cmsj:
the tl;dr is that if you put "Closes #foo" in your git commit message, when that gets merged, github will close the issue for you :)

Issue #960, opened 2016-08-23
hs.logger request


asmagill:
@lowne, since hs.logger was yours, got a question/request...

To keep labels consistent in the variety of things I work on, I often set a variable at the beginning of code I think I'm likely to reuse in other modules with the module's name and then use the variable instead of a hard coded name... In doing this with hs.logger, it's less obvious then I'd like because the label has its end truncated at 10 characters... I find myself doing this:

``` lua
local logLabel = (#USERDATA_TAG <= 10) and USERDATA_TAG
                                       or hs.utf8.codepointToUTF8(0x2026) .. USERDATA_TAG:sub(-7)
local log      = hs.logger.new(logLabel, "warning")
```

This pre-truncates the label from the beginning because most of my labels start with "hs._asm.", which is 8 characters right there, leaving me with only 2 to distinguish modules... my approach keeps the most likely different portion (the end) and truncates the most likely repetitive part... and because I like the ellipsis 1 character width, the Unicode character ends up taking up 3 spaces, rather than the 1 it visually does, so I have to use -7 rather than -9, though that's more of a nit.

So, I guess two requests/questions... would making hs.logger UTF8 aware (at least for strings that it truncates internally) break anything, and would changing the truncation to the beginning break anything (or would adding a flag indicating which truncation to do be better)?

I can take a look at this next week, if you're busy, but since it was your module to begin with, I figured you might have a better idea about making sure the requested changes don't break your intentions/expectations of the module.

lowne:
After quickly looking at the source (been a long time...):

- there's an undocumented `logger.idLength(nchars)` function ([here](https://github.com/Hammerspoon/hammerspoon/blob/master/extensions/logger/init.lua#L170)), which might be the easiest solution to your problem
- truncation currently is done via `string.format` [here](https://github.com/Hammerspoon/hammerspoon/blob/master/extensions/logger/init.lua#L158), so yea, it's raw bytes
- making it UTF-8 aware + fancy (and flaggable?) truncations doesn't seem trivial; if you have the time and inclination, maybe you could just redo whole thing in ObjC (IIRC that's been the repeatedly agreed-to conclusion wrt `logger` in various threads; moreover it seems likely that NSString/... could already have fancy truncation methods)

cmsj:
@asmagill looks like this was taken care of by #968?

asmagill:
yup

PR #968, opened 2016-08-26
hs.logger: add control over id display in logs

asmagill:
Addresses #960 

Adds hidden variables allowing control over display of logger id's:
* hs.logger.truncateID - a string which can be "head" or "tail", defaults to "tail".
* hs.logger.truncateIDWithEllipsis - a boolean indicating whether or not the truncation should be marked with "…"

Also uses utf8.len for length comparison when deciding whether or not to truncate id.

@lowne, I don't think I broke anything, but if you could give it a once over, I'd appreciate it.

E.g. before:
~~~
17:30:01     hotkey:         Created hotkey for S
17:30:01     hotkey:         Created hotkey for C
17:30:01     hotkey:         Created hotkey for ESCAPE
17:30:06 hs._asm.ko:     ASync Response A6B33690-3A93-485D-A56E-530B21A003CF Status:200 with result
17:30:06 hs._asm.ko:     http://kodi:••••••@latitude.local:8080 online
~~~

After with truncateID set to "head" and truncateIDWithEllipsis = true
~~~
17:28:43     hotkey:         Created hotkey for B
17:28:43     hotkey:         Created hotkey for I
17:28:43     hotkey:         Created hotkey for ⌘V
17:28:43     hotkey:         Created hotkey for ESCAPE
17:28:50 …_asm.kodi:     ASync Response 07D676B8-4D8D-4640-BAA2-27DF54184FFF Status:200 with result
17:28:50 …_asm.kodi:     http://kodi:••••••@latitude.local:8080 online
~~~

Should these new variables and idLength() already present be documented? I don't know how a general user would care about them, but... hidden features can be frustrating as well :-)

### DIFF
--- a/extensions/logger/init.lua
+++ b/extensions/logger/init.lua
@@ -189,7 +189,7 @@ local l = function(loglevel,lvl,id,...)
   if histSize>0 or loglevel>=lvl then return lf(loglevel,lvl,id,srep('%s',select('#',...),' '),...) end
 end
 
-logger.idlength=function(len)
+logger.idLength=function(len)
   if len==nil then return idlen end
   if type(len)~='number' or len<4 then error('len must be a number >=4',2)end
   len=min(len,40) idlen=len
--- a/extensions/logger/init.lua
+++ b/extensions/logger/init.lua
@@ -116,6 +116,26 @@ logger.history=function()
   return history
 end
 
+local formatID = function(theID)
+  if utf8.len(theID) > idlen then
+    if logger.truncateID == "head" then
+      theID = ssub(theID, -idlen)
+      if logger.truncateIDWithEllipsis then
+          theID = "…" .. ssub(theID, 2)
+      end
+    else
+      theID = ssub(theID, 1, idlen)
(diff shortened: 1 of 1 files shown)


cmsj:
@asmagill I was hoping to get 0.9.47 out tonight, but I'm running dangerously low on time at this point, so I think it would be better to push it to tomorrow. This is now the last item on the 0.9.47 milestone - how would you like to handle it absent feedback from @lowne (who is probably neck-deep in baby :D )

asmagill:
I am pretty confident that it doesn't break anything... I only modified code in the two places where the id label was being printed, so I'd say go ahead and add it.

However it is a really minor issue that only really affects developers, so waiting isn't a big deal either.

lowne:
Looks good to my 3am sleep-deprived eyes, fwiw. 
I should also mention that we're moving to another country, so I've been/will be necessarily mostly afk for a while.

cmsj:
@Lowne wow! Hope it goes well :)

PR #981, opened 2016-09-02
Fixed typo in hs.notify.register

junkblocker:


junkblocker:
Note that without this change, the register mechanism is broken. The response on button or label clicking fails.

Issue #976, opened 2016-08-30
10.12 - Console text input field shrinks when pressing up/down


asmagill:
Console text input field shrinks when pressing up/down arrows to scroll through recent command history.  Also tried with stock Hammerspoon and empty init, so it's not something I've added with my code to auto-save/track command history across reloads.

Tapping any "displayable" key (e.g. space) causes input field to grow back to regular size (with proper adjustments based upon number of lines in input)

PR #982, opened 2016-09-06
Console input field resizes properly in 10.12

asmagill:
Actually, two minor things in this pull:
* console input field resizes properly in 10.12 - issue #976
* added `hs.console.titleVisibility` so `hs.webview.toolbar` can mimic Safari like toolbar in the titlebar area when a toolbar is added to the console (a future pull will include this and additional tweaks to `hs.webview`).

I need someone running 10.11 to verify that the first doesn't change the console input field behavior in 10.11 -- my 10.11 machine ate itself over the weekend and I haven't had a chance to rebuild it yet.


cmsj:
@asmagill lgtm on 10.11 :)

Issue #979, opened 2016-08-30
Can't exit modal in 0.9.47


magnus-staberg:
The code below works in 0.9.46 but not 0.9.47:
```
modal = hs.hotkey.modal.new({'ctrl'}, 'M')
modal:bind({}, 'escape', function() modal:exit() end)
```
It gives
`hotkey: The hotkey was deleted, cannot enable it`
in console. The invocation of `modal:exit` is the culprit.

Am I missing something or is it a bug? 
Couldn't find anything in the release notes.

cro:
I'm seeing this behavior also.

asmagill:
My experience with this is that while the error message appears, it does clear the modal keys and execute `modal:exited()` (if it exists) properly... does this match your experience or are you seeing something not being executed properly or keys not unset?  I ask because it may affect what I should be looking for...

magnus-staberg:
I don't know if it executes `modal:exited()` but it doesn't exit properly.

asmagill:
Are you able to build a custom build of Hammerspoon and test the fix I posted in Pull #988?

Actually, it might be easier to just apply the change yourself, since it's just one line to the `hs.hotkey` init.lua file...  Try the following:

Locate the Hammerspoon application icon in your Finder... it's probably in "Applications" unless you've moved it somewhere else.  Right click on the icon and select "Show Package Contents".  Double click on the "Contents" folder, the "Resources" folder, the "extensions" folder, the "hs" folder, then the "hotkey" folder (they should show up in that order).  Select the "init.lua" file and open it up in an editor... double clicking on it might open one, or you can right click on it and select "Open With" and choose an editor... TextEdit works fine for this if you don't have anything else.

Look for the following code at around line 51... 
~~~lua
local function enable(self,force,isModal)
  -- force is there just in case it'll be needed down the line; not exposed as public API for now
  -- isModal is only used to differentiate loglevel (modals can cause a lot of shadowing/unshadowing)

  -- this ensures "nested shadowing" behaviour; i.e. can't re-enable a hotkey that is currently shadowed (unless force),
  -- must wait for the current top dog to get disabled
  if not force and self.enabled then log.v('Hotkey already enabled') return self end
  local idx = self.idx
  if not idx or not hotkeys[idx] then log.e('The hotkey was deleted, cannot enable it') return end
  for i=#hotkeys[idx],1,-1 do
    local hk=hotkeys[idx][i]
[...]

PR #988, opened 2016-09-13
fix hotkey.bind return value

asmagill:
Since hotkey now wraps everything in a table rather then use the raw userdata value, the bind wrapper should do the same rather than return the raw userdata.

Since hotkey has a lot of stuff added by @lowne, he should review to make sure that this doesn't break anything.

Should address #979 (it seems to on my system, but more testing is required and a clarification of the actual error behavior as requested in 979 already)


lowne:
lgtm, after a bit of digging around I found that the change was introduced in 477770c4 (by @asmagill  😜) but I'm not sure why (if it really has something to do with logging on the c side, it's beyond my abilities), so no objections to reverting it.

PR #998, opened 2016-09-23
Fix two hs.screen GC bugs - screen_gc wasn't actually releasing hs.screen userdata objects, screens_gc was expecting userdata even though it's a module level function

cmsj:
@asmagill I'd appreciate an eye on this. I'm pretty sure the `screens_gc` one is right, that should never have been looking for `LS_TUSERDATA` and the error is only hidden because Lua is in the process of being destroyed, but if you hook a breakpoint into `checkArgs` without this patch, you'll see it's expecting `LS_TUSERDATA`, but being given `LUA_TABLE` because `screens_gc` is the module-level GC, not the screen object GC, so our `checkArgs` call seems pointless and wrong (and probably my fault!)

The second one I'm less sure about - it looks like in `screen_gc` we should do `__bridge_transfer` because `new_screen()` is doing a `__bridge_retained`.

### DIFF
--- a/extensions/screen/internal.m
+++ b/extensions/screen/internal.m
@@ -684,7 +684,7 @@ void screen_gammaReapply(CGDirectDisplayID display) {
 }
 
 static int screen_gc(lua_State* L) {
-    NSScreen* screen __unused = get_screen_arg(L, 1);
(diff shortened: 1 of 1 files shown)


cmsj:
ping @asmagill :D

asmagill:
lgtm

cmsj:
thanks :)

PR #1024, opened 2016-10-04
make sure pasteboardItems isn't cleared between subsequent message calls

asmagill:
hs.pasteboard.contentTypes was invoking pasteboardItems message repeatedly rather than maintaining a strong reference, which allowed for the slight possibility of it being changed between object message sends.


asmagill:
@cmsj, this resulted in a stalled build... is there something I can change on my end to prevent/retry this, or do I just need to make a cosmetic tweak and update the request to get it to try again?

cmsj:
@asmagill in theory you should just be able to visit the build on travis's site and click the little refresh button somewhere over on the right and it'll re-run the build.

Issue #1029, opened 2016-10-06
Constant crashes on Sierra+0.9.48


sprig:
Hello,

I saw the issue #1023 but am opening a report since there the problem was apparently was related to the pasteboard and here it doesn't look like it from the report, nor from the symptoms.

[Almost?] Every time I leave my computer for a while and come back, hammerspoon crashes. 

I'm not certain whether the computer asleep and when exactly the crashes occur, but I suspect that it does sleep and that they occur as it wakes up. The time on this crash report supports this hypothesis. It could also be when I first try to use it after waking up the computer, since I notice this when I try to use a keybinding and see that it isn't working. 

I upgraded to Sierra and to 0.9.48 almost at the same time, so am not sure which of them started this, but it didn't occur on El Capitan+0.9.47. Finally, I have a different computer with Sierra and 0.9.48 and it does not happen there.

Let me know if I can provide any additional info. The report:

```
contact@surteesstudios.comProcess:               Hammerspoon [47164]
Path:                  /Applications/Hammerspoon.app/Contents/MacOS/Hammerspoon
Identifier:            org.hammerspoon.Hammerspoon
Version:               0.9.48 (0.9.48)
Code Type:             X86-64 (Native)
Parent Process:        ??? [1]
Responsible:           Hammerspoon [47164]
User ID:               501

Date/Time:             2016-10-06 12:48:37.603 +0300
OS Version:            Mac OS X 10.12 (16A323)
Report Version:        12
Anonymous UUID:       DOESNT-MATTER


Time Awake Since Boot: 170000 seconds

System Integrity Protection: disabled

Crashed Thread:        0  Dispatch queue: com.apple.main-thread

Exception Type:        EXC_BAD_INSTRUCTION (SIGILL)
Exception Codes:       0x0000000000000001, 0x0000000000000000
Exception Note:        EXC_CORPSE_NOTIFY

Termination Signal:    Illegal instruction: 4
Termination Reason:    Namespace SIGNAL, Code 0x4
Terminating Process:   exc handler [0]

Application Specific Information:
Crashing on exception: -[__NSCFType isEqualToString:]: unrecognized selector sent to instance 0x6080008490f0

Application Specific Backtrace 1:
0   CoreFoundation                      0x00007fff9f8e552b __exceptionPreprocess + 171
1   libobjc.A.dylib                     0x00007fffb3fbdcad objc_exception_throw + 48
[...]

PR #1031, opened 2016-10-07
detect bad role type in is_window

asmagill:
Should address #1029

### DIFF
--- a/extensions/uielement/internal.m
+++ b/extensions/uielement/internal.m
@@ -1,6 +1,6 @@
-#import <Cocoa/Cocoa.h>
-#import <Carbon/Carbon.h>
-#import <LuaSkin/LuaSkin.h>
+@import Cocoa ;
+@import Carbon ;
+@import LuaSkin ;
 #import "uielement.h"
 #import "../window/window.h"
 #import "../application/application.h"
@@ -10,7 +10,7 @@
 static const char* userdataTag = "hs.uielement";
 static const char* watcherUserdataTag = "hs.uielement.watcher.userdata";
 static const char* watcherTag = "hs.uielement.watcher";
-NSArray *eventNames;
+static NSArray *eventNames;
 
 static void new_uielement(lua_State* L, AXUIElementRef element) {
     LuaSkin *skin = [LuaSkin shared];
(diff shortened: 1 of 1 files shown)


extensions/uielement/internal.m
```diff
@@ -42,15 +42,20 @@ static bool is_window(AXUIElementRef element, NSString* role) {
...
+        return YES;
+      } else {
+        return NO;
+      }
   } else {
-    return NO;
+      [LuaSkin logBreadcrumb:[NSString stringWithFormat:@"%s:is_window AXRole is not a string type", watcherTag]] ;
+      return NO;
   }
 }
```
sprig: I think that you also need to amend the push_element function:

Changing
``` lua
// Use the Role of the element to decide which type of object to create: window, app, or plain uielement.
// Retains a copy of the element, if necessary.
static void push_element(lua_State* L, AXUIElementRef element) {
    NSString* role = get_prop(element, NSAccessibilityRoleAttribute, @"");

    if (is_window(element, role)) {
        new_window(L, (AXUIElementRef)CFRetain(element));
    } else if ([role isEqualToString: (NSString*)kAXApplicationRole]) {
        pid_t pid;
        AXUIElementGetPid(element, &pid);
        if (!new_application(L, pid)) {
            lua_pushnil(L);
        }
    } else {
        new_uielement(L, (AXUIElementRef)CFRetain(element));
    }
}
```
to
``` lua
// Use the Role of the element to decide which type of object to create: window, app, or plain uielement.
// Retains a copy of the element, if necessary.
static void push_element(lua_State* L, AXUIElementRef element) {
    NSString* role = get_prop(element, NSAccessibilityRoleAttribute, @"");

    if (is_window(element, role)) {
        new_window(L, (AXUIElementRef)CFRetain(element));
    } else if ([role isKindOfClass:[NSString class]] && [role isEqualToString: (NSString*)kAXApplicationRole]) {
        pid_t pid;
        AXUIElementGetPid(element, &pid);
        if (!new_application(L, pid)) {
            lua_pushnil(L);
        }
    } else {
        new_uielement(L, (AXUIElementRef)CFRetain(element));
    }
}
```
(single line change)

asmagill:
That's probably pretty close to what I will add, though I want to read up a little more and see if we can add some logging to identify the specific applications/windows with this problem... it may be that we should be filtering them another way or that it might be worth submitting a bug report to them about not using Accessibility correctly -- but first we have to identify them!

Keep me posted on how it works for you.

asmagill:
@sprig, this has additional logging that attempts to identify the role type and the application it's from... I'd appreciate it if you could test with this and, if it seems to fix the crashes, execute the following in the Hammerspoon console every so often and report what it displays: `hs.logger.printHistory(nil,nil,"uielement")`.

If it does continue to crash, your crash logs have been useful so far, but you might also check the system log for lines with "hs.uielement" in the message portion and copy these into here as well... by using the logVerbose method, it should put the log messages in both the system log *and* the Hammerspoon console log.

sprig:
@asmagill Thanks, will do

sprig:
Quick update:

Newly patched HS has been running since the last comment. No crashes so far.

``` lua
> hs.logger.printHistory(nil,nil,"uielement")


> 
```
(Nothing to report)

sprig:
Still no crashes and nothing from the logger

Issue #1022, opened 2016-10-04
hs.keycodes.methods() should include com.apple.keylayout.US


xream:
With tools like [hnakamur/inputsource](https://github.com/hnakamur/inputsource), we can get `com.apple.keylayout.US`, but `hs.keycodes.methods()` does not.

asmagill:
It's been a long time since I've looked closely at the `hs.keycodes` source, but have you looked at `hs.keycodes.layouts` and `hs.keycodes.currentLayout`?

~~~lua
> hs.inspect(hs.keycodes.layouts())
{ "U.S.", "Dvorak" }

> inspect(hs.keycodes.currentLayout())
"U.S."
~~~

I'm not entirely sure what the `hs.keycodes.methods` function is supposed to return (I've never used it and it returns an empty table on my machine), but it's querying property `kTISTypeKeyboardInputMode` while the code you refer to is querying for `kTISPropertyInputSourceID`, which appears to be a more formal version of what `hs.keycodes.layouts` and `hs.keycoes.currentLayout` return (property `kTISTypeKeyboardLayout`).

xream:
I have two input methods. `com.apple.keylayout.US` for English, `com.apple.inputmethod.SCIM.ITABC` for Chinese.
```
> hs.inspect(hs.keycodes.layouts())
{ "U.S." }
> inspect(hs.keycodes.currentLayout())
"U.S."
```
US is missing:
```
> inspect(hs.keycodes.methods())
{ "Pinyin - Simplified" }
```
Why this is nil?
> inspect(hs.keycodes.currentMethod())
nil
```

xream:
If you try any other language, like `com.apple.keylayout.Spanish`. `hs.keycodes.currentMethod()` is always nil. 
Maybe it's because my system local is Chinese? Only Chinese input methods are included in `hs.keycodes.methods()`.

asmagill:
Interesting... apparently some of the input methods only set certain properties and not others... Ok, I'll take a closer look tonight at the Text and Input Services API and see about adding support for some of the other properties available.

Yeah, I suspect some of it may be locale based.

asmagill:
Make that script, not locale, based.

asmagill:
[...]

PR #1033, opened 2016-10-08
hs.keycodes additions

asmagill:
Prompted by #1022 

* Adds `hs.keycodes.currentSourceID([sourceID])` to get/set keyboard input by layout or method source id
* add optional boolean `sourceID` parameter to `hs.keycodes.layouts` and `hs.keycodes.methods` to return names as source ids instead of localized string (default behavior is same as before)
* added some notes to clarify between layouts and methods in doc strings... I'm not 100% clear about it myself, so if anyone has better verbiage, then go for it.


asmagill:
addresses #1034 and Fabric report 422

PR #1038, opened 2016-10-11
Support colour images as menubar icons.

hikinggrass:
Currently menubar icons can only be specified as templates.
This makes sense if you want to support Dark Mode out of the box, as it forces you to use appropriate icons. But there are some use-cases where the usage of colour may be useful to the user. Examples for this include the Keyboard Layout Flag, the Network Activity indicator from Little Snitch or the CPU Usage indicator from Menu Meters.
This commit introduces an optional parameter to the setIcon function that defaults to the current behaviour.


cmsj:
thanks!

Issue #1061, opened 2016-10-27
Feature request: Style for hs.alert


junkblocker:
The current style for the alert is pretty good but when looking at a different computer through screen sharing I realized I couldn't tell which machine the alert was supposed to be for. For such cases and for choice in general it would be good if we could add configurable colors, font, and maybe even border radius style facilities as are available in default macOS notifications. Since @asmagill very nicely converted alerts to lua code it can be tackled by most anybody now :) We'd need to figure out how we want the API to look like before someone tackles that.

latenitefilms:
Great idea!

asmagill:
As suggested above, this is easy enough to add... what types of components do we want to make "settable"?  Background color is an obvious choice... font and font size as well?  I also think that the solution should allow for a module level default as well as specific alert overrides -- the idea of being able to make specific alerts red versus the default, even on the same machine, appeals to me.

My personal feeling (but I'm open to discussion) is that anything beyond that (positioning decision, frame shape, etc.) gets into enough situation specifics that the user is better off overriding the alert module themselves or writing their own alert code for their specific need... since it's only lua code, this is easy for anyone to do with a few pointers that we can either put into the documentation or add to a wiki page.

junkblocker:
@asmagill , agree completely. Color, font, background color are good enough for me and I wouldn't want to take layout calculations. For the rest we should make it extendable in lua code by user themselves.

latenitefilms:
One thing that could be handy though is defining which screen the alert appears on?

asmagill:
Have to think a little on that... I think we could probably easily calculate it based on the origin of `hs.screen.mainScreen`, but it will be later tonight before I can try this out.

asmagill:
closed with #1064

PR #1064, opened 2016-11-01
add minimal style attrbutes to hs.alert

asmagill:
Adds `hs.alert.defaultStyle` and an optional `style` argument to `hs.alert.show`

~~~
Variable: hs.alert.defaultStyle[]

A table defining the default visual style for the alerts generated by this module.

The following may be specified in this table (any other key is ignored):
 * fillColor   - a table as defined by the `hs.drawing.color` module to specify the background color for the alert, defaults to { white = 0, alpha = 0.75 }.
 * strokeColor - a table as defined by the `hs.drawing.color` module to specify the outline color for the alert, defaults to { white = 1, alpha = 1 }.
 * strokeWidth - a number specifying the width of the outline for the alert, defaults to 2
 * textColor   - a table as defined by the `hs.drawing.color` module to specify the message text color for the alert, defaults to { white = 1, alpha = 1 }.
 * textFont    - a string specifying the font to be used for the alert text, defaults to ".AppleSystemUIFont" which is a symbolic name representing the systems default user interface font.
 * textSize    - a number specifying the font size to be used for the alert text, defaults to 27.

If you modify these values directly, it will affect all future alerts generated by this module.  To adjust one of these properties for a single alert, use the optional `style` argument to the `hs.alert.show` function.
~~~

Per #1061


asmagill:
~~~
hs.alert.show(str, [style], [screen], [seconds]) -> uuid
Function
Shows a message in large words briefly in the middle of the screen; does tostring() on its argument for convenience.

NOTE: For convenience, you can call this function as `hs.alert(...)`

Parameters:
 * str     - The string to display in the alert
 * style   - an optional table containing one or more of the keys specified in [hs.alert.defaultStyle](#defaultStyle)
 * screen  - an optional `hs.screen` userdata object specifying the screen (monitor) to display the alert on.  Defaults to `hs.screen.mainScreen()` which corresponds to the screen with the currently focused window.
 * seconds - The number of seconds to display the alert. Defaults to 2.  If seconds is specified and is not a number, displays the alert indefinately.

Returns:
 * a string identifier for the alert.

Notes:
 * If you specify a non-number value for `seconds` you will need to store the string identifier returned by this function so that you can close it manually with `hs.alert.closeSpecific` when the alert should be removed.
 * Any style element which is not specified in the `style` argument table will use the value currently defined in the [hs.alert.defaultStyle](#defaultStyle) table.
~~~

Hold up... reading through this as I post it with the new additions I realize that I think I broke persistent alerts... will followup with a fix...

asmagill:
Ok, new argument parsing process as follows:

~~~
Notes:
 * The optional parameters are parsed in the order presented as follows:
   * if the argument is a table and `style` has not previously been set, then the table is assigned to `style`
   * if the argument is a userdata and `screen` has not previously been set, then the userdata is assigned to `screen`
   * if `duration` has not been set, then it is assigned the value of the argument
   * if all of these conditions fail for a given argument, then an error is returned
 * The reason for this logic is to support the creation of persistent alerts as was previously handled by the module: If you specify a non-number value for `seconds` you will need to store the string identifier returned by this function so that you can close it manually with `hs.alert.closeSpecific` when the alert should be removed.
 * Any style element which is not specified in the `style` argument table will use the value currently defined in the [hs.alert.defaultStyle](#defaultStyle) table.
~~~

extensions/alert/init.lua
```diff
@@ -11,6 +11,28 @@ local uuid    = require"hs.host".uuid
...
+---  * fillColor   - a table as defined by the `hs.drawing.color` module to specify the background color for the alert, defaults to { white = 0, alpha = 0.75 }.
+---  * strokeColor - a table as defined by the `hs.drawing.color` module to specify the outline color for the alert, defaults to { white = 1, alpha = 1 }.
+---  * strokeWidth - a number specifying the width of the outline for the alert, defaults to 2
+---  * textColor   - a table as defined by the `hs.drawing.color` module to specify the message text color for the alert, defaults to { white = 1, alpha = 1 }.
+---  * textFont    - a string specifying the font to be used for the alert text, defaults to ".AppleSystemUIFont" which is a symbolic name representing the systems default user interface font.
+---  * textSize    - a number specifying the font size to be used for the alert text, defaults to 27.
+---
+--- If you modify these values directly, it will affect all future alerts generated by this module.  To adjust one of these properties for a single alert, use the optional `style` argument to the [hs.alert.show](#show) function.
+module.defaultStyle = {
+    strokWidth  = 2,
```
cmsj: s/strok/stroke/ I think?

asmagill:
Yeah, that does look and work better, doesn't it? :-P

PR #1065, opened 2016-11-01
add :locations() to hs.network.configuration

Habbie:
I am uncertain about object lifetime management / memory leaks here. Beyond that, this works - tested on 10.10.

### DIFF
--- a/extensions/network/configurationinternal.m
+++ b/extensions/network/configurationinternal.m
@@ -328,6 +328,38 @@ static int dynamicStoreLocation(lua_State *L) {
     return 1 ;
 }
 
+/// hs.network.configuration:locations() -> table
+/// Method
+/// Returns all configured locations
+///
+/// Parameters:
+///  * None
+///
+/// Returns:
+///  * a table of key-value pairs mapping location UUIDs to their names
+///
+static int dynamicStoreLocations(lua_State *L) {
+    LuaSkin *skin = [LuaSkin shared] ;
+    [skin checkArgs:LS_TUSERDATA, USERDATA_TAG, LS_TBREAK] ;
+    SCPreferencesRef prefs = SCPreferencesCreate(NULL, CFSTR("Hammerspoon"), NULL);
+
+    CFArrayRef locations = SCNetworkSetCopyAll(prefs);
+    CFIndex i, c = CFArrayGetCount(locations);
+    NSMutableDictionary *dict = [[NSMutableDictionary alloc] init];
+
+    for (i=0; i<c; i++) {
+        SCNetworkSetRef location = CFArrayGetValueAtIndex(locations, i);
+        CFStringRef setID = SCNetworkSetGetSetID(location);
+        CFStringRef name = SCNetworkSetGetName(location);
+        dict[(__bridge NSString*) setID] = (__bridge NSString *)(name);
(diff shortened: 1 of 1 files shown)


extensions/network/configurationinternal.m
```diff
@@ -328,6 +328,38 @@ static int dynamicStoreLocation(lua_State *L) {
...
+/// Parameters:
+///  * None
+///
+/// Returns:
+///  * a table of key-value pairs mapping location UUIDs to their names
+///
+static int dynamicStoreLocations(lua_State *L) {
+    LuaSkin *skin = [LuaSkin shared] ;
+    [skin checkArgs:LS_TUSERDATA, USERDATA_TAG, LS_TBREAK] ;
+    SCPreferencesRef prefs = SCPreferencesCreate(NULL, CFSTR("Hammerspoon"), NULL);
```
asmagill: while it isn't *supposed* to happen, we've seen a few cases where core foundation objects which are never supposed to be NULL turn out to be NULL anyways, thus causing a crash... I would recommend wrapping from here to just before the return with `if (prefs) { ... } else { lua_pushnil(L) ; }

extensions/network/configurationinternal.m
```diff
@@ -328,6 +328,38 @@ static int dynamicStoreLocation(lua_State *L) {
...
+
+    CFArrayRef locations = SCNetworkSetCopyAll(prefs);
```
asmagill: likewise with this: `if (locations) { ... }` and move the `CFRelease(locations)` inside the then block.

asmagill:
Other than the recommendations made in the line comments, lgtm... we have to be extra defensive with core foundation objects since NULL will cause a crash while the fact that nil can receive messages allows us to be a little laxer with Objective-C objects.

Habbie:
Like this?

extensions/network/configurationinternal.m
```diff
@@ -328,6 +328,42 @@ static int dynamicStoreLocation(lua_State *L) {
...
+///
+/// Returns:
+///  * a table of key-value pairs mapping location UUIDs to their names
+///
+static int dynamicStoreLocations(lua_State *L) {
+    LuaSkin *skin = [LuaSkin shared] ;
+    [skin checkArgs:LS_TUSERDATA, USERDATA_TAG, LS_TBREAK] ;
+    SCPreferencesRef prefs = SCPreferencesCreate(NULL, CFSTR("Hammerspoon"), NULL);
+
+    if(!prefs) { lua_pushnil(L); CFRelease(prefs); return 1; }
```
cmsj: You don't need to CFRelease() here because `SCPreferencesCreate()` returned NULL.

extensions/network/configurationinternal.m
```diff
@@ -328,6 +328,42 @@ static int dynamicStoreLocation(lua_State *L) {
...
+
+    CFArrayRef locations = SCNetworkSetCopyAll(prefs);
+    if(!locations) { lua_pushnil(L); CFRelease(prefs); CFRelease(locations); return 1; }
```
cmsj: ditto here for `CFRelease(locations)`

Habbie:
Duh. Pushed!

PR #1066, opened 2016-11-01
nit

Habbie:


Habbie:
I've noticed lots of missing periods since - but those are at the ends of paragraphs so less noticable. Would be happy to do a run on those if you like, @cmsj.

asmagill:
+1 for doing a full review and cleanup of the documentation... I make adjustments as I see them when working on various modules, but we could definitely benefit from a formal review.

cmsj:
We're very inconsistent with this. I think I prefer. It having periods at the end of these things, but I'd take consistent use of one form over choosing which form I prefer :D

cmsj:
I meant to type "prefer not having periods"

asmagill:
Since this particular example is with to the LuaSkin docs, it reminded of something I've been meaning to ask... does anyone know of a tool to convert the headerdoc format we use for LuaSkin into a docset which can be loaded with Dash?  Or at the very least converted to json so we can include it in the built-in documentation?

cmsj:
@asmagill I don't know of such a tool, but it sure would be nice. Getting the docs to render at all was a total nightmare :)

PR #1067, opened 2016-11-01
raw data read/write to pasteboard

asmagill:
adds:
~~~
hs.pasteboard.readDataForUTI([name], uti) -> string
hs.pasteboard.readPListForUTI([name], uti) -> string
hs.pasteboard.writeDataForUTI([name], uti, data) -> string
~~~

Writing a property list is on hold until we can better understand the exact format expected; reading a property list is included for information purposes only at this time.  The docstrings reflect this.


cmsj:
lgtm

PR #1069, opened 2016-11-03
add :setLocation()

Habbie:


extensions/network/configurationinternal.m
```diff
@@ -300,6 +300,97 @@ static int dynamicStoreLocalHostName(lua_State *L) {
...
+///  * location - string containing UUID or name of new location
+///
+/// Returns:
+///  * bool
+static int dynamicStoreSetLocation(lua_State *L) {
+    LuaSkin *skin = [LuaSkin shared] ;
+    [skin checkArgs:LS_TUSERDATA, USERDATA_TAG, LS_TSTRING, LS_TBREAK];
+    
+    NSString *target ;
+    if (lua_gettop(L) == 2) {
```
Habbie: copied this `if` from other code.. do I really need it after `checkArgs`?

extensions/network/configurationinternal.m
```diff
@@ -300,6 +300,97 @@ static int dynamicStoreLocalHostName(lua_State *L) {
...
+    }
+    AuthorizationRef authorization = NULL;
+    AuthorizationFlags flags = kAuthorizationFlagDefaults;
+    OSStatus status = AuthorizationCreate(NULL,
+                                          kAuthorizationEmptyEnvironment,
+                                          flags,
+                                          &authorization);
+
+    if (status != errAuthorizationSuccess) {
+        lua_pushboolean(L, 0);
```
Habbie: The 'false' returns in here are pretty quiet, should I be throwing instead?

asmagill: While I will be the first to admit that I haven't been consistent on this either, it kind of depends... my current thinking is that if a failure to authorize is a valid state that the user could get with this function/method, then it should return a value of false or nil (depending upon whichever makes the most sense in context); if it indicates a problem that means absolutely nothing can go forward, then code execution should halt right then and there so that the user knows where they need to fix or adjust something for this to be useful.

However, others may argue that it's up to the calling code to check for `false` and throw an error or do something else if they want to... and a counter to that is that if the calling code truly wants to handle an error condition itself, then it can always wrap the function/method that throws an error in `pcall`, so... it comes down to developer preference since we don't have a style guide yet.

extensions/network/configurationinternal.m
asmagill: Unless you also mark the `LS_TSTRING` as optional (i.e. `LS_TSTRING | LS_TOPTIONAL`), checking the stack top with `lua_gettop(L)` isn't necessary -- it will always be 2 if it's made it that far.  So... if `nil` is a valid value for `target`, then add `| LS_TOPTIONAL` to the checkArgs method for the LS_TSTRING argument; if it isn't, then line 328 (and the corresponding close brace) isn't necessary but isn't going to crash either...

asmagill:
@habbie, do you want to make changes before this is merged?

cmsj:
Per IRC discussion, bumping this for 0.9.50

Habbie:
@asmagill I don't need to make changes before merge but if I find the time before @cmsj pulls the trigger I might remove the useless gettop. I discussed error handling with @cmsj and the consensus is "we don't know and we don't really have consistency today so this is fine". If people come complaining that this feature does not work we can then find out the reasons for failure - I haven't seen any other than "the location does not exist" which is obvious user error.

PR #1071, opened 2016-11-06
add hs.sharing module

asmagill:
Allow Hammerspoon to create Sharing Service requests with Hammerspoon data items.

e.g.

~~~lua
mailer = hs.sharing.newShare("com.apple.share.Mail.compose")
mailer:subject("Screenshot generated at " .. os.date()):recipients({ "user@address.com" })
mailer:shareItems({ [[
    Add any notes that you wish to add describing the screenshot here and click the Send icon when you are ready to send this

]], hs.screen.mainScreen():snapshot() })
~~~


cmsj:
Very cool! Haven't tested it yet, but from a quick scan it lgtm :)

What would we need to do to be a sharing target? Or perhaps better, to be able to present multiple sharing targets? I'm guessing it's all hard coded in the App plist?

asmagill:
I'm not sure regarding becoming a sharing target... Subclassing NSSharing and using NSSharingServicePickerDelegate allows you to insert your own sharing options into your own application, but I haven't found an example yet of adding one that other applications could see... granted, I haven't looked too hard yet!

cmsj:
Apple has a guide on App Extensions that includes sharing, and you basically add another target to your Xcode project and it all looks to be very predefined and static. Worth thinking about fur the future I guess, but not relevant to this PR, which is excellent :)

PR #1072, opened 2016-11-07
`setMuted` now returns `true` on success

aldur:
I noticed that any call of `setMuted` returned `false`, even on built-in devices.
Referring to the docs, False should only be returned if the devices can not be muted, so I assumed it was an error, and discovered the typo fixed by the commit.


cmsj:
thanks!

PR #1075, opened 2016-11-12
hs.location.geocoder and region monitoring

asmagill:
Almost a complete rewrite of hs.location

It remains backwards compatible, but adds `hs.location` objects which have independent callbacks.  The independent objects can also perform region monitoring which allows callbacks on entry and exit of specific regions.

The hs.location.geocoder submodule allows converting locations into addresses and addresses into locations through the CLGeocoder class.


asmagill:
@cmsj, objections to merging, or would you rather this wait until the next release version?

cmsj:
@asmagill I'd rather merge and get feedback :)

PR #1084, opened 2016-11-17
Add hs.network.ping

asmagill:
Now I believe it's ready.  The stop method will only stop the specific pingObject or echoRequestObject rather than all of them, and some additional bugs were fixed.

See #1077 if you want a more descriptive explanation of the module.


asmagill:
Objections?  I want to land this if there are no objections before landing spotlight since both modify project.pbxproj and I like to avoid merge conflicts when possible.

cmsj:
No objections :)

Issue #1011, opened 2016-09-30
Modifier key doesn't get released


generator85:
I'm binding my Harmony Hub keyboard commands to Plex shortcuts.

I simply to this to bind F7 to command+L (next subtitle):
```
hs.hotkey.bind({}, "f7", function()
	hs.eventtap.keyStroke({"cmd"}, "L")
end)
```

But it seems that the command key doesn't get released, since Plex keeps on switching subtitles. (just like when I manually press command+L on my keyboard and then release the L key while keeping the command key pressed).

cmsj:
I would generally always recommend using the `releasedfn` parameter of `hs.hotkey.bind()` rather than the `pressedfn` one, when you're emitting events. So, add a `nil` argument before your function.

Having said that, I did notice that the `hs.eventtap.keyStroke()` function doesn't wait between posting its keydown and keyup events, which may be related. The next release will wait 200ms between each event. If changing to `releasedfn` doesn't help, hopefully this delay, will :)

I'll go ahead and close this Issue for now, but please feel free to re-open it if it's not fixed after the next release is out.

JakeSc:
Hey @cmsj, I noticed this change after updating to the latest release (0.9.49). Now `hs.eventtap.keyStroke` events are delayed, and delays during key-repeats (key being held down) are very noticeable. Is there a way to disable this behavior?

Here's my particular use case. I have assigned hotkeys to the `ctrl`+`h`/`j`/`k`/`l`, so that pressing them moves `left`/`down`/`up`/`right`, just like in Vim. Now with the delay, this mapping feels much more laggy.

Here's my lua:
```
local move = function(direction) hs.eventtap.keyStroke({}, direction) end
local moveLeft = function() move("left") end
local moveDown = function() move("down") end
local moveUp = function() move("up") end
local moveRight = function() move("right") end

function mod.init()
	hs.hotkey.bind({"ctrl"}, "h", moveLeft, nil, moveLeft)
	hs.hotkey.bind({"ctrl"}, "j", moveDown, nil, moveDown)
	hs.hotkey.bind({"ctrl"}, "k", moveUp, nil, moveUp)
	hs.hotkey.bind({"ctrl"}, "l", moveRight, nil, moveRight)
end
```

Thanks!

generator85:
[...]

PR #1085, opened 2016-11-18
Add delay option on keyStroke function

kkamdooong:
I inspired from [this comment](https://github.com/Hammerspoon/hammerspoon/issues/1011#issuecomment-261114434).

I think if keyStroke function offers delay option, so happy who use keyStroke function


cmsj:
Thanks! I made some small tweaks to the docs to make them more consistent with our usual style, but otherwise this was very helpful :)

PR #1090, opened 2016-11-21
hs.spotlight module

asmagill:
I still need to review `hs.application` to see where it can benefit from supporting alternate names for applications (find and get, definitely, but maybe others as well)

Because Spotlight can be one of those things that some people like and others hate, I'm thinking something along the lines of adding an `hs.application.enableSpotlightForNameSearches` which will take a boolean... true saves a value with `hs.settings` so that the on each restart, Spotlight is used to build up the alternate names lookup table, while false (or the default of no setting) simply makes the lookup table empty and doesn't start the Spotlight search.

Thoughts?

### DIFF
--- a/Hammerspoon.xcodeproj/project.pbxproj
+++ b/Hammerspoon.xcodeproj/project.pbxproj
@@ -241,6 +241,8 @@
 		D64B03401BA954780006C468 /* LuaSkin.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 4F9850E81B722CBF00F7E909 /* LuaSkin.framework */; };
 		D64B03421BA954780006C468 /* webview.h in Headers */ = {isa = PBXBuildFile; fileRef = D64B03361BA9541E0006C468 /* webview.h */; };
 		D64B03471BA954A20006C468 /* usercontent.m in Sources */ = {isa = PBXBuildFile; fileRef = D64B03381BA954660006C468 /* usercontent.m */; };
+		D6A212EE1DE38F5F00998C0B /* LuaSkin.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 4F9850E81B722CBF00F7E909 /* LuaSkin.framework */; };
+		D6A212F41DE38F8200998C0B /* internal.m in Sources */ = {isa = PBXBuildFile; fileRef = D6A212E71DE38F3B00998C0B /* internal.m */; };
 		D6A318A51DDD833100D9B328 /* LuaSkin.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 4F9850E81B722CBF00F7E909 /* LuaSkin.framework */; };
 		D6A318AB1DDD836800D9B328 /* internal.m in Sources */ = {isa = PBXBuildFile; fileRef = D6A3189B1DDD830A00D9B328 /* internal.m */; };
 		D6A318AC1DDD836800D9B328 /* SimplePing.m in Sources */ = {isa = PBXBuildFile; fileRef = D6A3189D1DDD830A00D9B328 /* SimplePing.m */; };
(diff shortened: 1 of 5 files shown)


cmsj:
the idea of having the hs.application integration being configurable, is interesting. I think we should aim to have it on by default, but perhaps not initially. Something like `hs.application.find()` obviously cries out for searching more widely than it currently does.

asmagill:
The default would be to not fill the map since checking the setting would return nil -- effectively false.

The map filling has to be done in the background since Spotlight queries are asynchronous.  Even if we have `find` turn it on the first time it is used, that particular use of `find` will still return no result since the map won't have a chance to fill until the query callbacks occur the next time Hammerspoon is idle.

I could see having `find` (and whatever other functions benefit) displaying a log message the first time they fail to find a result check to see if Spotlight integration has been enabled and if it hasn't log a recommendation to the console telling the user how to turn it on (or off, if we make turning it on automatic in such a case, however, at that particular moment, the find will still have to return that nothing was found since the map will take time to build)

I guess it comes down to I question the idea of turning it on the first time it is needed because it won't change the result for that particular function call, only future ones.  Its only useful if it's already running.  But adding code that logs to the console that they can turn it on to maybe provide additional results does make sense... once per function per load of Hammerspoon (i.e. track if we've show the message in a local variable)?  Or once and then never again (i.e. track that we've shown the message in a setting)?

cmsj:
The search could be started the first time hs.application is loaded, or it could even just be a special case that the core app runs when it first loads. Neither guarantees that the results will be available in time for the first use of hs.application though, so I suppose it's going to be hard to make this seamless without blocking until results are in, which seems bad. Maybe it does need to stay as an optional feature for those functions, with some way to prime it and see if it's returned results yet?

asmagill:
It really only takes about half a second to start populating the map table... but it does (currently) have to be populated asynchronously.

Take a look at the `init.lua` for `hs.application` in the latest version of this pull... It checks a setting, if it is nil (which it will be for anyone who has not set a value), then it prints a message to the console the first time they use the `hs.application.find` function... if they use the function indicated and set it to true, then spotlight queries for the alternate names immediately and on every subsequent load of the `hs.application` module.  If they set it to false, then it suppresses the message, but does not use Spotlight.

I could invert the logic, but in general I prefer things that I "opt in" to rather than have to "opt out" of...

I could make a blocking version of the spotlight query, and it *usually* would be under a couple of seconds (in my experience)... I think I'd need to add a new flag because isGathering doesn't become true until after the query starts, so there is a small window when the following would fail because the `while` starts too fast:

~~~lua
a = hs.spotlight.new():....whatever...:start()
-- without hs.timer.usleep(10000) here (maybe a smaller number would work, but not much
-- smaller), the while is never entered because the query hasn't entered the gathering
-- phase yet (hasn't passed "didStart")
while(a:isGathering()) do 
  -- could put some timeout logic in here if we wanted as well
end
-- now `a` has some stuff in it, though it could still be updated with "didUpdate"
-- messages unless we also do a:stop()
~~~

Should I add a method which will allow code like the above?  I'm of two minds... on the one hand, it kind of misses the point and strength of Spotlight being async... otoh I sometimes hate having to rethink an approach to a problem because some specific part of it is async by nature.

asmagill:
What's the consensus?  Currently `hs.application.find` will not use spotlight until you enable it, but it will display a message to the console indicating that you *can* enable it when you use the function.  Enabling it persists across reloads/restarts.

And should I play around with adding a way to use spotlight in a non-asynchronous manner?  I think I know a way to do it, but it will be blocking during that time... usually it's pretty quick, I've never seen a query take more than 5 seconds to reach completion (and usually under 2) -- not counting updates after the initial gather, but... there is no guarantee... I'd probably add a user adjustable timeout to it as well, just to be safe.

cmsj:
@asmagill I don't feel like I've thought about this deeply enough to offer up a sensible opinion. I'm happy to defer to your judgement on the best way to proceed :)

latenitefilms:
@asmagill For what it's worth, I think the way you've tackled it makes sense to me! Personally, I think anything that affects speed/performance should always be opt-in, as opposed to opt-out.

chdiza:
> Enabling it persists across reloads/restarts

What happens if it's on and then the user disables spotlight outside of HS?

I always have spotlight either set to ignore my hard drive, or prevented from ever launching by disabling its Launch{Agent,Daemon}.  I would hope that under these circumstances HS will not be delayed in carrying out any tasks, e.g., by looking for a non-existence spotlight cache, or querying a non-responsive `md*` process.

asmagill:
As far as the lookup changes to `hs.application` are concerned, if the alternate names table is empty, then it's empty -- there are no alternate names to compare against so you'll get the same result you do now when trying `hs.application.find("AppleScript Editor")`... no result even when Script Editor is running.

On such a machine, you'd probably want to enter `hs.application.enableSpotlightForNameSearches(false)` so that it doesn't print the logging message even the one time per reload, but otherwise having an empty spotlight database or having disabled Spotlight won't matter since the query is run asynchronously.

cmsj:
@asmagill this seems like it's ready to land - anything else you want to do on it?

asmagill:
I have no current plans for more additions, so barring any other questions or suggestions, it's ready.

PR #1094, opened 2016-11-23
Port LuaCocoa's ObjC exception handler and add hs.crash.throwObjCException() to test it

cmsj:


cmsj:
As a quick way to show how this works, `hs.crash.throwObjCException("lolception", "This was deliberate")` in your init.lua will cause Hammerspoon to exit, but running the same from the Console (which runs entered code via `lua_pcall()` will print a traceback to the console).
To see the old behaviour, comment out `LUA_USE_OBJC_EXCEPTIONS` in `LuaSkin/lua-5.3.3/src/luaconf.h` and you'll get Hammerspoon exits in both of the above scenarios.

cmsj:
I know I've generally been against patching Lua in the past, and that still holds, but I wonder if this is useful enough to be worth doing.

asmagill:
How hard would it be to get the same in the init.lua file to dump the message to the console rather than crashing? I think a full blown crash is probably the worst thing we can do to a user... even a cryptic error message that they paste into an issue on this site is better than getting mad because "it blew up" and they have no idea why or anything to show for it.

I too am leery of patching the Lua source directly, but if it's the only way to add this, I think it would help greatly, especially when testing out new code... sometimes I would like to try a few parameter changes to see exactly what causes the exception and what doesn't and having to restart Hammerspoon each time is annoying.

Another idea/question came to mind just now as well... what would it take to add a facility that allowed users to attach comments to a crash report when a crash does occur?  Something like Apple's dialog which asks if you want to add comments when submitting a crash report during beta testing?  Does this require a separate "Helper Application", or can it be made part of the Crashlytics crash reporting process?  Without knowing what the user was just doing, it's not always obvious what caused the problem.

cmsj:
So to catch exceptions caused by init.lua we'd either wrap that section of MJLua.m with a `@try{}@catch{}` thing, or change the internals somehow so it ends up being called with pcall. I would tend to favour the former I think.

AFAIK there's no way to add annotations to a Crashlytics report after it's happened, but I can check again - there may even be a full-fledged crash report dialog hidden in there for us to use :)

cmsj:
I was wrong, it is possible to add annotations to a crash report before submitting it, but there's no precanned UI, so we'd need to do our own thing in `MJAppDelegate.m`'s `- (void)crashlyticsDidDetectReportForLastExecution:(CLSReport *)report completionHandler:(void (^)(BOOL submit))completionHandler`. That seems like fodder for a separate Issue :)

cmsj:
Hmm, not sure why I thought this would crash entirely when an exception is triggered from init.lua, that stuff is all inside a `pcall()` and testing again now to figure out why it was different, shows it indeed captures the exception and dumps it to the Console :)

So, since it seems like there's no additional work to be done here (I just pushed an update to add a test to LuaSkinTests), any objections if I land this?

asmagill:
lgtm

Issue #1081, opened 2016-11-16
Unify music player extensions


S1ngS1ng:
Hi. Thanks for merging my PR on VOX music player.

I noticed that, the `extension/itunes`, `extension/spotify` and `extension/vox` are sharing some similar APIs. I would like to wrap them up in **one file**, making it something like `hs.player.play()` for example.

Would you think it is a good idea? Or does it worth to do this?

asmagill:
If they can be closely integrated, then I think creating a single merged module makes good sense... and we should then encourage future player additions to extend that instead of creating yet another module name in the `hs` namespace.

For backwards-compatibility reasons, I'd also recommend replacing `itunes`, `spotify`, and `vox` (well, maybe/maybe not since its so new... your call) with "wrapper" modules... look at the code for `hs.javascript` and `hs.applescript` to see what I mean... they're both wrappers for `hs.osascript`.

cmsj:
I'm in favour of unifying the code, so long as it exports backwards-compatible API for people using the old module names.

cmsj:
I remain in favour of this happening, but it doesn't seem like anyone has the time to work on it (and this was actually the second issue we had open about this), so I'm going to close it out. If someone wants to work on this, please re-open :)

PR #1099, opened 2016-11-26
Add 'hs.deezer'

aldur:
Deezer music control extension (yet another one).

I'm aware of #1081, but ATM I don't feel I have enough Lua experience to provide an elegant solution to the problem.

Meanwhile, I needed something to control Deezer, so I adapted the code from the Spotify extension.
Feel free to keep this on hold until a general wrapper is ready, or merge whenever you are ready.

### DIFF
--- a/extensions/deezer/LICENSE
+++ b/extensions/deezer/LICENSE
@@ -1,0 +1,21 @@
+Released under MIT license.
+
+Copyright (c) 2016 Adriano Di Luzio
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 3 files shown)


S1ngS1ng:
I was considering wrap them up. Though, it is kinda tough for me 😞 , as the API varies from app to app.

cmsj:
I think this is fine to land - we should figure out a way to merge these music control extensions, but I don't see any point holding them up until that time.

cmsj:
(I've poked the Travis CI build to re-run for this, because I think it failed for other reasons. Once this has a green CI, it can merge :)

PR #1108, opened 2016-11-29
Luaskin table w call

asmagill:
This pull allows LuaSkin's `checkArgs:` method to recognize a table with a __call metamethod as a valid candidate for the LS_TFUNCTION type.

Note that this does not change the type actually on the stack... if your function/method (e.g. `hs.tasks.new`) does it's own check of the parameter type to determine internal logic, it will still see this argument as a table... this update *only* addresses LuaSkin's `checkArgs:`.

This pull was test with the `hs.timer` module and worked as expected:

~~~lua
> t = hs.timer.doAfter(1, function(...) print((hs.inspect(table.pack(...)):gsub("%s+", " "))) end)
{ n = 0 }

> t = hs.timer.doAfter(1, setmetatable({}, { __call = function(...) print((hs.inspect(table.pack(...)):gsub("%s+", " "))) end } ))
{ {}, n = 1 } -- the __call metamethod expects the table itself as the first argument (i.e. `self`)

> t = hs.timer.doAfter(1, setmetatable({}, { __notcall = function(...) print((hs.inspect(table.pack(...)):gsub("%s+", " "))) end } ))
[string "t = hs.timer.doAfter(1, setmetatable({}, { __..."]:1: ERROR: incorrect type 'table' for argument 2 (expected function)
stack traceback:
	[C]: in function 'hs.timer.internal.doAfter'
	[string "t = hs.timer.doAfter(1, setmetatable({}, { __..."]:1: in main chunk
	[C]: in function 'xpcall'
	...app/Contents/Resources/extensions/hs/_coresetup/init.lua:321: in function <...app/Contents/Resources/extensions/hs/_coresetup/init.lua:301>
~~~

A review of modules for which this is insufficient should be made at some point, but this should be independent of this pull, since each module will have to be reviewed to determine what, if any, implications adding additional logic may involve.

For the record, the following modules are known to require updates to take advantage of this pull (this is by no means complete, just what I encountered while trying to identify a reliable test):
* `hs.hotkey` - currently uses its own internal logic to identify such tables... a "reversion" is possible once this lands, but is certainly not required.
* `hs.eventtap` - does not currently use LuaSkin's `checkArgs:` method for parameter validation.
* `hs.task` - uses the parameter type as identified with `lua_gettype` in the `new` function for internal logic... while the `checkArgs:` method will accept a table with a __call metamethod as a function, the internal logic will still see it as a table and a fix should be considered.

### DIFF
--- a/LuaSkin/LuaSkin/Skin.h
+++ b/LuaSkin/LuaSkin/Skin.h
@@ -377,6 +377,20 @@ NSString *specMaskToString(int spec);
  */
 - (void)checkArgs:(int)firstArg, ...;
 
+/*!
+ @abstract Returns the effective Lua type for the item at the specified stack index.
+
+ @discussion This method returns the Lua type for the item at the specified index.
+
+ At present, the only difference between this and the Lua API function `lua_type` is that a table with a __call metamethod is considered a function and will return LUA_TFUNCTION, since [LuaSkin protectedCallAndTraceback:nresults:] can accept such a table as the function to invoke.
+
+ @param idx the index on lua stack which contains the data to return a type for
+
+ @returns An integer which will be one of the following: LUA_TNIL, LUA_TNUMBER, LUA_TBOOLEAN, LUA_TSTRING, LUA_TTABLE, LUA_TFUNCTION, LUA_TUSERDATA, LUA_TTHREAD, or LUA_TLIGHTUSERDATA.
+
+ */
+- (int)luaTypeAtIndex:(int)idx ;
+
 #pragma mark - Conversion from NSObjects into Lua objects
 
 /*! @methodgroup Converting NSObject objects into Lua variables */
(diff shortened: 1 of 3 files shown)


cmsj:
maybe we could have a LuaSkin method which we can use instead of `lua_type()` when what we actually want to know is whether a stack entry is callable? `[skin isCallableAtIndex:2]` or similar?

asmagill:
Maybe, but what about a more generic replacement for `lua_type` entirely... 

`[skin luaTypeAtIndex:2]` and possibly add a `withOptions:` variant with the LS_NSRawType option where it forgoes checking a table's metamethods and *just* checks to see if it's a table...

Are the core lua types identical to the LuaSkin variants?  I know we've added a few, but, for example, does LUA_TTABLE == LS_TTABLE?  If so, this could a drop-in replacement for the most part.

asmagill:
Ok, brain fart... we're returning the LUA_T* type value, not the LS_T* type... we can make it be a drop in replacement for lua_type...  Let me add that real quick...

PR #1129, opened 2016-12-09
Chooser cell type for choice with no subtext

spro:
This adds a different version of cell to make it look nicer when a chooser choice has no subtext:

There were also some simplifications to how the auto layout was set up, and some alignment adjustments while doing so.


latenitefilms:
Awesome work @spro !

asmagill:
lgtm

Issue #1113, opened 2016-11-30
A setter for chooser:selectedRow (or workaround?)


mikaylathompson:
Currently, `chooser:selectedRow()` is just a getter--would it be possible to add a setter? Something like `chooser:selectedRow(3)` or `chooser:selectRow(3)`.

My current workaround is that every time I want to change the selected item, I rotate the entire list of choices such that the element I want to be selected is now in position 1.  I'm pretty unhappy with this because it's both ugly and inefficient.

The workaround I'd be okay with is sending eventtap "down arrow" events. I would be totally happy with this, but it hasn't worked. (my code looks like `hs.eventtap.keyStroke({}, "down")`, but nothing happens)

Any other workarounds I should take a look at?

latenitefilms:
That's an interesting idea! Curious... what's the scenario when you'd want to "highlight" a chooser row? Given that moving the mouse over the chooser will affect what's highlighted, do you want to override this default behaviour? What happens when the user changes the chooser's query?

Regardless of why you want to do it, I can't get `hs.eventtap.keyStroke({}, "down")` to work with the console either, which is interesting and confusing, as it's definitely working in other apps - so your code is certainly correct.

The only hacky workaround I can think of if you don't want to mess with the `hs.chooser` code is to move the mouse over the item you want, then instantly move it back to it's original position. This will work - but it's not ideal obviously.

I'm also curious... are you actually using `chooser:selectedRow()` function?

mikaylathompson:
I'm trying to use it to make a cmd-tab app switcher replacement (which uses the functionality of the chooser, like searchability and the numerical shortcuts).  So I'd like to duplicate the functionality where if you press cmd-tab repeatedly, it moves down the list, and shift-cmd-tab moves up the list.

Good points on the mouse and the query. In my case, it would be totally fine to override the highlighted row with the mouse, and maybe to reset position on changing the query? I think on the query, I'd be fine with either keeping the position or resetting.

So my code would ideally be something like: `chooser:selectRow(chooser:selectedRow() + 1)`


Just another datapoint on the keyStroke not working: this line `hs.eventtap.keyStroke({"cmd"}, "1")`, works totally fine and selects the first item. And `hs.eventtap.keyStroke({}, "A")` is also fine and enters "A" into the query box. So it's something fairly specific to the arrow keys.

cmsj:
This should be addressed by #1140 :)

Issue #1058, opened 2016-10-26
Inspiration...


latenitefilms:
I would LOVE it if you could build something like this using Hammerspoon:

http://www.videocopilot.net/blog/2016/10/new-workflow-plug-in-fx-console-is-now-available/

This is a great example of the kind of user interfaces it would be awesome if you could build use Lua.

cmsj:
@latenitefilms have you seen hs.chooser? It's not quite as capable as the one shown in that video, but it's essentially the same kind of thing.

cmsj:
(I'm going to go ahead and close this now, please feel free to re-open if you think there is more we need to do here)

latenitefilms:
I didn't even think about **hs.chooser** - I've seen it in the API documents, but haven't played with it yet. Thanks @cmsj - I'll give it a bash!

latenitefilms:
@cmsj - I've finally had a proper play with **hs.chooser** - it's awesome! Thank you!

A few things I'd love to see added to it:

- Ability to "right click" on results. For example, I'd love the ability to right click on a result, which opens up a menu allowing me to "Remove Search Result", etc.

- I'd love a function that allows me to know if the **hs.chooser** is currently visible or not.

- Dark Mode doesn't look great for some reason on my machine (MacBook Pro, macOS 10.11.6, Hammerspoon 0.9.48)  - see attached screenshots:

<img width="1886" alt="lightmode" src="https://cloud.githubusercontent.com/assets/22286696/19756501/dc93f638-9c68-11e6-9479-9e65d585eddd.png">
<img width="1886" alt="darkmode" src="https://cloud.githubusercontent.com/assets/22286696/19756502/dcec3f32-9c68-11e6-8eb1-22a0a7cf085b.png">

- Also, I'm being REALLY picky - but if you look at the attached screenshot you can see that the UI elements don't really line up:

<img width="793" alt="ui-glitch" src="https://cloud.githubusercontent.com/assets/22286696/19761237/c1e23d0a-9c80-11e6-9f0d-f7aaec14cd7b.png">

I might try giving some of these a bash myself sometime!

cmsj:
[...]

Issue #782, opened 2016-02-06
hs.chooser callback for Return/Enter pressed


ventolin:
I'm experimenting with implementing a very simple Notational Velocity replacement using hs.chooser. Basically, right now I am scanning a directory for text files and presenting those files (with contents as subtext) via hs.chooser, then opening the selected file in an editor. This works great for files that already exist. But one thing I'd like to do to mimic Notational Velocity is to create a new file using the search string as the file name (if it doesn't exist), and then edit that.

What would be helpful is to have some kind of callback when the user hits Return/Enter and no choices are selected. Or alternately, perhaps a second argument to the main callback that describes whether the chooser was dismissed or not, along with an hs.chooser setting to make the chooser call that callback when Return/Enter is pressed (then I could check if choice == nil and dismissed == false, for example).

I can sort of do what I want, now, by checking the current query() value in the dismiss callback if the choice is nil, but that's less than ideal because it requires me to hit escape to create the new file (which is confusing) and also no longer allows me to actually cancel the whole process.

szymonkaliski:
Maybe related: https://github.com/Hammerspoon/hammerspoon/issues/748 ? If we decide to have separate hs.input you could use that to create notes, and hs.chooser to browse them. Otherwise if we implement this callback for "enter" with empty list, I'd love to have UI where I can have no rows, so we could use hs.chooser as pure input... Something to figure out.

cmsj:
I was trying to avoid complicating the conceptual aspect of chooser by not adding this, but I do think it's a valid use case

ventolin:
I had an idea that I might be able to do what I want by using queryChangedCallback and checking for Return/Enter press there. I haven't had time to try it out yet, though. If that works then perhaps this issue isn't that relevant.

ventolin:
So I was able to get something working the way I wanted. I'm using the queryChangedCallback to do some custom filtering, and if the filter results in an empty list (no files found with the given query, so the chooser list would be empty), I'm inserting an item into the chooser list to "Create new note," so that hitting enter will select that item, which acts as a command that uses the current query text as the new filename and creates the file. This is working well (I was afraid the filtering would be too slow, but it's fine so far), and I'm honestly not sure if my original feature request would make this much easier, because I'd still want to do the custom filtering.

So as far as I'm concerned, this issue could be closed and you could avoid complicating the chooser further. Thanks for considering it, though!

heptal:
[...]

PR #468, opened 2015-08-05
Extensions in xcode

cmsj:
This is my WIP PR for switching extensions over to building in Xcode. Closes #34 

@asmagill definitely needs to look at this and comment :D

(the good news is that I'm not changing the structure of the app bundle at all. I gave up trying to make it work, and am just using a script to copy the dylibs into the old places, with the old names, just with a .dylib extension instead of a .so)

PR #1140, opened 2016-12-15
additions to hs.chooser

asmagill:
Adds optional row parameter to `hs.chooser:selectedRow` per #1113 
Adds `hs.chooser:selectedRowContents` per #1058 

@cmsj, you should probably take a look at this... I tested this against a static table of items, but I know chooser has options for a dynamically generated list... I *think* it should work, but I haven't had a chance to build one yet.

### DIFF
--- a/extensions/chooser/HSChooser.h
+++ b/extensions/chooser/HSChooser.h
@@ -72,7 +72,7 @@
 
 // HSChooserTableViewDelegate
 - (void)tableView:(NSTableView *)tableView didClickedRow:(NSInteger)row;
-- (void)didRightClick;
+- (void)didRightClickAtRow:(NSInteger)row;
 
 // NSTextFieldDelgate
 
--- a/extensions/chooser/HSChooser.m
+++ b/extensions/chooser/HSChooser.m
@@ -270,12 +270,13 @@
     }
 }
 
-- (void)didRightClick {
+- (void)didRightClickAtRow:(NSInteger)row {
     if (self.rightClickCallbackRef != LUA_NOREF && self.rightClickCallbackRef != LUA_REFNIL) {
         // We have a right click callback set
         LuaSkin *skin = [LuaSkin shared];
         [skin pushLuaRef:*(self.refTable) ref:self.rightClickCallbackRef];
-        if (![skin protectedCallAndTraceback:0 nresults:0]) {
+        lua_pushinteger(skin.L, row + 1) ;
+        if (![skin protectedCallAndTraceback:1 nresults:0]) {
             [skin logError:[NSString stringWithFormat:@"%s:rightClickCallback error - %@", USERDATA_TAG, [skin toNSObjectAtIndex:-1]]] ;
             lua_pop(skin.L, 1) ; // remove error message
         }
--- a/extensions/chooser/HSChooser.m
+++ b/extensions/chooser/HSChooser.m
@@ -7,6 +7,7 @@
 //
 
 #import "HSChooser.h"
+#import "chooser.h"
 
 #pragma mark - Chooser object implementation
 
@@ -263,6 +264,7 @@
         [skin pushLuaRef:*(self.refTable) ref:self.completionCallbackRef];
         [skin pushNSObject:choice];
         if (![skin protectedCallAndTraceback:1 nresults:0]) {
+            [skin logError:[NSString stringWithFormat:@"%s:completionCallback error - %@", USERDATA_TAG, [skin toNSObjectAtIndex:-1]]] ;
             lua_pop(skin.L, 1) ; // remove error message
         }
     }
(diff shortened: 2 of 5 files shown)


asmagill:
@cmsj, when manually changing the query with `hs.chooser:query` it does not affect the items being displayed if the chooser is currently visible, and when the choices array is invalidated with `hs.chooser:refreshChoicesCallback` it resets the list to the entire (new) list, ignoring the current query string (again, if visible, when this is called).

(a) is this intended? if so, educate me please, since I'm missing the reason
(b) would calling `[chooser controlTextDidChange:nil]` at the end of `query` and after the choices callback completes be sufficient to address this?
(c) should it be added or made an option? or completely left as is?

asmagill:
For cabric #468:

hs.chooser isn't validating types of keys in dictionary used to generate list... should either force text and subtext to be string (and image to be image) or coerce when not.

Will add to this pull tonight when I have the time.

asmagill:
@cmsj, can you take a look at this pull, specifically my first two comments, when you get a chance?  I'd like to start landing the currently outstanding pulls and I need your input on these points to know if this specific pull is ready or needs more work.

cmsj:
@this PR looks fine, and I don't think it will break dynamic choice lists.

As for your questions - those behaviours are not intended, and I think (b) should do the trick. I also just noticed that `:query()` doesn't have/document an easy way to clear the search query. Setting it to `""` is probably sufficient, but we don't mention that, and it might be nicer to be able to pass a `nil`.

Sorry for the delay in checking this PR, and thanks!

junkblocker:
Umm, something is broken on HEAD and maybe dynamic choice lists ARE broken. I press a single character and ``queryChangedCallback`` gets called umpteen times. Code for reference is ventolin's notes module from https://github.com/Hammerspoon/hammerspoon/issues/782 .

cmsj:
@junkblocker can you try with this patch applied? I think these are the two lines that are causing the potential for loops:

```
diff --git a/extensions/chooser/HSChooser.m b/extensions/chooser/HSChooser.m
index 69c5aff..277b719 100644
--- a/extensions/chooser/HSChooser.m
+++ b/extensions/chooser/HSChooser.m
@@ -380,7 +380,7 @@ - (void)selectPreviousChoice {
 #pragma mark - Choice management methods

 - (void)updateChoices {
-    [self controlTextDidChange:[NSNotification notificationWithName:NSControlTextDidChangeNotification object:nil]] ;
+    //[self controlTextDidChange:[NSNotification notificationWithName:NSControlTextDidChangeNotification object:nil]] ;
     if (self.window.visible) {
         [self.choicesTableView reloadData];
     } else {
diff --git a/extensions/chooser/internal.m b/extensions/chooser/internal.m
index 21bac7a..2677082 100644
--- a/extensions/chooser/internal.m
+++ b/extensions/chooser/internal.m
@@ -259,7 +259,7 @@ static int chooserSetQuery(lua_State *L) {
                 lua_pushnil(L);
                 break;
         }
-        [chooser controlTextDidChange:[NSNotification notificationWithName:NSControlTextDidChangeNotification object:nil]] ;
+        //[chooser controlTextDidChange:[NSNotification notificationWithName:NSControlTextDidChangeNotification object:nil]] ;
     }
     return 1;
 }
```

junkblocker:
Yes @cmsj, that seems to fix it. Thanks!

cmsj:
@junkblocker thanks!

@asmagill I'm going to apply that partial reversion patch, I think we should probably leave things as they are in this regard - users should be able to determine when it's safe for them to call `:refreshChoicesCallback()`

asmagill:
@cmsj, how hard would it be to setup a "beta" track for releases?  This isn't the first time we've had to make a quick reversion because of changes which work in 90% of the cases, but not in that final 10% that no tester can always check out fully.  And unit testing only works if you know of an issue to test for; it's more suited to make sure future changes don't break things already working rather than testing every use case of something new... and considering we're doing more and more things with async threads and callbacks... well sometimes I'm at a loss for even *how* to build a certain type of test, though that may just be a lack in my knowledge that could be addressed by a tutorial or good reference.

In general I do like to leave a pull up for a few days (except for right before a release), but that relies on our end-users being able and willing to build their own local copies of Hammerspoon... it would be quicker and easier (for them) if some of our users could toggle a checkbox or a flag with `hs.setting` and get early releases for testing...

I don't know anything about Sparkle, so I don't know if this is a something relatively easy to add once we figure out the proper setup on github wrt to branches, etc., but I do think it's an idea worth considering.

latenitefilms:
@asmagill & @cmsj - On this note, if doing a "beta" track is too tricky, what would be really helpful is having a **build** folder in Master, that's always up-to-date. So for example, if @asmagill makes a tweak to `hs.chooser` and merges the changes with the Master, rather than having to build locally, or wait until a public release, we could just grab the "latest" build from the build folder. Thoughts?

cmsj:
@asmagill yeah I know the beta thing keeps coming up. I do like the idea, but I'm not super thrilled with the overhead. I'll open up a new issue and braindump what I think we'd need to do, to make it work.

PR #1141, opened 2016-12-15
add ability to encode/decode NSKeyedArchiver

asmagill:
Adds `hs.pasteboard.writeArchiverDataForUTI` and `hs.pasteboard.readArchiverDataForUTI` which allow reading serialized NSObject's into and out of the pasteboard.

### DIFF
--- a/extensions/pasteboard/internal.m
+++ b/extensions/pasteboard/internal.m
@@ -360,7 +360,7 @@ static int readItemForType(lua_State *L) {
     return 1 ;
 }
 
-/// hs.pasteboard.readPListForUTI([name], uti) -> string
+/// hs.pasteboard.readPListForUTI([name], uti) -> any
 /// Function
 /// Returns the first item on the pasteboard with the specified UTI as a property list item
 ///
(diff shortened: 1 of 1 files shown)


cmsj:
@asmagill I don't really understand the UTI and KeyedArchiver stuff, so I trust your decision on when this is ready to land!

asmagill:
These additions are primarily aimed at trying to provide at least minimal support for manipulating pasteboard data that doesn't quite fit the standard types (string/number, image, etc.)... most people will probably never use them.

In theory these should allow accepting and manipulating almost *anything* that *any* application can use a pasteboard for, though of course LuaSkin may still need a conversion function before the data can be of any use beyond a raw binary chunk... but the pasteboard code shouldn't need an update once we  add the conversion functions.  About the only remaining thing that I can think of that a pasteboard can support that Hammerspoon can't yet would be providing promised, rather than actual, data.

I need to review and probably redo some of the documentation with examples, but I'll wait for the dust to settle with the new documentation standard before doing so.

PR #1146, opened 2016-12-17
Add hs.canvas

asmagill:
Adds hs.canvas to Hammerspoon

Includes hs.drawing wrapper (disabled by default).  I need to review the documentation again and add a function to enable/disable the wrapper, but its almost ready


randomeizer:
Is there any documentation for this?

latenitefilms:
@randomeizer See: https://github.com/asmagill/hammerspoon_asm/tree/master/canvas

asmagill:
The URL @latenitefilms gives is probably the best source at the moment... like all Hammerspoon modules, this one will include documentation in the Hammerspoon Docset available with Dash, and include internal documentation once it's merged into core, but that URL is probably best for the moment... the main difference you'll need to keep in mind is that this pull installs the module as `hs.canvas` while the original github site and source (should you install it from there instead) installs it as `hs._asm.canvas`... otherwise they are identical minus a few documentation tweaks.

The examples site listed in the internal documentation for this pull points to https://github.com/asmagill/hammerspoon/wiki/hs.canvas.examples which has already had the `_asm` portion removed from the code examples given.

randomeizer:
Cool, thanks.

Issue #1131, opened 2016-12-09
Help with Wifi Watcher Function


nikitavoloboev:
I have this function in hammerspoon that works really great : 

```Lua
-- _wifi watcher
wifiwatcher = hs.wifi.watcher.new(function ()
    net = hs.wifi.currentNetwork()
    if net==nil then
        hs.notify.show("wifi disconnected","","","")
    else
        hs.notify.show("wifi connected","",net,"")
    end
end)
wifiwatcher:start()
```

The only thing is that I find annoying how every time I login, I get two notifications on the top since it connects to wifi every time on login. Is it possible to extend this function of wifi watching to only work when logged in already and not have these notifications shown on the login screen as they are quite useless to me.

Thank you for any help.

eoranged:
I don't undestand what *exactly* you're trying to achieve, but [hs.caffeinate.watcher](http://www.hammerspoon.org/docs/hs.caffeinate.watcher.html) should do the trick:

```
wifiwatcher = hs.wifi.watcher.new(function () ... end)
cwatcher = hs.caffeinate.watcher.new(function(state) 
   if state == hs.caffeinate.watcher.systemDidWake then
      wifiwatcher:start()
   elseif state == hs.caffeinate.watcher.systemWillSleep then
       wifiwatcher:stop()
   end
end)
cwatcher:start()
```

I haven't tested the code above and It might not work straight ahead, but It should give you the basic idea how this can be done.

asmagill:
I think I've identified a problem with `hs.wifi.watcher`... If you have multiple wifi watchers defined, then the callback for each watcher is getting called for the notification each watcher has registered for... simplified, if I register 2 watchers, each watcher gets called twice for every change, if I register 4, then each watcher gets called four times, etc.

@nikitavoloboev, do you have another wifiwatcher running for some additional purpose? Or maybe you've setup more than one in the console during your testing?

At any rate, the ultimate fix is to figure out why and then correct for it, but at least now I know the "what".

nikitavoloboev:
This are the only lines of code I have in my init.lua relating to hs.wifi.watcher

I also didn't type anything in the console related to it.

asmagill:
Ok, it seems to have something to do with the number of times `[CWInterface interface]` is invoked...

[...]

PR #1152, opened 2016-12-21
modernize hs.wifi and hs.wifi.watcher

asmagill:
Removes methods deprecated in 10.10

Adds watchers for link quality, BSSID, power, etc. in addition to default of SSID change.

Addresses #1131

### DIFF
--- a/extensions/wifi/internal.m
+++ b/extensions/wifi/internal.m
@@ -56,16 +56,16 @@ CWInterface *get_wifi_interface(NSString *theInterface) {
     if (_fnRef != LUA_NOREF) {
         LuaSkin *skin = [LuaSkin shared] ;
         lua_State *L = [skin L] ;
-        [[LuaSkin shared] pushLuaRef:refTable ref:_fnRef];
+        [skin pushLuaRef:refTable ref:_fnRef];
         if ([object isKindOfClass:[NSError class]]) {
-            [LuaSkin logInfo:[(NSError *)object localizedDescription]] ;
+            [skin logInfo:[(NSError *)object localizedDescription]] ;
             [skin pushNSObject:[(NSError *)object localizedDescription]] ;
         } else {
             [skin pushNSObject:(NSSet *)object] ;
         }
 
         if (![skin protectedCallAndTraceback:1 nresults:0]) {
-            [LuaSkin logError:[skin toNSObjectAtIndex:-1]] ;
+            [skin logError:[skin toNSObjectAtIndex:-1]] ;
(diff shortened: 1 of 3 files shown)


extensions/wifi/internal.m
```diff
@@ -110,14 +58,14 @@ - (void)invokeCallback:(id)object {
         lua_State *L = [skin L] ;
         [[LuaSkin shared] pushLuaRef:refTable ref:_fnRef];
         if ([object isKindOfClass:[NSError class]]) {
-            log_to_console(L, _cINFO, [(NSError *)object localizedDescription]) ;
+            [LuaSkin logInfo:[(NSError *)object localizedDescription]] ;
```
cmsj: This method looks like it has a lot of built-up LuaSkin weirdness. We have a `skin` object at the top, so the second `LuaSkin shared` is unnecessary, and these log calls could use the instance variants too.

asmagill:
holdover from copy/paste after pulling the outdated logging code... it's cleaner now :-)

asmagill:
@cmsj, noticed that your latest comment was tagged as a review, not a comment... How does that work?  Was the fact that you made a line-tem note sufficient to make it a review, or is their a field/toggle somewhere that I'm missing?

asmagill:
@cmsj, the travis log shows (among other things):

~~~
/Users/travis/.rvm/gems/ruby-2.0.0-p648/gems/xcpretty-0.2.3/lib/xcpretty/syntax.rb:24:in `highlight_with_formatter': undefined method `include?' for nil:NilClass (NoMethodError)
	from /Users/travis/.rvm/gems/ruby-2.0.0-p648/gems/xcpretty-0.2.3/lib/xcpretty/syntax.rb:15:in `highlight'
	from /Users/travis/.rvm/gems/ruby-2.0.0-p648/gems/xcpretty-0.2.3/lib/xcpretty/formatters/formatter.rb:171:in `format_failure'
...
~~~

Since it seems to be a ruby error, I'm uncertain... I can merge this locally with master and it builds fine, both App and documentation, so I don't know what to do next...

Is this because of changes you're making so Travis can support the documentation generation changes?

cmsj:
Shouldn't be anything to do with my docs work, none of that has hit master yet.

The error seems to be in the xcpretty stuff that we use in travis to make build logs not overflow the Travis limits. Perhaps we should have pinned that to a known working version. I'll try to investigate later today and when it's fixed I'll land this. 

As for the review thing - yes, starting a line comment these days starts a review.

Issue #1151, opened 2016-12-20
hs.webview -999 Error


latenitefilms:
Every now and again `hs.webview` will randomly give me the following error:

```
An Error code: -999 in NSURLErrorDomain occurred during navigation:

Description: The operation couldn’t be completed. (NSURLErrorDomain error -999.)
```

It only happens every now and again.

[This](http://stackoverflow.com/questions/16073519/nsurlerrordomain-error-code-999-in-ios) might be related?

Any suggestions on how I can fix it? Do I just need to put a delay somewhere in my code, or is this something that needs to be done "within" `hs.webview`?

latenitefilms:
[This](https://github.com/ShareKit/ShareKit/issues/56#issuecomment-2383297) might also be related?

Any ideas @asmagill or @cmsj ?

asmagill:
How often are you updating the URL?  And are the pages you're showing ones that use AJAX or some other javascript code to update themselves dynamically?

My first thought is to try something like `webview:stopLoading():url(thenewurl)` instead of just `webview:url(thenewurl)` and see if that makes a difference.

I need to see if I can contrive of a situation that *always* (or at least *usually*) produces this so I know what to test for before adding something to the module code... if you have a site that causes this more often than others, let me know and I'll take a look at it.

asmagill:
Ok, well `webview:url("http://www.yahoo.net"):url("http://www.google.com")` seems to trigger it pretty reliably, and sticking `stopLoading()` in between them doesn't help...

Need to do a little digging and decide whether it's better to make `url` blocking while it waits to ensure the page has completed (or cancelled) or if we can tie it to one of the delegate callbacks...

latenitefilms:
@asmagill - I'm using `hs.webview:html()` to generate the content dynamically.

I'm assuming maybe my code's taking too long to generate the HTML?

asmagill:
The latest pull listed above moves the actual meat of `url`, `reload` and `html` into an async queue running on a concurrent thread rather than the main one... the down side is now they return only one value (the webview object) since the navigation id won't be known until the old page is cancelled and the new one begins... if you need it, set up a `navigationCallback` and capture it during the `didStartProvisionalNavigation` message.

latenitefilms:
Legend, thanks so much! I've just added this updated `hs.webview` module to our code, so I'll let you know if we see any random -999 errors anymore.

PR #1153, opened 2016-12-21
hs.webview and -999 errors

asmagill:
This pull cleans up warnings during compile and moves the actual rendering for `hs.webview:url`, `hs.webview:reload`, and `hs.webview:html` into an async queue to prevent errors (-999) when trying to go to a new page when the old one hasn't finished rendering yet (or is doing an AJAX/Javascript request)

@cmsj, this does change the previous behavior in the following way (I think it's worth it, but want your input before merging this):

These methods no longer return the navigation id because it won't be known until the current page is cancelled and the new one is rendered by the threaded queue.  If you need the navigation id for your own tracking purposes, setup an `hs.webview:navigationalCallback` and capture it there, or at least wait until the navigational callback has passed at least the `didStartProvisionalNavigation` message and then use `hs.webview:navigationID`.

Addresses #1151


cmsj:
@asmagill I don't claim any great expertise on the webview module, but the change seems reasonable to me :)

PR #1154, opened 2016-12-22
hs.chooser ensure correct types for choices keys during display and search

asmagill:
addresses fabric issue 468

`text` and `subText` key values can still be any type, but for display and string search purposes they will now be converted to a string with `[NSString stringWithFormat:@"%@", value]` when they exist but are not strings.  This seemed easier (and maybe slightly more useful) then validating each dictionary to make sure `text` and `subText` were in fact strings within `hs.chooser:choices`.

`image` will be treated as nil for display purposes when it exists but is not an `hs.image` object.

This pull is separate from #1140 because I ran into merge conflicts when including it and trying a local merge back to master, but as separate pulls they both merge as separate merges back into master... whatever, this works :-)

### DIFF
--- a/extensions/chooser/HSChooser.m
+++ b/extensions/chooser/HSChooser.m
@@ -374,6 +374,7 @@
 #pragma mark - Choice management methods
 
 - (void)updateChoices {
+    [self controlTextDidChange:nil] ;
     if (self.window.visible) {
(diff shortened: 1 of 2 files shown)


asmagill:
meant to put the latest push in #1140, but it doesn't matter since both will be added for 0.9.51

PR #1155, opened 2016-12-22
[WIP] Replacement docstring processor

cmsj:


cmsj:
This isn't finished yet, I just thought I'd publish it now because I just got far enough to be able to recreate docs.json in a way that is at least compatible with making HTML docs and the sqlite database.

I know some folk use docs.json in other ways, so I'd be interested to hear if they have any problems. I've intentionally made the docs.json output be a superset of what was available before, so tools/modules/whatever can migrate over to the richer set of data that's available (which is very much up for discussion). Here is an example of what is available for a given function/method/variable/etc:

```json
      {
        "def": "hs.chooser.new(completionFn) -> hs.chooser object",
        "desc": "Creates a new chooser object",
        "doc": "Creates a new chooser object\n\nParameters:\n* completionFn - A function that will be called when the chooser is dismissed. It should accept one parameter, which will be nil if the user dismissed the chooser window, otherwise it will be a table containing whatever information you supplied for the item the user chose.\n\nReturns:\n* An `hs.chooser` object",
        "name": "new",
        "parameters": [
          "* completionFn - A function that will be called when the chooser is dismissed. It should accept one parameter, which will be nil if the user dismissed the chooser window, otherwise it will be a table containing whatever information you supplied for the item the user chose."
        ],
        "returns": [
          "* An `hs.chooser` object"
        ],
        "signature": "hs.chooser.new(completionFn) -> hs.chooser object",
        "stripped_doc": "",
        "type": "Constructor"
      }
```

While not used in this example, `stripped_doc` would include anything between the one-line `desc` and the start of the Parameters/Returns/Notes text.

I would now consider the `def` item as deprecated, and probably the `doc` one too, but I'm happy to talk about those.

I would also like to introduce an Example section under Notes, which is optional, but would (unsurprisingly) contain an example of using a particular piece of API. Something along these lines would also seen sensible at the module level.

cmsj:
(Note that there is one python dependency currently required to be installed, `docopt`)
(Also note that this is a heck of a lot more code than gencomments/genjson are, partially because I have intentionally tried to keep the code free of fancy magic, and partially because much more work is being done to parse the contents of the docstrings)

junkblocker:
@cmsj, I like to `pip install --user` this kind of stuff. Also depending on whether `pip2` or `pip3` was installed/updated last `pip` command gets overwritten with that version. And homebrew likes to mess with python too. So this install part is not making me entirely comfortable.

cmsj:
@junkblocker well, I've been careful to use `#!/usr/bin/env python` as the hashbang, and we had requirements before for building docs in Ruby. I know that homebrew likes to mess with Python and I deeply, profoundly and angrily disagree with that, and always stamp on it when it installs a new Python on my machines.

However we generate the docs, we need *some* kind of language and libraries to help, maintaining the entire stack by hand would be terrible. So, I'm not really sure what other options there are here?
(I could drop docopt as a dependency, it's doing pretty trivial work here, but the HTML generation needs *some* kind of templating library, so there will soon be a hard dependency)

junkblocker:
@cmsj, ok, let's enumerate options.

1. `docopt` is MIT licensed and single source file at a glance, so we can include it in if really needed but `getopt` in the standard library is fairly easy to use.
2. The very basic `string.template` is in the standard library. It meets the 'some kind of templating library' need. If we have more specific requirements we can look for similar single source + permissive license solutions.

`#!/usr/bin/env python` is usually essential but for this brew mess I think it might be better to say `#!/usr/bin/python` instead.

cmsj:
@junkblocker I'm movable on docopt, since the options are so simple, but `string.template` is going to make the HTML writing significantly less friendly. I would strongly prefer to use something like jinja2.

Switching to `usr/bin/python` is fine for the python interpreter, but it would mean `pip` could be coming from anywhere when the requirements are installed.

I have a suggestion though - if I ditch docopt and make the jinja module only be imported when generating HTML docs, we should be fine. Folk doing local builds are unlikely to need to build the html docs (it's not in the default Makefile set afair, just the JSON part is). Travis has a more predictable environment, and I will take care of my environment.

asmagill:
I know python about as well as I know ruby (i.e. enough to work on someone else's code, but not enough to build my own things from scratch) so I don't have strong opinions either way... I'm guessing Apple doesn't include pip in it's default python installation so we can't invoke it with a hard path?

What if you did include the `--user` argument like @junkblocker mentions... does python look wherever this would install things by default or do you have to set environment variables to support user installed packages first?

Let me know when you do settle on whatever package does the conversion to html... I'd like to keep `hs.doc.hsdocs` in sync if possible so our docs look the same... or we could include the HTML in the application itself as well (we wouldn't need the SQLLite database that Dash does) and forgo the live conversion that currently occurs.

cmsj:
@asmagill I'm doing no dependency installation at any point, it's up to the user to do, if they want to build that part - same as is currently true with the Ruby dependencies listed in `scripts/docs/Gemfile`.

I'm not against including the HTML docs, although it would mean putting the jinja2 dependency on anyone doing local builds.

As for pip - you're right that Apple doesn't include it, although they do include easy_install which can do the work of installing dependencies, just not as well as pip.

cmsj:
I don't want to seem unsympathetic, I do feel the pain of unusual Python setups on macOS, but I tend to lean on the side of saying that anyone who wants to do local builds of Hammerspoon (a pretty advanced thing to be doing to a pretty advanced tool) who also has a weird Python setup, ought to be capable of coping with it, or at least should be pretty easy to help if they're stuck.

@asmagill something I'd like to do with the HTML docs is make them a lot prettier and more functional, which lends support to bundling them in the app for hsdocs. Are there going to be any issues there with loading included CSS/JavaScript files?

asmagill:
No issues if the pages and resources (javascript, etc.) are all provided.  Currently hsdocs converts the markdown with the same code that the ruby tool uses (the C portion of the relevant ruby module is included as `hs.doc.markdown`) but if we had a set of static files which could be served directly, it would be even simpler.  I'd probably keep the markdown conversion module as part of Hamemrspoon -- I also use it to visually test local documents I intend to host at GitHub -- but it would no longer be necessary.

cmsj:
Awesome. I'm not really sure what I'm going to do with the HTML yet, I'm so badly out of date on web stuff. I'm looking at Bootstrap, but it's a disappointingly large amount of CSS/JS/fonts.

junkblocker:
@cmsj, I am fine with docs build being considered a advanced use case and thus modules required as a prerequisite for that, just shouldn't try to install modules for the user during it. Thanks!

cmsj:
@junkblocker I'm not going to install any modules, nor have we ever.

I think there is some merit to the idea of shipping the HTML docs in the app, so I can't promise that we'll avoid needing module installs for regular builds, forever. As I said above, I'm sympathetic to people whose Python setup has become unmanageable, but I'm not at all convinced that we should avoid adding value to Hammerspoon because those people may find it hard to do a local build.

cmsj:
@asmagill so, this branch should now be at feature-parity with the old docs generator, and is fully integrated into the Makefile. I would appreciate any testing you can do with the bits of yours that use docs.json (note that you'll need to do something like ```sudo -H pip install -r requirements.txt``` in the top level directory).

cmsj:
What I'd like to do here is get this branch fixed up so it's as close to 100% compatible with the old generator as is humanly possible, then land it.

From there, I want to start pulling in stuff like: https://github.com/Hammerspoon/hammerspoon/pull/1157/commits/d66dd9adab2806e98b5c6cfd2b9314ef881fd585

(which splits out the Parameters/Returns/Notes data into their own sections, and is trivially extensible to add an Example section. It also adds in a link from each API item to its source code, because why not :)

asmagill:
@cmsj, an initial build with this pull seems to work well with both the built in text documentation and web browser versions... I'll run some more tests tonight, but I think it's fine.

cmsj:
This has now been merged.

Issue #1149, opened 2016-12-20
`hs.chooser` possible improvements


aldur:
Yesterday I was [poking around](https://aldur.github.io/articles/hammerspoon-emojis/) with `hs.chooser` and I believe it has great potential.

On this topic, I'd like to propose some improvements, as:
- Fuzzy searching.
- Hiding empty rows while searching (i.e. reducing the height of the window).
- Ability to specify additional search keywords that do not need to appear in the UI.
- "Modal" items: i.e. items that, once selected, clear the current choices and present other, related, options.

Some of the improvements could probably be achieved through current API:
e.g. hiding empty rows by dynamically changing `hs.chooser:rows([numRows])`, or modal choices.
I haven't tried this yet, however.

What do you guys think?

cmsj:
An option to enable fuzzy searching seems like a good idea, and I'm mildly sure that the modal idea could be achieved with an option that keeps the chooser open after a selection has been made, allowing the code to set new choices and clear the search field (which would also need some additional API).

I'm not sure I understand what the benefit of hidden search terms would be - do you have something specific in mind for that?

Hiding empty rows is probably a good idea, although I suspect probably the hardest of these things to implement.

latenitefilms:
The benefit I can see of "hidden" search field, would be the ability to "favourite" things. For example, maybe you have two effects called "Blur Alpha" and "Blur Motion" - but really when you're typing "bl...", you always REALLY want the "Blur Motion" option, so that comes to the top of the list, even though it's not alphabetically at the top. Or can this already be done with the existing code?

latenitefilms:
In terms of "modal" items - @aldur, I'm doing [something similar](https://github.com/fcpxhacks/fcpxhacks/blob/develop/src/hs/fcpxhacks/modules/hacksconsole.lua) to this currently, by having different "modes" in the chooser. Basically, you can turn on a "Remove from List" mode via right-clicking on the chooser, that allows you to delete items from the chooser, and a "Restore to List" to get items back. However, with the next update to `hs.chooser`, I'll be able just right click specific chooser items and delete them directly without having "modes", which will be pretty handy too. "Modes" still have their use though, as they allow you to delete multiple things really quickly. Anyway - food for thought.

aldur:
[...]

PR #1159, opened 2016-12-24
Let `hs.chooser` match macOS configured theme (light/dark)

aldur:
By default, the chooser now matches macOS preferences (and also changes its color on settings updates).
In case the user explicitly sets the color, it sticks to it, ignoring the preferences.

@cmsj, @asmagill: It's been a while since I last used Objective C, so advices/reviews/thoughts are welcome.

### DIFF
--- a/extensions/chooser/HSChooser.h
+++ b/extensions/chooser/HSChooser.h
@@ -55,6 +55,9 @@
 // A pointer to the hs.chooser module's references table
 @property(nonatomic) int *refTable;
 
+// Keep track of whether we are observing macOS interface theme (light/dark)
+@property(nonatomic) BOOL isObservingThemeChanges;
+
 // Initialiser
 - (id)initWithRefTable:(int *)refTable completionCallbackRef:(int)completionCallbackRef;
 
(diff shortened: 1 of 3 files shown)


aldur:
Partially refers to #1149.

asmagill:
A possible suggestion for an addition:  if an explicit `nil` is passed into `hs.chooser:bgDark`, turn the observer back on... probably need to add a property to track whether the observer is active or not so it doesn't create a new observer if one is already in place.

I might move the removal of the observer from a dealloc method into the `userdata_gc` function since we don't care about it anymore once Lua/Hammerspoon forgets about the object, but that's more of a style issue... either way works fine in this context.

Otherwise, looks good to me!

aldur:
I was implementing the explicit `nil` to turn the observer back on, but then I noticed that this would change previous behaviour (returning the current setting):

```
/// hs.chooser:bgDark([beDark]) -> hs.chooser object or boolean
...
/// Parameters:
///  * beDark - A optional boolean, true to be dark, false to be light. If this parameter is omitted, the current setting will be returned
...
/// Returns:
///  * The `hs.chooser` object or a boolean, true if the window is dark, false if it is light
```

What do you think about it?

aldur:
Nevermind, an explicit `nil` is a different case, not an omitted parameter. :)

aldur:
Rebased against current `master` and implemented the explicit `nil`/observer property.

cmsj:
Thanks @aldur. I made a few changes on the way in, but nothing super dramatic :)

PR #1165, opened 2016-12-26
Add JIS key codes

rizumita:
Add yen / underscore / keypad comma / eisu / kana key codes.

### DIFF
--- a/extensions/keycodes/internal.m
+++ b/extensions/keycodes/internal.m
@@ -28,6 +28,7 @@ int keycodes_cachemap(lua_State* L) {
         kVK_ANSI_Grave, kVK_ANSI_Equal, kVK_ANSI_Minus, kVK_ANSI_RightBracket,
         kVK_ANSI_LeftBracket, kVK_ANSI_Quote, kVK_ANSI_Semicolon, kVK_ANSI_Backslash,
         kVK_ANSI_Comma, kVK_ANSI_Slash, kVK_ANSI_Period, kVK_ISO_Section,
+        kVK_JIS_Yen, kVK_JIS_Underscore, kVK_JIS_KeypadComma, kVK_JIS_Eisu, kVK_JIS_Kana,
     };
 
     // NOTE: It appears that TISCopyCurrentKeyboardInputSources() can return NULL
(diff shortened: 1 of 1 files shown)


cmsj:
@asmagill I think you understand this stuff more than I do. This seems like a reasonable change to me - can you think of any problems with it? There may also be some ISO related codes that can be added.

extensions/keycodes/internal.m
```diff
@@ -168,6 +169,12 @@ int keycodes_cachemap(lua_State* L) {
     pushkeycode(L, kVK_DownArrow, "down");
     pushkeycode(L, kVK_UpArrow, "up");
 
+    pushkeycode(L, kVK_JIS_Yen, "¥");
+    pushkeycode(L, kVK_JIS_Underscore, "underscore");
+    pushkeycode(L, kVK_JIS_KeypadComma, "padcomma");
```
asmagill: Is there a reason to label this "padcomma" instead of "pad,"?  Other typographical's on the keypad are done with the character itself and not the name (e.g. "pad.")

asmagill:
Looks fine to me.  I've been meaning to review `hs.keycodes` and see about adding some items that I've come across that we're currently not including... iirc this addresses the primary ones I was thinking about which correspond to actual keys -- I can't test them since I don't have an international keyboard, but code wise it looks good.

There are a few other labels that I may add later as well once I figure out the proper eventtap sequences for us to reliably detect/generate events for things like brightness, volume, etc. (unfortunately it's not as simple as just sticking the keycode number into `hs.eventtap.event.newKeyEvent`) but  I think this pull takes care of the labels corresponding to actual physical keys.

rizumita:
@asmagill I changed padcomma to `pad,` because I thought it was suited too.

asmagill:
closing and reopening to force travis to rebuild with new travis settings

tanakahisateru:
@rizumita This is what just I wanted! Just a thing what I suggest is `"yen"` instead of `"¥"` for font compatibility.

cmsj:
@tanakahisateru agreed, that seems like a good change. I've merged this PR with ¥ switched to `yen`. Thanks everyone!

Issue #1164, opened 2016-12-26
hs.alert centre text


latenitefilms:
If you have multiple lines of text, it would be great if you could centre things within a `hs.alert`.

Maybe something that can be added to `hs.alert.defaultStyle`?

asmagill:
You can do this yourself by adding the following line of code into your `init.lua` file:

~~~lua
require("hs.alert").defaultStyle.paragraphStyle = { alignment = "center" }
~~~

Unless there is a differing consensus, I'm reluctant to make this the default behavior since it will differ from the historical behavior of `hs.alert` which predates the current model which supports this type of style modification.

asmagill:
Ok, correction... this *won't* work... yet.

I forgot that I merged the text and the box styles into one table and don't use it as a straight `hs.styledtext` table... I'll get a fix up tonight that will allow this to work... sorry for the confusion!

latenitefilms:
Thanks @asmagill !

PR #1167, opened 2016-12-27
Alert with more style

asmagill:
Adds support for messages which are already `hs.styledtext` objects and adds the `textStyle` key to the `hs.alert.defaultStyle` table to support more styles when providing `message` as a string.

E.g.  To center the text of an alert you could do one of the following:

~~~lua
-- this will make all alerts with multiline text center each line independently:
hs.alert.defaultStyle.textStyle = { paragraphStyle = { alignment = "center" } }

-- this will do so for just this one alert:
hs.alert("my multi-line\nmessage", { textStyle = { paragraphStyle = { alignment = "center" } } })
~~~

See https://github.com/asmagill/hammerspoon_core/tree/master/alert if you want to try it out without doing a full rebuild.

Addresses #1164


latenitefilms:
Perfect! Works great. Thanks @asmagill !

cmsj:
@asmagill is this landable? it seems ok from a quick read.

PR #1172, opened 2016-12-29
Change how keycodes are mapped

### LINKED ISSUE #1169: Possible bug identified in keycodes_cachemap when binding keys with dead-key mappings
I have a configuration where the mapping for key code '41' is incorrect.

I have a .keylayout file with the following entry:

```
<keyMapSet id="ANSI">
		<keyMap index="0">
                         <!-- Lots of other bindings -->
                        <key code="41" action="o"/>
```

In the action mappings I have:

```
      <actions>
                <!-- Lots of action mappings -->
		<action id="o">
			<when state="none" output="o"/>
			<when state="Dead Key State 0" output="ó"/>
		</action>
```

In Hammerspoon init.lua:

`print(41,'=', hs.keycodes.map[41])`

Hammerspoon Console:

`41	=	ó`

I noticed this when trying to bind "o":

```
*** ERROR: ...oon.app/Contents/Resources/extensions/hs/hotkey/init.lua:409: Invalid key: o - this may mean that the key requested does not exist in your keymap (particularly if you switch keyboard layouts frequently)
stack traceback:
	[C]: in function 'error'
	...oon.app/Contents/Resources/extensions/hs/hotkey/init.lua:23: in upvalue 'getKeycode'
	...oon.app/Contents/Resources/extensions/hs/hotkey/init.lua:190: in function 'hs.hotkey.new'
	...oon.app/Contents/Resources/extensions/hs/hotkey/init.lua:409: in method 'bind'
	/Users/Attensi/.hammerspoon/init.lua:25: in main chunk
	[C]: in function 'xpcall'
	...app/Contents/Resources/extensions/hs/_coresetup/init.lua:481: in function 'hs._coresetup.setup'
	(...tail calls...)

```

In trying to identify the cause of the issue I've looked at `hammerspoon/extensions/keycodes/internal.m`

In the `keycodes_cachemap` method, the keyTranslateOption argument in the`UCKeyTranslate` calls might be incorrect from what I can gather from https://developer.apple.com/reference/coreservices/1390584-uckeytranslate?language=objc and http://stackoverflow.com/questions/27735217/how-to-use-uckeytranslate

```
for (int i = 0 ; i < (int)(sizeof(relocatableKeyCodes)/sizeof(relocatableKeyCodes[0])) ; i++) {
            UCKeyTranslate(keyboardLayout,
                           relocatableKeyCodes[i],
                           kUCKeyActionDisplay,
                           0,
                           LMGetKbdType(),
                           kUCKeyTranslateNoDeadKeysBit,
                           &keysDown,
                           sizeof(chars) / sizeof(chars[0]),
                           &realLength,
                           chars);

            const char* name = [[NSString stringWithCharacters:chars length:1] UTF8String];

            pushkeycode(L, relocatableKeyCodes[i], name);
        }
```

It seems that `kUCKeyTranslateNoDeadKeysMask` should be used instead of `kUCKeyTranslateNoDeadKeysBit`?

And possibly also `kUCKeyActionDown` should be used instead of `kUCKeyActionDisplay` ?

https://developer.apple.com/reference/coreservices/1390619-key_actions?language=objc

P.S. I'm relatively green with regards to Hammerspoon and custom .keylayout files on macOS.

sindrijo:
Using mask kUCKeyTranslateNoDeadKeysMask  instead of flag
kUCKeyTranslateNoDeadKeysBit

https://developer.apple.com/reference/coreservices/1390568-key_translati
on_options_mask?language=objc

Also using kUCKeyActionDown instead of kUCKeyActionDisplay which is
said to be unreliable
http://stackoverflow.com/questions/27735217/how-to-use-uckeytranslate

Fixes #1169

### DIFF
--- a/extensions/keycodes/internal.m
+++ b/extensions/keycodes/internal.m
@@ -43,10 +43,10 @@ int keycodes_cachemap(lua_State* L) {
         for (int i = 0 ; i < (int)(sizeof(relocatableKeyCodes)/sizeof(relocatableKeyCodes[0])) ; i++) {
             UCKeyTranslate(keyboardLayout,
                            relocatableKeyCodes[i],
-                           kUCKeyActionDisplay,
+                           kUCKeyActionDown,
                            0,
                            LMGetKbdType(),
-                           kUCKeyTranslateNoDeadKeysBit,
+                           kUCKeyTranslateNoDeadKeysMask,
                            &keysDown,
                            sizeof(chars) / sizeof(chars[0]),
                            &realLength,


asmagill:
closing and reopening to force travis to rebuild with new travis settings

cmsj:
I'm not sure I 100% understand this change, possibly because I use a fairly simple British keyboard layout, but it seems reasonable. I'm going to merge it and we can see if anyone notices any issues :)

Thanks!

sindrijo:
@cmsj I was having trouble mapping some keys on a custom keyboard layout I have (modified colemak), I think the fact that the code was working before for most other keys was a lucky coincidence.

The usage of `kUCKeyTranslateNoDeadKeysBit` was incorrect as it was not shifted, correct usage would have been `1 << kUCKeyTranslateNoDeadKeysBit` which is equivalent to `kUCKeyTranslateNoDeadKeysMask` according to a comment in the stack-overflow link.

Issue #1163, opened 2016-12-26
Crash on startup when non-latin input method activated


Zert:
Hammerspoon crashes when non-latin input method activated (also had this problem with Mjolnir)

```
Welcome to the Hammerspoon Console!
You can run any Lua code in here.

-- Lazy extension loading enabled
-- Loading ~/.hammerspoon/init.lua
-- Loading extension: fnutils
-- Loading extension: hotkey
*** ERROR: ...oon.app/Contents/Resources/extensions/hs/hotkey/init.lua:342: Invalid key: i - this may mean that the key requested does not exist in your keymap (particularly if you switch keyboard layouts frequently)
stack traceback:
	[C]: in function 'error'
	...oon.app/Contents/Resources/extensions/hs/hotkey/init.lua:23: in upvalue 'getKeycode'
	...oon.app/Contents/Resources/extensions/hs/hotkey/init.lua:190: in function 'hs.hotkey.internal.new'
	...oon.app/Contents/Resources/extensions/hs/hotkey/init.lua:342: in function 'hs.hotkey.internal.bind'
	/Users/zert/.hammerspoon/init.lua:13: in local 'fn'
	...on.app/Contents/Resources/extensions/hs/fnutils/init.lua:57: in function 'hs.fnutils.map'
	/Users/zert/.hammerspoon/init.lua:41: in main chunk
	[C]: in function 'xpcall'
	...app/Contents/Resources/extensions/hs/_coresetup/init.lua:481: in function 'hs._coresetup.setup'
	(...tail calls...)
```

Steps to reproduce:
1. Quit Hammerspoon if already running
2. Open any application with input area, iTerm2 for example
3. Switch to non-latin input method, like Greek
4. Run Spotlight by hotkey
5. Input method in Spotlight is also Greek, so switch to English
6. Write `hammerspoon` and press Enter
7. You will see a window with the error message pasted above

This crash reproduced in 100% cases for any non-latin input method (Greek, Russian, Serbian, etc).

asmagill:
This occurs because `hs.hotkey` is actually internally converting the "i" into the corresponding keycode by looking it up in `hs.keycodes.map`.  When the Greek layout is enabled, "i" is not a valid key:

~~~lua
> hs.keycodes.currentLayout(), hs.keycodes.map["i"], hs.keycodes.map[34]
U.S.	34	i

> hs.keycodes.currentLayout(), hs.keycodes.map["i"], hs.keycodes.map[34]
[...]

PR #1194, opened 2017-01-19
fallback to ANSI Standard US Keyboard layout in map

asmagill:
includes __index metatable to `hs.keycodes.map` to fallback to ANSI standard US keyboard layout if character isn't found.

ANSI fallback table is hard coded because `/System/Library/Frameworks/Carbon.framework/Versions/Current/Frameworks/HIToolbox.framework/Versions/Current/Headers/Events.h` file is only present if the user has Xcode installed; if Apple does change this (unlikely, but not impossible) we may need to update it in the future.

### DIFF
--- a/extensions/keycodes/init.lua
+++ b/extensions/keycodes/init.lua
@@ -2,6 +2,66 @@
 ---
 --- Convert between key-strings and key-codes. Also provides funcionality for querying and changing keyboard layouts.
 
+-- fallback table based on ANSI-Standard US Keyboard as defined in /System/Library/Frameworks/Carbon.framework/Versions/Current/Frameworks/HIToolbox.framework/Versions/Current/Headers/Events.h as of macOS 10.12, Xcode 8.
+local fallbackKeyMap = {
+    ["0"]        = 0x1d,  ["1"]        = 0x12,  ["2"]        = 0x13,  ["3"]        = 0x14,
+    ["4"]        = 0x15,  ["5"]        = 0x17,  ["6"]        = 0x16,  ["7"]        = 0x1a,
+    ["8"]        = 0x1c,  ["9"]        = 0x19,  ["a"]        = 0x00,  ["b"]        = 0x0b,
+    ["\\"]       = 0x2a,  ["c"]        = 0x08,  [","]        = 0x2b,  ["d"]        = 0x02,
+    ["e"]        = 0x0e,  ["="]        = 0x18,  ["f"]        = 0x03,  ["g"]        = 0x05,
+    ["`"]        = 0x32,  ["h"]        = 0x04,  ["i"]        = 0x22,  ["j"]        = 0x26,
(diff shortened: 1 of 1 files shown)


asmagill:
Should usage of fallback table be optional?  This can easily be done with a setting stored in the hammerspoon defaults.  If so, default on or default off (when the key is missing from the defaults, as would be the initial case)?

asmagill:
addresses #1163

randomeizer:
What's the downside of leaving it on by default?

cmsj:
lgtm

cmsj:
(I would probably go for having it on by default)

asmagill:
So, should I add a way to disable the fallback, but make it an opt out?

@randomeizer, the downside is that if you're trying to assign to a key which does not exist in your current layout, it will print a message to the console but assign the callbacks to whatever the corresponding key location in the ANSI standard layout and continues executing your startup code. An error causes execution to halt and is generally noticeable almost immediately.

It's a trade off... I suppose I could add code so that it also uses `hs.notify` so that a visual indicator occurs even if the console isn't currently open.  Thoughts?

randomeizer:
I guess question is, is it an error to use the ANSI codes? If so, a visual
warning might be appropriate. Otherwise printing to the console at 'warn'
or 'info' would be preferred.

cmsj:
@asmagill let's try with it just always on and see what the reaction is. If we can avoid the opt-out, that's one fewer conditional branch to look after :)

As for the warning, I think the console is fine, it would be consistent with everything else we warn/error about.

asmagill:
The changes are all in Lua, and it currently just uses `print` to output the possible messages:

~~~lua
    ...
    print("-- hs.keycodes.map:key '" .. tostring(key) .. "' not found in active keymap; using ANSI-standard US keyboard layout as fallback, returning " .. tostring(newKey))
    ...
    print("-- hs.keycodes.map:key '" .. tostring(key) .. "' not found in active keymap or ANSI-standard US keyboard layout")
    ...
~~~

The first is the one most likely to occur -- it indicates that the fallback table was used.  The second indicates that even the fallback table didn't recognize the key and then returns nil (resulting in the original behavior when a key couldn't be identified, usually an error in the calling code).

I could add an `hs.logger` instance for `hs.keycodes` or use `hs.luaSkinLog` (which is the logger instance used by the LuaSkin log methods) if you think that's better... it would provide a timestamped history retrievable with `hs.logger.printHistory`, but I suspect few people actually use `hs.console.clearConsole` very often, so I don't know that it's worth the extra code.

Unless you think we should change to using logger, this is ready to go.

randomeizer:
I would kind of prefer a logger, and the ability to set the printed log
level, since we may be totally fine with using ANSI. If it's not a big, why
clog the console?

asmagill:
You can suppress the warnings from automatically printing to the console with `hs.keycodes.log.level = 1` and you can display them at will with `hs.logger.printHistory(nil, nil, "hs.keycodes")`.

PR #1195, opened 2017-01-19
allow referencing of canvas element by id in lua code

asmagill:
e.g. for

~~~lua
c = require("hs.canvas")
a = c.new{ x = 100, y = 100, h = 100, w = 100 }:show()
a[1] = { type = "rectangle", id = "part1", fillColor = { blue = 1 } }
a[2] = { type = "circle", id = "part2", fillColor = { green = 1 } }
~~~

You can refer to `a[2]` as `a.part2`.  Useful when you don't know the index of an element or the indexes might change because of things being inserted or removed.


cmsj:
I have no objections to this, but I'm not very familiar with hs.canvas, so I defer to you about when it's ready to land :)

PR #1198, opened 2017-01-20
Pass the menu item table to the item callback.

deryni:
This lets the callback know what menu item was clicked, store extra
info in the item table for the callback, etc.

### DIFF
--- a/extensions/menubar/internal.m
+++ b/extensions/menubar/internal.m
@@ -98,6 +98,7 @@ typedef NS_ENUM(NSInteger, NSStatusBarItemPriority) {
 @interface HSMenubarCallbackObject : NSObject
 @property lua_State *L;
 @property int fn;
+@property int item;
 @end
 @implementation HSMenubarCallbackObject
 // Generic callback runner that will execute a Lua function stored in self.fn
@@ -135,7 +136,9 @@ typedef NS_ENUM(NSInteger, NSStatusBarItemPriority) {
         lua_pushboolean(L, isFnKey);
         lua_setfield(L, -2, "fn");
 
-        fn_result = [skin protectedCallAndTraceback:1 nresults:1];
+        [skin pushLuaRef:refTable ref:self.item];
+
+        fn_result = [skin protectedCallAndTraceback:2 nresults:1];
     } else {
         // event is very unlikely to be nil, but we'll handle it just in case
         fn_result = [skin protectedCallAndTraceback:0 nresults:1];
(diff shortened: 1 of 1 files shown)


deryni:
The impetus for this change was for an SSH host menu I was working on for my [sshchooser](https://github.com/deryni/sshchooser) module.

Currently I need to use a [callback generator function](https://github.com/deryni/sshchooser/blob/master/sshchooser.lua#L249-L253) to generate unique callbacks for each menu item.

With this change I could have a single function instead.

```lua
local function menu_cb(flags, tab)
    return sshfns[sshfn](tab.title)
end
```

deryni:
Given that `fn` is in the table being stored in the delegate would it make sense to stop storing `fn` in the delegate directly and just use it from the stored table?

(That would mean that modifications to the table would change the function the callback called which might be problematic for hammerspoon but I don't know what the general feeling is on that sort of thing.)

asmagill:
I like the idea... I too prefer when the callback function has enough information in its arguments that it doesn't need to rely on upvalues or use a wrapper function, but I have a couple of questions:

What happens when the menu items are generated by a function rather then as a set table?  It looks like it might work ok, since the function will return a table and that's what your addition is working from, but I haven't tried it out to make sure the stack indexes line up.

And when do the table items you're pushing onto the registry get released?  Unless I'm missing something, I don't see a corresponding `luaUnref:ref:` in the new code.

cmsj:
@asmagill good point about the function scenario.

@deryni did your testing of this include the situation where the menu table is returned by a function, rather than statically defined?

deryni:
@asmagill You didn't miss anything. I didn't see where the `fn` ref was being cleaned up in my original looking and so didn't add anything for it. I think I see it now (`erase_menu_items`, right?) so I'll add the unref there.

@cmsj No, I didn't test that case but I can do that. I agree with @asmagill's assessment that it should work out just fine though as `parse_table` is just being handed the index of the table to parse and the relative indexes on top of that should work out the same way in either case.

deryni:
Some quick testing looks like the function case works just fine.

I also ran some tests to ensure the unref was working and while I did stop seeing the size of the menubar's `refTable` stop growing I didn't actually see a memory leak without the unref (though I suppose maybe I didn't push it hard enough but I did see the memory climb as a rapidly clicked the function-based menu to generate garbage tables but then it would drop back down as soon as I stopped clicking).

Having some computer issues but I'll get the unref change pushed up as soon as I can.

PR #1203, opened 2017-01-22
added print table to core

z3t0:
just added a function `hs.printTable(t)` which recursively prints out the table `t`. This should be **extremely** helpful in debugging.


cmsj:
Does this offer any specific advantages over hs.inspect()?

z3t0:
Oh was not aware of that.

PR #1213, opened 2017-01-26
Add getDuration() to itunes and spotify extensions.

drn:
I've been manually triggering applescript tells when pulling current duration in some of my scripts. I thought it would be useful as part of the extensions.

One oddity was that when testing, iTunes returns duration in seconds while Spotify returns duration in milliseconds even though the Spotify.sdef defines `duration (integer, r/o) : The length of the track in seconds.


extensions/itunes/init.lua
```diff
@@ -281,6 +281,20 @@ function itunes.setPosition(p)
...
+--- Gets the duration (in seconds) of the current song
+---
+--- Parameters:
+---  * None
+---
+--- Returns:
+---  * The number of seconds long the current song is, 0 if no song is playing
+function itunes.getDuration()
+  local duration = tonumber(tell('duration of current track'))
+  return duration ~= nil and duration or 0
```
z3t0: Just my 2c but would it be better ti return `nil` instead of `0` ? Also what if the song is actually at `0`. Just something that came to mind...

drn: I can't think of a situation where a song will ever be 0 seconds long - seems like it that would be a broken file, no?

The reason why I converted the nil to a 0 was so that the method always returned a consistent type. That means that I could operate directly with `hs.itunes.getDuration()` without having to worry about exceptions. Like: `local minutes = hs.itunes.getDuration() / 60`

Happy to change it if you think returning nil is better though!

z3t0: No it's fine your explanation makes perfect sense... not sure what I was thinking. It's pretty late here so I misread the function as returning the current time of the song

cmsj:
@aldur could you check if Deezer also supports this AppleScript interface? If so we should add it there too.

drn:
I've never heard of Deezer before, but I downloaded it and it does - I'll add it / test it
screen shot 2017-01-26 at 2 02 03 pm

Side note - I checked out Vox yesterday and it does not have an interface for this.

drn:
@cmsj - I've updated the commit to include support for deezer

drn:
@cmsj - do you need me to do anything else before this gets merged in?

PR #1218, opened 2017-01-31
Fix accessing nil hs.application:name

aldur:
I noticed that after the `hs.spotlight` addition I was getting some tracebacks in the console related to `hs.application.find` and an attempt to address a nil variable (sorry, I recompiled HS and did not copy the logs).

After a bit of digging I found the culprit: `hs.application:name` seems to be nil in some cases, and this PR should fix the issue.

On a broader note: is `hs.application:name` allowed to be nil by definition? Is this an issue rooted elsewhere?
Pinging @asmagill for this one.

### DIFF
--- a/extensions/application/init.lua
+++ b/extensions/application/init.lua
@@ -148,8 +148,12 @@ function application.find(hint,exact)
   else for _,a in ipairs(apps) do local aname=a:name() if aname and aname:lower():find(hint:lower()) then r[#r+1]=a end end end
 
   if spotlightEnabled then
-      for i, v in ipairs(table.pack(realNameFor(hint, exact))) do
-          for _, a in ipairs(apps) do if v:lower() == a:name():lower() then r[#r+1]=a end end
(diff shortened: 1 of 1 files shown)


asmagill:
I'd be interested in knowing what application returns `nil` for the name, but looking at the underlying objective-c method which returns the name, `[NSRunningApplication localizedName]`, there is no prohibition in the documentation or header file about its return value being non-null, so... without more info I can't say whether this is "normal" acceptable behavior or indicative of something else.

I did once notice the spotlight code generating a lot of nil results when I had to kill an `mds` process to unmount a network share, but haven't had a chance to recreate the situation to figure out exactly what was going on.  Not sure if this is related, but thought I'd put it out there in case anyone has any ideas.

The code change looks good to me.

aldur:
@asmagill Yes, I was interested into finding it out too, but `hs.inspect` did not reveal anything interesting. As for the reproducibility: it happened to me a few times over the day, so if you any ideas on the "how" I can dig deeper.

Issue #1210, opened 2017-01-24
Regression in function hs.wifi.watcher in 0.9.52


lavir:
The Hammerspoon v.0.9.52 cannot init the hs.wifi.watcher

Message from MacOS (10.11.6) logs:
`24/01/17 08:27:55,460 Hammerspoon[1046]: (missing delegate):log level 2: hs.wifi.watcher:initManager unable to register for event type scanCacheUpdated: The operation couldn’t be completed. (com.apple.wifi.request.error error 4.)
`

It unable to register for all types of events, not only for 'scanCacheUpdated'

cmsj:
@asmagill at a glance, I'm wondering if this is a 10.11 vs 10.12 thing?

asmagill:
According to the docs, `CWEventTypeScanCacheUpdated` has been around since 10.10.  @lavir, is your configuration file available online or can you post the relevant section here?  I may have access to a 10.11 machine I can test on later this week.

lavir:
If you meant my lua script, then the relevant part is below:
```
local cbWifi = function ()
    net = hs.wifi.currentNetwork()
    print("wifi" ..net)

    if net==nil then
        hs.notify.show("wifi disconnected","","","")
    else
        hs.notify.show("wifi connected","",net,"")
    end
end

wifiwatcher = hs.wifi.watcher.new(cbWifi)
wifiwatcher:start()
```
Also I did fresh install of MacOS week ago, and then moved Hammerspoon.app and prefs from backup. Maybe this also affected on this issue. I'll clean up prefs and other files which are relates to Hammerspoon. 
I'll come back if I'll notice some changes.

lavir:
Ok, I did clean up. Didn't help.
BTW I have noticed a message in MacOS Logs
`25/01/17 08:55:46,953 airportd[62]: ERROR: Hammerspoon (29427) is not entitled for com.apple.wifi.events, will not register for event type 6
`
I have allowed access the Hammerspoon in Accessibility Prefs

z3t0:
Pretty sure this is a [bug](https://forums.developer.apple.com/thread/11307)

asmagill:
[...]

PR #1225, opened 2017-02-07
hs.wifi.watcher: use old style notifications for CWInterface

asmagill:
Addresses #1210


cmsj:
Shame we need this, but +1

Issue #1238, opened 2017-02-23
hs.host.cpuUsage() memory leak and invalid idle reporting


cmsj:
Per some other examples that use host_processor_info(), we seem to be missing a call to vm_deallocate(), I think.

See https://gist.github.com/mmisono/351441

asmagill:
It's entirely possible... iirc, the original source of the code this grew from was a cmd line utility and I've noticed that these often don't clean up properly after themselves assuming that the normal task tear-down will take care of it for them.

asmagill:
I was thinking of the memory statistics... the cpuUsage stuff comes from http://stackoverflow.com/questions/6785069/get-cpu-percent-usage and they do perform a vm_dealloc in the code snippet; sorry I missed that!

asmagill:
Changing this to address google issue at https://groups.google.com/forum/#!topic/hammerspoon/SbhLIQoW9Zw as well.

A side by side comparison of top versus hs.host.cpuUsage confirms that our code is indeed wrong... the samples I've seen all show a differential between runs because `host_processor_info` reports everything in ticks, and I *thought* that as long as we calculated the percentages against the current total, it would work out... and my earlier tests were just single snapshots against what was a fairly idle machine so the numbers were actually close to what top reported.  A continuous test (see https://github.com/asmagill/hammerspoon-config/blob/master/_scratch/idleWatcher.lua if your curious about the code I'm using now) shows that our function does not track properly at all.

I've downloaded Apple's top implementation source code and will see exactly how they're doing it since this seems to be what most of us (myself included) are comparing against.

asmagill:
Ok thinking about it now, that was pretty stupid of me... what its currently reporting is the percentage the cpu(s) have spent in such states since the machine was started up, not a snapshot in time... The tick counts are constantly being incremented and never decremented.

Got a fix in mind and came across `host_statistics` which looks like it might result in cleaner code and take care of the dealloc for us.  I should have a pull up this weekend.

Top uses a 1 second delay between samples, which is too long for Hammerspoon to block for something like this... I'm thinking 1/10 of a second (initial tests suggest that this gives good results that don't jump around *too* much -- smaller delays tend to jump quite a bit) but I'll add a parameter to the function so you can decide on a case by case basis if you're willing to block longer) unless someone has other thoughts on what the delay between samples should be by default?

asmagill:
[...]

PR #1246, opened 2017-03-04
properly sample cpu ticks with `hs.host.cpuUsage`

asmagill:
Fixes `hs.host.cpuUsage` to perform a proper sample over a specified period of time, both blocking and non-blocking versions provided.  Also addresses memory leak identified in #1238 and code cleaned up to reduce compile time warnings.

- - -

hs.host.cpuUsage([period], [callback]) -> table
Function
Query CPU usage statistics for a given time interval using [hs.host.cpuUsageTicks](#cpuUsageTicks) and return the results as percentages.

Parameters:
 * `period`    - an optional value specifying the time between samples collected for calculating CPU usage statistics.
   * If `callback` is not provided, this is an optional integer, default 100000, specifying the number of microseconds to block between samples collected.  Note that Hammerspoon will block for this period of time during execution of this function.
   * If `callback` is provided, this is an optional number, default 1.0, specifying the number of seconds between samples collected.  Hammerspoon will *not* block during this time period.
 * `callback` - an optional callback function which will receive the cpu usage statistics in a table, described below, as its sole argument.

Returns:
 * If a callback function is not provided, this function will return a table containing the following:
   * Individual tables, indexed by the core number, for each CPU core with the following keys in each subtable:
     * user   -- percentage of CPU time occupied by user level processes.
     * system -- percentage of CPU time occupied by system (kernel) level processes.
     * nice   -- percentage of CPU time occupied by user level processes with a positive nice value. (See notes below)
     * active -- For convenience, when you just want the total CPU usage, this is the sum of user, system, and nice.
     * idle   -- percentage of CPU time spent idle
   * The key `overall` containing the same keys as described above but based upon the average of all cores combined.
   * The key `n` containing the number of cores detected.
 * If a callback function is provided, this function will return a placeholder table with the following metamethods:
   * `hs.host.cpuUsage:finished()` - returns a boolean indicating if the second CPU sample has been collected yet (true) or not (false).
   * `hs.host.cpuUsage:stop()`     - abort the sample collection.  The callback function will not be invoked.
   * The results of the cpu statistics will be submitted as a table, described above, to the callback function.

Notes:
 * If no callback function is provided, Hammerspoon will block (i.e. no other Hammerspoon activity can occur) during execution of this function for `period` microseconds (1 second = 1,000,000 microseconds).  The default period is 1/10 of a second. If `period` is too small, it is possible that some of the CPU statistics may result in `nan` (not-a-number).

 * For reference, the `top` command has a default period between samples of 1 second.

 * The subtables for each core and `overall` have a __tostring() metamethod which allows listing it's contents in the Hammerspoon console by typing `hs.host.cpuUsage()[#]` where # is the core you are interested in or the string "overall".

- - -

hs.host.cpuUsageTicks() -> table
Function
Returns a table containing the current cpu usage information for the system in `ticks` since the most recent system boot.

Parameters:
 * None

Returns:
 * A table containing the following:
   * Individual tables, indexed by the core number, for each CPU core with the following keys in each subtable:
     * user   -- number of ticks the cpu core has spent in user mode since system startup.
     * system -- number of ticks the cpu core has spent in system mode since system startup.
     * nice   --
     * active -- For convenience, when you just want the total CPU usage, this is the sum of user, system, and nice.
     * idle   -- number of ticks the cpu core has spent idle
   * The key `overall` containing the same keys as described above but based upon the combined total of all cpu cores for the system.
   * The key `n` containing the number of cores detected.

Notes:
 * CPU mode ticks are updated during system interrupts and are incremented based upon the mode the CPU is in at the time of the interrupt. By its nature, this is always going to be approximate, and a single call to this function will return the current tick values since the system was last rebooted.
 * To generate a snapshot of the system's usage "at this moment", you must take two samples and calculate the difference between them.  The [hs.host.cpuUsage](#cpuUsage) function is a wrapper which does this for you and returns the cpu usage statistics as a percentage of the total number of ticks which occurred during the sample period you specify when invoking `hs.host.cpuUsage`.

 * Historically on Unix based systems, the `nice` cpu state represents processes for which the execution priority has been reduced to allow other higher priority processes access to more system resources.  The source code for the version of the [XNU Kernel](https://opensource.apple.com/source/xnu/xnu-3789.41.3/) currently provided by Apple (for macOS 10.12.3) shows this value as returned by the `host_processor_info` as hardcoded to 0.  For completeness, this value *is* included in the statistics returned by this function, but unless Apple makes a change in the future, it is not expected to provide any useful information.

 * Adapted primarily from code found at http://stackoverflow.com/questions/6785069/get-cpu-percent-usage

### DIFF
--- a/extensions/host/init.lua
+++ b/extensions/host/init.lua
@@ -7,6 +7,7 @@
 
 local host = require "hs.host.internal"
 local fnutils = require "hs.fnutils"
+local timer   = require "hs.timer"
 
 local __tostring_for_tables = function(self)
     local result = ""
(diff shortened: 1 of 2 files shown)


cmsj:
Slightly convoluted retrofitting of usefulness to the existing API call, but it seems sane enough :)
+1 from me.

PR #1249, opened 2017-03-07
Enable scrolling with Ctrl-P/N/V and PageUp/PageDown in chooser

knu:
Adding these shortcuts allows you to move through choices using the keyboard even more easily.

### DIFF
--- a/extensions/chooser/HSChooser.m
+++ b/extensions/chooser/HSChooser.m
@@ -100,6 +100,12 @@
 
     [self addShortcut:@"Up" keyCode:NSUpArrowFunctionKey mods:NSFunctionKeyMask|NSNumericPadKeyMask handler:^{ [_self selectPreviousChoice]; }];
     [self addShortcut:@"Down" keyCode:NSDownArrowFunctionKey mods:NSFunctionKeyMask|NSNumericPadKeyMask handler:^{ [_self selectNextChoice]; }];
+    [self addShortcut:@"p" keyCode:-1 mods:NSControlKeyMask handler:^{ [_self selectPreviousChoice]; }];
+    [self addShortcut:@"n" keyCode:-1 mods:NSControlKeyMask handler:^{ [_self selectNextChoice]; }];
+
+    [self addShortcut:@"PageUp" keyCode:NSPageUpFunctionKey mods:NSFunctionKeyMask handler:^{ [_self selectPreviousPage]; }];
+    [self addShortcut:@"PageDown" keyCode:NSPageDownFunctionKey mods:NSFunctionKeyMask handler:^{ [_self selectNextPage]; }];
+    [self addShortcut:@"v" keyCode:-1 mods:NSControlKeyMask handler:^{ [_self selectNextPage]; }];
 }
 
 - (void)windowDidResignKey:(NSNotification *)notification {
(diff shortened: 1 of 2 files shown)


asmagill:
In the interests of parity, what is the opposite to Ctrl-V (in emacs, right?)?

Otherwise, I don't see a problem with adding this.  @cmsj?

knu:
@asmagill It's Meta+V in Emacs, but conflicts with the key for pasting (⌘V).  My personal experience tells that you don't need scroll back as much as you need to scroll forward, so I added it.  Maybe it should be assigned to something else (like Ctrl+Shift+V), or it could be just dropped for consistency.

asmagill:
Agree with not changing ⌘V, and in theory Alt-V could be something someone might want to search for in the text (Alt-V == √)

I'm ok with Ctrl-Shift-V or leaving it off entirely, which ever you think would be most obvious/expected by users.  I don't think Ctrl-V needs to be removed, though... while I couldn't remember the page-back key sequence for Emacs, I (and I assume many like me) will know ctrl-v and try it, so...

The other approach might be to add the ability to add/remove hotkeys and then leave a few defaults but let each user setup the rest the way they like.  But that's probably a bit more than you intended with this pull and can be added later if someone doesn't like or wants to change these.

latenitefilms:
Awesome work @knu - the page up/down shortcuts are a great idea, that'll come in very handy! Thank you!!

For what it's worth, I agree with @asmagill that it would be great if `hs.chooser` had some default shortcut keys built-in, but they could be customised or disabled if required. My thought with Hammerspoon is that things like shortcuts keys should never be forcefully "baked in". Ideally you should always be able to change things like this using Lua, as opposed to having to re-compile Hammerspoon yourself.

knu:
@asmagill @latenitefilms Thank you for sharing your thoughts.  Actually the shortcut keys added by this PR are meant to be part of the reasonable built-in default bindings.

Seems the UI is already customizable by Hammerspoon itself, by adding hotkeys for the chooser right before calling `chooser:show()` and disabling them in `completionFn`.  However, I've found one problem; you can call `chooser:selectedRow(row)` to move the selection, but the scroll position does not follow even if the selection moves out of the visible area.  So, once it is fixed and a callback for adding hotkeys is added (maybe `showCallback` or something) you can say with confidence that key bindings are fully customizable.

cmsj:
knu: if the list isn't scrolling to selected rows, that's a bug. Would you mind filing an issue for that?

knu:
@cmsj I added a commit to fix it.

Here's an example configuration for reproduction.  The following code enables a hotkey Cmd+Ctrl+[.] that opens a chooser to input the selected text to the frontmost window, in which you can navigate between choices up and down using Ctrl+[.]/[,].  You can see hs.chooser:selectedRow(row) does not update the scroll position even if the new selection is out of visible area.
```lua
function withRepeat(fn)
  return fn, nil, fn
end

do
  local choices = {}
  for i = 1,20 do
    table.insert(choices, { text = "choice"..i })
  end
  local hotkeys, window
  local chooser = hs.chooser.new(function (choice)
      hs.fnutils.each(hotkeys, function (hotkey) hotkey:disable() end)
      window:focus()
      if choice then
        hs.eventtap.keyStrokes(choice.text)
      end
  end)
  chooser:searchSubText(true)
  chooser:choices(choices)

  hotkeys = {
    hs.hotkey.new({"ctrl"}, ",", withRepeat(function ()
          local row = chooser:selectedRow()
          if row >= #choices then
            chooser:selectedRow(1)
          else
            chooser:selectedRow(row + 1)
          end          
    end)),
    hs.hotkey.new({"ctrl"}, ".", withRepeat(function ()
          local row = chooser:selectedRow()
          if row <= 1 then
            chooser:selectedRow(#choices)
          else
            chooser:selectedRow(row - 1)
          end          
    end)),
  }

  hs.hotkey.bind({"cmd", "ctrl"}, ".", function ()
      window = hs.window.focusedWindow()
      hs.fnutils.each(hotkeys, function (hotkey) hotkey:enable() end)
      chooser:show()
  end)
end
```

knu:
I've filed #1264 which makes it easier for user to customize key bindings in the chooser UI.

cmsj:
thanks :)

PR #1251, opened 2017-03-09
Adopt new Vox Applescript dictionary (fixes #1250)

### LINKED ISSUE #1250: Vox now uses `player state` instead of `playerState`
Seems like new versions of Vox use `player state` instead of `playerState`
this makes the `.getPlayerState` method return `nil`.

Could we please update:
```
Resources/extensions/hs/vox/init.lua
296:  return tell('playerState')
```

to:
```
Resources/extensions/hs/vox/init.lua
296:  return tell('player state')
```

Thanks! 🙂

aldur:
This breaks backwards compatibility with previous versions of Vox.
If it is worth it we can manage to check for Vox's version and then use the correct Applescript dictionary string.

### DIFF
--- a/extensions/vox/init.lua
+++ b/extensions/vox/init.lua
@@ -293,7 +293,7 @@ end
 ---  * 0 for paused
 ---  * 1 for playing
 function vox.getPlayerState()
-  return tell('playerState')
+  return tell('player state')
 end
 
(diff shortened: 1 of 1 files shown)


aldur:
Closing this to leave space to other contributors too (#1256).

PR #1264, opened 2017-03-22
Add methods to hs.chooser for user customization

knu:
This adds some methods to hs.chooser for user customization as suggested in #1249.

- `hs.chooser:showCallback([fn])`: Called when show() is called, and can be used as a hook where user can add custom hotkeys.
- `hs.chooser:select([row])`: Closes the chooser by selecting a row.
- `hs.chooser:cancel()`: Cancels the chooser.

This also includes a fix for a bug where `hs.chooser:selectedRow(row)` does not update the scroll position.

### DIFF
--- a/extensions/chooser/HSChooser.h
+++ b/extensions/chooser/HSChooser.h
@@ -47,6 +47,7 @@
 @property(nonatomic, retain) NSArray *filteredChoices;
 
 // Lua callback references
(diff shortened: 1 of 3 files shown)


latenitefilms:
Awesome!!

cmsj:
thanks :)

PR #1266, opened 2017-03-24
Improve pasteboard management

knu:
Currently hs.pasteboard.writeDataForUTI() always clears the pasteboard before writing given data, so you cannot store multiple values with differnt UTIs.

This PR adds an optional flag `add` which allows you to optionally "add" a new value in addition to existing values with different UTIs.

New utility functions readAllData() and writeAllData() are also added so you can easily save and restore what's in the pasteboard.

### DIFF
--- a/extensions/pasteboard/init.lua
+++ b/extensions/pasteboard/init.lua
@@ -12,4 +12,49 @@ require("hs.sound")
 require("hs.styledtext")
 require("hs.drawing.color")
 
+-- Public interface ------------------------------------------------------
+
+--- hs.pasteboard.readAllData([name]) -> table
+--- Function
+--- Returns all values in the first item on the pasteboard in a table that maps a UTI value to the raw data of the item
+---
+--- Parameters:
+---  * name - an optional string indicating the pasteboard name.  If nil or not present, defaults to the system pasteboard.
+---
+--- Returns:
+--   a mapping from a UTI value to the raw data
+module.readAllData = function (name)
+  local contents = {}
+  for _, uti in ipairs(module.contentTypes(name)) do
+    contents[uti] = module.readDataForUTI(name, uti)
+  end
+  return contents
+end
+
+--- hs.pasteboard.writeAllData([name], table) -> boolean
+--- Function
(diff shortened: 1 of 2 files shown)


latenitefilms:
Awesome.

cmsj:
lgtm, paging Dr @asmagill for a second opinion :)

asmagill:
lgtm.

For consistency, this should probably be added to `hs.pasteboard.writePListForUTI` and `hs.pasteboard.writeArchiverDataForUTI` as well.

asmagill:
Just had a thought but don't have time to test it myself... what happens if you attempt to append to a pasteboard that Hammerspoon doesn'tt currently have ownership of?  (The `clearContents` method assigns Hammerspoon as the "owner" of the pasteboard, which is why we've never really worried about it before)

asmagill:
(should say I don't have time right now... I can later tonight if no one else does first)

knu:
Oh, if that is a problem, writeAllData() should become a primitive instead of allowing for adding one by one, I guess?

asmagill:
The docs for NSPasteboard's  `setData:forType:` describe the following for the return value:

> Return Value
>
> YES if the data was written successfully, otherwise NO if ownership of the pasteboard has changed. Any other error raises an NSPasteboardCommunicationException.

And I just confirmed, we do return the boolean result, so maybe an addition to the docs indicating that `false` return value most likely means that the pasteboard has changed ownership and that you will have to start over since your previous additions are no longer there should  be sufficient.

knu:
@asmagill Thanks for your feedback, including the detailed investigation.  I've added the `add` parameter to other write* functions and updated the documentation.

cmsj:
@asmagill can I ask you to merge this in when you're happy with it please? :)

knu:
I'd much appreciate if this could be merged by the next release.  I desperately need a way to save and restore the contents of pasteboard.

cmsj:
@knu I would like to do a release fairly soon, but there is no fixed timeline, and we can wait for this to land. It looks fine to me, but I'd like a +1 from @asmagill since he knows more about the pasteboard code than I do.

PR #1281, opened 2017-03-28
hs.noises use internal queue for internal callback

asmagill:
Per docs for AudioQueueNewInput:

> inCallbackRunLoop	
> The event loop on which the callback function pointed to by the inCallbackProc parameter is to be called. If you specify NULL, the callback is called on one of the audio queue’s internal threads.

More reliable sound detection when Hammerspoon is under load.

### DIFF
--- a/extensions/noises/internal.m
+++ b/extensions/noises/internal.m
@@ -103,7 +103,7 @@ void AudioInputCallback(void * inUserData,  // Custom audio metadata
   status = AudioQueueNewInput(&recordState.dataFormat,
                               AudioInputCallback,
                               (__bridge void *)self,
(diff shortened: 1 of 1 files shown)


cmsj:
Any objections from @trishume to this?

trishume:
Seems good to me, that was the first (and still the only) OSX audio code I ever wrote, I don't claim to have known what I was doing.

asmagill:
@trishume, I stumbled across this by accident and was happy with the results... I'm trying to do as much as possible in alternate queues and threads to keep Hammerspoon responsive with large configurations (like mine), so this was a fortuitous find.

PR #1286, opened 2017-03-29
Added hs.uploadCrashData()

latenitefilms:
- Added a hs.uploadCrashData([state]) function

### DIFF
--- a/Hammerspoon/MJLua.m
+++ b/Hammerspoon/MJLua.m
@@ -63,6 +63,21 @@ void MJLuaSetupLogHandler(void(^blk)(NSString* str)) {
 
 @end
 
+/// hs.uploadCrashData([state]) -> bool
+/// Function
+/// Set or display the "Upload Crash Data" status for Hammerspoon.
+///
+/// Parameters:
+///  * state - an optional boolean which will set whether or not Hammerspoon should be upload crash data via Crashlytics.
+///
+/// Returns:
+///  * True if Hammerspoon is currently (or has just been) set to upload crash data or False if Hammerspoon is not.
+static int core_uploadCrashData(lua_State* L) {
+    if (lua_isboolean(L, -1)) { HSSetUploadCrashData(lua_toboolean(L, -1)); }
+    lua_pushboolean(L, HSUploadCrashData()) ;
+    return 1;
+}
+
 /// hs.autoLaunch([state]) -> bool
 /// Function
 /// Set or display the "Launch on Login" status for Hammerspoon.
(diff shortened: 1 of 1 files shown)


cmsj:
I'm slightly curious what the rationale is for adding this to the API - are we expecting people to want to toggle the crash uploading frequently?

(Also if we do add it to the API, I think it should have a note that asks people to please leave crash reporting on if at all possible, because it helps us a lot)

latenitefilms:
Really, it's for my selfish own purposes, where I'm controlling all the Hammerspoon preferences through my own Preferences Panel. I just figured, if I need, maybe someone else will later down the line.

No stress at all if you don't want to include this - feel free to close if you feel it's not appropriate.

cmsj:
@latenitefilms it's a tough call. I like exposing things in the API, I just want to keep getting as many crash reports as possible ;)

@asmagill any thoughts here?

latenitefilms:
I really doubt adding this to the API will give you LESS crash reports. Users are far more likely to just un-click the "Send crash data" option when the Preferences panel popup appears after installing and running, than looking through the API documentation, realising you can disable the crash logs, then entering than code in their `init.lua` and executing it. Feels like way too many steps than just ticking an option on a popup!

But like I said, no stress at all if you don't think it's appropriate or necessary.

asmagill:
How hard would it be to make the toggle temporary (i.e. until the next time the Hammerspoon application is reloaded or restarted)?  I suspect restarted may be easier, but even so... that way one could disable logging for a portion that you know is buggy/being worked on during development without permanently changing one's preference to send bug reports.

Conversely if I normally *don't* send crash reports (for whatever reason), I can temporarily turn them on when I want to generate a crash report so we can get it and look into a problem a user may be having with something.

I too like having options... so my gut says to add it, but... I also understand your argument, @cmsj, so... I'm fine either way.  We can always make it a hidden function or third party module that overrides the built in `hs.crash` module if the user installs it into `~/.hammerspoon/hs/crash`, though then we have to keep it current with future updates to Crashlytics or new functions that may be added... (or put it in a different module... perhaps `hs.crash.advanced`)

cmsj:
Ok, if we can have a note in there encouraging people to please upload crashlogs if they can, and remove the errant "be" in the parameter doc, this can land :)

latenitefilms:
@cmsj - Awesome, thank you! I've updated the in-line documentation. Feel free to tweak as necessary!

cmsj:
thanks :)

Issue #1239, opened 2017-02-24
accessibilityState Watcher?


latenitefilms:
Any chance of adding a `accessibilityState` watcher to Hammerspoon? It would be great to know if a user (or the OS) has turned off the `accessibilityState` mid-session.

I guess you could just use a `hs.timer` but I assume a proper "watcher" would be less CPU intensive?

latenitefilms:
Thoughts @asmagill & @cmsj ?

cmsj:
I don't have a problem with doing this. We're doing it in the core of Hammerspoon anyway.

latenitefilms:
Awesome - thanks @cmsj !

latenitefilms:
@cmsj Sorry - I've tried to give this a bash myself, but unfortunately it's a bit out of my skill-set, so if anyone else is able to do, that would be amazing!

@asmagill or @knu ?

PR #1288, opened 2017-03-30
Added hs.accessibilityStateCallback

latenitefilms:
Added `hs.accessibilityStateCallback` to address #1239.

Please note, I really have absolutely no idea what I'm doing in Xcode, so this code definitely needs reviewing.


latenitefilms:
Pinging @cmsj & @asmagill for review.

asmagill:
To make it more obvious what the function actually does (and does not) do, I might suggest adding something like the following to the documentation for `hs.accessibilityStateCallback`:

~~~
---
--- Notes:
--- * The function will not receive any arguments when called.  To check what the accessibility state has been changed to, you should call [hs.accessibilityState](#accessibilityState) from within your function.
~~~

(The \[...\]\(...\) structure allows the documentation generator to create links and the style above should work since these two items will be documented in the same html/docset file.)

latenitefilms:
@asmagill - Great idea! I've updated the commit.

Issue #1293, opened 2017-03-31
hsdocs launches a server open to the world


junkblocker:
Only figured this when shodan came knocking and LittleSnitch alerted me. The documentation gives the impression that it is listening on local interfaces only when it is not true. ``hsminweb`` documentation is cautious enough but ``hsdocs`` doesn't really connect the two. Also, rather than or maybe in addition to an ``accessList`` to do filter based deny, I'd prefer an option to listen on ``::1/127.0.0.1``. ``hsdocs`` doesn't currently give me a public API to do access control either.

asmagill:
i don't think the current hs.httpserver allows us to bind to a specific interface/address; hs.httpserver.hsminweb, which the docs are based on, does support an access filter which can block/allow by ip address -- which i thought i had enabled but apparently not. i'll get this added as the default and add a function for toggling/editing it.

cmsj:
@junkblocker you have your Mac on a public IP? Brave :)

junkblocker:
@cmsj, well, it's an old laptop I thought was running minimal stuff sufficiently protected but yeah :) I should switch over to the linux box.

cmsj:
Fixed (in that we can now get/set the interface. It's still going to listen on all interfaces by default)

asmagill:
looks like hsdocs will still need to invoke this (and this call is simpler then the access list specification i was considering).

if no one beats me to it, i can do it later tonight.

junkblocker:
Same for hsminweb.

asmagill:
As far as I can tell from looking at the server code, only one interface can be specified, so it's all of them or just one.  @cmsj, can you confirm or correct this, since you know the code better than I do?

Questions:

* should `hs.httpserver.hsminweb` default to localhost interface only?  Since `hs.httpserver` defaults to open, I'm leaning towards this also defaulting to open since `hsminweb` is just a more standards compliant wrapper to `hs.httpserver`.

* should `hs.doc.hsdocs` default to localhost interface only?  Since this is a specific application of `hs.httpeserver.hsminweb` and is for Hammerspoon documentation, I'm leaning towards it being restricted by default.

* the docs do mention that the interface needs to be set before the web server is started; however `setInterface` will allow you to set an interface after the server has started without warning (and `getInterface` will report the changed value.  Should a warning be issued?  At least we should probably add to the `setInterface` docs that if you change the interface on a running server, you will need to stop and restart it before the change goes into effect.

[...]

PR #1297, opened 2017-04-01
add `interface` to hsminweb and hsdoc

asmagill:
further addresses #1293 

Note that this pull requires a build from the current (as of its submission) master -- the latest formal release does not support `hs.httpserver.setInterface` so this pull will not work for that even though the changes in this pull are only to lua files.

Adds

>hs.httpserver.hsminweb:interface([interface]) -> hsminwebTable | current-value
>Method
>Get or set the network interface that the hsminweb web server will listen on
>
>Paramaters:
> * interface - an optional string, or nil, specifying the network interface the web server will listen on.  An explicit nil specifies that the web server should listen on all active interfaces for the machine.  Defaults to nil.
>
>Returns:
> * the hsminwebTable object if a parameter is provided, or the current value if no parameter is specified.
>
>Notes:
> * See `hs.httpserver.setInterface` for a description of valid values that can be specified as the `interface` argument to this method.
> * the interface can only be specified before the hsminweb web server has been started.  If you wish to change the listening interface for a running web server, you must stop it with [hs.httpserver.hsminweb:stop()](#stop) before invoking this method and then restart it with [hs.httpserver.hsminweb:start](#start).

and

>hs.doc.hsdocs.interface([interface]) -> currentValue
>Function
>Get or set the network interface that the Hammerspoon documentation web server will be served on
>
>Paramaters:
> * interface - an optional string, or nil, specifying the network interface the Hammerspoon documentation web server will be served on.  An explicit nil specifies that the web server should listen on all active interfaces for the machine.  Defaults to "localhost".
>
>Returns:
> * the current, possibly new, value
>
>Notes:
> * See `hs.httpserver.setInterface` for a description of valid values that can be specified as the `interface` argument to this function.
> * A change to the interface can only occur when the documentation server is not running. If the server is currently active when you call this function with an argument, the server will be temporarily stopped and then restarted after the interface has been changed.

The difference in behavior between `hs.httpserver.hsminweb:interface` and `hs.doc.hsdocs.interface` when a server is active (i.e. already started) is intended -- `hsminweb` aims to give you complete control over a (mostly) standards compliant web server; the internal documentation is intended to be available to the newest of users and should be as easy to access and change as possible.

This can be altered if preferred.

### DIFF
--- a/extensions/doc/hsdocs/init.lua
+++ b/extensions/doc/hsdocs/init.lua
@@ -359,6 +359,48 @@ end
 
 -- Public interface ------------------------------------------------------
 
+--- hs.doc.hsdocs.interface([interface]) -> currentValue
+--- Function
+--- Get or set the network interface that the Hammerspoon documentation web server will be served on
+---
+--- Paramaters:
+---  * interface - an optional string, or nil, specifying the network interface the Hammerspoon documentation web server will be served on.  An explicit nil specifies that the web server should listen on all active interfaces for the machine.  Defaults to "localhost".
+---
+--- Returns:
+---  * the current, possibly new, value
+---
+--- Notes:
+---  * See `hs.httpserver.setInterface` for a description of valid values that can be specified as the `interface` argument to this function.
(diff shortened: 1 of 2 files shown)


cmsj:
Closes #1293

PR #1301, opened 2017-04-03
Add modifier keycodes to hs.keycodes.map

knu:
Does it make sense to add these?

### DIFF
--- a/extensions/keycodes/init.lua
+++ b/extensions/keycodes/init.lua
@@ -77,7 +77,8 @@ end
 ---     pad0, pad1, pad2, pad3, pad4, pad5, pad6, pad7, pad8, pad9,
 ---     padclear, padenter, return, tab, space, delete, escape, help,
 ---     home, pageup, forwarddelete, end, pagedown, left, right, down, up,
----     capslock, shift, rshift, cmd, rcmd, alt, ralt, ctrl, rctrl, fn
+---     shift, rightshift, cmd, rightcmd, alt, rightalt, ctrl, rightctrl,
+---     capslock, fn
 
 local keycodes = require "hs.keycodes.internal"
 keycodes.map = attachFallbackTable(keycodes._cachemap())
--- a/extensions/keycodes/init.lua
+++ b/extensions/keycodes/init.lua
@@ -76,7 +76,8 @@ end
 ---     f16, f17, f18, f19, f20, pad, pad*, pad+, pad/, pad-, pad=,
 ---     pad0, pad1, pad2, pad3, pad4, pad5, pad6, pad7, pad8, pad9,
 ---     padclear, padenter, return, tab, space, delete, escape, help,
----     home, pageup, forwarddelete, end, pagedown, left, right, down, up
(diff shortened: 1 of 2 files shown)


knu:
A typical use case for these codes is to check which key is actually pressed for a flagsChanged event and only react to the right or left shift key.

asmagill:
My thoughts:

* For consistency with the way we name things, `rcmd`, `rshift`, `ralt`, and `rctrl` should be camel cased.  Further, my preference would be to spell out right (e.g. `rightCmd` instead of `rCmd`, etc.)
* The docs either within `hs.keycodes.map` or `hs.hotkey.bind` should make it clear that these specific keymappings cannot be used in hotkey binding -- they are strictly for use with `hs.eventtap`

Since a future update I am working on for `hs.eventtap` was going to introduce these anyways, I have no objection to adding them now.

knu:
@asmagill OK, thanks for sharing your thoughts and a future plan.  I just named them after the existing mnemonics such as "forwarddelete", "pageup" and "padenter" which don't use upper case letters.  Perhaps we should make them all camel cased and make lookup case-insensitive by using a metatable, so common mistakes to spell one letter key in upper case would be covered.  What do you think?

cmsj:
@knu I like the sound of that :)

asmagill:
forgot that hs.keycodes.map is one of the few places we haven't properly camel-cased yet... sounds like a good compromise, so 👍

knu:
Oh wait, there's reverse lookup also.  Returning the camel-cased name  for a keycode might break user code.  While case-insensitivity may be great, should we keep compatibility and have lower-case names in the map for now?

cmsj:
@knl you make a good point, let's not break things. Sorry for the mixed signals on this!

knl:
@cmsj I think you meant @knu :)

cmsj:
@knl sorry :)

asmagill:
Ok... then I think the question becomes `ralt` or `rightalt`... I favor the later, but either is fine.

I may consider amending the metafunctions later to allow referencing them by name in a camelCased fashion because I think it looks and reads better, but I agree that the reverse should stay the same lowercased version for backwards compatibility.

knu:
@asmagill My bad, the forward lookup (name -> keycode) was already made case-insensitive.
https://github.com/Hammerspoon/hammerspoon/blob/67eb825ee6753f6a77b13b0017a48d6f886f3e35/extensions/keycodes/init.lua#L49-L52

So I just changed the `r`- names to `right`- to suit your tastes.

knu:
I thought for a moment of making the new keys camel-cased like `rightCmd`, but that would introduce inconsistency users would hardly welcome, so why not just leave the consistency argument for now and revisit until the keycodes/hotkey framework gets a revamp?

cmsj:
@knu it looks like, at the moment, some of the key names in the docstring comment, don't match the names that are going in the table? (e.g. "rcmd" vs "rightcommand")

knu:
@cmsj Thanks, fixed!

cmsj:
Thanks :)

Issue #1098, opened 2016-11-24
Update system image list for 10.12 in hs.image


asmagill:
Just noticed that a bunch of new image names were added in 10.12... should be added to hs.image's system image list at some point.

cmsj:
+1

Is this something we could script?

asmagill:
They're all from NSImage.h, so... probably, but not sure if its something we can do at compile time or if its something we'll have to do occasionally by hand...

PR #1302, opened 2017-04-03
add 10.12 images to hs.images.systemImageNames

asmagill:
addresses #1098


cmsj:
@asmagill the CI failure here looks like a legit crash relating to pushing the image table, FYI.

asmagill:
sigh... let me guess... travis compiles with Xcode 8 but runs the tests in a 10.11 environment?  Guess it's better to find out now that I'm going to have to wrap all of the constants to verify existence then to crash for users not using 10.12... ok, I'll see if I can get to it tonight or tomorrow.

Issue #1177, opened 2017-01-01
Feature Request - Better AppleScript/Automator Integration


latenitefilms:
It would be amazing if Hammerspoon had better integration with AppleScript & Automator.

For example, it would be amazing if Hammerspoon had it's own AppleScript Dictionary, and if you could manipulate this dictionary via LUA code.

For example, the Terminal App has it's own "Terminal Suite" of AppleScript commands, such as `do script`. What I would like to do is use Hammerspoon to define my own set of AppleScript Commands, which trigger LUA functions.

That way, I could build up my own Automator workflows in Automator, but they actually really just trigger Hammerspoon functions.

Is this remotely possible? @cmsj or @asmagill ?

nikitavoloboev:
Yeah I wold really love to have something like [external triggers](https://www.alfredapp.com/help/workflows/triggers/external/) in hammer spoon. Right now the only way to call lua bindings from hammer spoon is using hotkeys.

latenitefilms:
@nikitavoloboev - You can actually tell Hammerspoon what to do via [URL events](http://www.hammerspoon.org/docs/hs.urlevent.html), or [CLI](http://www.hammerspoon.org/docs/hs.ipc.html) as well, but I do think an AppleScript interface would be really handy too.

cmsj:
I like the idea of this, although before people get too excited, I would note that the AppleScript dictionary for an app has to be defined statically as an XML file in the .app bundle, and is fairly rigid, so we'll have to do something pretty much exactly like `hs.urlevent` - we'll offer up a single AppleScript command (maybe `trigger` :) and it will have to take probably just one string argument, possibly with URL encoding so it can be split up later. I haven't looked at how return values work, but again I think the declaration is pretty static ahead of time, so we may just have to return a string and the  user can decide what they want to pack into it.

latenitefilms:
@cmsj - Maybe this is TOO crazy, but I was thinking we could just manipulate the dictionary files that are WITHIN the Hammerspoon bundle? For example, if I wanted to add a new AppleScript Action to the Dictionary, we could just modify the `.scriptSuite` and `.scriptTerminology` XML files within the Application Bundle to add this new Action.

However, if modifying the contents of the Hammerspoon Bundle is getting too dodgy, then I'd be happy with just having some un-changable Hammerspoon AppleScript Commands.

Maybe something like like this to start...

```
Hammerspoon Suite > application > triggerFunction([globalFunction])
[...]

PR #1306, opened 2017-04-04
Add AppleScript Dictionary to Hammerspoon

latenitefilms:
This is just an initial proof of concept, as discussed in #1177.


latenitefilms:
@cmsj - As discussed in #1177, I've thrown together a proof of concept that allows you to `execute` Lua code from AppleScript. I really know nothing about Objective-C or Xcode, so this is very much a copy and paste job. I think this kind of functionality could be really useful though, so I'd love yours and @asmagill brains to have a look and see what magic you can add to this!

cmsj:
@latenitefilms do you have an example of what it looks like in applescript to execute some code in Hammerspoon?

cmsj:
One thing that slightly concerns me here is that this is loaded unconditionally, which gives the user no way to block AppleScript from being able to execute arbitrary things with Hammerspoon. It's not a huge concern because if someone can execute malicious AppleScript, they can do more powerful things than even Hammerspoon can do, but it's worth considering nonetheless :)

latenitefilms:
@cmsj - No worries! I've added a `hs.appleScript()` function which allows you to enable/disable the AppleScript `executeLua` command (with the intention for this to also enable/disable future AppleScript commands we add). I've also added some example code to this Hammerspoon function, so that you can play/test the AppleScript functionality yourself. Here's a very basic example too:

```
tell application "Hammerspoon"
	executeLua "print [[This will write stuff to the Hammerspoon Console]]"
end tell
```

The usual disclaimer - I know next-to-nothing about Objective-C or Xcode, so this is again, very much just a copy and paste job.

I assume you can just modify stuff in my branch?

I've tried to add a `dockIcon` property as well, but haven't got that working yet. Once I work out how to do this, I'm hoping to "expose" most of the Core Hammerspoon functionality via AppleScript.

I also still haven't worked out how to actually return data back to AppleScript. Maybe this is something you can help with @cmsj or @asmagill ? Here's an [example](https://www.objc.io/issues/14-mac/scripting-data/) of how it can be done. This [documentation](https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/ScriptableCocoaApplications/SApps_about_apps/SAppsAboutApps.html#//apple_ref/doc/uid/TP40001976) might also be helpful?

latenitefilms:
For anyone playing at home - I've updated the code.

You can learn more about how it works, and see some example AppleScript code, by entering the following command into the Hammerspoon console:

`help("hs.allowAppleScript")`

@cmsj & @asmagill - This branch as it currently stands should be all good for `0.9.53`. However, the question now is how much of the Hammerspoon API do we want to "expose" through AppleScript (especially given `execute lua code` can basically trigger anything you can do in Hammerspoon anyway). I would still like to expose more of the Core Hammerspoon functionality to keep things nice and clean - but it's not really urgent or a priority.

cmsj:
FWIW your branch seems to also include some parts of your CommandPost fork. Since this is a fairly small PR I can merge it in by hand, but it's something to watch out for :)

latenitefilms:
Opps - apologies! I think I must have accidentally synced the wrong way at some point.

I can't work out how to undo the last commit, but [this](https://github.com/latenitefilms/hammerspoon/tree/4d677ab304178866e496b4976435a5b6678eea47) "version" should be correct?

cmsj:
So reading through this - the question about what to expose directly to AppleScript...

From the point of view of keeping the implementation of this as simple as possible in Hammerspoon, I can't help but think that having just one command to execute Lua, would be a winner - we expose so much core functionality of the app to Lua, that people will be able to do anything they want with that.

I get that it's cleaner to have other things exposed to AppleScript directly, but is it worth the maintenance overhead of having more ObjC code around?

latenitefilms:
Fine by me!

Let me know if you need me to do anything to close up this issue.

latenitefilms:
Legend, thanks @cmsj !

PR #1312, opened 2017-04-08
hs.urlevent Tweaks

latenitefilms:
- Replaced `print` messages with `hs.logger`
- Scheme is now taken from bundle identifier
- Bundle ID is no longer hardcoded in `internal.m`


latenitefilms:
@cmsj - This is probably not relevant to anyone but me, so feel free to close if you don't think it's worthwhile adding to Hammerspoon.

PR #1317, opened 2017-04-11
add hs.host.locale submodule

asmagill:


knu:
Nice!  This will make it possible to conditionalize language/locale specific tweaks and greatly help make your Hammerspoon work universally available rather than keeping it as local hack!

Issue #1196, opened 2017-01-19
hs.alert doesn't display on correct screen in certain display configurations


latenitefilms:
If you have your Displays arranged where one is above the other (as opposed to next to one another - see attached screenshot), currently `hs.alert` doesn't display the message on the correct screen.

Pinging @asmagill - as I know you've been working on `hs.alert` recently.

<img width="780" alt="screen shot 2017-01-20 at 12 58 35 am" src="https://cloud.githubusercontent.com/assets/22286696/22109393/a86afbe2-deab-11e6-9acc-7285e21161e0.png">

latenitefilms:
Reckon we could add this to the 0.9.53 wish-list @cmsj ?

asmagill:
I'm going to try looking into this tomorrow, but just so I know what I'm trying to fix:

The virtual menubar in your setup above is in the lower monitor, correct?

And what behavior are you expecting versus seeing?  According to the alert code, it *should* be displaying the alert on the screen with the currently focused window... if it's doing that but isn't what you expect, then we can consider offering an option to always use the primary display; if it *isn't* doing that, then I need to dig into why...

latenitefilms:
I'm away from my laptop currently, but from memory, if I have an external display ABOVE my laptop screen, and had the console active on this above screen, alerts were still appearing on the laptop screen and not the "active" external display, which is not what I expected to happen based on the documentation. Will test again tonight.

latenitefilms:
Attached is my Display Arrangement.

**macOS:** 10.12.3
**Hammerspoon:** 0.9.52
**Hardware:** MacBook Pro (Retina, 15-inch, Early 2013)

If I have the Hammerspoon Console open in the top display, and enter `hs.alert("test")` it will appear on the bottom screen (even though my mouse is on the top screen, and so is the console (which is the "Active Window").

Same result if I enter `hs.alert.show("Test", nil, hs.screen.mainScreen(), 3)`.

Same result if I enter `hs.alert.show("Test", nil, hs.screen.primaryScreen(), 3)`.

Same result if I enter `hs.alert.show("Test", nil, hs.screen.allScreens()[1], 3)`.

Same result if I enter `hs.alert.show("Test", nil, hs.screen.allScreens()[2], 3)`.

I'd guess that maybe [this code](https://github.com/Hammerspoon/hammerspoon/blob/master/extensions/alert/init.lua#L99) has something to do with it?
 
[...]

PR #1318, opened 2017-04-11
hs.alert: handle multiple screens better

asmagill:
Address #1196 

It's going to be at least a day before I can fully test this on a multi-monitor setup, but I think this adds the appropriate logic that was completely missing before (my bad!).  @latenitefilms if you get a chance to check it out first let me know.


latenitefilms:
Legend, thanks mate! Seems to work great for me!

@cmsj - Merge away!

PR #1320, opened 2017-04-11
Added Callback for when Text or Files are dropped onto the Hammerspoon Dock Icon

latenitefilms:
Added `hs.textDroppedToDockIconCallback(value)` and `hs.fileDroppedToDockIconCallback(value)` which allow users to be notified when a file or text is dragged onto the Hammerspoon Dock Icon.


latenitefilms:
@cmsj & @asmagill - This all seems to work fine on my machine. Feel free to tweak anything around as needed.

Thanks so much for all your help and advice pulling this together @cmsj !!

latenitefilms:
Legend, thanks @cmsj !!

cmsj:
@latenitefilms did you notice that I added in a Service for sending files to Hammerspoon too? :)

latenitefilms:
I did - VERY appreciative of all your code improvements! Lots of things for me to keep in mind next time I do something similar. Thank you!

nonissue:
Maybe I should open an issue, but this functionality doesn't work on my machines and from a quick search of the hammerspoon repo, I can't find any mention of either the  hs.textDroppedToDockIconCallback or hs.dockIconClickCallback methods. When I run either in my console, I get:
```
> hs.textDroppedToDockIconCallback
nil
```

```
> hs.textDroppedToDockIconCallback("test")
[string "return hs.textDroppedToDockIconCallback("test..."]:1: attempt to call a nil value (field 'textDroppedToDockIconCallback')
stack traceback:
	[string "return hs.textDroppedToDockIconCallback("test..."]:1: in main chunk
	[C]: in function 'xpcall'
	...app/Contents/Resources/extensions/hs/_coresetup/init.lua:417: in function <...app/Contents/Resources/extensions/hs/_coresetup/init.lua:397>
```

Am I missing something silly @latenitefilms?

latenitefilms:
@nonissue - `hs.textDroppedToDockIconCallback` is a optional variable.

So if you want to use it, you'd do something like this:

```lua
hs.textDroppedToDockIconCallback = function(value)
print(string.format("Text dropped to dock icon: %s", value))
end)
```

nonissue:
Oh amazing, that totally works. Thank you!

PR #1327, opened 2017-04-13
Make checking return value of hs.eventtap.event:getFlags() easier

knu:
It hasn't been a very easy task to check return value of `hs.eventtap.event:getFlags()`, because it is a table of key-values which cannot be simply tested against another list of values in Lua, even if you use `hs.fnutils`.

For example, to tell if pressed modifiers are exactly "cmd" and "shift", you have to say something like this:
```lua
local flags = e:getFlags()
if flags.cmd and flags.shift and not flags.ctrl and not flags.alt and not flag.fn then
  -- Listing all of existing modifiers..
end
```
Which is way too far from optimal.

This PR adds two methods to return value of hs.eventtap.event:getFlags().

- contain(mods) -> boolean

  Returns true if the modifiers contain all of given modifiers

- containExactly(mods) -> boolean

  Returns true if the modifiers contain all of given modifiers exactly and nothing else

```lua
if e:getFlags():containExactly{"cmd", "shift"} then
  -- Easy, no variable assignment!
end
```

The names I propose here are just some random choices, but what do you guys think about the idea of making it easier to test for pressed modifiers?

### DIFF
--- a/extensions/eventtap/event.m
+++ b/extensions/eventtap/event.m
@@ -1,6 +1,8 @@
 #import "eventtap_event.h"
 #import <IOKit/hidsystem/ev_keymap.h>
 
+#define FLAGS_TAG "hs.eventtap.event.flags"
+
 CGEventSourceRef eventSource;
 
 static int eventtap_event_gc(lua_State* L) {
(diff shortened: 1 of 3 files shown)


latenitefilms:
Awesome! Great idea!

cmsj:
@asmagill I think you've done much more work on eventtap than I have recently, what's your view on this PR?

asmagill:
lgtm

knu:
Thanks for your approval.  The commit accidentally included a couple of irrelevant changes, so I've removed them and then rebased on the current master.

PR #1332, opened 2017-04-14
Fix the key name for kVK_ANSI_Quote in the kTISPropertyUnicodeKeyLayoutData fallback table

knu:
It is for the single quote, not the double quote.  Because of this typo, you get the following results when you specify the quote key:

```
> hk = hs.hotkey.bind({"ctrl"}, "'", function () hs.eventtap.keyStroke({}, "`") end)
2017-04-14 18:57:57: 18:57:57 ** Warning:hs.keycode: key ''' not found in active keymap; using ANSI-standard US keyboard layout as fallback, returning '39'
2017-04-14 18:57:57:              hotkey: Enabled hotkey ⌃"
```

As you can see, the quote key only works thanks to the fallback to the ANSI/US layout, and the human representation for the key code 39 is incorrect.

### DIFF
--- a/extensions/keycodes/internal.m
+++ b/extensions/keycodes/internal.m
@@ -100,7 +100,7 @@ int keycodes_cachemap(lua_State* L) {
         pushkeycode(L, kVK_ANSI_Minus, "-");
         pushkeycode(L, kVK_ANSI_RightBracket, "]");
         pushkeycode(L, kVK_ANSI_LeftBracket, "[");
-        pushkeycode(L, kVK_ANSI_Quote, "\"");
+        pushkeycode(L, kVK_ANSI_Quote, "'");
         pushkeycode(L, kVK_ANSI_Semicolon, ";");
         pushkeycode(L, kVK_ANSI_Backslash, "\\");
         pushkeycode(L, kVK_ANSI_Comma, ",");


knu:
I found this bug when I was looking into [a question on SO](http://stackoverflow.com/questions/43383348/remap-apostrophe-to-other-keys-using-hammerspoon/43410169).

Looks like the fallback was implemented after the last release, so I hope this will get merged before the next release or you cannot specify the quote key by the name ("'") but only by the code (39/0x27).

cmsj:
@asmagill any thoughts here? IIUC, this is more correct because these codes are supposed to represent the unshifted character of each physical key, so `'` would be right for a US/ANSI keyboard.
Little worried about breaking existing configs, but, the current value seems wrong and that's confusing for all future users.

asmagill:
lgtm

I think this is an edge case -- the hard coded map in `internal.m` is used only when `TISGetInputSourceProperty(currentKeyboard, kTISPropertyUnicodeKeyLayoutData)` returns NULL, i.e. the current layout has no `kTISPropertyUnicodeKeyLayoutData` property.

knu:
I see, it is an edge case but applies to all East-Asian language environments:
- https://github.com/ianyh/Amethyst/issues/170
- https://lists.apple.com/archives/cocoa-dev/2009/Aug/msg01510.html
- https://github.com/davedelong/DDHotKey/commit/d5e6ab5bad050d21d93781a5e4e8c6e60ca449e6

I know you know that but these are some pointers I collected for future reference.

asmagill:
Edge case may be a bit strong if an entire class of input sources are affected, but... I'm not sure what if anything we can do about it...  Further digging into the Script.h file for the CoreServices framework has the following in its comments:

> *    3. smKCHRCache - To obtain the key layout data for the keyboard
> *    layout currently in use, use
> *    TISCopyCurrentKeyboardLayoutInputSource to get that input source,
> *    then pass it to TISGetInputSourceProperty with the
> *    kTISPropertyUnicodeKeyLayoutData key (this returns 'uchr' Unicode
> *    layout data only; it will not return any data for keyboard
> *    layouts that only have 'KCHR' data). 

So, having the fallback in `internal.m` seems warranted.  And having it corrected especially so :-)

PR #1344, opened 2017-04-19
Added force option to hs.ipc.cliUninstall

latenitefilms:


latenitefilms:
Ready for idiot checking @asmagill & @cmsj !

cmsj:
I think one problem you're going to have here is that there's no way to tell if a boolean parameter is intended to be the `silent` option or the `force` option.

extensions/ipc/init.lua
```diff
@@ -231,23 +231,25 @@ module.cliInstall = function(path, silent)
...
 ---
 --- Notes:
 ---  * This function is very conservative and will only remove the tool if it was installed by this instance of Hammerspoon. If you have more than one copy of Hammerspoon, this will be detected and they will not remove each others' tools.
-module.cliUninstall = function(path, silent)
+---  * If the optional force parameter is used, the remove tool will run without any checks.
+module.cliUninstall = function(path, silent, force)
     local path = path or "/usr/local"
     local silent = silent or false
-    if module.cliStatus(path, silent) == true then
+    if module.cliStatus(path, silent) == true or force then
```
asmagill: to be pedantic, if force is true, this will still run the check (and display output, depending upon silent) but will ignore the results.  To match what you describe in the documentation string, you should make it `force or module.cliStatus(path, silent)` or change the documentation string.  You should probably also state that to specify the `force` parameter, you must also specify the silent argument (i.e. it's not optional if you want to specify `force`.

asmagill:
I'm ok with adding a force variable... it really should only matter to people who build multiple Hammerspoon binaries...

TBH, though, I'm not positive that mismatched versions really matter much, though... the tool hasn't changed much since its inclusion -- I think I recall one update that made it more robust with non-unicode ascii, but nothing else since that has actually changed the `hs` code itself, so even if you're running it from a different compilation, it will connect to the "currently running" Hammerspoon application.

(Caveat: Mismatched versions *will* matter if you change the registered port name in `hs.ipc`'s internal.m file and in the `hs` source code)

That said, I have had a couple of ideas on some additions, though I've no immediate plans to implement them, just sort of a "maybe someday wish list":

1. if Hammerspoon isn't running, optionally prompt to start it, kind of like the "Install Command Line Tools" that occurs when you try to run `clang` but haven't install Xcode.
2. a persistent command history, maybe even tying it into the Console's history (which can be made persistent with some Lua code)

latenitefilms:
> I think one problem you're going to have here is that there's no way to tell if a boolean parameter is intended to be the silent option or the force option.

Fixed in the branch.

> if Hammerspoon isn't running, optionally prompt to start it, kind of like the "Install Command Line Tools" that occurs when you try to run clang but haven't install Xcode.

Great idea!

cmsj:
@asmagill agreed that mismatched versions aren't an issue, but I think one place that actually might affect people is if they have installed the symlinks, then moved Hammerspoon.app. At that point, `hs` won't work, and the uninstall command will refuse to do anything.

I think I'm leaning towards not having a force option, and just unconditionally removing the existing symlinks when someone asks us to uninstall. Any dissent?

cmsj:
I have pushed such a change, but still invite dissent :)

latenitefilms:
Works for me!

Issue #1305, opened 2017-04-04
Release 0.9.53


cmsj:
It feels like we're on the home stretch for a bunch of work, so let's get our ducks in a row for 0.9.53.

Usual drill, please nominate anything for the milestone that you want/need urgently fixed/added

latenitefilms:
**My Wishlist:**

- Fix up `hs.webview.toolbar` `priority` documentation and/or the way it's handled. See: #1300
- Add `hasMoved` and `gotFocus` callbacks to `hs.webview`. See: #1259
- Add `hs.changeUpdateInterval(value)`, and allow `hs.updateAvailable()` to return the new version (if possible?). See: #1221
- `hs.alert` doesn't display on correct screen in certain display configurations. See: #1196
- `hs.drawing:bringToFront(true)` doesn't work for full-screen window. See: #1184
- Add User Locale to `hs.host`. See: #1170
- Displaying Menubar Title & Icon at same time. See: #1015

**...and some others, that might not be suitable for `0.9.53`:**

- Add Hammerspoon Bundles. See: #1182
- Add [`hs._asm.touchbar`](https://github.com/asmagill/hammerspoon_asm/tree/master/touchbar) to Hammerspoon core. See: #1078
- Add [`hs._asm.axuielement`](https://github.com/asmagill/hs._asm.axuielement) to Hammerspoon core.
- Add API to allow drag & drop support to Hammerspoon Dock Icon and/or Menubar. See: #1057
- Add extension that allows reading/writing to plist files. Discussed [here](https://github.com/Hammerspoon/hammerspoon/issues/1130#issuecomment-266266474).
- Improve `hs.application` bundle handling. See: #1001

MichaelHindley:
Fixing the problem with repeating keys which make any type of navigation or repeating tasks unusably inconsistent: #1156, #1179, #1178, #1308

latenitefilms:
@MichaelHindley - Curious... have all your tests been on macOS 10.12? Have you tested on 10.11 as well? Any difference?

MichaelHindley:
@latenitefilms no, I don't have access to a 10.11 machine, and I was not a user of hammerspoon at that time, so I can't answer to the timeline aspect of this issue.

MichaelHindley:
sorry was a bit quick, yes, ALL tests were on 10.12 :)

latenitefilms:
[...]

PR #1347, opened 2017-04-21
Refactor hs.doc and add support for Spoon bundles

asmagill:
Currently supports documentation in the console for Hammerspoon, Lua 5.3, and installed Spoon bundles.  Documentation is added to the console functions/metamethods (i.e. using as `hs.doc.help("whatever")` or as `hs.doc.whatever`) as each file is registered; thus the delay in parsing the core documentation set occurs only once (this particular delay has always occurred during the startup of Hammerspoon, but now it doesn't have to be repeated when adding new files)

Still in progress:
* ~~it's a little brittle... check out `hs.doc.hs.spaces` -- since there is no documentation defined between `hs` and `hs.spaces.watcher`, this currently breaks.  The simple fix would be to add a place holder for `hs.spaces`, but I am holding off on this because the code needs to be able to survive such a situation if/when third party module writers (i.e. not spoons) take advantage of this as well.~~
* ~~`hs.doc.hsdocs` has not been updated to include the Spoon documentation yet~~
  * ~~in addition, the pull down menu displayed in the search bar of the webview toolbar (when using the built in webview to view the docs rather than an external browser) is empty~~
* ~~`hs.doc.builder` is not yet documented~~
  * in addition, it's not quite an identical parser when compared to `build_docs.py` -- `stripped_doc` is not created and the other new fields have not been fully verified yet.  This does not prevent its use for `hs.doc` or `hs.doc.hsdocs` (they were not previously aware of these new fields), but may impact future plans, hence the attempt to make this parser as identical as possible.
* ~~Other tweaks to `hs.doc.hsdocs` to support injected table of contents per https://github.com/Hammerspoon/hammerspoon/issues/1305#issuecomment-294372238~~
* Probably more, but I need a break :-)

However, with all that said, it is ready for some initial testing by others.


asmagill:
This is probably ready for 0.9.53, though it would help if someone other then myself could test it a bit.

The javascript that I want to inject as described in a comment in #1305 is going to take some work... it breaks the dark-mode option provided by `hs.doc.hsdocs` and changes the font sizes a bit more than I like as it moves some things a bit... I like the idea of a left column listing the current pages methods/functions/etc, and I like the ability to hover over an icon and get a table of contents (though if you're using the built in browser, this can be done with the search field in the toolbar already so it's really more of a boon to those using external browsers), but it needs some thought and experimentation and discussion regarding the changes before being added.

@cmsj, adding spoon support to `hs.doc.hsdocs` was actually easier then I expected... and it's making me rethink how the console documentation support is coded... I suspect I made it harder then it has to be.  I hope to review and possibly simplify the code this weekend, but it does work now (for me at least), so... if you're ready to release and we get some feedback that I haven't broken anything else, then don't let my "but wait, I can do it better!" 😜 stand in the way.

asmagill:
Oh and I'll also note that builder will undergo modification in the future to bring it more in line with `build_docs.py`, but it's currently sufficient for `hs.doc` and `hs.doc.hsdoc` (i.e. the json code it creates per the Spoon documentation example is sufficient since we don't use the additional fields internally at present)

latenitefilms:
@asmagill - Looking at your recent commits, I wonder if it's worth adding something to `hs.webview` that allows the `hs.webview` to automatically remember the `hs.webview` position in Objective-C land rather, than have to manually set up watchers in Lua land?

asmagill:
@cmsj, I'm not going to worry about this for 0.9.53, but it occurs to me that if I were to add a `docs.json` file for one of my modules (e.g. `hs._asm.axuielement`) and register it with `hs.doc.registerJSONFile`, then in the visual documentation provided by `hs.doc.hsdocs`, my module would appear within the `API` tab, mixed in with the core modules.

While this mixing is useful for the console command line help (e.g. typing in `hs.help.hs._asm.axuielement.whatever`), it's perhaps less useful in a visual web based display... should a `Third-Party` tab be added for any documentation which does not belong to a core module or to a spoon?

asmagill:
And because the text describing this last pull doesn't really convey what is meant, a picture to speak the other 900 and some odd words that pictures are said to carry:

screen shot 2017-04-22 at 1 58 08 pm

cmsj:
@asmagill I'm not sure that third party modules are sufficiently common for it to really matter that much. I guess it's probably not too hard to figure out if a module is third party from its path, but I think it's fine for it to be in the API tab - it is providing API afterall :)

asmagill:
should spoon sorting be case insensitive?

cmsj:
I would prefer people to CamelCase their Spoon names, but yes I think the sorting should be case insensitive. I'll take a look at getting that done in `build_docs.py` too

asmagill:
Ok, I've added the sidebar containing the entities defined in the current module, one of the changes https://github.com/Hammerspoon/hammerspoon/issues/1305#issuecomment-294372238 suggests.  

As I discussed above, the popup list of other modules that you can jump to is not as useful if you're using the built in webviewer to view the documentation, and the other changes involve style changes that I believe are outside the scope of `hs.doc.hsdocs` since the purpose is to match the online documentation as closely as possible.  If anyone disagrees or has other thoughts, I'm open to reconsidering (but probably not until 0.9.54).

As an example, this is what it looks like once you've enabled this with `hs.doc.hsdocs.moduleEntitiesInSidebar(true)`:

screen shot 2017-04-22 at 5 55 08 pm

(It should be noted that in both of the screen shots I've included in this thread, I have dark mode enabled; the effects would be inverted if you have set `hs.doc.hsdocs.browserDarkMode` to false or do not have dark mode enabled in your General preferences.)

asmagill:
@latenitefilms, I'm not sure adding frame saving to `hs.webview` is worth it... if you look at the code change which implements (let's be honest, fixes) frame saving for the documentation browser, it's really very little code for when you want to do it, but if we made it part of the module, other questions arise...

For example, the webview requires a unique name.  For a traditional application where you're using the tools provided by Xcode, this makes sense -- one is added automatically for every window you create in the interface builder, even if you don't explicitly set one, and macOS's NSWindow and NSApplication infrastructure take care of most of the details implementing the saving and re-applying for you.

Hammerspoon, however, doesn't use the Interface builder, so we don't know in advance how many or what windows we'll have, so we can't leverage that support.

So, we still need a way to assign a name to the webview, but the webview window exists before it has a name... so what should happen when we assign the name?  Automatically move to the saved frame for that name, even if it's currently visible somewhere else?  What about when we change it's name... should it adjust to whatever's saved for the new name?  Or, for both cases, overwrite what may be saved if it's already been made visible?

The truth is, I can imagine cases where I'd want each and every variation of what I've just described above, so since the code is now simpler with `hs.webview:windowCallback`, it's my opinion we should leave it as is so the coder can implement exactly what they want for a specific use case.

It might be a good choice for a documentation addition or for an examples page or even for a spoon which "wraps" webview and handles this for you behind the scenes (you could even "wrap" individual methods by modifying the table returned by `hs.getObjectMetatable("hs.webview")` if you're careful, and effectively extend `hs.webview` to include modified/new methods.)

As always, I'm open to discussion, but that's where I am with regards to the question at present.

latenitefilms:
@asmagill - I could be completely wrong, but I just assumed if we used the same code (which is only a few lines) that we're now using for the Console to save/restore position with the webview's, macOS would just automatically take care of the rest? I would ASSUME that if you changed the webview name, it would just create a new entry in the plist with the new name?

Also, for what it's worth, I would personally love to be able to add third party extensions to the Hammerspoon docs!

asmagill:
The console *was* designed by the interface builder, so there is never a point in time when it doesn't have the name the OS uses to save/restore it's frame.  Webview's will have a period, even if it's measured in milliseconds, between when it's initial frame is programmatically set by the `new` function and when it has a name assigned to it, even if we add it as a parameter to `hs.webview.new`.

Plus, while I haven't looked at the specific code that was added to correct this for the console, IIRC from my reading about this in the past, it's in the application delegate when the application has finished loading that the frame positions are initially read in for the windows which were defined by the interface builder.  We're way past that method being invoked when a webview is created.

Of course, I may be remembering wrong... it *was* almost 9 months ago when I was working on possibly adding media subviews to canvas that I read about this, so...

latenitefilms:
@asmagill - Ah right, no worries - that all makes sense. We'll just keep doing it the way we're doing it then! Thanks for the detailed explanations!

cmsj:
@asmagill lgtm, although it should add this patch to match up with a change I just made in `build_docs.py`:

```diff
diff --git a/extensions/doc/hsdocs/index.md b/extensions/doc/hsdocs/index.md
index a8ea2430..b2ff753a 100644
--- a/extensions/doc/hsdocs/index.md
+++ b/extensions/doc/hsdocs/index.md
@@ -5,6 +5,7 @@ Resource                 | Link
 Website                   | http://www.hammerspoon.org/
 GitHub page               | https://github.com/Hammerspoon/hammerspoon
 Getting Started Guide     | http://www.hammerspoon.org/go/
+Spoon Plugin Documentation| https://github.com/Hammerspoon/hammerspoon/blob/master/SPOONS.md
 Official Spoon repository | http://www.hammerspoon.org/Spoons
 IRC channel               | irc://chat.freenode.net/#hammerspoon
 Mailing list              | https://groups.google.com/forum/#!forum/hammerspoon/
```

cmsj:
(which I can add after this merges. I'll see about doing a release later tonight :)

asmagill:
The only addition I think I'd like to squeeze in is in response to #1352.  I think I have the size/topLeft/frame methods ready, but I'm still not sure about the crash he mentions regarding the toolbar reassignment... I think I can get it addressed today as well, but it's going to be a few hours (family beckons and is sending me the frowny faces right now 😝)

latenitefilms:
Amazing - thanks so much for all your constant help and support @cmsj & @asmagill !!

latenitefilms:
@asmagill - For what it's worth, if I'm able to change the size/frame of the webview even without `hs.webview.windowMasks` containing `resizable` - then I shouldn't have to apply the toolbar more than once, so HOPEFULLY this will mean no bugs/crashes. I think the only reason Hammerspoon was struggling was because I was adding a toolbar to a webview that already previously had a toolbar (and the only reason the toolbar disappeared was because of changing the `windowStyle` after the webview was already visible).

PR #1349, opened 2017-04-21
Added Shortcut to Trash Preferences

### LINKED ISSUE #1348: Hold Command+Option on Launch to Trash Hammerspoon Preferences
It would be great to have a really easy way to trash the Hammerspoon Preferences.

For example, Apple's Final Cut Pro allows you to hold down Command+Option when launching the app from the Dock to trash the preferences. The user will be presented with the attached dialog box to confirm.

screen shot 2017-04-21 at 10 03 01 pm

latenitefilms:
* Added Shortcut to Trash Preferences when Command+Option keys are held
down when launching Hammerspoon from the Dock
* Closes #1348

### DIFF
--- a/Hammerspoon/MJAppDelegate.m
+++ b/Hammerspoon/MJAppDelegate.m
@@ -132,7 +132,7 @@ static BOOL MJFirstRunForCurrentVersion(void) {
         [alert addButtonWithTitle:@"Continue"];
         [alert addButtonWithTitle:@"Delete Preferences"];
         [alert setMessageText:@"Do you want to delete the preferences?"];
-        [alert setInformativeText:@"Deleting the preferences will reset all application settings to their defaults."];
+        [alert setInformativeText:@"Deleting the preferences will reset all Hammerspoon settings (including everything that uses hs.settings) to their defaults."];
         [alert setAlertStyle:NSWarningAlertStyle];
         
         if ([alert runModal] == NSAlertSecondButtonReturn) {
--- a/Hammerspoon/MJAppDelegate.m
+++ b/Hammerspoon/MJAppDelegate.m
@@ -122,8 +122,34 @@ static BOOL MJFirstRunForCurrentVersion(void) {
     return YES;
 }
 
-- (void)applicationDidFinishLaunching:(NSNotification *)aNotification {
 
+- (void)applicationDidFinishLaunching:(NSNotification *)aNotification {
+    
+    // User is holding down Command (0x37) & Option (0x3A) keys:
+    if (CGEventSourceKeyState(kCGEventSourceStateCombinedSessionState,0x3A) && CGEventSourceKeyState(kCGEventSourceStateCombinedSessionState,0x37)) {
+        
+        NSAlert *alert = [[NSAlert alloc] init];
+        [alert addButtonWithTitle:@"Continue"];
+        [alert addButtonWithTitle:@"Delete Preferences"];
+        [alert setMessageText:@"Do you want to delete the preferences?"];
+        [alert setInformativeText:@"Deleting the preferences will reset all application settings to their defaults."];
+        [alert setAlertStyle:NSWarningAlertStyle];
+        
+        if ([alert runModal] == NSAlertSecondButtonReturn) {
+            
+            // Reset Preferences:
+            NSDictionary * allObjects;
+            NSString     * key;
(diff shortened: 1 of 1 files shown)


asmagill:
It should be noted that this will also purge anything the user or code has saved with `hs.settings`.

latenitefilms:
@asmagill - I've updated the alert message to make it a bit clearer.

Thoughts @cmsj ?

cmsj:
@latenitefilms what's the motivation for this? I'm not sure I've ever had someone ask how to delete their Hammerspoon app settings.

latenitefilms:
It's basically just a fast way of Trashing Preferences. I always Trash the preferences prior to testing a release version, just to make sure that everything works properly. There's been plenty of times when I think everything is working great, but then when I Trash Preferences, I release that I wrongly ASSUMED a setting would already exist when it doesn't.

This would mainly be for people regularly making tweaks to the Hammerspoon app (or fork). However it could also be a useful tool for users who just want to "reset" all their settings when something's going funky.

It's just easier than remembering the two Terminal command to delete a plist file and refresh the plist cache.

latenitefilms:
If you're happy to add this feature, you could probably merge into this upcoming release? I don't feel like it'll do any harm - and is really just an extra handy tool for Hammerspoon developers.

If you're not interested in adding just close the issue.

Thanks!

latenitefilms:
Third opinion @asmagill ?

asmagill:
I don't have a strong opinion one way or the other... I think I've needed to do this only once or twice myself (though there have been a couple more times when my testing *should* have included clearing or at least moving the preferences out of the way temporarily...)

Generally I think I have enough things in my settings now that I don't see a likelihood of my taking advantage of this, but it's a distinct enough key combination, and it does prompt before actually doing the purge, so... I see no harm in adding it... 

Just to confirm, hitting Return triggers "Continue", rather than "Delete Preferences", correct?

What *would* interest me would be a feature allowing you to chose between multiple sets of preferences (or at least create a temporary one in memory only that doesn't overwrite the primary with any changes, so you can test in a "pristine" state), but that's beyond this pull's original scope and I'm honestly not sure if it's possible without significant changes to the Preferences pane or how/where we store stuff manipulated with `hs.settings`.

asmagill:
Though now that I think of it... now that we include `hs.sqlite3`, would it be worth exploring modifying `hs.settings` so it could be toggled to use that rather than the userdefaults?  That would provide a proper separation between Hammerspoon preferences and things that use `hs.settings` and make multiple or even isolated sets possible...

latenitefilms:
> Just to confirm, hitting Return triggers "Continue", rather than "Delete Preferences", correct?

Yes, the default button is "Continue", to avoid any accidental deletion of preferences.

> What would interest me would be a feature allowing you to chose between multiple sets of preferences (or at least create a temporary one in memory only that doesn't overwrite the primary with any changes, so you can test in a "pristine" state), but that's beyond this pull's original scope and I'm honestly not sure if it's possible without significant changes to the Preferences pane or how/where we store stuff manipulated with hs.settings.

Looking at the **Build Settings > Packing > Info.plist File** section in Xcode, it looks like you can provide a different .plist file for Debug versus Release - so I think this does what you want to do natively?

> Though now that I think of it... now that we include hs.sqlite3, would it be worth exploring modifying hs.settings so it could be toggled to use that rather than the userdefaults? That would provide a proper separation between Hammerspoon preferences and things that use hs.settings and make multiple or even isolated sets possible...

This is probably more a discussion for @cmsj - however, personally, I like that everything is stored in the one plist file - it keeps things neat and tidy, so I'm happy to leave it as-is.

Personally, I'd much rather use your brain power and spare time towards building a proper plist reader/writer extension! :)

asmagill:
>Looking at the Build Settings > Packing > Info.plist File section in Xcode, it looks like you can provide a different .plist file for Debug versus Release - so I think this does what you want to do natively?

Not really... I almost never run a formal release version of Hammerspoon unless I'm trying to verify that a change I'm working on has broken something from the current release.

The sqlite idea was just that... an idea.  If I find that I actually start using sqlite for anything that I can't/don't want to use settings for, I may write a "wrapper" to simplify things and give it a similar interface, but... I'm in no rush.

cmsj:
Given that this is a small patch, is very self-contained, and it nicely prompts the user, I'm going to land it :)

PR #1367, opened 2017-04-25
adds in page search for hsdocs web pages for module details

asmagill:
When viewing the web pages displaying module details from `hs.doc.hsdocs`, Command-F will toggle a search field at the top-right.  You can enter a term to search for in the page and use Enter/Shift-Enter or Cmd-G/Cmd-Shift-G to search forwards or backwards within the page.

Works for both the `hs.webview` viewer and in Safari.

### DIFF
--- a/extensions/doc/builder.lua
+++ b/extensions/doc/builder.lua
@@ -34,24 +34,27 @@ local sections = {
     Method      = 8,
 }
 
---- hs.doc.builder.genComments(path) -> table
+--- hs.doc.builder.genComments(path, [recurse]) -> table
 --- Function
 --- Generates a documentation table for Hammerspoon modules or Spoon bundles from the source files located in the path(s) provided.
 ---
 --- Parameters:
 ---  * where - a string specifying a single path, or a table containing multiple strings specifying paths where source files should be examined to generate the documentation table.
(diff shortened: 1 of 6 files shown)


asmagill:
To make it more clear what this adds:

screen shot 2017-04-25 at 10 00 16 pm

Not sure what the best way to document this is... just a blurb in the header for the docs for the `hs.doc.hsdocs` module?

asmagill:
Will require #1373 and #1374.  Adds submodule list to module pages (so you don't have to keep going back and forth to the index) and an in-browser help screen that you can toggle with the new button at the top right of the toolbar.

Now if I can just figure out how to get the LuaSkin docs in there... though that's a whole new level of parsing since the documentation formats are incompatible, so... not in this pull.

Still want to work on word wrap for long lines
Cmd-R (undocumented and may be removed) will reload the page (useful if you just added a third-pary docs.json file), but also brings up Hammerspoon console... considering an eventwatcher to intercept this, but haven't decided for sure yet.

New help page (non-dark mode image):
screen shot 2017-04-27 at 12 26 12 am

asmagill:
Also the find-in-page and help only load if you use the webview browser -- if you view in an outside browser, use the features of that browser instead.

asmagill:
Objections to landing, @cmsj?  I'll also need to land #1373 and #1374.

cmsj:
weird, I thought I commented on this earlier. No objections :)

PR #1370, opened 2017-04-25
Push values for relocatable key codes only if translation succeeds

### LINKED ISSUE #1366: Bindings to § stop working on 0.9.53
I have a couple binding on paragraph key:
```
function getCurrentLayout()
    if (hs.keycodes.currentLayout() == "U.S.") then
        return "en"
    else
        return "ru"
    end
end

function swapBackquote()
    if (getCurrentLayout() == 'en') then
        hs.eventtap.keyStrokes("`")
    else
        hs.eventtap.keyStrokes("ё")
    end
end

function swapTilda()
    if (getCurrentLayout() == 'en') then
        hs.eventtap.keyStrokes("~")
    else
        hs.eventtap.keyStrokes("Ё")
    end
end

hs.hotkey.bind(nil, '§', swapBackquote, nil, swapBackquote);
hs.hotkey.bind('shift', '§', swapTilda, nil, swapTilda);
```

After updating to 0.9.53 they stop working - callbacks aren't called.

knu:
This should fix #1366.

### DIFF
--- a/extensions/keycodes/internal.m
+++ b/extensions/keycodes/internal.m
@@ -42,20 +42,20 @@ int keycodes_cachemap(lua_State* L) {
         UniCharCount realLength;
 
         for (int i = 0 ; i < (int)(sizeof(relocatableKeyCodes)/sizeof(relocatableKeyCodes[0])) ; i++) {
-            UCKeyTranslate(keyboardLayout,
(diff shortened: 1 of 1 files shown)


Habbie:
I tested this and it works. Thank you!

PR #1374, opened 2017-04-26
programmatically select searchfield for text input

asmagill:
Working on toolbar and hsdocs browser, this will be part of keyboard shortcuts for the viewer, but is independent


asmagill:
This update allows you to programmatically put the cursor into the search field of a toolbar with a search field in it (see the `hs.doc.hsdocs` browser for an example) and also set whether hitting enter in this box moves focus away from the field after the callback.

Objections?

cmsj:
@asmagill sorry, missed this one. No objections :)

PR #1375, opened 2017-04-27
Tweaks to build_docs.py

latenitefilms:
A few minor tweaks to `build_docs.py` as suggested by @randomeizer.


asmagill:
I don't know python that well, but the regular expression added looks good to me. Since the python doc builder is yours @cmsj, your thoughts?

randomeizer:
To clarify, it is fixing the case where there is additional information after a field of constant value is declared. For example, we are providing some type information about fields, like so:

```lua
--- foo.bar.field <boolean>
```

Without the adjusted pattern it will lose 'field' altogether in the table of contents/etc.

cmsj:
I'm 95% sure on this, I just need to check a couple of bits and I'll merge it.

asmagill:
+1

This fixes a link issue in the built in docs -- if you view the web docs for hs.canvas:behavior (e.g. `hs.doc.hsdocs.help("hs.canvas.behavior")`) and then click on the `windowBehavior` link, it fails before adding this fix because the anchor name it should jump to has the extra `[]` at the end of it.  This patch removes that from the field the docs generates the anchors from.

PR #1386, opened 2017-05-01
Add otherMouse... events to hs.eventtap.event

### LINKED ISSUE #1385: Missing values in hs.eventtap.event.types
The following event types are described in the documentation for `hs.eventtap.event`, but aren't actually present in the `hs.eventtap.event.types` table:

* `otherMouseDown`
* `otherMouseUp`
* `otherMouseDragged`

duskwuff:
Fixes #1385.

### DIFF
--- a/extensions/eventtap/event.m
+++ b/extensions/eventtap/event.m
@@ -816,6 +816,12 @@ static void pushtypestable(lua_State* L) {
     lua_pushstring(L, "tabletPointer") ;            lua_rawseti(L, -2, kCGEventTabletPointer);
     lua_pushinteger(L, kCGEventTabletProximity);    lua_setfield(L, -2, "tabletProximity");
     lua_pushstring(L, "tabletProximity") ;          lua_rawseti(L, -2, kCGEventTabletProximity);
+    lua_pushinteger(L, kCGEventOtherMouseDown);     lua_setfield(L, -2, "otherMouseDown");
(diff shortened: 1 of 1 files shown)


cmsj:
lgtm

@asmagill any objections?

asmagill:
These are already in there as "middleMouse..." (see https://github.com/Hammerspoon/hammerspoon/blob/master/extensions/eventtap/event.m#L799-L804)

The naming was inherited from the Hydra/Mjolnir origins, but I agree that "otherMouse..." is more correct.

My thought is to just correct the labels, but as this would break backwards compatibility, I'll defer to your judgement... I've never used these particular labels myself, but that doesn't mean that someone else hasn't.

duskwuff:
Oh, weird! [The documentation](http://www.hammerspoon.org/docs/hs.eventtap.event.html#types) said they were supposed to be called "otherMouse", so I totally didn't notice that they were present as "middleMouse".

The simplest solution would probably be to leave the forward mappings in place (so that `types.middleMouseDown` still works), but change the reverse mappings to the "otherMouse" names (so `types[25] = "otherMouseDown"`, for instance). This works with my commit, but would probably be a little cleaner without the redundant field-setting on lines 799/801/803.

asmagill:
Pull the reverse mappings for middleMouse* and then I say lets land this... the code order *should* make this unnecessary, but I agree that we should be as clean about it as we can.

cmsj:
Thanks! (I pulled the reverse mappings for middleMouse* on the way to landing this)

Issue #1389, opened 2017-05-02
Feature request - more methods for hs.eventtap.event


duskwuff:
The following CGEvent methods are not available through `hs.eventtap.event`. Having them available would make it much easier to synthesize specific types of events.

- [ ] CGEventCreate
- [ ] CGEventSetType
- [ ] CGEventGetLocation and CGEventSetLocation

(I'm specifically trying to synthesize "swipe" events to remap buttons 4 and 5 on a USB mouse to back/forward actions. There's probably a lot of other things these methods would be useful for, though!)

asmagill:
These are part of a replacement I want to write for eventtap, but getting that right is taking longer than I thought without breaking things or introducing other issues that I don't have a fix for yet.

I'll see about getting these added to the existing core module later tonight/tomorrow.

Issue #1278, opened 2017-03-27
ROpt/LOpt & RCommand/LCommand Support


laktak:
Would it be possible to add support for the left/right option and command keys?

For example I'd like to bind Right-Option + Cursor-Left while leaving Left-Option + Cursor-Left alone.

asmagill:
Not directly, no.

Couple of "messy" possibilities right now...

* Set up the hotkey as you normally would then use `hs.eventtap.checkKeyboardModifiers(true)` and look at the `_raw` field and logically AND the value against one of the following:
~~~
deviceRightAlternate      64
deviceRightCommand        16
deviceRightControl        8192
deviceRightShift          4
~~~

Down side, you capture `opt-right` no matter which option key is used, but you can have it do something different for each.  Some people have had some luck with trying to pass the typed key back out of a hotkey callback, others have had nothing but problems.  See https://github.com/Hammerspoon/hammerspoon/issues/894 (and the link to the Google group thread in it) for one such discussion.  You can also find others by searching in these issues for `hotkey` and `newKeyEvent`.


* You can watch for `flagsChanged` events with `hs.eventtap` and with a similar ANDing against `event:getRawEventData().CGEventData.flags` in the event callback (after verifying it's the appropriate modifier flag(s) with `event:getFlags()` and *only* then enabling the hotkeys (and also disabling them whenever the modifier keys change again -- either a release or additional modifiers)


I am working on an update to `eventtap` (and maybe eventually `hotkey` if I can address some timing issues) that should make this kind of thing cleaner, but its a significant change in the code base to address a wide range of oddities and limitations we've discovered with `eventtap`... I have no timeframe atm.

Issue #1235, opened 2017-02-19
Invalid key: cmd - this may mean that the key requested does not exist in your keymap


aristidesfl:
Trying to toggle modifiers like command, is there a way to do this?

```
local cmdOn = hs.eventtap.event.newKeyEvent({}, 'cmd', true)
local cmdOff = hs.eventtap.event.newKeyEvent({}, 'cmd', false)
hs.hotkey.bind({}, 'f', cmdOn:post, cmdOff:post, nil)
```

aristidesfl:
The reason why this is necessary is because something like the following is not enough to emulate ⌘⇥ application switching:

```
hs.hotkey.bind({}, 'f', keyCode('tab', {'cmd'}), nil, nil)
```

aristidesfl:
[`CGEventCreateKeyboardEvent`](https://developer.apple.com/reference/coregraphics/1456564-cgeventcreatekeyboardevent) supports pressing modifier keys. Is there a reason why Hammerspoon doesn't?

deryni:
You can't grab `cmdOn:post` like that. It isn't a curried function, but you could wrap it in a function easily enough.

```lua
function cmdDown()
    return cmdOn:post()
end
```

But I would imagine that you would need to send `cmd-tab` and then a synthetic `cmd`-release instead of three events ... but maybe not.

Though actually it looks like `hs.eventtap.keystrokes` is what you'd want to use here. (Though I'm not sure I see how you could get arbitrary `⌘⇥ ⇥` sequences that way.)

aristidesfl:
The code you shared also doesn't work cause you are calling `post()` when returning. Maybe you meant something like this?
```
local function keyDown(key)
  local keyEvent = hs.eventtap.event.newKeyEvent({}, key, true)
  return function ()
    keyEvent:post()
  end
end

local function keyUp(key)
  local keyEvent = hs.eventtap.event.newKeyEvent({}, key, false)
  return function ()
    keyEvent:post()
  end
end


hs.hotkey.bind({}, 'f', keyDown('cmd'), keyUp('cmd'), nil)

```

The problem remains that `cmd` is not in the list of allowed keys. Tried with keycodes like `55` and others, but also no effect. Do you know what is the actual keycode of `cmd` key?

I actually need to manipulate `cmd` independently, since I would need for example the following sequence: `cmd:on, tab:keystroke, tab:keystroke, cmd:off` in toder to go to the second app in the list.

[...]

PR #1396, opened 2017-05-03
hs.eventtap.event additions

asmagill:
This should be considered an early release for review and comments... I want to give it a more thorough testing before it makes it into core, but I welcome any comments or testing that others may wish to offer as well.

1. cleans up many of the compile time warnings from this module
2. includes support for `hs.eventtap.event.newKeyEvent` to create modifier key only events as described in #1278, #1235 (others ?)
3. includes requested methods from #1389 
4. includes methods for serializing events as strings and recreating them (allowing recording, passing to another computer over the network, etc.)
5. allows getting/setting raw flag values and mask table to build/disect them directly

This is minimally tested at the moment, so expect some additional commits before its ready.


knu:
This looks great.  If it is recommended you break up a key combination into separate events of modifier and normal keys up and down, why don't we have a new helper function that would return a list (table) of events for generating a given key combination?  I often need one typically when building a return value from an eventtap handler to replace the original key event with.

It can be implemented like this:
```lua
function module.event.newKeyEvents(modifiers, character)
  local codes = fnutils.map({table.unpack(modifiers), character}, getKeycode)
  local n = #codes
  local events = {}
  for i, code in ipairs(codes) do
    events[i] = module.event.newKeyEvent(code, true)
    events[2*n+1-i] = module.event.newKeyEvent(code, false)
  end
  return events
end
```

asmagill:
Objections?  This could be included in the next release for #1433.

cmsj:
No objections :)

PR #1405, opened 2017-05-04
New hs.spoons module

zzamboni:
For now the same contents as SpoonUtils from https://github.com/Hammerspoon/Spoons/pull/8,
restructured as a module.


cmsj:
Merged (although I removed the `printList()` function, I don't think it's really necessary). Thanks!

zzamboni:
Fully agree about `printList()`. I'm working on some additional functions to install Spoons, should be submitted shortly :)

PR #1409, opened 2017-05-07
Fix handling of parameters in hs.pasteboard.readString()

### LINKED ISSUE #1403: hs.pasteboard.read* functions have unexpected behavior
This code will print the string in the pasteboard
```
function test1()
  type = hs.pasteboard.typesAvailable()
  str = hs.pasteboard.readString()
  print(str)
end
test1()
```

But this code will print a table
```
function test2()
  local type = hs.pasteboard.typesAvailable()
  str = hs.pasteboard.readString()
  print(str)
end
test2()
```

When use `hs.pasteboard.readImage()`, similarly the test1() works fine but test2() print a table

knu:
This should fix #1403.


asmagill:
lgtm

Issue #1411, opened 2017-05-08
hs.pathwatcher - Only trigger when file has completed copying?


latenitefilms:
@asmagill and/or @cmsj - is there anyway to make `hs.pathwatcher` only trigger when a file has FINISHED being copied to a watch folder?

knu:
Unfortunately no, since currently it does not pass eventFlags to a callback function.

knu:
I guess the condition for checking if a regular file that has been opened for writing is closed would be this:
```
eventFlags[i] & (kFSEventStreamEventFlagItemIsFile | kFSEventStreamEventFlagItemCreated | kFSEventStreamEventFlagItemModified) == kFSEventStreamEventFlagItemIsFile | kFSEventStreamEventFlagItemCreated | kFSEventStreamEventFlagItemModified
```

If you also need to detect when a file is moved to the folder you are watching, you'll have to cover events with `kFSEventStreamEventFlagItemRenamed`.

latenitefilms:
AMAZING! Thank you!!

I'll test out later tonight. Thanks again!

latenitefilms:
Thanks so much for your help with this @knu! HUGELY appreciated!

At the moment, `eventFlags` seems to be returning a table with a single number - for example: 

`{ 67584 }`

Would it be better to return a table with strings of the various constants? For example:

`{ "itemCreated", "itemIsFile", "itemModified" }`

It would probably also be worthwhile adding some more documentation to `hs.pathwatcher` to explain what all the callback results actually mean (which I guess would just be a lot of copying and pasting between the Apple Docs?).

Thoughts @asmagill  & @cmsj ?

Either way, I'd love to see this in Hammerspoon!

latenitefilms:
Also, if I rename a file, I get:

`{ 67584, 67584 }`

...the first value in the table which relates to the original filename, and the second value relates to the new file name.

I wonder if `hs.pathwatcher` should be smart enough to only release a single callback when something is renamed, rather than two? Or should this "logic" be done in Lua-land as opposed to in the Objective-C extension?

latenitefilms:
FYI: [hs.speech:setCallback](http://www.hammerspoon.org/docs/hs.speech.html#setCallback) is probably a good example of how I would imagine the callback to work/be documented.

knu:
As hs.pathwatcher has been there at least for a couple of years and it deals with filesystems, I don't think we should break compatibility and risk user's files in any way.

[...]

PR #1412, opened 2017-05-08
hs.patchwatcher: Pass event flags to a callback function in addition to changed files

knu:
`hs.patchwatcher.eventFlags` is the new table containing flag constants.

Event flags are requisite when you need to, for example, detect a new file opened for writing being closed.

This PR is an answer to #1411.

### DIFF
--- a/extensions/pathwatcher/internal.m
+++ b/extensions/pathwatcher/internal.m
@@ -14,6 +14,40 @@ typedef struct _watcher_path_t {
     bool started;
 } watcher_path_t;
 
+#if MAC_OS_X_VERSION_MAX_ALLOWED < 1090
+static const FSEventStreamEventFlags kFSEventStreamEventFlagOwnEvent           = 0x00080000;
+#endif
+#if MAC_OS_X_VERSION_MAX_ALLOWED < 101000
+static const FSEventStreamEventFlags kFSEventStreamEventFlagItemIsHardlink     = 0x00100000;
+static const FSEventStreamEventFlags kFSEventStreamEventFlagItemIsLastHardlink = 0x00200000;
+#endif
+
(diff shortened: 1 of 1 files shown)


latenitefilms:
FYI: 

screen shot 2017-05-09 at 12 14 35 pm

knu:
Ah, of course.  Fixed.

extensions/pathwatcher/internal.m
```diff
@@ -72,6 +72,8 @@ void event_callback(ConstFSEventStreamRef __unused streamRef, void *clientCallBa
 ///   * ownEvent (OS X 10.9+)
 ///   * itemIsHardlink (OS X 10.10+)
 ///   * itemIsLastHardlink (OS X 10.10+)
+///
+/// See [the official documentation](https://developer.apple.com/reference/coreservices/1455361-fseventstreameventflags/kfseventstreameventflagitemrenamed?language=objc) for details.
```
latenitefilms: I think you probably meant to link to this instead?

https://developer.apple.com/reference/coreservices/1455361-fseventstreameventflags/

knu: Exactly!

asmagill:
@knu, decide on if or how you want to handle the suggestions in #1411 re how the flags are included in the callback (your thoughts @cmsj?) but otherwise lgtm.

knu:
@asmagill I agree that table lookups would be user-friendlier than bitwise operations, so I've changed hs.pathwatcher to pass flag tables instead to a callback function.

latenitefilms:
Perfect! Thank you!

All looks good to me!

asmagill:
lgtm

PR #1416, opened 2017-05-09
Don't assume that we can always make a UTF8 string out of the HTTP re…

cmsj:
…sponse body. Determine behaviour from the Content-Type response header

Ping @asmagill for sanity check :)

(this is in response to an issue raised on IRC by @zzamboni)

### DIFF
--- a/extensions/http/internal.m
+++ b/extensions/http/internal.m
@@ -11,12 +11,6 @@ typedef struct _webSocketUserData {
     void *ws;
 } webSocketUserData;
 
-// Lua response body struct
-typedef struct _luaResponseBody {
-    void *data;
-    unsigned long len;
-} luaResponseBody;
-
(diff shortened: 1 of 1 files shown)


asmagill:
FYI, LuaSkin's pushNSObject does handle NSData objects, so I might have made `responseBodyToLua` return a type of `id` and made the object returned NSString, if text, or NSData if not.

However, coding choices aside, looks good to me.

zzamboni:
The patch seems to work fine for both `get()` and `asyncGet()`.

cmsj:
@asmagill doh, that's a very good point! I'll update the PR :)

cmsj:
@zzamboni would very much appreciate if you could test the updated PR here. It should work exactly as the previous version, just less code and more @asmagill awesomeness \o/

zzamboni:
@cmsj the updated patch also seems to work OK!

Issue #1356, opened 2017-04-23
command line tool additions/changes


asmagill:
Since this was originally mentioned in a pull request, I thought I'd list it here as a feature request as a reminder or in case someone else wants to tackle it before I do:

From #1344:
... I have had a couple of ideas on some additions, though I've no immediate plans to implement them, just sort of a "maybe someday wish list":

1. if Hammerspoon isn't running, optionally prompt to start it, kind of like the "Install Command Line Tools" that occurs when you try to run clang but haven't install Xcode.
2. a persistent command history, maybe even tying it into the Console's history (which can be made persistent with some Lua code)
3. lua/Hammerspoon errors should be output through stderr rather then combined into the stdout stream (at least when using the `-c` option) with an appropriate return code for shell scripts to detect

zzamboni:
Other wishes for `hs` from me:
- add a `-f` option that reads the code from a file
- allow passing command-line parameters to the executed code, which should get stored in the `arg` table as for a regular Lua script.

szymonkaliski:
Would it be possible to have `hs` work as replacement for `hs.console`? I'd much prefer to run `hs` in my `tmux` split and see all the logs, than opening new window...

asmagill:
@szymonkaliski just to clarify... you want a mode (perhaps a specific command line argument to `hs`) which causes output traditionally bound for the console to be sent to the command line tool's instance instead/in-addition?  A `tee` for the console?

szymonkaliski:
Yeah exactly :)

cmsj:
I like the idea of hs acting as a remote console, but I would go for sending it messages in addition to the built-in console

asmagill:
To allow bidirectional communication between Hammerspoon and the command line tool which can be initiated by either end is a significant departure from the original... How important is backwards compatibility?

Specifically:
1. if the new version of the CLI attaches to the old version of `hs.ipc` should it work in a "legacy" mode? This is probably easy if desired.
2. if the old version of the CLI attaches to the new version of `hs.ipc` should it work in a "legacy" mode? A little more difficult, but should also be doable if desired.
3. anything printed in the old CLI instance is also printed to the console... should this continue to be a default for the new one?  This will have to be a toggle to allow console messages to be replicated to the CLI instance since having both on would create an infinite loop... should it be the default, though?
[...]

Issue #1433, opened 2017-05-22
Release 0.9.56


cmsj:
Usual drill folks, let's get nominations for what people feel is important for the next release :)

latenitefilms:
**My Wishlist for 0.9.55:**
- Extension for Reading/Writing External Property List Files. See: #1432
- `hs.notify.warnAboutMissingFunctionTag` - as a callback instead? See: #1415
- Better Error Handling for `hs.fs.tagsAdd()`. See: #1402
- Path to Resource URL? See: #1401
- `hs.hotkey.allowed(mods, key)`. See: #1391
- `hs.chooser` weirdness. See: #1393
- AppleScript should raise an error if Lua code has errors. See: #1324

**My Wishlist for future versions:**
- `hs.keycodes` should be able to return a keycode map for the current layout. See: #1307
- Add Ability to add Custom Items to Dock Icon Contextual Menu. See: #1350
- `hs.fs` - Support for reading/writing Finder comments? See: #1400
- `hs.fs` - Support for Reading/Writing Extended Attributes? See: #1394
- Add [`hs._asm.touchbar`](https://github.com/asmagill/hammerspoon_asm/tree/master/touchbar) to Hammerspoon core. See: #1078
- Add `[hs._asm.axuielement](https://github.com/asmagill/hs._asm.axuielement)` to Hammerspoon core.
- Add API to allow drag & drop support to Hammerspoon Dock Icon and/or Menubar. See: #1057
- `hs.webview` - Dark UI option? See: #1321

zzamboni:
My list:
- hs.spoons: https://github.com/Hammerspoon/hammerspoon/pull/1427 and https://github.com/Hammerspoon/hammerspoon/pull/1427
- Handling binary data in hs.http responses: https://github.com/Hammerspoon/hammerspoon/pull/1416
- @asmagill's hs.ipc improvements from https://github.com/Hammerspoon/hammerspoon/issues/1356
- Themeable console: https://github.com/Hammerspoon/hammerspoon/pull/1363

I'll add more as I think of them.

cmsj:
@zzamboni fwiw #1416 #1427 and #1363 are merged already :)

zzamboni:
@cmsj I know, just making it explicit :)

Ninlives:
Is there any plan for the real touch bar customization?

latenitefilms:
[...]

PR #1439, opened 2017-05-29
hs.ipc v2

asmagill:
Per discussion in #1356 

Replaces existing hs.ipc and cli tool.


asmagill:
Has anyone had problems with this?  I believe it's ready as well.

asmagill:
This has worked reliably for me now for the past month... it is a significant change in how the command line tool works, but my tests show that it falls back gracefully (minus custom handlers, but I'm not sure if anyone was using those) if there is a version mismatch.

Unless there are objections, I believe this too can be merged per #1433

cmsj:
+1 from me

Issue #1391, opened 2017-05-02
hs.hotkey.allowed(mods, key)


latenitefilms:
Following on from #1311, although the new error messages are really handy, I would like to avoid these errors in the console.

What are the chances of adding a `hs.hotkey.allowed(mods, key)` function that returns `true` if the modifier/keycode combination can be used without conflicting with macOS, otherwise `false`?

Thoughts @asmagill & @cmsj ?

asmagill:
This is something I've thought about... a possible solution would be to check the results of `CopySymbolicHotKeys` (see `/System/Library/Frameworks/Carbon.framework/Frameworks/HIToolbox.framework/Headers/CarbonEvents.h`) , but I've not run any real tests to see if it's truly thorough or not.

I won't have time to work on it myself for a while, but I add the following in case anyone wants to follow up:

The following is a snippit from https://github.com/asmagill/hammerspoon_asm/tree/master/extras that shows what the result structure looks like (it would need some parsing, but should be doable):

~~~objc
static int hotkeys(lua_State *L) {
    LuaSkin *skin = [LuaSkin shared] ;
    CFArrayRef hotkeys = NULL ;
    OSStatus status = CopySymbolicHotKeys(&hotkeys) ;
    if (status != noErr) {
        lua_pushinteger(L, status) ;
    } else if (hotkeys) {
        [skin pushNSObject:(__bridge_transfer NSArray *)hotkeys withOptions:LS_NSDescribeUnknownTypes] ;
    } else {
        lua_pushnil(L) ;
    }
    return 1 ;
}
~~~

This returns the following on my machine:

~~~lua
> inspect(_xtras.hotkeys())
{ {
    kHISymbolicHotKeyCode = 53,
    kHISymbolicHotKeyEnabled = true,
    kHISymbolicHotKeyModifiers = 2304
  }, {
    kHISymbolicHotKeyCode = 48,
    kHISymbolicHotKeyEnabled = true,
    kHISymbolicHotKeyModifiers = 256
  }, {
    kHISymbolicHotKeyCode = 48,
    kHISymbolicHotKeyEnabled = true,
    kHISymbolicHotKeyModifiers = 768
  }, {
    kHISymbolicHotKeyCode = 1,
    kHISymbolicHotKeyEnabled = true,
    kHISymbolicHotKeyModifiers = 4352
  }, {
    kHISymbolicHotKeyCode = 53,
    kHISymbolicHotKeyEnabled = true,
[...]

Issue #1426, opened 2017-05-15
hs.logger updates?


asmagill:
In conjunction with #1425, hs.logger may need some review, but it is listed here as a separate issue in case there are other hs.logger specific tweaks to be added to the mix

* the initial history size is 0.  (see https://github.com/Hammerspoon/hammerspoon/blob/master/extensions/logger/init.lua#L68) As such,  even if the first line in your `init.lua` file changes this, some informational logs that occur during Hammerspoon's initial loading will be lost as we move core module status notification from `print` to `hs.logger`.

cmsj:
I think it makes sense to increase this if we're going to move core messages into hs.logger. We're talking about strings here, so almost any number is perfectly reasonable :)

asmagill:
* LuaSkin log methods need an overhaul -- some thoughts
  * first, remove [LuaSkin logAtLevel:withMessage:fromStackPos:] -- it never did what I had hoped (we're usually not in a good place for a useful traceback when its invoked) and I think only one module I've written actually uses it (and with no useful information added because of it) -- and I doubt anyone else has used it at all
  * modify the delegate so its callback keeps a table of `hs.logger` objects for each module (will probably need some mods to `hs.logger` for this as well since we'll want the lua side of each module to use it as well) or create one if it doesn't exist.  This will allow proper directing of lua skin log messages if we also:
  * add a tag argument to the log methods which will contain the USERDATA_TAG which most modules have (and the few that don't can have added easily) so that the delegate callback can pick the right object to log the messages with
  * add support for varargs to the log methods similar to NSLog?  A large portion of our log messages are sent with [NSString stringWithFormat:@"...", ...] as the string.

@cmsj, how set do you consider the delegate and logging methods of LuaSkin?  If I start playing around with it trying to implement some of the above next week, will I break anything else you're using LuaSkin for or planning to use it for?

Issue #1425, opened 2017-05-15
Change autoload functionality


asmagill:
Per a variety of discussions, I am entering this as a new issue so it can be tracked/addressed directly:

It has been proposed:

* Remove the ability to disable autoloading of modules.  This has always been an undocumented feature and untested in quite some time.  At present it is known that at least one one scenario of requiring `hs.window`, `hs.application`, and `hs.uielement` will fail to load properly without autoloading.

* move autoloading console messages into `hs.logger`

* examine other uses of `print` for status in core modules and move these into `hs.logger` as well.

Have I stated the intention correctly?

cmsj:
seems like an accurate description to me, and I am +1 for all of these things :)

asmagill:
@cmsj, should the print statements in `setup.lua` be handled through the default logger instance or should using the use of `hs.logger` only start when `hs._coresetup` is loaded?

cmsj:
@asmagill hmm, I think it might well be a good idea to use hs.logger only after _coresetup is done - better to sacrifice a few lines of logging flexibility than to introduce dependencies to _coresetup imo.

asmagill:
`hs._coresetup` already requires logger when it establishes the logging callback delegate currently used by LuaSkin.

This proposal just makes that use universal for any status/debug/detail output generated by Hammerspoon or it's modules so that the user can make a choice as to whether they want to see it or not.

Coupled with the LuaSkin logging changes proposed in #1426, you would be able to select this on a per module basis.

cmsj:
Good point, +1 :)

PR #1444, opened 2017-05-30
adds hs.hotkey.assigned

asmagill:
sort of addresses #1391 

This isn't really a conclusive test (for example, both `Cmd-F5` and `Ctrl-1` are defined on my system and will return a table with this function; however, if I create a Hammerspoon hotkey for `Cmd-F5`, it works, while a Hammerspoon hotkey for`Ctrl-1` fails.)

What's the group consensus?  It's useful to see if you're trying to overwrite a system hotkey, but... it isn't a reliable way to tell if you *can* or not.

> Function: hs.hotkey.assigned(mods, key) -> table | false
>
> Examine whether a potential hotkey is in use by the macOS system such as the Screen Capture, Universal Access, and Keyboard Navigation keys.
>
> Parameters:
>  * mods - A table or a string containing (as elements, or as substrings with any separator) the keyboard modifiers required,
>    which should be zero or more of the following:
>    * "cmd", "command" or "⌘"
>    * "ctrl", "control" or "⌃"
>    * "alt", "option" or "⌥"
>    * "shift" or "⇧"
>  * key - A string containing the name of a keyboard key (as found in [hs.keycodes.map](hs.keycodes.html#map) ), or a raw keycode number
>
> Returns:
>  * if the hotkey combination is in use by a system function, returns a table containing the following keys:
>    * keycode - the numberic keycode for the hotkey
>    * mods    - a numeric representation of the modifier flags for the htokey
>    * enabled - a boolean indicating whether or not the key is currently enabled
>  * if the hotkey combination is not in use by the operating system, returns the boolean value `false`
>
> Notes:
>  * this is provided for informational purposes and does not provide a reliable test as to whether or not Hammerspoon can use the combination to create a custom hotkey -- some combinations which return a table can be over-ridden by Hammerspoon while others cannot.  At this time there is no known reliable test except to assign the hotkey with `hs.hotkey.enable` and see if it works.

### DIFF
--- a/extensions/hotkey/init.lua
+++ b/extensions/hotkey/init.lua
@@ -70,10 +70,13 @@ local function enable(self,force,isModal)
   local returnVal = self._hk:enable() --objc
   if returnVal ~= nil then
     log[isModal and 'df' or 'f']('Enabled hotkey %s%s',self.msg,isModal and ' (in modal)' or '')
-  end
     tinsert(hotkeys[idx],self) -- bring to the top of the stack
 --   return returnVal
     return self
+  else
+    self.enabled = false
+    return nil
+  end
 end
 
 --- hs.hotkey:disable() -> hs.hotkey object
@@ -178,7 +181,7 @@ end
 ---  * repeatfn - A function that will be called when a pressed hotkey is repeating, or nil
 ---
 --- Returns:
----  * A new `hs.hotkey` object
+---  * A new `hs.hotkey` object or nil if the hotkey could not be enabled
 ---
 --- Notes:
 ---  * You can create multiple `hs.hotkey` objects for the same keyboard combination, but only one can be active
@@ -219,7 +222,7 @@ function hotkey.new(mods, key, message, pressedfn, releasedfn, repeatfn)
   return hk
 end
 
---- hs.hotkey.assigned(mods, key) -> table | false
+--- hs.hotkey.systemAssigned(mods, key) -> table | false
 --- Function
 --- Examine whether a potential hotkey is in use by the macOS system such as the Screen Capture, Universal Access, and Keyboard Navigation keys.
 ---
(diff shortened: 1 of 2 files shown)


cmsj:
I don't have any objections, but I wonder if this could be implemented as a wrapper around `hs.hotkey.new()` that tries to bind the hotkey, and removes it if it succeeded?

asmagill:
@cmsj, as noted above, just because this function returns a table doesn't reliably determine whether or not Hammerspoon *can* override it.  If this function returns `false`, then the hotkey is safe to use... if it returns a table, the hotkey may or may not be safe to use.

Since Hamemrspoon *can* over-ride some of the combinations this returns a table for, do we really want to prevent it?  And if we do add the check, how many existing user setups will break because of it?

At the very least, if you do want the check added, there should be someway to tell `.new` or `.bind` to ignore the results of the check.

cmsj:
@asmagill I didn't mean that we would bake this check into .new()/.bind(), more that .assigned() (or maybe .isAvailable() ?) would call .new() and see if it works or not, unbinding it if it did work.

asmagill:
Because of the heavy additions made to `hs.hotkey` to handle stacking assignments of the same hotkey and adding the inline labels/help, it's not intuitively obvious, but this can already be done with the existing infrastructure with something like (I've not tested this as a function, but the doing this by hand with a combination that I know works and a combination that I know fails works as expected, so I am fairly confident the final version will be pretty similar):

~~~lua
hotkey.isAssignable = function(...)
    local k = hotkey.new(...)
    local prevLevel = hs.luaSkinLog.level
    -- supress luaSkinLog error if binding fails
    hs.luaSkinLog.level = 0
    local status = k._hk:enable()
    if status then k._hk:disable() end
    hs.luaSkinLog.level = prevLevel
    return status and true or false
end
~~~

Should this be added instead?  Or are both worth having?  The earlier version could still be used to determine if the OS is using the hotkey combination -- just because someone *can* override some of these doesn't necessarily mean that they will *want* to.

asmagill:
@cmsj, clarify your position on my previous comment -- this could be added to #1433 fairly easily.

cmsj:
@asmagill yeah that's the kind of thing I meant :)

asmagill:
Unless you object, I'm going to move the function which checks against system definitions to `hs.hotkey.systemAssigned` and add the example function above as `hs.hotkey.assigned`.

Should `hs.hotkey.new` (which is already a wrapper to `hs.hotkey.bind(...):enable()`) do something similar and return `nil` on failure?  This would be a change in behavior because currently `new` returns a hotkey table object, just one that isn't/can't be enabled if binding fails.

cmsj:
@asmagill hmm, that's a tough one. I think returning nil would definitely be more correct, but nobody is going to be checking the return type, so it might make a lot of things explode.

I *think* it's probably better to move the API towards being useful though, so let's try it and see what happens :D

asmagill:
`hs.hotkey.systemAssigned` and `hs.hotkey.assignable` added

`hs.hotkey.new(...):enable()` (also known as `hs.hotkey.bind(...)`) will return `nil` if activation fails, but still displays the error messages to the console (`assignable` suppresses them)... this is because the most common use of `hs.hotkey.bind` is to simply execute it and ignore the return value.

If you really want to suppress the messages no matter what, either check with `hs.hotkey.assignable` before binding or wrap your hotkey assignments with:

~~~lua
local prevLevel = hs.luaSkinLog.level
hs.luaSkinLog.level = 0

-- do hotkey assignments as usual

hs.luaSkinLog.level = prevLevel
~~~

Note that in the future, it is hoped that these will be sent to a `hs.hotkey` specific log, so suppressing them can be more local than by temporarily suppressing *all* error messages coming from objc (hence the reassignment to `prevLevel` after assignments) but #1425 and #1426 aren't going to be completed in time for the next release.

PR #1457, opened 2017-06-10
Logging updates to LuaSkin

asmagill:
* makes run-of-the-mill logs from `hs.socket` less obtrusive
* remove –logAtLevel:withMessage:fromStackPos:

Basically prep work for #1426 and #1425; more to come, not ready for merge yet.


asmagill:
This does clean up the unnecessary log display for hs.socket (moves connection success, etc. to a lower level), and removes an unnecessary method from LuaSkin, so I believe it should be added per #1433, even though the additional log changes haven't been made yet... Objections?

Issue #1458, opened 2017-06-11
invalid percentage string specified for field w of frame for canvas element 


ashfinal:
I wrote a quite nice desktop widget, which works well on my computer, but not on my friend's. 
It says: 
```
2017-06-09 21:42:18: 21:42:18 ERROR:   LuaSkin: hs.canvas:invalid percentage string specified for field w of frame for element 6
2017-06-09 21:42:18: ********
2017-06-09 21:42:18: ********
2017-06-09 21:42:18: 21:42:18 ERROR:   LuaSkin: hs.timer callback error: ...oon.app/Contents/Resources/extensions/hs/canvas/init.lua:525: bad argument #3 to 'elementAttribute' (invalid argument type for frame specified)
stack traceback:
	[C]: in method 'elementAttribute'
	...oon.app/Contents/Resources/extensions/hs/canvas/init.lua:525: in metamethod '__newindex'
	/Users/user/.hammerspoon/widgets/timelapsed.lua:116: in function 'updateElapsedCanvas'
	/Users/user/.hammerspoon/widgets/timelapsed.lua:122: in function </Users/hien/.hammerspoon/widgets/timelapsed.lua:122>
2017-06-09 21:42:18: ********
```
Here's [the code](https://github.com/ashfinal/awesome-hammerspoon/raw/master/widgets/timelapsed.lua) and [background image resouce](https://raw.githubusercontent.com/ashfinal/awesome-hammerspoon/2ec05404/resources/timebg.png) if you want to test it on your own computer.

And there is more info if you're interested. https://github.com/ashfinal/awesome-hammerspoon/issues/34

asmagill:
The conversion from percentages uses the NSNumberFormatter class and attempts to identify the number as matching the current locale's decimal format (NSNumberFormatterDecimalStyle) or as the current locale expresses percentages (NSNumberFormatterPercentStyle).

So, what is your friend's locale, and does temporarily changing it to match yours fix the issue?

In any case, this *could* be a problem for this very reason in the future, so I'll see about adding code in the next couple of days so that this conversion code falls back to `en_US` if the current locale's conversions yield nothing.

ashfinal:
Thank you for the reply. It may need some time to confirm this.

biocyberman:
@asmagill It is indeed the cause of the problem. I changed `region` setting and it works

ashfinal:
Nice!

PR #1459, opened 2017-06-12
canvas: add percentage conversion fallback

asmagill:
Fallback to en_US locale for percentage conversion if using the current locale fails.

Test case: set a locale that uses "," as the decimal separator (e.g. "fr-US"); if you set a percentage through lua division (e.g. `x = tostring(10/100)`), lua isn't locale aware and generates "0.10" which the current code fails on.

This fix adds a fallback to "en_US" if the current locale doesn't recognize the percentage string given.

Addresses #1458


asmagill:
@cmsj, objections?

cmsj:
@asmagill lgtm

Issue #1466, opened 2017-06-21
Hammerspoon crashes when using hs.base64.encode on an image


zzamboni:
```
> img=hs.image.imageFromURL("http://www.hammerspoon.org/images/hammerspoon.png")


> img
hs.image:  (0x608000442188)

> hs.base64.encode(img)
-- Loading extension: base64
2017-06-21 09:24:30.215 hs[47154:10290197] *** CFMessagePort: dropping corrupt reply Mach message (0b000100)
error communicating with Hammerspoon: message port was invalidated
error unregistering CLI instance with Hammerspoon: message port invalid
```

asmagill:
admittedly the crashing is an issue I'll have to look into, but `hs.base64.encode` isn't intended for converting userdata objects (which `img` is in your example).  If you're needing to convert an image into a string (for storage, transport, etc.), use `hs.image.encodeAsURLString`

zzamboni:
@asmagill thanks! `encodeAsURLString` is exactly what I was looking for - I remembered seeing it somewhere but couldn't find it. Works perfectly, but I agree it would be nice to fix this crash.

cmsj:
+1 for leaving hs.base64 away from trying to deal with random userdata objects
+1 for fixing the crash when someone tries it anyway :D

PR #1478, opened 2017-07-02
base64.encode triggers tostring for non-string input

asmagill:
addresses #1466 

this fix invokes the "__tostring" metamethod on anything that isn't a string or number, so using an image like `hs.base64.encode(hs.image.imageFromName(hs.image.systemImageNames.ActionTemplate))` actually encodes the string "hs.image: NSActionTemplate (0x6000010542a8)" (the address will differ).

Is this sufficient or should it generate an error instead?

### DIFF
--- a/extensions/base64/init.lua
+++ b/extensions/base64/init.lua
@@ -47,7 +47,7 @@ end
 --- Returns:
 ---  * A string containing the decoded data
 module.decode = function(data)
-    return module._decode(data:gsub("[\r\n]+",""))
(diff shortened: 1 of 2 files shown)


cmsj:
@asmagill I think it might be best to generate an error, otherwise people might think they are getting a base64 encode of the image, and then be surprised when it doesn't decode to an image.

asmagill:
Changes requested have been applied.

asmagill:
Objections?  I think this should be included in the next release for #1433.

cmsj:
No objections

Issue #1494, opened 2017-07-18
Can't create middle mouse button press with 0.9.55


2shortplanks:
In 0.9.54 the way to press the middle mouse button down was to start with this:

```
ev = hs.eventtap.event.newMouseEvent(
            hs.eventtap.event.types.middleMouseDown,
            hs.mouse.getAbsolutePosition(),
            {}
        )
```

This no longer works in 0.9.55, giving this error:

```
...n.app/Contents/Resources/extensions/hs/eventtap/init.lua:131: bad argument #1 to '_newMouseEvent' (number expected, got nil)
stack traceback:
	[C]: in function 'hs.eventtap.event._newMouseEvent'
	...n.app/Contents/Resources/extensions/hs/eventtap/init.lua:131: in function 'hs.eventtap.event.newMouseEvent'
	[string "ev = hs.eventtap.event.newMouseEvent(..."]:1: in main chunk
	[C]: in function 'xpcall'
	...app/Contents/Resources/extensions/hs/_coresetup/init.lua:401: in function <...app/Contents/Resources/extensions/hs/_coresetup/init.lua:381>
```

Whoops, I think 198997a818b5097173914a771f2dd62353f89660 caused this.  It looks like the intent was to remove the mapping from event code to the old `middleMouseDown` string (so it could be mapped to the `otherMouseDown` string instead) but instead the commit accidentally removed mapping `middleMouseDown` to the event code instead...breaking newMouseEvent.

To compound matters, this doesn't work either:

```
ev = hs.eventtap.event.newMouseEvent(
            hs.eventtap.event.types.otherMouseDown,
            hs.mouse.getAbsolutePosition(),
            {}
        )
```

We get this error instead:

```
2017-07-18 12:18:39: Error: unrecognized mouse button eventtype: 25
```

Presumably this is because the [newMouseEvent code](https://github.com/Hammerspoon/hammerspoon/blob/a796c2718ff670e131b6776f2dc0091ba6342af0/extensions/eventtap/init.lua#L125) is still using the accidentally removed `middleMouseDown` in the if statement.  This would be fixed by restoring the `middleMouseDown` -> `25` mapping in `hs.eventtap.event.types`, but probably should be changed over to the new string for completeness also.

(I'd send you a pull request for this, but I haven't got round to working out how to build hammerspoon in Xcode yet)

PR #1495, opened 2017-07-19
clean up missing other/middle click changes in hs.eventtap

asmagill:
Should address #1494, though I won't be able to fully test it until tomorrow.

Moves `hs.eventtap.middleClick` to `hs.eventtap.otherClick` (though the former name can still be used as it references the same function) and adds support in the new function for specifying the specific "other" mouse button.


cmsj:
darn it, I thought we got this right the first time. Thanks for taking it on

extensions/eventtap/init.lua
```diff
@@ -177,29 +177,40 @@ function module.rightClick(point, delay)
...
 --- Returns:
 ---  * None
 ---
 --- Notes:
----  * This is a wrapper around `hs.eventtap.event.newMouseEvent` that sends `middlemousedown` and `middlemouseup` events)
-function module.middleClick(point, delay)
+---  * This is a wrapper around `hs.eventtap.event.newMouseEvent` that sends `otherMouseDown` and `otherMouseUp` events)
+---  * For backwards compatibility, `hs.eventtap.middleClick` can also be used to invoke this function.
+---
+---  * OS X recognizes up to 32 distinct mouse buttons, though few mouse devices have more than 3.  The left mouse button corresponds to button number 0 and the right mouse button corresponds to 1;  distinct events are used for these mouse buttons, so you should use `hs.eventtap.leftClick` and `hs.eventtap.rightClick` respectively.  All other mouse buttons are coalesced into the `otherMouse` events and are distinguished by specifying the specific button with the `mouseEventButtonNumber` property, which this function does for you.
```
2shortplanks: OS X is now called macOS

asmagill: Simple enough to adjust when I see if other changes are also required.

@cmsj, doing a pattern search on `docs.json` with `\s[oO][sS]\s?[xX]\s` shows that it appears 208 times in the documentation... do we care enough to open an issue to change this in the future, or do we just want to try and be cognizant of the name change going forward?

extensions/eventtap/init.lua
```diff
@@ -177,29 +177,41 @@ function module.rightClick(point, delay)
     module.event.newMouseEvent(module.event.types["rightMouseUp"], point):post()
 end
 
---- hs.eventtap.middleClick(point[, delay])
```
mongito: What about keeping the middleClick wrapper (now just calling the other click) for backwards compatibility and as a usability matter as middleClick is a well known event (unambiguous name for clicking with mouse button 2).

asmagill: if you leave the button number off of `hs.eventtap.otherClick`, it defaults to 2, which is the middle click.

asmagill: @mongito or are you referring to leaving in the documentation for middleClick?  My thought was keep it as a valid function, hence setting middleClick = otherClick, so existing code doesn't break, but encourage future use of the correct function by documenting only the new function since it's more in line with apple's naming conventions.

mongito: I was referring mainly about the function, but documentation should still exist for it. I see your point in aligning with apple's naming conventions, maybe you can refer to that in the middleClick documentation.

asmagill:
If travis doesn't barf, this is good to go

Issue #1227, opened 2017-02-08
Feature Request - hs.dialog()


latenitefilms:
I currently use `hs.osascript.applescript()` to create things like dialog boxes, colour pickers, file browsers, etc, which works great - HOWEVER, because of [the weirdness](https://github.com/Hammerspoon/hammerspoon/issues/1178) that happens between Hammerspoon & AppleScript, I'd love to come up with something a bit more reliable. For example, when I create a dialog box, for some strange reason on Sierra, I need to click the dialog box button a few times before it actually gets clicked. Maybe it's also related to [this](https://github.com/Hammerspoon/hammerspoon/issues/1180)?

What are the chances of someone here making up a extension that allows you to create various dialog boxes using Objective-C?

These links might be relevant:

https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/Dialog/Tasks/UsingAlerts.html
https://developer.apple.com/library/content/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/UsingtheOpenandSavePanels/UsingtheOpenandSavePanels.html
https://developer.apple.com/reference/appkit/nscolorpanel

For anyone who's interested, you can see how we're doing it with AppleScript [here](https://github.com/CommandPost/CommandPost/blob/develop/src/cp/dialog/init.lua). It would be really awesome to do something similar, but use Objective-C instead.

Thoughts?

In terms of "controls" it would be amazing to be able to offer all the same functionality you currently get with AppleScript - for example:
<img width="525" alt="screen shot 2017-02-08 at 10 47 17 pm" src="https://cloud.githubusercontent.com/assets/22286696/22736059/a1f02e86-ee50-11e6-8e2e-4dbbc61e65d8.png">
<img width="525" alt="screen shot 2017-02-08 at 10 47 19 pm" src="https://cloud.githubusercontent.com/assets/22286696/22736060/a2495966-ee50-11e6-8e13-1834e3201d83.png">
<img width="525" alt="screen shot 2017-02-08 at 10 47 22 pm" src="https://cloud.githubusercontent.com/assets/22286696/22736061/a24da7f0-ee50-11e6-98a2-8767be88a93e.png">

cmsj:
I think it's super unlikely that anyone has the time at the moment to work on something like those Dialog Maker examples, but I can see us slowly building up an `hs.dialog` with really simple single-purpose dialogs (so like a text entry one, a color picker one, a file requester one, etc).

[...]

PR #1504, opened 2017-07-27
Issue #1227 - Initial Release of hs.dialog

latenitefilms:


latenitefilms:
@asmagill & @cmsj - Some questions for you:

I believe it's possible to have a `NSAlert` only impact a specific window - for example:

`[alert beginSheetModalForWindow:window modalDelegate:self didEndSelector:@selector(alertDidEnd:returnCode:contextInfo:) contextInfo:nil];`

Given this, do you reckon it would be possible for me to code something so that the user could supply `hs.webview:hswindow()` and the alert would only "block" that specific webview? Any ideas on how I can achieve this? I can't think of any examples of existing extensions that accept a `hs.window` as a parameter in Objective-C land - but if there are some examples, let me know so I can attempt some copying and pasting!

The original reason I wanted to build this extension was to avoid AppleScript (which seems to cause all kinds of delays/issues with Hammerspoon) - but also to avoid blocking (as currently if I have an AppleScript alert displayed - our `hs.menubar` won't respond until the alert is closed). Obviously the code currently committed doesn't solve this. @cmsj mentioned "for non-alerts you'd likely want a simple NSWindow with a textfield for the message, and various buttons" - so I'm thinking of having a `hs.dialog.displayMessage()` that basically does the same thing as `hs.dialog.displayAlertMessage()` except that it's non-blocking and offers a callback. Do you know of any examples to help me pull this together?

Note to self:
- [Customising NSAlert: accessory views](http://1klb.com/posts/2013/12/18/customising-nsalert-accessory-views/)
- [NSAlert – Alerts in Cocoa, Objective-C](http://www.knowstack.com/nsalert-cocoa-objective-c/)
- [NSAlert.h](https://sourceforge.net/p/lazarus-ccr/svn/364/tree/bindings/pascocoa/appkit/NSAlert.inc)

asmagill:
First, using an `hs.window` object in another module is probably going to be problematic... `hs.window` doesn't deal with NSWindow or NSPanel objects but rather is a thin wrapper over an AXUIElementRef -- the reason for this is that Hammerspoon only has access to the NSWindow and NSPanel objects that it itself creates -- to work with the windows of other applications, we have to use the accessibility API.

`hs.drawing`, `hs.canvas`, and `hs.webview` (maybe a couple of others that create UI elements do as well, but it's more hidden in their cases) do all use NSWindow/NSPanel objects which another module could gain access to in a way that you could use `[alert beginSheetModalForWindow:window modalDelegate:self didEndSelector:@selector(alertDidEnd:returnCode:contextInfo:) contextInfo:nil];` with.

However, it might be easier to just create your own NSWindow object and attach the sheet to it... that way if you find that the window needs to have certain properties set (is there a minimum size/how does the window size affect the appearance of the sheet, etc) you can just set them yourself.

I do recall having to create modal sheets for `hs.webview` for the javascript dialog support and HTTP Basic Auth login dialogs... I admit that the `hs.webview` code isn't pretty, but you may find some pointers/ideas in there that could help.

As to creating dialog boxes from scratch with their own arbitrary (or even a specific set of predefined) layouts is something I've thought about doing, but never really tried my hand at it... I may have saved some links (I have a bookmark folder of things that look interesting to me as possible additions or inspiration for Hammerspoon modules that I want to consider but don't want to work on right away) that could be useful... let me review them and see what jumps out.

latenitefilms:
Sorry for the delayed reply - busy week! Thanks HEAPS for all the information @asmagill - HUGELY appreciated!

Maybe rather than the `hs.dialog.displayAlertMessage` accepting a `hs.window` it should accept a `hs.webview` instead, and this functionality should only work for webviews (as this is all I really need it for anyway)? Any hints on how I can achieve this in Objective-C land (i.e. if someone "submits" a `hs.webview` in Lua-land, how do I get a `window` object from this in Objective-C land)?

asmagill:
Is there a particular reason to use `hs.webview`?  Or do you just need a window to attach the sheet to?

I ask because if you need to add any property to the webview object (like a callback for the alert's response) then you'll have to add this to the webview module itself.  You could use a class category to add any necessary methods, and this wouldn't necessarily require modifying or adding to the webview module code, but if they override any of the existing methods, it gets ... complicated.

Unless you need to leverage something specific about the `hs.webview` module, I think it will be a whole lot easier to build your own NSWindow/NSPanel and use that instead as the anchor... as the webview code is fairly complex, it might be easier to look at an example that's a little simpler... check out https://github.com/asmagill/hammerspoon_asm/tree/master/avplayer which may be easier to follow since it's simpler.  Don't worry about any of the `HSAVPlayerView` code -- you need a window, not a view.

latenitefilms:
Let's say for example, I've built a Preferences panel using `hs.webview` (see attached screenshot). When a user clicks one of the buttons, I want it to display an alert that PREVENTS the user from doing any further interaction with the Preference panel. I'm currently using an AppleScript dialog that obviously doesn't do this - but I'm HOPING I can do this using the methods explained above.

Does that make sense @asmagill ?

screen shot 2017-08-03 at 7 27 36 am

asmagill:
Hmm... Ok probably the way I would approach this is something like the following (I've not tested this, so I can't promise you won't need to tweak it some, but I'm trying to avoid adding code to `hs.webview` -- it's already pretty large and complex in its own right)...

First, in your code, create a class which is a subclass of NSObject... in this new class, you'll need two properties: an `int` to hold the lua reference for the callback function and an `NSAlert` to hold the actual alert object itself.  This object will be what comprises the userdata object that your module creates and methods are called from.  Then, to attach it to the webview window, I *think* something like the following should work:

~~~objc
static int alert_runFromWindow(lua_State *L) {
    LuaSkin* skin = [LuaSkin shared] ;
    [skin checkArgs:LS_TUSERDATA, USERDATA_TAG, LS_TUSERDATA, "hs.webview", LS_TBREAK] ;
    AlertObj* obj = ... whichever method you choose to get your object from the userdata ... ;
    NSWindow *win = [skin toNSObjectAtIndex:2] ; // the built in conversion tools will actually
        // return an HSWebViewWindow object, but since this is a subclass of
        // NSPanel which is itself a subclass of NSWindow and we're not actually
        // going to do  anything with the webview specific methods it should work
    [obj.alert beginSheetModalForWindow:win completionHandler:^(NSModalResponse returnCode){
        ... code which invokes callback with the return code ...
    }] ;
    lua_pushvalue(L, 1) ;
    return 1 ;
}
~~~

If you need an example of creating a subclass of NSObject and using it as a userdata, I think I recall `hs.sharing` as being relatively straightforward... You could also take a look at https://github.com/asmagill/hammerspoon_asm/blob/master/internal.m_template which is the template I use a lot that has examples of the support functions which make userdata objects work within lua.

If this does in fact work, then the good news is that we can add a new type of "window" to attach it to (like `hs.canvas` for example) by copying this functionand  just changing the `checkArgs:` line and *maybe* the `NSWindow* win` line depending upon how that particular module encapsulates it's window object.

latenitefilms:
Amazing - thank you! Will have a play and see what I can come up with!

latenitefilms:
FYI: Although I haven't spent a huge amount of time on it, I was struggling a bit with your suggestions @asmagill in regards to "attaching" an alert to a webview. I'm still happy to keep bashing along with it - but feel free to jump in and make any tweaks you want, as you'll be a lot faster than me!

Also, @cmsj - feel free to jump in too, and add a colour picker function if you want (as discussed in #1517).

I do think `hs.dialog`could be REALLY handy for the wider Hammerspoon community.

latenitefilms:
@asmagill - I've managed to get the alert appearing in the `hs.webview` which is AWESOME! But I can't work out how to get the callback working. Any chance you could have a quick look and let me know what I need to go to finish this off? Thanks mate!

https://github.com/Hammerspoon/hammerspoon/pull/1504/commits/c71c5df04b42c604ed89c8937e3a3f6046a7dcf8#diff-d062c31625ef9604fbdc551cfe5592e4R170

extensions/dialog/internal.m
```diff
@@ -0,0 +1,388 @@
...
+            [alert setAlertStyle:NSWarningAlertStyle];
+        }
+    }
+    
+    [alert beginSheetModalForWindow:webview completionHandler:^(NSModalResponse result){
+        if (result == NSAlertFirstButtonReturn) {
+            if (buttonOne == nil) {
+                [LuaSkin logInfo:@"hs.dialog.webviewAlert() - Default button pressed during callback."];
+                
+                // @asmagill - what do I need to put here to trigger the callback?
```
asmagill: Something along the lines of

~~~objc
    lua_pushvalue(L, 2) ; // push the function passed in as an argument to the top of the stack
    [skin pushNSObject:defaultButton]; // or whatever arguments you want passed to the callback
    if (![skin protectedCallAndTraceback:1 nresults:0]) { // returns NO on error, so we check if the result is !YES
        [skin logError:[NSString stringWithFormat:@"hs.dialog:callback error - %s", lua_tostring(L, -1)] ; // -1 indicates the top item of the stack, which will be an error message string in this case
        lua_pop(L, 1) ; // remove the error from the stack to keep it clean
    }
~~~

should probably work.

Since you're calling the same callback with basically the same format, just a different argument, I'd probably use the if-then-else-etc to set an NSString containing the name of the button (i.e. defaultButton or buttonOne or etc.) and then only do the callback code once with that variable after the if-then-else block, just to reduce repetition and chance of typos.

asmagill: The only thing to watch for is I don't know what thread alert will run its code it... I *think* the main one, but if you experience some unexpected crashes, check the Console and User crashlog... if the crash is in any thread but thread 0, wrap the whole thing in 

~~~objc
    ...  the above code for doing the callback ...
}) ;
~~~

latenitefilms: @asmagill - I tried using `lua_pushvalue(L, 2)`, but it just gives me: 

```
2017-08-30 14:29:15: ********
2017-08-30 14:29:15: 14:29:15 ERROR:   LuaSkin: hs.dialog:callback error - attempt to call a nil value
stack traceback:
2017-08-30 14:29:15: ********
```

I've updated the branch to reflect the changed code. Any ideas? I'm probably just doing something silly.

Although I haven't seen any crashes, I did try wrapping everything in `dispatch_async` - but it gave me some Xcode error, so I'll just leave it for now.

asmagill:
My bad... I forgot that the NSAlert block is running asynchronously so it's not "in" the function anymore when the callback occurs.  Simplest way, store the function in the refTable before invoking the alert, then pull it back out within the block... see the comments I'm about to add since it involves putting code in two different places...

extensions/dialog/internal.m
```diff
@@ -0,0 +1,388 @@
...
+ */
+static int webviewAlert(lua_State *L) {
+    
+    NSString* defaultButton = @"OK";
+    
+    LuaSkin *skin = [LuaSkin shared];
+    [skin checkArgs:LS_TUSERDATA, "hs.webview", LS_TFUNCTION, LS_TSTRING, LS_TSTRING, LS_TSTRING | LS_TOPTIONAL, LS_TSTRING | LS_TOPTIONAL, LS_TSTRING | LS_TOPTIONAL, LS_TBREAK];
+    
+    NSWindow *webview = [skin toNSObjectAtIndex:1];
+    //NSObject *callback = [skin toNSObjectAtIndex:2];
```
asmagill:
~~~objc
    lua_pushvalue(L, 2) ; // copy the function to the top of the stack
    int callbackRef = [skin luaRef:refTable] ; // store what's at the top of the stack in the registry and save it's reference number
~~~

extensions/dialog/internal.m
```diff
@@ -0,0 +1,388 @@
...
+        else if (result == NSAlertSecondButtonReturn) {
+            button = buttonTwo;
+        }
+        else
+        {
+            [LuaSkin logError:@"hs.dialog.webviewAlert() - Failed to detect which button was pressed."];
+            lua_pushnil(L) ;
+        }
+        
+        lua_pushvalue(L, 2) ; // push the function passed in as an argument to the top of the stack
```
asmagill: remove this lua_pushvalue line and replace it with:

~~~objc
    [skin pushLuaRef:refTable ref:callbackRef] ; // put the saved function back on the stack
    [skin luaUnref:refTable ref:callbackRef] ; // remove the stored function from the registry; normally we'd do this in a userdata's __gc, but since we're not using a userdata for this, we need to do so within the NSAlert callback context... could go at the end, but since it won't affect the stack at all, I'd do it here so we don't forget
~~~

extensions/dialog/internal.m
asmagill: Note `luaRef:` will pull off the top value of the stack, so the net effect of these two lines is to leave the stack of arguments as-is.

latenitefilms:
Thanks so much for all your help @asmagill - HUGELY helpful!

Any chance you could also help me with the Colour Picker code?

https://github.com/Hammerspoon/hammerspoon/pull/1504/commits/990843727f9a5219a19f288e97772242886ead01#diff-d062c31625ef9604fbdc551cfe5592e4R15

asmagill:
Will test this tonight/tomorrow and take a look at the color code then as well.  Looking forward to trying this out.

asmagill:
I thought I remembered working on a color panel object at one time, and I did... the code can be found at https://github.com/asmagill/hammerspoon_asm/tree/master/panels and you're more then welcome to take any or all of it and do what you want. (I did just clean it up a bit and updated some deprecated names, but otherwise have no immediate plans for it, so have at it).

The color stuff I'm pretty happy with, though by all means adjust or throw out anything you want to change.  I was never completely happy with the font panel stuff, and the callback for style changes isn't as clear as I'd like when removing styles, so I'd definitely recommend reviewing it pretty closely and maybe tweaking it if you want to include it as well.

And it's not documented.  Sorry about that, but a little experimentation (I've included a compiled build) and review of the code should make it pretty obvious what it does.

Programming notes (why I wrote it the way I did, besides the fact that my coding style has evolved some):

The color and font panels use a single shared object each for a given application -- the expectation is that you'll never have more then one open at a time (in fact, you have to jump through hoops to even *have* two separate color panels open at the same time and I'm not sure its really recommended or formally supported).  For this reason, it didn't seem to make sense to go the userdata object with constructor approach.

However, the panels don't support blocks, so the callbacks require an Objective-C object as the target for updates...  Usually that's what we stick into a userdata.  In the case with this code, the "receiver" object is created in the lua module init function (stored in `cpReceiverObject` and `fpReceiverObject`) and cleaned up in the module's global meta table __gc function (`releaseReceivers`).  Technically we could get away without releasing them -- on reload, they would get recreated and the old ones would be released through ARC, but it's good practice in case the same approach is used with something that could have side effects if you don't remove the callbacks, etc. manually when you no longer need them around.

At any rate, take a look at the code and see if it helps you with the color panel support you desire.

asmagill:
Some ideas/thoughts re the webview alert (and this is your module, so don't feel that you need to do these if you disagree):

1. allow buttons to be specified as `nil` -- this would allow you to create a single button alert with a style other than Warning -- you'd still need the place holders (the logic gets real ugly real fast to try and "guess" what the user intended, so I agree on requiring *something* in each spot) but it would offer more flexibility
2. allow "warning", "informational", or "critical" instead of having to type in the whole Objective-C string name... and maybe print a warning (or error, but at least a warning to the console) if they type in a style other than a "correct" one... granted it's just an icon difference ATM, but Apple may change that and if I want critical, but type in "NSCriticlAlertStyle" instead, I might not notice if I wasn't looking too closely.
3. a wrapper function that creates a webview object and tears it down, just to display an alert... e.g.
~~~lua
function wrapperAlert(loc, callback, ...)
    loc.h, loc.w = 10, 10 -- small enough to be hidden by the alert panel
    local wv = hs.webview.new(loc):show()
    hs.dialog.webviewAlert(wv, function(...)
        wv:delete()
        callback(...)
    end, ...)
end
~~~

Just some thoughts, but it's looking good!

asmagill:
And don't allow empty strings for button labels... I just did so and created an alert I can't get rid of without restarting (reloading isn't enough) Hammerspoon!

latenitefilms:
Legend, thanks so much!

I've had a quick look through all your panels code on my phone, and I THINK I get most of it, so I'll try and bring the same functionality across to this new extension. I still don't really fully understand Objective-C or LuaSkin properly - so we'll see how we go!

I agree with your thoughts and suggestions - so will try and do these things as well.

I'm obviously MORE than happy for you to tweak anything you want - or completely abuse my code. 99% of my code is copying and pasting from existing extensions anyway - so it's most likely all your own code anyway - just re-ordered and renamed.

Also, I realise my code is pretty basic, in that there are some messy if...then...else stuff going on - so feel free to suggest better ways to do anything. I'm SLOWLY learning how Objective-C works - but don't quite "get it" yet.

latenitefilms:
Any suggestions on how I fix these errors?

screen shot 2017-08-31 at 9 32 01 pm

asmagill:
#### The answer, I think:

 I *think* this is occurring because the warning/error flags in the `dialog` target differ from those used in the other targets (usually I copy an existing target when adding a new module, then modify the couple of sections necessary for the new code and rename the target... it looks like you created a new one or allowed XCode to auto-create it -- not wrong, just not the way I usually do it, so I think a few of the settings differ and this may be one of them... e.g.

For the dialog target:
dialog

For another (not dialog) target:
notdialog

I think it's just the one section that needs to be changed, but let me know if it doesn't seem sufficient and we can dig a little deeper; or instead you can change every case of `_someVar` to `self.someVar` in the class implementation, your choice.

#### Explanation of the underlying issue (you don't have to read this, but it should help in understanding, if you're interested):

When you define a property in an objective-c class interface, there are some things going on behind the scenes that have changed somewhat over the years... a line such as `@property int callbackRef ;` is actually doing a few things automatically for you:
1. Creates an instance variable within the class to hold the actual value: `int _callbackRef ;`
2. Automatically synthesizes a getter: `- (int)callbackRef { return self->_callbackRef ; }`
3. Automatically synthesizes a setter: `- (void)setCallbackRef:(int)value { self->_callbackRef = value ; }`
4. And some other things we don't care as much about like, especially when the property is for another objective-c object and not a simple type like `int`

An instance variable is *not* accessible from outside of the class methods and must be accessed through its getter/setter methods (which is why outside of the class definition you'll see `objc.callbackRef` which is really a shorthand for `[obj callbackRef]` and `objc.callbackRef = int` which is shorthand for `[obj setCallbackRef:int]`.

But inside, it is common practice (though not universal and some programming style choices do recommend against, though I personally disagree with most of them -- google if you care about it more) to access the properties through instance variables rather than through getter/setter methods because (a) it's slightly faster (by an unimaginably small and irrelevant amount 90% of the time, unless you're doing real-time audio or video -- see `hs.noises`), and (b) if you define a property as read-only (e.g. `@property (readonly) int callbackRef ;`) then it can *only* be set or changed within the class through it's instance variable name.

The other style you see in the code presented is `self->_instanceVarName`... this is used within blocks (i.e. within a `dispatch_async...` function) to explicitly let the compiler know that we *know* that it's going to have to create a reference to the object itself when creating the block -- as a simplified explanation, the block can be thought of as a separate function while `self->` is like a lua up-value, so we're telling the compiler that yes, we know we need to keep access to this non-local (to the block) variable... I'm going to guess that if you opt for the `self.callbackRef` approach rather than changing the warnings, you would loose this warning about outside references within blocks... personally I like this warning because it keeps me aware of everything's scope, but as I alluded to above, there are different schools of thought.

latenitefilms:
Legend, thanks! Will try push out an update for you to review later today. Stay tuned!

latenitefilms:
@asmagill - I've merged your code into `hs.dialog`. The color panel stuff all seems to work really well, but I'm having trouble with the font panel stuff. Any suggestions?

cmsj:
(just a note to say that -Weverything and -Werror are preferred, since they help us avoid niggling little inconsistencies in the code. However, I do recognise that they are a huge pain to have turned on, so feel free to not worry about that for now :)

asmagill:
Just an FYI, ignore as you wish since it's my preferences here and not by any stretch anything official, but FWIW, the warning flags I use during development are usually:

~~~
-Weverything                          | A good starting point...
-Wno-objc-missing-property-synthesis  | We want auto synthesis by default, so why warn that it's happening?
-Wno-implicit-atomic-properties       | We want atomic properties by default, so why warn that it's happening?
-Wno-direct-ivar-access               | What started this discussion on warnings!
-Wno-cstring-format-directive         | Since lua uses cstrings, I don't want NSString complaining about the %s format directive
-Wno-padded                           | I don't want to worry about structure/alignment padding -- memory isn't that tight
-Wno-covered-switch-default           | I usually do put a default in switch's that cover everything in case Apple later adds a new enum
-Wno-missing-prototypes               | Because most of my interfaces and implementations are in the same file I don't want to have to declare every method in the interface that is never called from outside of the file itself
-Werror-implicit-function-declaration | A missing function should cause an error not just a warning
~~~

I don't usually include `-Werror` because during development I often want to compile as a spot check and I already *know* that certain sections are incomplete; however, I do almost religiously tweak and modify until the module compiles with 0 warnings under these settings before releasing the code or adding it to the core module set.

With very rare exceptions, if my code compiles with no warnings with these settings, it compiles fine within the Hammerspoon XCode workspace... there is enough inherited code from way-back-when in Hammerspoon that it would be a major challenge to go back and enforce this for everything, though as always, it *would* be a good idea at some point!

asmagill:
I'd go with a lua rather than objective-c wrapper; leave the function which takes a webview object as a parameter exposed as I can envision wanting to make the dialog seem attached to an existing web "application" (and I have some long term ideas about building a more functional web browser out of the webview module where save/open dialogs would be useful), but I think a wrapper should also be available for when you just want to display a specific dialog and not have to think about what it's attached to.

Re the text list, I don't have any immediate plans to address this specifically, but a gui building module I hope to start soon should make this easy, I just don't have a time frame yet... The idea is to use a variety of "widget" like subcomponents and let you plug them into a grid and then have the module build a view to display what has been described (sorta like tkinter's grid if you're familiar with python) -- it's the first step in what I envision being an eventual merging of canvas, webview, and an avplayer module in my wip repo, but it will also come with a set of built in widgets as well for buttons, checkboxes, lists, pop-up menus, etc. first.

Are you hoping for the dialog module to be added to 0.9.56 or do you envision it evolving a bit more before its ready?  I ask because if I merge #1515 there will be a conflict with `project.pbxproj` that will have to be manually handled... it's pretty simple -- for this particular conflict, 99% of the time you just remove the extraneous text git puts in to mark the conflicted areas because we want both of the changes that it presents and order doesn't matter because they're independent hash values or target blocks on separate lines --  if you think this will be ready soon, I'll take care of the massaging myself; on the other hand if you expect this to hang around in the pulls for a while yet, you'll need to do it when you are ready to merge into Hammerspoon (and I'll be happy to help if it's not obvious what you need to do).

latenitefilms:
> I'd go with a lua rather than objective-c wrapper; leave the function which takes a webview object as a parameter exposed as I can envision wanting to make the dialog seem attached to an existing web "application" (and I have some long term ideas about building a more functional web browser out of the webview module where save/open dialogs would be useful), but I think a wrapper should also be available for when you just want to display a specific dialog and not have to think about what it's attached to.

Great, done!

> Re the text list, I don't have any immediate plans to address this specifically, but a gui building module I hope to start soon should make this easy, I just don't have a time frame yet...

Given this is already built into AppleScript, I was kinda hoping there would already be an Objective-C equivalent (like NSAlert), but if not, I guess I could just do a Lua/AppleScript version for the time being?

> Re the text list, I don't have any immediate plans to address this specifically, but a gui building module I hope to start soon should make this easy, I just don't have a time frame yet... 

We've built something similar using HTML, JavaScript and `hs.webview` for CommandPost which seems to work pretty well. We basically do all the UI creation in HTML/CSS/JS - which gives you a huge amount of freedom. 

> Are you hoping for the dialog module to be added to 0.9.56 or do you envision it evolving a bit more before its ready?

I would LOVE to get `hs.dialog` into the master branch ASAP - so that I can merge the code into CommandPost and start using it - so yeah, the sooner I can get it across the better.

The main thing we need to fix at the moment is the font panel, as that doesn't really work as expected for me.

> I ask because if I merge #1515 there will be a conflict with project.pbxproj that will have to be manually handled

Well, if you merge #1515 into `master`, then I can just update my branch and fix any conflicts? As I'm constantly having to manage conflicts when I merge the main Hammerspoon code into CommandPost-App, I'm pretty good at fixing conflicts, I think.

latenitefilms:
@cmsj - I tried just using `-Weverything` and `-Werror`, and changing every case of `_someVar` to `self.someVar`, but I couldn't get it to work. There was also an issue with using `default` in a `switch` that I'm not sure how best to solve - so I've given up changing the "Other warning flags". However, feel free to jump in and correct yourself!

@asmagill - I've tried playing around with the `hs.dialog.font` code, but for whatever reason I can't get the callback to return anything but `nil`. Let me know if you have any ideas on how to fix.

latenitefilms:
@asmagill - I'm still not getting any font information on the callback - I always just get `nil` for the first parameter:

```
> hs.dialog.font.callback(function(a,b) print("FONT CALLBACK:\nSelected Font: " .. hs.inspect(a) .. "\nPanel Closed: " .. hs.inspect(b)) end)
function: 0x6001e8e47fd0
> hs.dialog.font.show()

2017-09-03 10:52:43: FONT CALLBACK:
Selected Font: nil
Panel Closed: false
2017-09-03 10:53:43: FONT CALLBACK:
Selected Font: nil
Panel Closed: false
2017-09-03 10:53:46: FONT CALLBACK:
Selected Font: nil
Panel Closed: false
2017-09-03 10:53:49: FONT CALLBACK:
Selected Font: nil
Panel Closed: true
```

MacBook Pro (15-inch, 2017) running macOS 10.12.5.

asmagill:
looks like maybe it doesn't know about the NSFont object type... try adding `require("hs.styledtext")` somewhere in the module's init.lua (doesn't matter where and you don't have to capture it -- we just need to make sure it loads its data type conversion routines).

latenitefilms:
Bingo!

Legend, thank you!

latenitefilms:
@asmagill - The only thing left to fix now is adjusting font colours from the font panel doesn't always work - I assuming there's some kind of conflict between `hs.dialog.color` and `hs.dialog.font` because they share the same color panel? I wonder if we should make it so that `hs.dialog.font` has it's own completely seperate color panel than `hs.dialog.color` (i.e. if you click the color chooser icon in `hs.dialog.font` it should open a SECOND color panel window, if one has already been created by `hs.dialog.color`)?

I know you mentioned earlier that having multiple color panels is a bit of a pain - but maybe this might be helpful? https://github.com/DrummerB/BFColorPickerPopover

asmagill:
Note to self and @latenitefilms ; reloading HS when font panel is open does not close it; module gc should probably close these (assuming color panel is the same way).  I'll look into tomorrow (about to nod off) if you don't before.

asmagill:
I'm at a loss as to how to capture color data when being changed through the font panel... changing the text color triggers up to 4 callbacks and doesn't result in a change; changing the background color doesn't trigger anything.

I'm tempted to recommend removing the `textColorEffect` and `documentColorEffect` modes from the constants table and turning off their flags in `allEffects` and `allModes`, but that's just a thought.

As to multiple color pickers, the only examples I'm finding are using the NSColorWell view class, not the NSColorPanel class.  The NSColorWell view class make sense to add to the GUI builder I have in mind -- it can't stand on its own without a window around it, which what the builder will provide.

So for the moment, I'm not sure that the font panel is ready for prime-time, but that's your call since its your module.

latenitefilms:
@asmagill - Does this help at all? 

http://www.cocoabuilder.com/archive/cocoa/172684-how-to-receive-text-color-changes-from-nsfontpanel.html

asmagill:
If you look at the code you'll see that we already do have a handler for `changeAttributes:` and adding a `changeColor:` method doesn't help -- it never gets called.

latenitefilms:
@asmagill - Please excuse me if you already know all this, or have already tried all this, but just so that I can wrap my head around everything... Also, please ignore the fact I don't know any of the correct terminology!

When I change the colour in the Font Panel, it seems to trigger `changeAttributes:` - in which the `obj` returns a `NSFontEffectsBox` (the header file of which can be found [here](https://github.com/pombredanne/osx_headers/blob/master/Frameworks/AppKit/NSFontEffectsBox.h)).

From what I can gather, you need to use `convertAttributes` - which you seem to be already doing? - but based on [this](https://stackoverflow.com/questions/33438648/nstextview-nsfontmanager-and-changeattributes), I'm assuming you need to actually specify what attributes you want to "access"?

Given this, I'm GUESSING that maybe there's something wrong with this line of code?

`self->_attributesDictionary = [obj convertAttributes:self->_attributesDictionary] ;`

This might also be relevant?

> When changeAttributes: is called, it invokes a new method
> 
> - (NSDictionary *)convertAttributes:(NSDictionary *)attributes;
> 
> on the sender of changeAttributes:, successively with each set of
> attributes used in the current selected range (or in the typing
> attributes). The sender then makes whatever attribute changes it
> desires and return a converted dictionary. The sender is expected to
> leave untouched attributes it is not interested in.

To be honest, I don't actually need this Font Panel for my purposes, so it doesn't really bother me if it doesn't make it into the next Hammerspoon release, but we seem SO CLOSE, that it would be a shame not to solve this problem and just add the font panel to the next release.

latenitefilms:
Also maybe relevant? 

https://developer.apple.com/documentation/appkit/nstextview/1449270-selectedtextattributes?language=objc

asmagill:
Personally I'd recommend pulling the font stuff then... I've had very little luck finding good documentation for it and it acts funky at times, some attribute changes being reflected, others not, and some (especially wrt color) that sorta work but sorta don't (i.e. trigger a callback but don't provide any information).

Re the specific line you mention, as I recall, `convertAttributes:` is the only way to get from the sender exactly what attributes were changed when the method (`changeAttributes:`) is invoked.  We're essentially appending any new changes to the currently recorded changes so that the callback to lua has *all* of the current changes in it's table argument, but if you wanted to check what changed *this time*, you'd have to compare `self-> _attributesDictionary` to the returned dictionary...

If I do get the gui builder module going to the point where text blocks are added as an input type, I may revisit this then, but until then, I personally see no need/use for the font panel until someone comes up with a specific use case and I don't have the time right now anyways...

latenitefilms:
@asmagill - No worries - agreed. I've removed `hs.dialog.font` for now and we can treat this as a seperate issue.

@cmsj - With `hs.dialog.font` now removed, I believe this extension is good to go, unless you have any objections?

latenitefilms:
@cmsj - Let me know if you need me to do anything else with this one to wrap it up?

asmagill:

cmsj:
I haven't really played with this extension in depth, but since it's new and I trust you guys, +1 from me :)

PR #1508, opened 2017-08-05
Fix double margin between windows when using grid

jimeh:
Currently the grid margin value applies in all directions. Which means get twice as much space inbetween windows, as you get between windows and the screen edge. This change ensures that if you set the margin to `{ w = 10, h = 10}` you get a 10 pixel gap between windows and screen edges, and also between different windows.

Before:

menubar_and_untitled_3_and_untitled_2

After:

menubar_and_untitled_3_and_untitled_2 2

Apologies for not adding any tests for this change, the grid extension currently has no tests, and I'm brand new to Lua, so I don't really know where to start.

If lack of tests is a showstopper for getting this merged in, please point me in the right direction to get started with putting together some tests :)

### DIFF
--- a/extensions/grid/init.lua
+++ b/extensions/grid/init.lua
@@ -319,6 +319,30 @@ function grid.set(win, cell, scr)
     w = cell.w * cellw - (margins.w * 2),
     h = cell.h * cellh - (margins.h * 2),
   }
+
+  -- ensure windows are not spaced by a double margin
+  if cell.h < screengrid.h and cell.h % 1 == 0 then
+    if cell.y ~= 0 then
+      newframe.h = newframe.h + margins.h / 2
+      newframe.y = newframe.y - margins.h / 2
+    end
+
+    if cell.y + cell.h ~= screengrid.h then
+      newframe.h = newframe.h + margins.h / 2
+    end
+  end
+
+  if cell.w < screengrid.w and cell.w % 1 == 0 then
+    if cell.x ~= 0 then
+      newframe.w = newframe.w + margins.w / 2
+      newframe.x = newframe.x - margins.w / 2
+    end
+
+    if cell.x + cell.w ~= screengrid.w then
+      newframe.w = newframe.w + margins.w / 2
+    end
+  end
+
   win:setFrameInScreenBounds(newframe) --TODO check this (against screen bottom stickiness)
   return grid
 end


asmagill:
@cmsj, I don't use grid spacing (I think it's set to 0 in my setup for the few times I do use the grid module), but the code looks good to me.

jesseleite:
Hmm, I set this:

```
hs.grid.setMargins({w=10, h=10})
```

But am still getting this:

screen shot 2018-12-06 at 10 54 02 am

Am I missing something?

PS. Thank you for your work on this!

jimeh:
@jesseleite are you using the latest version of Hammerspoon? I believe 0.9.56 (15 Sep 2017) is the first release that included the changes in this PR. And 0.9.73 is the latest release as of writing.

jesseleite:
`brew cask install hammerspoon`'d today

screen shot 2018-12-06 at 11 53 42 am

jimeh:
@jesseleite That's really bizarre. Are you sure you don't have a older grid library in your config that might be overriding things? Is your hammerspoon config on GitHub maybe somewhere?

You can also double check the relevant code is in your version of Hammerspoon by looking at `/Applications/Hammerspoon.app/Contents/Resources/extensions/hs/grid/init.lua` and making sure it contains the changes from this PR.

jesseleite:
Just uploaded [my (ugly) hammerspoon config](https://github.com/jesseleite/dotfiles/commit/429686cb8495e40d8f9416850669110b523a3f80).  Don't judge me, still learning 😬

Will take a look at your second point there too, thanks.

jimeh:
@jesseleite Okay, I know what's causing your issue. When using `hs.grid.show()` to set the window placements via the interactive grid UI, it seems the logic I added in this PR is bypassed somehow, so it behaves like it used to previously with the double margins. I don't use the grid UI much myself, hence I hadn't noticed it >_<

I'll try and submit a new PR to fix it soon :)

jesseleite:
@jimeh Ah!  In the long run I don't want to use the grid UI either, I'm just new to hammerspoon so I'm seeing what's available.  That said, it makes sense to patch it for users who do use the UI.  Thanks for your time!

jimeh:
@jesseleite If it helps you get going, here's my Hammerspoon config: https://github.com/jimeh/dotfiles/tree/master/hammerspoon

And yeah, it should definitely be fixed, cause it's not expected behavior :)

jesseleite:
Hey @jimeh, love how your window management stuff works, thanks for sharing!

I'm working on multi-window layouts using `hs.layout.apply()`, and [got it working with grid cell positions here](https://github.com/jesseleite/dotfiles/blob/426e8daac7e3305dc2276ed8b9930b5516435704/hammerspoon/init.lua#L18-L57).  However, because this doesn't call `grid.set()`, it doesn't respect grid margins.  Do you have any advice on how to respect grid margins in this kind of situation?

PS. I apologize if this isn't the right place to ask this kind of stuff. If you're open to chatting on Discord or something, hit me up at jesseleite#8296

jesseleite:
That last post got me thinking, [I guess I could skip layout.apply() and just use grid.set()](https://github.com/jesseleite/dotfiles/blob/eb1e67d9da2bfe0d9390686c5f0ec503850c0990/hammerspoon/init.lua#L18-L58) 😅

Issue #1415, opened 2017-05-09
hs.notify.warnAboutMissingFunctionTag - as a callback instead?


latenitefilms:
Rather than printing to the console, it would be awesome if `hs.notify.warnAboutMissingFunctionTag` was a callback instead.

asmagill:
That portion is actually pure lua, so you can take a crack at it if you're interested.  Look around line 100 of notify's init.lua `module.__tag_handler`.

My recommendation would be to test if it's a function and if so, call the function; otherwise test for truthiness to print or falseness to be silent.

latenitefilms:
Sweet - yeah, I'll give this a bash!

Sorry - yes, a lot of the issues I've posted recently are more "notes to self", rather than expecting other people to do. Generally I'll only tag other people if it's something that I don't think I can do myself.

PR #1514, opened 2017-08-18
hs.notify.warnAboutMissingFunctionTag can be a function

asmagill:
Addresses #1415

### DIFF
--- a/extensions/notify/init.lua
+++ b/extensions/notify/init.lua
@@ -39,7 +39,11 @@ module.activationTypes = ls.makeConstantsTable(module.activationTypes)
 
 --- hs.notify.warnAboutMissingFunctionTag
 --- Variable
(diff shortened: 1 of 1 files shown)


latenitefilms:
Amazing, thank you!

asmagill:
@cmsj, regularly I see failures during the UDP tests... usually it will clear if I tell it to rebuild, though sometimes I have to do so more than once.  Do you know if travis can only handle one of our builds using UDP at a time?  If so, is there any way we can detect other builds and stagger or pause them automatically?

cmsj:
@asmagill I'm not really sure why they fail randomly. I made some changes recently that were supposed to fix it, but they didn't, so I'm thinking at this point probably the easiest option is to SKIP_IN_TRAVIS() that particular test :/

asmagill:
@latenitefilms does this do what you want/need?  If so, I'll go ahead and merge it

latenitefilms:
Looks good to me, thanks again!!

PR #1518, opened 2017-08-23
Changed copyright date from 2015 to 2017

DAlperin:


asmagill:
@cmsj, are there other places we need to update this as well?

cmsj:
This should be it I think. Thanks @DAlperin - I've not been keeping up on this copyright/license when updating Lua.

DAlperin:
Great!

PR #1521, opened 2017-08-29
Update documentation for hs.network.configuration:setLocation

natemara:
The previous documentation stated that you could pass in a location UUID or name, but this is not accurate. You must pass in a name, or this method will fail.

### DIFF
--- a/extensions/network/configurationinternal.m
+++ b/extensions/network/configurationinternal.m
@@ -316,10 +316,10 @@ SCPreferencesCreateWithOptions      (
 /// Switches to a new location
 ///
 /// Parameters:
-///  * location - string containing UUID or name of new location
+///  * location - string containing name of new location
 ///
 /// Returns:
-///  * bool
+///  * bool - true if the location was successfully changed, false if there was an error
 static int dynamicStoreSetLocation(lua_State *L) {
     LuaSkin *skin = [LuaSkin shared] ;
     [skin checkArgs:LS_TUSERDATA, USERDATA_TAG, LS_TSTRING, LS_TBREAK];
--- a/extensions/network/configurationinternal.m
+++ b/extensions/network/configurationinternal.m
@@ -316,7 +316,7 @@ SCPreferencesCreateWithOptions      (
 /// Switches to a new location
 ///
 /// Parameters:
-///  * location - string containing name of new location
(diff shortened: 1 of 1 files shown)


natemara:
If you look at [Line 376](https://github.com/Hammerspoon/hammerspoon/pull/1521/files#diff-8c565509a4004416624b3096ebe6c1eaR376), you can see that only the name is compared for equality, not the UUID. If the original documentation was wanted, we should change the behavior of this method to match the documentation.

asmagill:
If I'm understanding the location code, it looks like none of the functions which allow name or UUID actually work with the UUID... I'll take a closer look at this tonight and see if it's more reasonable to update the documentation or make the code do what the documentation indicates...

I don't normally use multiple locations (well not in the last couple of years), do you know if there is a reason why the UUID might be preferable other than its guaranteed uniqueness?  Does the UUID change if you change a location's name?  If not, that could be a reason to allow its use.

natemara:
I checked, and the UUID does remain constant if you change the name of a location.

asmagill:
Ok, I misspoke, it looks like only `setLocation` is a concern here... I'm not going to get a chance to try this out myself tonight, but if you get a chance, try changing lines 376-377 to what follows this paragraph and let me know if that allows you to select a location by its UUID as well as by its name... I think I'd rather support both instead of dropping it from the docs -- the UUID seems more useful because of its persistence.

~~~objc
        CFStringRef name = SCNetworkSetGetName((SCNetworkSetRef)item);
        CFStringRef uuid = SCNetworkSetGetSetID((SCNetworkSetRef)item);
        if ((CFStringCompare(name, (CFStringRef)target, 0) == kCFCompareEqualTo) || (CFStringCompare(uuid, (CFStringRef)target, 0) == kCFCompareEqualTo)) {
~~~

natemara:
Those lines do seem to work, so I wrapped them in a commit for your viewing pleasure. I didn't have anything against using UUIDs in this method, I was just very confused when the documentation lied to me.

asmagill:
Works for me; I'll take a close look at it tonight and merge it soon if there are no other concerns.  Objections @cmsj?

cmsj:
(no objections :)

Issue #1324, opened 2017-04-12
AppleScript should raise an error if Lua code has errors


latenitefilms:
Currently the AppleScript `execute lua code` function (see: #1177) will simply return the result from `MJLuaRunString`. This works great if you feed it good code (i.e. `execute lua code "1+1"` will return `2`), but it should really generate an error if there's a Lua error (i.e. currently `execute lua code "1++1"` will return a string value of `"[string \"1++1\"]:1: unexpected symbol near '1'"`). This will require either duplicating and modifying, or just modifying `MJLuaRunString`.

FYI: `HSAppleScript.m` already has an example of how you can raise an error in AppleScript.

cmsj:
As mentioned on IRC, I don't want to tackle this in 0.9.53 because this is the wrong time to be messing with MJLuaRunString.

asmagill:
I haven't looked closely at the applescript integration, but it may be better to build your own runstring specifically for use by the applescript bridge. `hs.ipc` basically does this to avoid having to put a lot of conditionals into MJLuaRunString and the `runString` function returned by `hs._coresetup`.

latenitefilms:
Fine by me - but I fear that might be outside of my skillset, so might have to leave for you guys to finish. Unless I can just duplicate `MJLuaRunString`?

asmagill:
I should be able to take a look at this in a day or so, if no one can any sooner.

cmsj:
I believe this is now done.

PR #1526, opened 2017-09-03
Issue #1324 - AppleScript should raise an error if Lua code has errors

latenitefilms:
- Work in progress


asmagill:
@latenitefilms I haven't been following this line of AppleScript support... can you provide a link or short sample of how I can test it on my end?  I'll do some digging later, but if you have something handy it would save some time.

Thanks!

latenitefilms:
Away from laptop currently but the original issue #1324 should explain everything?

Basically I want AppleScript to trigger an error rather than receiving a text string containing an error message.

asmagill:
Yeah, I understand that it's Hammerspoon being accessed through AppleScript to perform some action... I was hoping for a simple applescript snippit I could run through osascript or a small file I could load into the AppleScript editor to try it out...

AppleScript and I have a love/hate relationship - I love to hate it mostly because I never really took the time to learn it very well and didn't want to have to try and come up with something from scratch...

Ok, I'll bite the bullet in a bit and see what I can come up with.

latenitefilms:
There's an example here: http://www.hammerspoon.org/docs/hs.html#allowAppleScript

asmagill:
SMH! You'd think I would have checked the documentation, wouldn't you?... thanks!

latenitefilms:
@asmagill - FYI: I tried duplicating `MJLuaRunString`, but I keep getting stuck on the following error:

screen shot 2017-09-03 at 8 23 25 pm

Any ideas?

asmagill:
The `evalfn` has to refer to a valid Lua function -- what's currently being pulled from the registry is undefined because `evalfn` is never assigned a value.

For the traditional `MJLuaRunString`, it looks for the second returned argument from `hs/_coresetup/init.lua` (line 566) and assigns it at line 667 of `MJLua.m`

Let me elaborate a little on what I meant in https://github.com/Hammerspoon/hammerspoon/issues/1324#issuecomment-323647668...

The traditional runstring, which is defined in `hs._coresetup` returns a string containing the visual (printed or returned) results of the lua string it is sent to parse.  This returned string is a string ***whether or not it was an error or a valid response***.  Once the string is back in `MJLua.m`, there is **no** possible way to determine if it was generated by an error or by a print statement or by a return statement.

You stated that you wished for the AppleScript integration code to be able to distinguish between a valid response (i.e. print or return) an an error, thus the suggestion that a special runstring function be created and registered for use specifically by the AppleScript integration code so that it *could* determine if a response was valid or not...  But you need to create that runstring command and somehow register it so that it's reference value can be stored in `evalfn` for recall when executing an AppleScript request.

This pull looks like you have some of the infrastructure in place, but you still need to write the lua portion and then register the lua function for use. As for writing the lua portion, take a look at `hs._coresetup`'s init.lua around line 395.  One change I would probably make in the new function is to return two values instead of one (a boolean indicating valid (true) or error (false), and a string containing the results... in the replicated runstring function, line 398 might end `... return false, tostring(err)` and line 406 might be `return results[1], str`)

Of course you'll want to define this new lua function somewhere other than `hs._coresetup` -- you don't want to break everything else by modifying the existing runstring.

asmagill:
@latenitefilms status?

latenitefilms:
Sorry, I was sure I replied!

All looks good to me! Thanks for all your hard work!

latenitefilms:
@cmsj - This is good to go, if you agree.

asmagill:

cmsj:
+1 from me

PR #1527, opened 2017-09-05
Issue #1528 - hs.dialog.font - Add Font Panel Support

latenitefilms:


asmagill:
Just an update: I haven't looked at the font stuff in a bit, but will likely do so when I get to the textview/scrollview element of the guitk module I'm currently working on.  I should have a better idea of how this can be leveraged for independent use once I'm more familiar with how it interacts with the views it was designed for in the first place.

latenitefilms:
@asmagill - Amazing, thank you!! I'm very, very excited about the `guitk` module! Looking awesome so far!

latenitefilms:
@asmagill - Did you mean to merge this in?

asmagill:
I don't see any font related code in master, https://github.com/Hammerspoon/hammerspoon/blob/master/extensions/dialog/internal.m.

I'm going to guess that because this commit is a direct ancestor to commit c4439e7, git considers that it has been merged (even though the very next merge is to remove a bunch of it)... I think you'll need to create a new branch and add the font stuff back if/when we get to a point where it can be used safely and effectively.

Some one more familiar with git should be able to confirm/refute this, but looking at the commit history of #1504, it's the only explanation I can come up with.

latenitefilms:
No problemo! All makes sense - thanks @asmagill !

PR #1529, opened 2017-09-05
allow hs.application:getMenuItems on a background thread

asmagill:
optional callback allowed for hs.application:getMenuItems allowing its use in a non-blocking manner

### DIFF
--- a/extensions/application/internal.m
+++ b/extensions/application/internal.m
@@ -1021,18 +1021,19 @@ id _getMenuStructure(AXUIElementRef menuItem) {
     }
 }
 
-/// hs.application:getMenuItems() -> table or nil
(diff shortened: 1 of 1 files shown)


cmsj:
Are we sure this is safe to do on a background thread? If so, +1 from me

asmagill:
I can't find a definitive answer, but its use in this function and in the `hs._asm.axuielement` module hasn't posed a problem for me.  In both cases, the background thread is only used to read (not set) accessibility elements in *other* applications, so basically all it's doing in the background queue is opening an IPC connection to the other app and waiting for a response.

cmsj:
@asmagill ah good point, that makes perfect sense :)

asmagill:

PR #1532, opened 2017-09-07
Add hs.streamdeck - full support for Elgato Stream Deck

cmsj:


cmsj:
(No idea why the title was wrong on this PR)

@asmagill it's been a while since I wrote an extension from scratch, and I'm not sure I've ever properly used all of the Lua<->NSObject helpers before. I wonder if you have some time to review this for me?

tl;dr it adds support for Elgato's Stream Deck, which is really just a USB HID device with 15 buttons that I want to use to trigger Hammerspoon stuff, but the cool thing about it is that each button has LCD behind it, so images can be dynamically displayed on the buttons 😁

(see https://twitter.com/cmsj/status/905908483052236800 for the first pic of a bunch of random hs.image objects rendered to some of the buttons)

asmagill:
@cmsj, the LuaSkin object registration looks good -- I don't have this device so I can't fully test it, but I don't see anything that raises any concerns.

The one note I will make in how you're managing the objects themselves is the use of `selfRef` as an actual lua reference to the object itself versus `selfRefCount` as a count of the number of times the object has a userdata in lua-land...

If you maintain an actual lua reference to the object as a property of the object, then Lua can never collect the object even if every lua variable containing the userdata becomes nil -- there is still one reference in the registry.  For drawing, canvas, webview, etc. this makes sense -- we don't want the stand-alone object to disappear without specific user action... which is why these modules also have a `delete` method which explicitly invokes the gc method to clear the lua reference from the registry.

Utilizing a reference counter that is incremented every time `pushOBJECT_TO_LUA` is called and decrements in every time `userdata_gc` is called we allow lua's own garbage collection to work normally -- while the `userdata_gc` function is called for every userdata that is "collected", we can make sure that the callback references are cleared and cleanup code for the object is run only when the *last* userdata is collected by wrapping the objective-c cleanup stuff with  `if (obj.selfRefCount == 0) { ...clear callback ref, do objc cleanup, etc. }` after the decrement.

The pros of `selfRef`:
* what an object represents (e.g. webview, drawing, etc.) stays around even if the user chooses not to save at least one userdata in a global variable
* every lua userdata for the same object is the exact same userdata value in lua -- you can use the userdata as a key in a table of key-value pairs even if the specific instance of the userdata you're using is from a callback rather than the one originally saved during its construction.

The cons of `selfRef`:
* without an explicit `delete` method, you can never have the object fully collected.

The pros of `selfRefCount`:
* lua garbage collection and ARC are in sync -- the last collection of the userdata by lua occurs as the retain count for the object itself reaches 0 -- no explicit delete is required.

The cons of `selfRefCount`:
* the userdata "wrapper" around the object is unique each time the object is pushed onto the Lua stack -- you can't user the userdata as a key in a table of key-value pairs because each wrapper has a different hash value in lua; they can be checked for "equality", though, through the use of an `__eq` meta method, though what you're really telling Lua is that userdata X and userdata Y wrap the same objective-c object.
* each userdata wrapper is unique -- this means that if you store extra data in the userdata uservalue (`hs.application` does this for the application PID) or with `debug.getuservalue` or `debug.setuservalue`, the data is unique to the wrapper, not the underlying objective-c object being wrapped.

And to give an example of how they might both be used together, you can check out `hs.speech` -- this module uses both a counter and a self ref, but the self ref is *only* set during speech playback and as soon as the delegate receives the message that the playback has completed, the self ref is released.  This has the effect that normal collection behavior occurs *except* when speech is actually occurring... if speech is in progress, an extra reference is made so that the object can't be collected -- if it were to get collected during playback, playback would abruptly stop in the middle.

asmagill:
@cmsj see my comment above and decide if you are ready for this to be merged or if you want to make any changes.

cmsj:
@asmagill I think I would like to rework the selfref stuff, so don't block on this if you need to land something that changes the xcode project files, I'll rebase after.

cmsj:
Switched to using the `selfRefCount` model. Thanks @asmagill!

Issue #1533, opened 2017-09-12
LuaSkin additions for increasing/decreasing object reference count


asmagill:
@cmsj, in my recent work to attempt to reduce replicated code by creating wrapper modules that can use objects from other modules (most recently the touchbar bar and touchbar items and everything about guitk, but going back this could also be applied to canvas and webview and toolbar and maybe even a re-written menubar) I'm finding that I need a way to increase the lua reference count on an object from another module that is required, but is considered a "sub component" of the module the user may be interacting with and not realize that they need to retain a global reference to both.

Put more simply, in the touchbar example, you have a userdata representing the touchbar and then a set of userdata's representing the touchbar items.  If an item is added to the touchbar, then it needs to stay viable as long as the touchbar is viable *even if the user forgets to save the item's userdata*.  Saving the touchbar in a global makes a kind of sense; saving the items -- why should I have to when they are "part of" the touchbar object?

I propose adding something along the following to LuaSkin:

~~~
-(int)retainObject:object refTable:refTable ;
-(void)releaseObject:object refTable:refTable ;
-(void)releaseReference:ref refTable:refTable ;
~~~

*-- edit: fixed name typo*

Since my current use case is that I'm needing the objective-c side to make sure that the object doesn't get collected by Lua (and I don't care where/how it's represented in Lua, just that Lua doesn't take it away), I'll most likely only use the first two methods, but I can envision maybe wanting to release based on a stored reference value, hence the third method.

Thoughts?  Objections?  Better name suggestions?

asmagill:
After some more thought, I think the additions to the interface would be more like this:

~~~objc
- (void)luaRetain:(int)refTable forObject:(id)object ;
- (void)luaRelease:(int)refTable forObject:(id)object ;
- (int)luaRef:refTable forObject:(id)object ;
~~~

If you use `luaRetain:` then you must use `luaRelease:` since LuaSkin will be keeping a dictionary per refTable mapping the reference number to the object and we don't care *which* particular reference for a specific object goes away, just that *one* of them does, while the version which returns an integer can just leverage the existing `- (int)luaUnref:(int)refTable ref:(int)reference` since we stored the specific reference number like we do for callback functions.

asmagill:
I'd also like to add a way to check if LuaSkin currently has a convertor for an object, something like:

~~~objc
- (BOOL)canPushObject:(id)object ;
~~~

cmsj:
I believe this can be closed now :)

PR #1539, opened 2017-09-14
Lua skin additions per #1533

asmagill:
following the convention of using NSObject in selectors as opposed to just Object, but otherwise in line with #1533


asmagill:
@cmsj, your thoughts?  I want to test this for a day or so to make sure I haven't introduced anything that breaks reload/restart behavior but since nothing actually uses it yet, if you agree I'd like this to get into 0.9.56 if possible so I can take advantage of it with the new modules I am working on (touchbar and guitk).

cmsj:
@asmagill lgtm.

cmsj:
Hmm, this may have been a premature merge - if you need to make changes feel free to push them directly to master.

PR #1541, opened 2017-09-14
Add hs.mouse.scrollDirection(). Closes #1536

### LINKED ISSUE #1536: Determine Mouse Scroll Direction?
I couldn't find it, so I'm assuming there's not already any API for determining or setting your mouses scroll direction (i.e. Natural vs Normal)?

Maybe a good addition to `hs.mouse`?

FYI: `defaults read ~/Library/Preferences/.GlobalPreferences com.apple.swipescrolldirection` returns 1 if "Natural" or 0 is "Normal".

cmsj:
Closes #1536 
I'm away from a useful computer at the moment - @latenitefilms would you be able to check if this works? If not I can do it next week :)

### DIFF
--- a/extensions/mouse/internal.m
+++ b/extensions/mouse/internal.m
@@ -111,6 +111,24 @@ static int mouse_mouseAcceleration(lua_State *L) {
     return mouseTrackpadAcceleration(L, CFSTR(kIOHIDMouseAccelerationType));
 }
 
+/// hs.mouse.scrollDirection() -> string
+/// Function
+/// Gets the system-wide direction of scolling
+///
+/// Paramters:
+///  * None
+///
+/// Returns:
+///  * A string, either "natural" or "normal"
+static int mouse_scrollDirection(lua_State *L) {
+    LuaSkin *skin = [LuaSkin shared];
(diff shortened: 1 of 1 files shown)


extensions/mouse/internal.m
```diff
@@ -111,13 +111,32 @@ static int mouse_mouseAcceleration(lua_State *L) {
...
+}
+
 //Note to future authors, there is no function to use kIOHIDTrackpadAccelerationType because it doesn't appear to do anything on modern systems.
 
 static const luaL_Reg mouseLib[] = {
 // Note that .get and .set are no longer documented. They should stick around for now, as they are used by our init.lua
     {"getAbsolutePosition", mouse_get},
     {"setAbsolutePosition", mouse_set},
     {"trackingSpeed", mouse_mouseAcceleration},
+    {"scrollingDirection", mouse_scrollDirection},
```
latenitefilms: @cmsj - I think this should be `scrollDirection` to match the documentation?

latenitefilms:
@cmsj - Apart from the one comment I've added to the code, all looks good. Although, I would still love it if it could do a callback, as discussed in #1536.

PR #1542, opened 2017-09-15
add KVO watcher for hs.settings

asmagill:
But I'm not sure it's worth adding... it doesn't work for keys with a period (".") in the name, so it doesn't really help with #1536 

the added function:

~~~
Function: hs.settings.watchKey(identifier, key, [fn | nil]) -> identifier | current value

Get or set a watcher to invoke a callback when the specified settings key changes

Parameters:
 * identifier - a required string used as an identifier for this callback
 * key        - the settings key to watch for changes to
 * fn         - the callback function to be invoked when the specified key changes.  If this is an
explicit nil, removes the existing callback.

Returns:
 * if a callback is set or removed, returns the identifier; otherwise returns the current callback
function or nil if no callback function is currently defined.

Notes:
 * the identifier is required so that multiple callbacks for the same key can be registered by
separate modules; it's value doesn't affect what is being watched but does need to be unique
between multiple watchers of the same key.

 * Does not work with keys that include a period (.) in the key name because KVO uses dot notation
to specify a sequence of properties.  If you know of a way to escape periods so that they are
watchable as NSUSerDefault key names, please file an issue and share!
~~~

### DIFF
--- a/extensions/settings/internal.m
+++ b/extensions/settings/internal.m
@@ -1,5 +1,47 @@
-#import <Cocoa/Cocoa.h>
-#import <LuaSkin/LuaSkin.h>
+@import Cocoa ;
+@import LuaSkin ;
+
+// establish a unique context for identifying our observers
+static const char * const USERDATA_TAG = "hs.settings" ;
+static void *myKVOContext = &myKVOContext ; // See http://nshipster.com/key-value-observing/
+static int refTable = LUA_NOREF ;
+
+@interface HSUserDefaultKVOWatcher : NSObject ;
+@property NSMutableDictionary *watchedKeys ;
+@end
+
+@implementation HSUserDefaultKVOWatcher
+- (instancetype)init {
+    self = [super init] ;
+    if (self) {
+        _watchedKeys = [[NSMutableDictionary alloc] init] ;
+    }
+    return self ;
+}
+
+- (void)observeValueForKeyPath:(NSString *)keyPath ofObject:(id)object change:(NSDictionary *)change context:(void *)context {
+//     [LuaSkin logWarn:[NSString stringWithFormat:@"in observeValueForKeyPath for %@ with %@", keyPath, change]] ;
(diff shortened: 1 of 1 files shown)


cmsj:
https://mjtsai.com/blog/2016/02/22/use-and-misuse-of-nsuserdefaults/ seems to suggest that you can't KVO notifications from another app.

That said, I still think this is a good addition - it could be a nice way to keep different parts of your config decoupled if they need to care about some shared settings.

asmagill:
I think I remember reading somewhere that that changed in 10.12, but I can't find it now...

At any rate, watching "NSCloseAlwaysConfirmsChanges" and then changing the relevant setting in the General System Preferences panel is caught, so it at least works sometimes for global items... just not ones with a period in the key name.

Having a period in the key name of a value stored in "org.hammerspoon.Hammerspoon" is also not watchable... it seems to be because of KVO's use of dot notation for indicating the path to a property on an object.

cmsj:
@asmagill shall we toss this in for 0.9.56?

asmagill:
I'm ambivalent... the original intent was to be able to watch global settings... and you can, some of them, but not, it seems some of the newer multi-part named ones (I almost wonder if that's on purpose by Apple)... on the other hand it is useful in some cases and can definately be used to watch Hammerspoon's own settings (again, as long as they don't have periods in their name) for user defined puposes across multiple functions/files or watchers... but that was why `hs.watchable` was added... so, I don't know... I don't anticipate using it myself, but if you or @latenitefilms still see value in it, then I have no objections, just disappointment that it's not more all encompassing.

asmagill:
And who knows,maybe someone out there knows how to escape the periods and will chime in with an update, so I guess I'm for it...

cmsj:
ack, thanks. I'll merge this and then cut a release (hopefully, I'm on my laptop rather than my iMac, so it might all explode and need to wait until Sunday/Monday)

PR #1545, opened 2017-09-15
Added hs.dialog.font

latenitefilms:
This is still very much a work-in-progress.


latenitefilms:
@asmagill - Did you come up with any brains waves for this when working on `hs.guitk`?

asmagill:
Not really... by setting a textfield element to editable and styleEditable, you can pull up the font dialog for the textfield by right clicking in it and selecting it from the Font submenu, but everything "just works" for modifying the text in the textfield -- I don't need to capture or track anything special.

I do have plans, after guitk is in core, to add support for a scrolling text area that supports rulers, scrollbar, etc. and it may be a bigger issue then; then again it may just work there too since that's the environment it was designed for and this module attempts to use it outside of its intended target.

So, sorry, but I don't have any suggestions yet!

Issue #1548, opened 2017-09-16
`hs -c "hs.reload()"` spews errors and doesn't reload


chdiza:
```bash
$ hs -c "hs.reload()"
error communicating with Hammerspoon: message port was invalidated
error unregistering CLI instance with Hammerspoon: message port invalid
```
And then, HS doesn't actually reload.

asmagill:
Interestingly for me I get the same output, Hammerspoon starts to reload and then crashes.

However, `hs -c "hs.timer.doAfter(1, hs.reload)"` works as expected.

I'll have to look into it.

daviehh:
can replicate with the newest version 0.9.56, it used to work okay before the update.

cmsj:
Looks like the crash is happening in `hs.ipc:sendMessage()` - perhaps because it's trying to send a message using a lua_State object that doesn't exist anymore due to the reload?

asmagill:
The crash is occurring in `ipc_callback`because the external tool `hs` is sending the reload command, then immediately following it with a request for a clean disconnect, but the lua state is in the process of being torn down... checking the state with `lua_status` in `ipc_callback` doesn't detect this (it's still considered valid during teardown), so I'm going to have to think a little before proposing and testing a solution.

In the mean time, if you need to do a reload through the command line tool, I suggest using `hs.timer.doAfter(1, hs.reload)`

asmagill:
Clarification - that's not quite accurate... the first call to `ipc_callback` is to set up the environment for the cli tool, the second is the actual command `hs.reload` and the call to `[skin protectedCallAndTraceback:nresults:]` never returns, which apparently is preventing the ipc module from being properly collected during the destroy process.

I need to look closer at how the original version differs... on the surface, I would expect both versions to have failed this way, so I must be missing something.

asmagill:
@cmsj, you're partially right that ipc_sendmessage is at fault... `hs.reload` does not actually do the reload... it queues it on the runloop and then returns (e.g. `hs.reload() ; t = os.time() ; hs.timer.usleep(10000000) ; print(os.time() - t)` will delay the reload for 10 seconds).  This means that any code which follows the reload command but does not trigger adding something else to the runloop queue will get executed before the reload actually takes place.  In our case it is `ipc_sendmessage` that adds to the runloop.

[...]

PR #1556, opened 2017-09-25
CL: *BugFix* wraps hs.reload so it doesn't crash when invoked through hs.ipc cli tool

asmagill:
Addresses #1548 

@cmsj, can change log messages be put into pull title or messages or does it have to be in a commit?


cmsj:
@asmagill the CL would need to be in the commit message - GH pull titles are not visible in the git history.

asmagill:
@cmsj, objections to merging?

Issue #1574, opened 2017-10-08
hs.notify : can we get a callback when the user clicks cancel?


alice0meta:
if the user has set Hammerspoon notifications to Alert in the Notification Center pane of System Preferences, then the notification has two buttons (the action button and the other button). We can then get a callback only if the user clicks the content or the action button - the other button is a hardwired cancel button. Can we get a callback when the user clicks the other button?

asmagill:
I don't believe so; cancel causes the notification center to clear the notification and the Hammerspoon delegate is never invoked.

If you know of an application or code that does capture the "cancel" action, I can take a look at it, but so far I don't see anything that looks promising.

alice0meta:
Is it possible to at poll for it?

asmagill:
ATM, no... however, looking at `NSUserNotificationCenter`'s `deliveredNotifications` property may provide a way for us to check and see if the userdata is still in the list.

It may be few days before I can get to this, so if someone wants to take a crack at it sooner, I won't object.

latenitefilms:
This would be awesome if possible!! Thanks @asmagill !

asmagill:
@latenitefilms, @alice0meta can either of you confirm if pull #1578, which has been merged into the master branch of Hammerspoon now, provides you with what you need to poll for this?

latenitefilms:
@asmagill - I've been meaning to test this out. Will try and get to it ASAP.

latenitefilms:
@cmsj & @asmagill - This issue can now be closed thanks to #1578

PR #1578, opened 2017-10-10
Add new functions and methods to hs.notify

asmagill:
Adds:
* hs.notify.deliveredNotifications -- returns an array of all Hammerspoon notifications still in the users notification center
* hs.notify.scheduledNotifications -- returns an array of all Hammerspoon notifications scheduled for delivery that have not yet been delivered

* hs.notify:additionalActions -- adds pop-up menu to action button of alert type notifications
* hs.notify:additionalActivationAction -- the additional action selected by the user

* hs.notify:hasReplyButton -- adds single line text input option to alert type notifications
* hs.notify:responsePlaceholder -- subdued text to display as placeholder in reply field
* hs.notify:response -- users typed response

Also adds __eq metamethod so notifications can be compared for equality

### DIFF
--- a/extensions/notify/init.lua
+++ b/extensions/notify/init.lua
@@ -77,9 +77,9 @@ module.warnAboutMissingFunctionTag = true
 ---   * actionButtonTitle   - see [hs.notify:actionButtonTitle](#actionButtonTitle)
 ---   * hasActionButton     - see [hs.notify:hasActionButton](#hasActionButton)
 ---   * otherButtonTitle    - see [hs.notify:otherButtonTitle](#otherButtonTitle)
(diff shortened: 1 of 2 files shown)


asmagill:
(oh, and for what its worth, this is no longer compatible with 10.9, but since Hammerspoon's minimum target is 10.10, this shouldn't be a problem.)

asmagill:
addresses #1574 (see docs for hs.notify.deliveredNotifications)

latenitefilms:
Awesome. Will have a play ASAP!

asmagill:
@cmsj, objections to merging?

latenitefilms:
@asmagill - I finally got around to looking at this - awesome work! Thank you - works a treat!

Issue #1569, opened 2017-10-04
HS is littering `default.profraw` files to disk


chdiza:
For some reason starting sometime within the last two weeks, HS has been dumping files called `default.profraw` on my disk.  Running `file` on those reveals that they're some kind of debugging tool for llvm (`LLVM raw profile data, version 4`).

I'm on 10.12.6.  My `.hammerspoon` is a symlink to a folder called `hs` that lives elsewhere, and it's in this `hs` folder that the `default.profraw` files are usually getting dumped.

I can reproduce this every single time on more than one machine by doing the following:
```
$ open -a Hammerspoon
[wait for it to finish loading]
$ osascript -e "tell app \"Hammerspoon\" to quit"
$ ls ~/.hammerspoon
default.profraw
[the usual files]
```

Sometimes this file gets dumped elsewhere on the file system, but it's usually to `~/.hammerspoon`.  I've never seen this before, and it's driving me nuts.  I built HS locally from source, but I didn't turn on any kind of debugging flag.  I just run `make` in the source dir.  Maybe this is Yet Another Thing Xcode 9 Sees Fit To Screw Up?

I've ruled out that it might be something in my `init.lua` or other things it sources (like my fake geeklets), by moving the contents of `~/hammerspoon` elsewhere and following the reproduction steps.  I still get littered.

In case it might be connected: There have been some occasions within the last week where on one of my machines my system comes to a crawl, and Activity Monitor has HS running at 200% CPU.  That also has never happened before, and I have no idea why HS would be doing that.  But this phenomenon is not a necessary precursor of the `default.profraw` litter (and anyway, the litter happens on two different machines while the CPU spike is only on one).

chdiza:
It's got to do with Xcode 9.  If I go back and rebuild HS using Xcode 8.3.3 and their associated CommandLineTools, I don't get the litter.  As soon as I point `xcode-select` at Xcode 9 and reinstall *their* CLT, I get the litter.

(If I don't match the CLT version to the Xcode.app version, then when I run `make` I see a complaint about some kind of framework mismatch.  So to rule that out from being a factor, I made the two versions match.)

So I'm hoping that either (a) there's some kind of setting that can be set in the `makefile` to get rid of the litter, or (b) there's some other way to build HS locally that doesn't cause the litter.

cmsj:
[...]

Issue #1555, opened 2017-09-24
Hammerspoon debug build enabled instruments profiling?


asmagill:
@cmsj, I can't tell if this is something you enabled or if it's a new default since I installed XCode 9, but I've started seeing `default.profraw` created in the working directory every time I launch Hammerspoon or use the `hs` command line tool... any thoughts?  I'm going to try and track down the specific flag that enables this in the application build scheme, but if you've enabled it on purpose, I'll make the change strictly local until I get to the point where I want to use it in my own tests.

cmsj:
@asmagill I don't think that was me, at least not intentionally. I try to keep my instruments/malloc_guard/address_sanitiser stuff on a separate build scheme which isn't shared.

asmagill:
I haven't tested it fully, but disabling "Gather coverage data" in the scheme editor for the Hamemrspoon scheme within XCode seems to work, at least for me.  I don't want to submit a pull until you've verified it doesn't affect something you want @cmsj, but @chdiza in the mean time, this is what I did:

1. Open  `Hammerspoon.xcworkspace` in XCode
2. Near the upper left corner of the window you will see a scheme designation of "Hammerspoon > My Mac" -- click on "Hammerspoon" and scroll al the way to the bottom and select "Edit Scheme..."
3. In the window that opens, click on the "Test" line in the column on the left
4. Uncheck "Gather coverage data"
5. Click on "Close"
6. Close XCode
7. Build as normal

You'll have to remove any currently existing `default.profraw` files, but so far I haven't seen them come back (and it used to everytime I quit and restarted Hammerspoon).

<img width="1635" alt="screen shot 2017-10-04 at 4 11 15 pm" src="https://user-images.githubusercontent.com/8139480/31200349-3ce78f6c-a920-11e7-9f8e-227d41412a6e.png">

chdiza:
Thansk for the tip!  It seems to be working over here too.

cmsj:
@asmagill if you want to submit a PR, go ahead - the only thing I care about here really is whether the Travis builds still collect coverage data, which I think they should, but the build results of a PR will tell us :)

PR #1579, opened 2017-10-11
disable the dropping of default.profraw

asmagill:
Turns off coverage data collection in Debug builds

Addresses #1569 and #1555


cmsj:
Well, Codecov's comment here suggests that this doesn't break the coverage data generation for CI :)

PR #1589, opened 2017-10-28
add `noaction` command to hs.window.layout

freewizard:
add `noaction` command to ignore selected windows in certain case e.g. you may ignore the current focused window and move the rest.

add `hasTitlebar` filter to hs.window.filter, so that you can skip edge case like some auto-complete pop-ups

add `frontmost` selector to hs.window.layout, it's like `focused` but only apply to frontmost app

### DIFF
--- a/extensions/window/filter.lua
+++ b/extensions/window/filter.lua
@@ -224,6 +224,7 @@ local function checkWindowAllowed(filter,win)
     if filter.allowScreens and not filter._allowedScreens[win.screen] then return false,'allowScreens' end
     if filter.rejectScreens and filter._rejectedScreens[win.screen] then return false,'rejectScreens' end
   end
+  if filter.hasTitlebar~=nil and win.hasTitlebar~=filter.hasTitlebar then return false,'hasTitlebar' end
   local approles = filter.allowRoles or windowfilter.allowedWindowRoles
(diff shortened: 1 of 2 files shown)


asmagill:
@cmsj, I don't use hs.window.filter, so I'm not exactly sure how to test this; however the code changes look ok to me

extensions/window/filter.lua
```diff
@@ -562,7 +564,7 @@ function WF:setAppFilter(appname,ft,batch)
           logs=sformat('%s%s={%s}, ',logs,k,first)
         else logs=sformat('%s%s=%s, ',logs,k,v) end
         filter.allowRoles=r
-      elseif k=='visible' or k=='fullscreen' or k=='focused' or k=='currentSpace' or k=='activeApplication' then
+      elseif k=='visible' or k=='fullscreen' or k=='focused' or k=='currentSpace' or k=='activeApplication' or 'hasTitlebar' then
```
asmagill: shouldn't that be `or k == 'hasTitlebar'`? As is, the line will always evaluate to true because a string constant is "true" in lua.

asmagill:
Ok, @freewizard take a look at the line comment above and once clarified, I see no reason not to merge this.

freewizard:
fixed. thanks @asmagill

PR #1593, opened 2017-10-30
[READY]: Issue #1422 - New hs.midi Extension

### LINKED ISSUE #1422: hs.midi - Add Extension for MIDI Control
Could this be implemented into a Hammerspoon extension?

https://github.com/charlieroberts/midiStroke

latenitefilms:
- Closes #1422


extensions/midi/internal.m
```diff
@@ -0,0 +1,139 @@
+@import Cocoa ;
+@import LuaSkin ;
+
+// FIXME: I have no idea why I need to give it the full path here:
+#import "/Users/chrishocking/Github/hammerspoon/Pods/MIKMIDI/Source/MIKMIDI.h"
```
latenitefilms: @asmagill @cmsj - Any idea why I can't just use `#import <MIKMIDI/MIKMIDI.h>`?

asmagill:
I won't have a chance to look into this at all for a few days, but I wonder if it might be easier to install MIKMIDI through the use of CocoaPods?  I did find it at https://cocoapods.org/pods/MIKMIDI, so unless you're making changes to the core library, it should be doable, and would remove the onus of getting the library to compile and be in the right place from you -- all you'd have to worry about is the Hammerspoon functions that interface with the MIKMIDI pod.

Unfortunately I have absolutely zero experience using or integrating CocoaPods, but I know @cmsj has done so for many of the extensions he's added, so I suspect it's not hard, just something I've never done before.

latenitefilms:
Thanks @asmagill - however I did actually use CocoaPods for the initial install, but for some reason the path suggested in the setup guide doesn’t work so I’m guessing there’s some trick to it. The Framework however all compiles fine and seems to embed itself into Hammerspoon without any issues. Any suggestions @cmsj?

latenitefilms:
@asmagill - Never mind, I worked it out. I just needed to set the `USER_HEADER_SEARCH_PATHS` value for the midi target. Thanks for all your help!

latenitefilms:
@asmagill - I'm SLOWLY getting this extension working. No pressure or rush, but if you do have a free moment and feel like offering some advice/direction on how best to finish things off, I'd be very much appreciative. Currently I'm struggling to work out the best way to implement the callback function.

latenitefilms:
@armadsen - Any chance you'd be interested in helping me port MIKMIDI to a Hammerspoon extension?

armadsen:
I'm not familiar with Hammerspoon, but I'm certainly happy to answer questions about MIKMIDI.

latenitefilms:
@armadsen - Amazing! Thank you so much - HUGELY appreciated!

Apologies in advance - I'm completely new to Objective-C, so sorry if these are very basic questions. Currently I'm at the "copy and paste" level of programming, where I'm basically just making my Objective-C code work by tearing apart example code.

You can see my latest code [here](https://github.com/latenitefilms/hammerspoon/blob/04ffbc90e05870b3ad02328e2d75572cec684da2/extensions/midi/internal.m).

Basically, my question is, how can I return the `channel`, `note`, `velocity` and `value` from a `MIKMIDIChannelVoiceCommand`?

Here's what I've got so far:

```
static int midi_callback(lua_State *L) {
    [skin checkArgs:LS_TUSERDATA, USERDATA_TAG, LS_TFUNCTION | LS_TNIL, LS_TBREAK];
    
    HSMIKMIDIDevice* midiDevice = (__bridge HSMIKMIDIDevice*)(*(void**)luaL_checkudata(L, 1, USERDATA_TAG));
    
    // In either case, we need to remove an existing callback, so...
    midiDevice.callbackRef = [skin luaUnref:refTable ref:midiDevice.callbackRef];
    if (lua_type(L, 2) == LUA_TFUNCTION) {
        lua_pushvalue(L, 2);
        midiDevice.callbackRef = [skin luaRef:refTable];
    }
    
    // Setup MIDI Device:
    MIKMIDIDevice *device = midiDevice.device;
    NSArray *source = [device.entities valueForKeyPath:@"@unionOfArrays.sources"];
    MIKMIDISourceEndpoint *endpoint = [source objectAtIndex:0];
    
    // Setup MIDI Device Manager:
    MIKMIDIDeviceManager *manager = [MIKMIDIDeviceManager sharedDeviceManager];
    NSError *error = nil;
    
    // Setup Event:
    [manager connectInput:endpoint error:&error eventHandler:^(MIKMIDISourceEndpoint *source, NSArray<MIKMIDICommand *> *commands) {
        for (MIKMIDIChannelVoiceCommand *command in commands) {
            
            LuaSkin *skin = [LuaSkin shared] ;
            if (midiDevice.callbackRef != LUA_NOREF) {
                lua_State *_L = [skin L];
                [skin pushLuaRef:refTable ref:midiDevice.callbackRef];
                
                // Device Name:
                NSString *deviceName = [device name];
                
                // Description:
                NSString *description = [command description];
                
                // Time Stamp:
                NSDateFormatter *dateFormatter = [[NSDateFormatter alloc] init];
                dateFormatter.dateFormat = @"HH:mm:ss.SSS";
                NSString *timestamp =[dateFormatter stringFromDate:[command timestamp]];
                
                // Command:
                // FIXME: This should return a string like "MIKMIDICommandTypeNoteOff" instead of a number value.
                MIKMIDICommandType commandType = [command commandType];
                NSString *command = [NSString stringWithFormat:@"%lu", (unsigned long)commandType];
                
                // Channel:
                // FIXME: No idea how to get the channel data:
                NSString *channel = @"Unsure";
                
                // Note:
                // FIXME: No idea how to get the note data:
                NSString *note = @"Unsure";
                
                // Velocity:
                // FIXME: No idea how to get the velocity data:
                //NSString *velocity = [NSString stringWithFormat:@"%lu", (unsigned long)self.velocity];
                NSString *velocity = @"Unsure";
                
                 // Value:
                 // FIXME: No idea how to get the value data:
                 NSString *value = @"Unsure";
                 
                // Push Values:
                lua_pushvalue(L, 1);
                [skin pushNSObject:deviceName];
                [skin pushNSObject:description];
                [skin pushNSObject:timestamp];
                [skin pushNSObject:command];
                [skin pushNSObject:channel];
                [skin pushNSObject:note];
                [skin pushNSObject:velocity];
                [skin pushNSObject:value];
                
                if (![skin protectedCallAndTraceback:9 nresults:0]) {
                    const char *errorMsg = lua_tostring(_L, -1);
                    [skin logError:[NSString stringWithFormat:@"%s: %s", USERDATA_TAG, errorMsg]];
                    lua_pop(_L, 1) ; // remove error message from stack
                }
            }
            
            NSLog(@"%@", command);
        }
    }];
    
    lua_pushvalue(L, 1);
    return 1;
}
```

I've tried looking through the MIKMIDI example code and documentation, but I haven't wrapped my head around it fully yet.

Any tips or suggestions would be great appreciated!

Basically all I'm trying to do is build a Hammerspoon extension that sends Hammerspoon MIDI information, when a user presses a key or moves a slider on a MIDI device - so it SHOULD be fairly straight forward.

Thanks so much for your time!

latenitefilms:
@asmagill - I'm slowly getting my head around this, however, I have no idea about things like garbage collection, etc. If you get a chance, would you mind having a quick look at what I've done and offer any words of wisdom? Thanks in advance!

asmagill:
Will probably be this weekend, but I'll be happy to take a look at it

asmagill:
First observations (after about 10 seconds and trying to verify it sees my usb-midi device -- it does, btw, but haven't done anything beyond discovery yet):

* `hs.midi.getDevices` doesn't exist -- `hs.midi.devices` appears to the the actual function... change docs or change function name, whichever you prefer

* and in reviewing the code to confirm the name difference, saw a note about KVO, so for reference, I got my introduction (or at least the one I understood) at http://nshipster.com/key-value-observing/.  See `hs.settings`, which uses KVO for its watchers.  You could also check out my in progress touchbar module (specifically wrt to items) or the avplayer element of guitk -- they also use KVO.

Do you want me to post notes, comments, bugs, etc. as I find them or wait until I've done a more complete review and list them all at once?

asmagill:
Line 343 pushes the first item on the stack *at the time of callback* -- this means that it is pushing the callback function onto the stack a second time, not the object userdata object as I think you intended. To push the userdata object with your current code, you'll need to replicate lines 114-118 here, changing `midiDevice`... which will get you into what I predict will be then next problem -- the callback will work for a little while and then will suddenly stop (or immediately stop if you invoke garbage collection after the callback has been invoked at least once).

I can go into more detail if you like, but in a nutshell, the way Hamemrspoon uses most userdata object, the userdata is basically a lua pointer to an objective-c object, not the object itself.  You can have multiple lua pointers (userdata's) pointing to the same objective-c object (e.g. when I invoke `new`, it returns a userdata, that's 1.  When the callback is invoked with what I think you meant to do based on the documentation, the first argument to the callback will be 2.  The userdata objects are "different" in lua because they were created at different times, but they point to the same object internally.

Now, when one of the userdatas goes out of scope and lua garbage collection kicks in, `userdata_gc` will be invoked -- this is independent of ARC so the object still exists in memory, but one of lua's pointers to it is now defunct -- the problem is that your garbage collection function does nothing to detect whether or not it is the *last* lua pointer to the object, so when it clears the callback ref, etc... *all* of the userdata objects pointing to the same objective-c object will have the callback removed.

Ok, I went into a little more detail than I originally intended, but I hope that makes clear what I predict will be an issue once you've either cleared up the mistake at line 343 or changed it so that the callback function *never* passes the userdata object as an argument (this is not my personal favored approach, but some people do prefer it: only `new` ever creates a new userdata object -- nothing else *ever* does. The down side is that if you're using multiple instances that point to different objects (say 2 different MIDI devices) at the same time, they can't use the same callback function because the callback has no way to determine which userdata stored in an upvalue (or global value) is the correct one to use. Well, you *could* use the same one if you identified the proper useradata another way or wrapped the callback before assignment so that the upvalue gets assigned the proper userdata before the wrapped function is actually assigned as the callback)

To keep track of multiple userdata objects referring to the same objective-c object, you'll either have to add a reference counter so that each time the same (objective-c) object creates a new userdata, this gets incremented, or you'll need to store a reference to the first one created and always use that so that the userdata passed back is actually always the same one.  See https://github.com/asmagill/hammerspoon_asm/blob/master/internal.m_template for the template I use which uses a reference counter and `hs.speech`'s internal.m shows the use of an actual selfRef to the object which is created whenever speech is occurring and removed once it has stopped to make sure that the speech doesn't get terminated because the userdata gets collected and the user didn't think to save it somewhere.

If this is as clear as mud, it took me a while as well and I can try and do a more thorough explanation with examples if you want.

asmagill:
(actually in my description above of the possible fix, you don't need to change `midiDevice` as you use the same name in both new and the callback block)

asmagill:
In the expanded metadata for systemMessage, it's been a while since I've reviewed the MIDI spec, but I think dataByte1 and dataByte2 are integers, not floats... if you confirm this, you should use `lua_pushinteger` rather than `lua_pushnumber` (if you're reusing code you found on the internet, you should know that prior to lua5.3 (5.2?) there was no distinction made between the two so even if the `lua_pushinteger` macro was enabled, most people didn't use it since it was just a wrapper to `lua_pushnumber` that reduced compiler warnings)

asmagill:
And since your documentation asks, I think I would like the ability to send MIDI commands added at some point as well.  No rush, get this working for your needs right now. One of us can get to it whenever.

latenitefilms:
Amazing! Thanks so much!!

I actually used your template and tutorial to get started, so I completely get what you’re saying. Still not entirely sure how to implement, but will give it a bash when back in front of the laptop and let you know when I get stuck. Thanks again!!

latenitefilms:
@asmagill - Hope all is well! I've tried to take on all your advice. I'm still struggling with the KVO code (it just produces errors) - and I'm still not sure I've set up everything correctly (i.e. currently I have multiple `MIKMIDIDeviceManager`'s, which I'm not sure is best practise?), so I'd love your advice and pointers.

latenitefilms:
@asmagill - After much trial and error, I've managed to get the KVO code working. Now it seems all the code is functional - but I'd love for you (or @cmsj ) to review it, and make sure I've done it all correctly - especially in terms of garbage collection, etc. Feel free to change around anything you think could be done better, or add/remove things as you see fit. I know you've got a million other projects happening at the moment (`hs.guitk` is looking AWESOME!) - but I've love to get this in the Hammerspoon master branch sooner rather than later so that I can start using it in CommandPost. Thanks so much for all your help and advice - HUGELY appreciated!!

asmagill:
I will try to look at it tomorrow; just merged it with my local repo and will build it tonight before heading off to bed shortly.

latenitefilms:
Legend, thanks @asmagill - very much appreciated!

FYI: I managed to spend a bit more time on this tonight, and have added initial support for sending commands. I'm open to suggestions on everything - so feel free to tweak/change/adjust anything you think needs work.

latenitefilms:
@asmagill - AMAZING!! Thank you so much - HUGELY appreciated!

All your changes make complete sense. Thank you!

In regards to:

> // FIXME: do we need to remove an existing block if one already is attached to the end point? Simple testing suggests not, but what does the library docs say?

The [doc](http://cocoadocs.org/docsets/MIKMIDI/1.6.1/Classes/MIKMIDIDeviceManager.html#//api/name/connectInput:error:eventHandler:) says:

> Used to connect to a single MIDI input/source endpoint. Returns a token that must be kept and passed into the `-disconnectConnectionforToken:` method.

...so I'm guessing we should actually be storing the "token", and using `disconnectConnectionforToken` in Garbage Collection? I'm thinking we should also put in some error detection, so if `connectInput:error:eventHandler:` returns a `nil`, we present the error message to the end user?

latenitefilms:
FYI: @asmagill - I THINK I've now addressed the callback garbage collection issue. Feel free to change anything I've done incorrectly.

latenitefilms:
@asmagill - I've added MIDI Synthesis support, and also tweaked the `hs.midi:sendCommand` function to match similar functionality to `hs.midi:callback` to keep things consistent.

The only thing left to do now is finish off `hs.midi:callback` (as I couldn't really work out how to handle `MIKMIDIPolyphonicKeyPressureCommand`, etc - but I'll have another play tonight), and test for bugs.

latenitefilms:
@asmagill & @cmsj - I've just tested out this extension, and as far as I can tell, everything seems to work as intended. Thanks SO MUCH for all your amazing help and support @asmagill - HUGELY appreciated! Feel free to make any tweaks or changes as required, but if you're both happy, I think this is all good to go into the Hammerspoon core.

latenitefilms:
FYI: One thing that will need fixing (not sure why it's broken)...

```
scripts/docs/bin/build_docs.py -o build/ --html Hammerspoon/ extensions/
Traceback (most recent call last):
  File "scripts/docs/bin/build_docs.py", line 617, in <module>
    main()
  File "scripts/docs/bin/build_docs.py", line 609, in main
    arguments.title, results)
  File "scripts/docs/bin/build_docs.py", line 527, in write_html
    write_templated_output(output_dir, template_dir, title, data, "html")
  File "scripts/docs/bin/build_docs.py", line 495, in write_templated_output
    data = process_markdown(data)
  File "scripts/docs/bin/build_docs.py", line 334, in process_markdown
    item["doc_gfm"] = md(item["doc"])
  File "/Library/Python/2.7/site-packages/mistune/__init__.py", line 970, in __call__
    return self.parse(text)
  File "/Library/Python/2.7/site-packages/mistune/__init__.py", line 980, in parse
    out = self.output(preprocessing(text))
  File "/Library/Python/2.7/site-packages/mistune/__init__.py", line 1029, in output
    out += self.tok()
  File "/Library/Python/2.7/site-packages/mistune/__init__.py", line 1039, in tok
    return getattr(self, 'output_%s' % t)()
  File "/Library/Python/2.7/site-packages/mistune/__init__.py", line 1101, in output_list
    body += self.tok()
  File "/Library/Python/2.7/site-packages/mistune/__init__.py", line 1039, in tok
    return getattr(self, 'output_%s' % t)()
  File "/Library/Python/2.7/site-packages/mistune/__init__.py", line 1110, in output_list_item
    body += self.tok()
  File "/Library/Python/2.7/site-packages/mistune/__init__.py", line 1039, in tok
    return getattr(self, 'output_%s' % t)()
  File "/Library/Python/2.7/site-packages/mistune/__init__.py", line 1062, in output_code
    self.token['text'], self.token['lang']
  File "scripts/docs/bin/build_docs.py", line 319, in block_code
    lexer = get_lexer_by_name(lang, stripall=True)
  File "/Library/Python/2.7/site-packages/pygments/lexers/__init__.py", line 115, in get_lexer_by_name
    raise ClassNotFound('no lexer for alias %r found' % _alias)
pygments.util.ClassNotFound: no lexer for alias u'hs.midi.deviceCallback(function(devices)' found
make: *** [build/html] Error 1
```

asmagill:
Lines 227-229 of internal.m, the doc parser isn't liking

     ///    ```hs.midi.deviceCallback(function(devices)
     ///         print(hs.inspect(devices))
     ///    end)```

Changing it to:

     ///    ```
     ///    hs.midi.deviceCallback(function(devices)
     ///         print(hs.inspect(devices))
     ///    end)
     ///    ```

works for me, though.

asmagill:
Something's not quite right here... I'm attached to a Kawai CN43 keyboard/piano, and when I setup a basic callback with no frills, I get this:

~~~lua
> m:callback(function(o,n,t,d,md) print(timestamp(), string.format("%s: %s (%s)", t, d, finspect(md))) end)
hs.midi: KAWAI CN43 (0x60000024fb08)
2017-11-09 09:28:28.7476	systemMessage: <MIKMIDISystemMessageCommand: 0x60c000228920> time: 09:28:28.747 command: 255 
	data: <fe> ({ dataByte1 = 0, dataByte2 = 0, timestamp = "09:28:28.747" })
2017-11-09 09:28:29.0075	systemMessage: <MIKMIDISystemMessageCommand: 0x60c00022afe0> time: 09:28:29.007 command: 255 
	data: <fe> ({ dataByte1 = 0, dataByte2 = 0, timestamp = "09:28:29.007" })
2017-11-09 09:28:29.2676	systemMessage: <MIKMIDISystemMessageCommand: 0x60c0002277a0> time: 09:28:29.267 command: 255 
	data: <fe> ({ dataByte1 = 0, dataByte2 = 0, timestamp = "09:28:29.267" })
...
~~~

(btw, timestamp is a function I wrote that prints os.date in a specific format I like for logs; leave it out if you want to test for yourself)

Now system message does correspond to 0xff in the MIKMIDI source, but the first byte of the data is 0xfe, which is actually a keep alive, which is what I would expect to be repeated ad-nauseum. And according to the MIDI spec, 0xff is supposed to be a system reset anyways, at least when sent.

When I filter these out and press a note on the keyboard like so:

~~~lua
> m:callback(function(o,n,t,d,md) if not d:match("command: 255%s+data: <fe>") then print(timestamp(), string.format("%s: %s (%s)", t, d, finspect(md))) end end)
hs.midi: KAWAI CN43 (0x60000024fb08)
2017-11-09 09:46:09.9974	noteOn: <MIKMIDINoteOnCommand: 0x608000230180> time: 09:46:09.995 command: 159 channel 0 note: 21 velocity: 73 
	data: <901549> ({ channel = 0, note = 21, timestamp = "09:46:09.995", velocity = 73 })
2017-11-09 09:46:10.4767	noteOn: <MIKMIDINoteOnCommand: 0x608000231c40> time: 09:46:10.476 command: 159 channel 0 note: 21 velocity: 0 
	data: <901500> ({ channel = 0, note = 21, timestamp = "09:46:10.476", velocity = 0 })

~~~

And look at the data portion, 0x90 corresponds to a noteOn for channel 0, 0x15 is note 21 and the last byte is in fact the velocity (I vaguely recall from messing with MIDI some years back that some devices don't use a noteOff; instead they reissue the noteOn with a velocity of 0, as shown above, so that doesn't jump out at me as odd).

So I'm pretty sure that the basic parser is correct but I'm not sure that it's handling system messages correctly... I haven't had a chance to dig up a good description of the actual bytes transmitted via the spec yet, just a high level "message X means Y" kind of thing so far, but... does the MIKMIDI home page give any examples of parsing system messages? I'd like something to compare against, even if I need to compile something here to test at a lower level.

armadsen:
@asmagill: The weird response to keep alive messages is due to those not being explicitly supported by the MIKMIDI command parsing and representation. I've just added support via a new `MIKMIDICommand` subclass `MIKMIDISystemKeepAliveCommand`. It's in MIKMIDI master right now. Try updating and see if the behavior is what you expect.

latenitefilms:
Legend, thanks @armadsen ! HUGELY appreciated!

FYI: I had to change `#import <MIKMIDI/MIKMIDI.h>` to `#import "MIKMIDI/MIKMIDI.h"` in `MIKMIDISystemKeepAliveCommand.h` to make it work.

@asmagill - I've updated to MIKMIDI v1.6.2 and added support for `MIKMIDICommandTypeSystemKeepAlive` in the callback function.

Let me know if that fixes thing for you?

asmagill:
What threw me was that I thought your code was supposed to already detect a keep alive (somewhere in the first switch in `midi_callback` there is a case to set `command` to "systemKeepAlive") but it still invoked the callback with the message "systemMessage"... apparently more was needed to properly detect it.

I'll pull the update tonight and give it a go; it was more of an observation, not sure how often I'm actually going to *care* about the system messages -- receiving and sending noteOn/noteOff will be the bulk of what I care about -- but figured on making sure there wasn't a bug somewhere in case someone does care, or my needs change.

Thanks to both of you!

latenitefilms:
@asmagill - Possibly related - I just hooked up my laptop to a Yamaha electronic piano, and I get a constant stream of:

```
2017-11-10 14:05:20: object: table: 0x60000026bc80
2017-11-10 14:05:20: deviceName: USB MIDI 1x1
2017-11-10 14:05:20: commandType: systemMessage
2017-11-10 14:05:20: description: <MIKMIDISystemMessageCommand: 0x604001224c20> time: 14:05:20.394 command: 255 
	data: <f8>
2017-11-10 14:05:20: metadata: {
  dataByte1 = 0,
  dataByte2 = 0,
  timestamp = "14:05:20.394"
}
```

...when using the following Lua code:

```
midiDevice = hs.midi.new("USB MIDI 1x1")
midiDevice:callback(function(object, deviceName, commandType, description, metadata)
		print("object: " .. tostring(object))
		print("deviceName: " .. deviceName)
		print("commandType: " .. commandType)	
		print("description: " .. description)
		print("metadata: " .. hs.inspect(metadata))
end)
```

I'm guessing the `data: <f8>` in the `description` refers to `MIKMIDICommandTypeSystemTimingClock = 0xf8` - which makes sense - however, I wonder then why the `commandType` isn't returning "systemTimingClock"?

If I look at what `[command commandType]` is returning using `NSLog`, it's 255 (which matches `MIKMIDICommandTypeSystemMessage = 0xff`) - so I'm wondering if there's actually a bug in MIKMIDI where it's incorrectly assigning `MIKMIDISystemMessageCommand` to things that should actually be `MIKMIDICommandTypeSystemTimingClock`?

Searching for `MIKMIDICommandTypeSystemTimingClock` in the MIKMIDI code reveals very little, so I'm GUESSING maybe @armadsen was treating `MIKMIDICommandTypeSystemTimingClock ` the same way as `MIKMIDISystemKeepAliveCommand`, in that it wasn't explicitly supported?

Sorry this comment must be really confusing @armadsen as we're mixing Objective-C with Lua, but if you have any ideas or suggestions let me know!

asmagill:
I think you may be right... I can confirm that my particular message is now properly reported as "systemKeepAlive", so unless you made notable changes to your code around detecting this, it must have been on his end.

Right now I'm looking at your implementation of sending a "sysEx" command... I think it's going to have to be changed to treat the input as an NSData instead of an NSString, but I'll get back to you on that, likely with a new pull later tonight/tomorrow morning...

It's kind of nice taking a break from `guitk`, though I'm almost done with the docs and once I give @cmsj another day or so to comment on the pending pulls I commented on yesterday, I want to land them then add a pull for `guitk`... it's going to involve a lot of changes to the `project.pbxproj` file so I want to land as much as possible to clear out as many potential conflicts as I can first (though I think this may be the only one in that category of conflict ATM).

asmagill:
Then again, looking closer at your code, maybe its the spaces that need to be removed... I'll get back to you that.

latenitefilms:
Ah, yes, you're absolutely right. I've just made a tweak - and now the below example code now works:

```
midiDevice = hs.midi.new(hs.midi.devices()[2])
midiDevice:callback(function(object, deviceName, commandType, description, metadata)
	if commandType == "systemKeepAlive" then return end	
	if not string.find(description, "f8") then	
		print("object: " .. tostring(object))
		print("deviceName: " .. deviceName)
		print("commandType: " .. commandType)	
		print("description: " .. description)
		print("metadata: " .. hs.inspect(metadata))
	end
end)
midiDevice:sendSysex("f07e7f06 01f7")
```

asmagill:
Interestingly, the CN43 doesn't respond to the identity request sequence... but I am able to trigger other changes via sysEx commands, so I have to assume that it works.

Regarding the `availableCommands` method: did you get that from the macOS examples in the MIKMIDI repository? I think that's just a list that is shown in the sample program -- what it is supposed to respond with, according to http://midi.teragonaudio.com/tech/midispec/identity.htm is some info about the device's identity, not it's command set... 

At the very least, the name and description of the method is unclear since it just returns the identity request command only.

asmagill:
Also, I've noted that

~~~lua
m = hs.midi.new("KAWAI CN43"):callback(function(o,n,t,d,md) if t ~= "systemKeepAlive" then print(timestamp(), string.format("%s: %s (%s)", t, d, finspect(md))) end end)
~~~

crashes Hammerspoon, but

~~~lua
m = hs.midi.new("KAWAI CN43") ; m:callback(function(o,n,t,d,md) if t ~= "systemKeepAlive" then print(timestamp(), string.format("%s: %s (%s)", t, d, finspect(md))) end end)
~~~

doesn't. Probably something that needs a few ms to settle before it becomes able to handle callbacks, but I'm not sure what yet.

Can you check and see if you get the same crash so I can see whether or not it's just something screwy on my end? (My HS is highly modified at the moment, so that's always a possibility)

latenitefilms:
Replies as follows:

> Regarding the availableCommands method: did you get that from the macOS examples in the MIKMIDI repository?

Yes - it was basically just a copy and paste job without really understanding what it did at the time. I've since removed `hs.midi:availableCommands()` and added `hs.midi:identityRequest()` - and hopefully the documentation is a bit clearer.

> Can you check and see if you get the same crash so I can see whether or not it's just something screwy on my end?

This works fine for me:

`m = hs.midi.new(hs.midi.devices()[2]):callback(function(o,n,t,d,md) if t ~= "systemKeepAlive" then print(timestamp(), string.format("%s: %s (%s)", t, d, finspect(md))) end end)`

...so I'm guessing it must be something specific to your config?

_FYI: I'm on a MacBook Pro (15-inch, 2017) running macOS 10.13._

Finally, as far as I can tell, "systemTimecodeQuarterFrame", "systemSongPositionPointer", "systemSongSelect", "systemTuneRequest", "systemTimingClock", "systemStartSequence", "systemContinueSequence" and "systemStopSequence" are all detected as a "systemMessage" `commandType` in terms of the callback, which is probably not ideal - but certainly not a show-stopper, as anyone who needs to detect these things, can always work it out based on the incoming `description`. As discussed above, I'm ASSUMING this is because @armadsen hasn't explicitly supported these commands in the Objective-C code. Like you @asmagill - I'm really only going to use `noteOff`, `noteOn` and `controlChange` anyway for the stuff I want to do (which is basically just trigger things when buttons are pressed or knobs are turned), so I don't think this should necessarily hold up the release of the extension.

This test code seems to work fine on my machine - so I THINK everything is working the way it should:

```
hs.midi.deviceCallback(function(devices) 
	print(hs.inspect(devices))
end)
midiDevice = hs.midi.new(hs.midi.devices()[2]):callback(function(object, deviceName, commandType, description, metadata)
	print("object: " .. tostring(object))
	print("deviceName: " .. deviceName)
	print("commandType: " .. commandType)	
	print("description: " .. description)
	print("metadata: " .. hs.inspect(metadata))
end)
midiDevice:identityRequest()
midiDevice:sendCommand("noteOn", {
	["note"] = 72,
	["velocity"] = 50,
	["channel"] = 0,
})
hs.timer.usleep(500000)
midiDevice:sendCommand("noteOn", {
	["note"] = 74,
	["velocity"] = 50,
	["channel"] = 0,
})
hs.timer.usleep(500000)
midiDevice:sendCommand("noteOn", {
	["note"] = 76,
	["velocity"] = 50,
	["channel"] = 0,
})
midiDevice:sendCommand("pitchWheelChange", {
	["pitchChange"] = 1000,
	["channel"] = 0,
})
hs.timer.usleep(100000)
midiDevice:sendCommand("pitchWheelChange", {
	["pitchChange"] = 2000,
	["channel"] = 0,
})
hs.timer.usleep(100000)
midiDevice:sendCommand("pitchWheelChange", {
	["pitchChange"] = 3000,
	["channel"] = 0,
})
```

Feel free to do any tweaks or changes you want prior to pushing into the Hammerspoon core! I feel like we're pretty close now!

Thanks so much for all your help and support - HUGELY appreciated! I've certainly learnt a MASSIVE amount trying to make this all work - so again, thank you.

asmagill:
@cmsj, objections?

asmagill:
@latenitefilms I wonder if it might be worth adding a way to get at the raw midi data from within the callback function without having to parse the description string... possibly an additional argument or field added to the metadata table... just the raw unfiltered unexamined contents of the actual data in the MIKMIDICommand.data property (LuaSkin would push this as a string, though it would not be a UTF8 compliant string, so the receiver, if it wanted to parse it, would have to do so byte-by-byte... or just ignore it in favor of the "cleaned up" version provided by the other arguments).

Should someone care about the system events mentioned above, this would be easier to work with then having to parse the hex portion of the description, then convert the hex string into two char chunks, then convert to decimal to get the byte value, etc.

Just a thought, at the moment I wouldn't take advantage of this, so it's not a show stopper and can always be added later.

latenitefilms:
I have no issue if you wanted to add a `data` field to the `metadata` table. However, shouldn't this information already be contained with `dataByte1` and `dataByte2`?

I'm MORE than happy for you to make tweaks to this as you see fit.

asmagill:
Take a look at the output I provided before keep alives were fully recognized or at the example you provided that shows the timing clock sync message — the initial command byte isn’t included in the meta data table... and in theory a full sysex message may contain more than 2 bytes of data and I’m not exactly sure how one could currently get at that data outside of the description field and string manipulation as outlined above.

It’s one of those details/features that I think *should* be in there at some point for completeness and because we never know what someone else might come up with that we haven’t thought of... on the other hand since we don’t need it yet, there is no rush and it may be worth digging a little deeper to see if there is a better or cleaner way to provide access to the raw unfiltered data, perhaps even a different callback all together that actually streams data byte-by-byte untouched as it comes in (I haven’t looked too closely at the MIKMIDI code so it may be doing some clean up before your code is invoked that might be worth bypassing for true “raw” access).

It’s just a thought and the more I think about it, it might be better to wait until someone with a need or a more complete understanding of the MIDI spec expresses an interest.

latenitefilms:
No problemo! I'll have a look next time I'm near a MIDI device and see what I can come up with.

latenitefilms:
@asmagill - I THINK I've done what you requested. As always, feel free to tweak and change as required.

asmagill:
I think it looks good but I won't be able to test it until the weekend.

latenitefilms:
@asmagill - No problemo! Once you've tested it, if all happy, then with @cmsj's blessing, I think it's all good to merge! Thanks so much for all your help!!

latenitefilms:
@asmagill - Do you reckon we need a `hs.midi:delete()` function as well, or is setting a `hs.midi` object to `nil` sufficient?

asmagill:
For the way you have coded `hs.midi`, setting the object to nil and then allowing garbage collection to do the rest is sufficient. You're not creating something that isn't properly cleared on the macOS side when lua garbage collection occurs.

*The rest is my opinion and you can safely skip it*

The need/option to add a delete method is somewhat of a philosophical question... what follows are my thoughts and are by no means the only ones or even the ones Hammerspoon generally follows, though I do tend to code this way unless the consensus desires otherwise -- again, I speak for myself here, not Hammerspoon in general:

Generally the objects the lua/hammerspoon userdata represents fall into one of two categories:
1. objects whose existence only makes sense when the user/program is interacting with them directly in some way -- once the user can no longer programmatically interface with the object (all of the variables referencing it are set to nil), the need to exist on the macOS side disappears as well.
2. objects which represent something that is in some sense "independent" of whether or not the user is still programmatically adjusting them. This also includes objects which have specific teardown requirements that ARC (or more generally objective-c garbage collection) can't handle alone either.

Category 1 objects might be things like a file/path watcher or in this case a midi connection -- when the program no longer holds a reference to the userdata object, the program can no longer start or stop watching, can no longer adjust the settings (if applicable), etc. so my thinking is if the user doesn't want to keep a reference, they must no longer need the object.

The downside to Category 1 objects is that they disappear at some point (sometimes seemingly randomly, though it can always be attributed to garbage collection even if it takes multiple passes) unless you save them in a global variable or as an upvalue for something else that isn't being collected.

Category 2 objects are things that represent windows (macOS sees all drawings, canvas objects, and webviews as windows, even if they don't necessarily have the "decorations" we tend to associate with application windows), menus, etc.  After they're set up, the program/user may never need to touch them again, but they still maintain an existence that the user can interact with (or at least see). If the userdata is collected, the "object" still persists.

The downside to Category 2 is that once the last reference to the userdata is collected, you can't remove the window, etc. without restarting Hammerspoon.

My general feeling is that anything which *can* clean up after itself, should (category 1). Anything which requires special steps to clean up or for which the containing userdata is basically a place holder after the side effects of its creation is in effect (category 2) requires an explicit delete, but I have to remember that its on me to clean it up if I want it to go away before restarting Hammerspoon.

latenitefilms:
@asmagill - Sweet, thanks for the detailed reply! All makes sense. I just need to remember to `collectgarbage()` after setting to `nil` then.

Let me know how you go testing that last commit, and if all happy, I think this one is good to merge away!

asmagill:
If the issue is that you want the callback to stop, just set the callback to nil. You *can* invoke `collectgarbage` whenever you want, but it will also be done automatically as Hammerspoon runs... them more active your Hammerspoon is (timers, background watchers, etc.) the more often it will run.

OTOH, you're using this with network based midi "devices", aren't you? Do you find that you need a connection to be completely dead (i.e. collected) before you can open a new one to the same "device"? If so, in my mind that justifies adding a disconnect or delete method... I'd just make it clear in the documentation that the method isn't required unless you need to fully disconnect before garbage collection would get around to it.

I'm home now and have a fully up to date version of Hammerspoon and the midi extension running, so I'll take a closer look at the latest changes tonight/tomorrow and let you know if I run into any problems!

latenitefilms:
Aha, I didn't even think about making the callback `nil` - that makes sense. Thanks! I was just setting the `hs.midi` instance itself to `nil` and was wondering why the callback was still working (until I did `collectgarbage()` obviously). That'll solve my issue.

I have tested `hs.midi` with an iOS app (PadMIDI) using the "network" MIDI device, which works great - however, I didn't even think about what happens if you want to use a different "network" device, etc. Will have to have a play. I've mainly being using it with physical MIDI keyboards.

cmsj:
@asmagill sorry for the slow reply. I don't have any objections to this in principle, but I don't have any MIDI devices to test it with, so I am very happy to defer to your excellent judgement wrt when this is ready to merge :)

latenitefilms:
@cmsj - If you do want to have a play there are plenty of free MIDI apps on the iOS App Store, which work with this. I use "PadMIDI".

latenitefilms:
@asmagill - I wanted to get your advice on something if you have time. I'm currently using a MIDI controller slider to "drive" a virtual slider in Final Cut Pro using GUI Scripting. It works - however, because the GUI Scripting code takes a few micro-seconds to do it's thing - it's all a bit laggy, as all the callbacks will build up, and it'll take a while for them all to execute. Any suggestions on how best to solve this? Basically I need to come up with a way to skip unnecessary callbacks (being the knob is still moving), but always make sure the "last" callback is triggered so that the virtual slider always matches the real slider when the user stops moving things. Any tips?

latenitefilms:
@asmagill - Someone just sent me through the following bug report. Any ideas? They're running macOS 10.11.6.

```
2017-12-08 08:39:14: *** ERROR: NSRangeException: *** -[__NSArrayM objectAtIndex:]: index 0 beyond bounds for empty array
(
	0   CoreFoundation                      0x00007fff96c27452 __exceptionPreprocess + 178
	1   libobjc.A.dylib                     0x00007fff8e8e6f7e objc_exception_throw + 48
	2   CoreFoundation                      0x00007fff96b3e775 -[__NSArrayM objectAtIndex:] + 245
	3   internal.so                         0x00000001098477db midi_callback + 492
	4   LuaSkin                             0x0000000101bb1920 luaD_precall + 695
	5   LuaSkin                             0x0000000101bafe45 luaV_execute + 7345
	6   LuaSkin                             0x0000000101bb1a76 luaD_call + 64
	7   LuaSkin                             0x0000000101bb1ab7 luaD_callnoyield + 21
	8   LuaSkin                             0x0000000101bb96bb luai_objcttry + 28
	9   LuaSkin                             0x0000000101bb1f85 luaD_pcall + 116
	10  LuaSkin                             0x0000000101bc2588 lua_pcallk + 122
	11  LuaSkin                             0x0000000101bbf0ea luaB_xpcall + 115
	12  LuaSkin                             0x0000000101bb1920 luaD_precall + 695
	13  LuaSkin                             0x0000000101bafe45 luaV_execute + 7345
	14  LuaSkin                             0x0000000101bb1a76 luaD_call + 64
	15  LuaSkin                             0x0000000101bb1ab7 luaD_callnoyield + 21
	16  LuaSkin                             0x0000000101bb96bb luai_objcttry + 28
	17  LuaSkin                             0x0000000101bb1f85 luaD_pcall + 116
	18  LuaSkin                             0x0000000101bc2588 lua_pcallk + 122
	19  CommandPost                         0x0000000101803d19 MJLuaInit + 1695
	20  CommandPost                         0x0000000101800198 -[MJAppDelegate applicationDidFinishLaunching:] + 2908
	21  CoreFoundation                      0x00007fff96bb2b0c __CFNOTIFICATIONCENTER_IS_CALLING_OUT_TO_AN_OBSERVER__ + 12
	22  CoreFoundation                      0x00007fff96bb2a9f ___CFXRegistrationPost_block_invoke + 63
	23  CoreFoundation                      0x00007fff96bb2a17 _CFXRegistrationPost + 407
	24  CoreFoundation                      0x00007fff96bb2782 ___CFXNotificationPost_block_invoke + 50
	25  CoreFoundation                      0x00007fff96b6f592 -[_CFXNotificationRegistrar find:object:observer:enumerator:] + 1922
	26  CoreFoundation                      0x00007fff96b6e7e5 _CFXNotificationPost + 693
	27  Foundation                          0x00007fff8707ef9a -[NSNotificationCenter postNotificationName:object:userInfo:] + 66
	28  AppKit                              0x00007fff91a47355 -[NSApplication _postDidFinishNotification] + 297
	29  AppKit                              0x00007fff91a470bf -[NSApplication _sendFinishLaunchingNotification] + 203
	30  AppKit                              0x00007fff91901d99 -[NSApplication(NSAppleEventHandling) _handleAEOpenEvent:] + 557
	31  AppKit                              0x00007fff91901843 -[NSApplication(NSAppleEventHandling) _handleCoreEvent:withReplyEvent:] + 250
	32  Foundation                          0x00007fff870cd54d -[NSAppleEventManager dispatchRawAppleEvent:withRawReply:handlerRefCon:] + 290
	33  Foundation                          0x00007fff870cd3c7 _NSAppleEventManagerGenericHandler + 102
	34  AE                                  0x00007fff905e2261 _Z20aeDispatchAppleEventPK6AEDescPS_jPh + 531
	35  AE                                  0x00007fff905e1fe8 _ZL25dispatchEventAndSendReplyPK6AEDescPS_ + 31
	36  AE                                  0x00007fff905e1f04 aeProcessAppleEvent + 288
	37  HIToolbox                           0x00007fff8ffaeaf9 AEProcessAppleEvent + 55
	38  AppKit                              0x00007fff918fd290 _DPSNextEvent + 2245
	39  AppKit                              0x00007fff918fc226 -[NSApplication _nextEventMatchingEventMask:untilDate:inMode:dequeue:] + 454
	40  AppKit                              0x00007fff918f0d80 -[NSApplication run] + 682
	41  AppKit                              0x00007fff918ba368 NSApplicationMain + 1176
	42  libdyld.dylib                       0x00007fff92d5b5ad start + 1
)
```

latenitefilms:
@asmagill - I've just pushed what I THINK is a bug fix. Feel free to tweak or change as required.

asmagill:
The fix looks like what I would do but I won't be able to test it myself until this weekend.  Do you know the conditions under which the crash occurred (i.e. can we replicate it so I can do a proper before and after test)?

latenitefilms:
Unfortunately not - I haven't been able to reproduce the issue on any of our machines. However, I can try reaching out to the person who submitted the bug report, and see if this fix solves their issue.

latenitefilms:
@asmagill - I've just had this bug pop up on one of our machines (macOS 10.11.6) - reckon it's related to the above bug, or something else?

```
2017-12-11 10:32:54: 10:32:54 ERROR:   LuaSkin: hs.application.watcher callback error: NSRangeException: *** -[__NSArrayM objectAtIndex:]: index 0 beyond bounds for empty array
(
	0   CoreFoundation                      0x00007fff92265452 __exceptionPreprocess + 178
	1   libobjc.A.dylib                     0x00007fff88877f7e objc_exception_throw + 48
	2   CoreFoundation                      0x00007fff9217c775 -[__NSArrayM objectAtIndex:] + 245
	3   internal.so                         0x000000010aa6f7db midi_callback + 492
	4   LuaSkin                             0x0000000107380920 luaD_precall + 695
	5   LuaSkin                             0x000000010737ee45 luaV_execute + 7345
	6   LuaSkin                             0x0000000107380a76 luaD_call + 64
	7   LuaSkin                             0x0000000107380ab7 luaD_callnoyield + 21
	8   LuaSkin                             0x00000001073886bb luai_objcttry + 28
	9   LuaSkin                             0x0000000107380f85 luaD_pcall + 116
	10  LuaSkin                             0x0000000107391588 lua_pcallk + 122
	11  LuaSkin                             0x0000000107375713 -[LuaSkin protectedCallAndTraceback:nresults:] + 203
	12  watcher.so                          0x000000010be9c726 -[AppWatcher callback:withEvent:] + 541
	13  watcher.so                          0x000000010be9c9a2 -[AppWatcher applicationActivated:] + 65
	14  CoreFoundation                      0x00007fff921f0b0c __CFNOTIFICATIONCENTER_IS_CALLING_OUT_TO_AN_OBSERVER__ + 12
	15  CoreFoundation                      0x00007fff921f0a9f ___CFXRegistrationPost_block_invoke + 63
	16  CoreFoundation                      0x00007fff921f0a17 _CFXRegistrationPost + 407
	17  CoreFoundation                      0x00007fff921f0782 ___CFXNotificationPost_block_invoke + 50
	18  CoreFoundation                      0x00007fff921ad592 -[_CFXNotificationRegistrar find:object:observer:enumerator:] + 1922
	19  CoreFoundation                      0x00007fff921ac7e5 _CFXNotificationPost + 693
	20  Foundation                          0x00007fff86aa1f9a -[NSNotificationCenter postNotificationName:object:userInfo:] + 66
	21  AppKit                              0x00007fff97eacbcb applicationStatusSubsystemCallback + 782
	22  LaunchServices                      0x00007fff927c45f4 ___LSScheduleNotificationFunction_block_invoke_2 + 47
	23  CoreFoundation                      0x00007fff921fa40c __CFRUNLOOP_IS_CALLING_OUT_TO_A_BLOCK__ + 12
	24  CoreFoundation                      0x00007fff921d9c55 __CFRunLoopDoBlocks + 341
	25  CoreFoundation                      0x00007fff921d99bb __CFRunLoopRun + 2347
	26  CoreFoundation                      0x00007fff921d8e28 CFRunLoopRunSpecific + 296
	27  HIToolbox                           0x00007fff95712935 RunCurrentEventLoopInMode + 235
	28  HIToolbox                           0x00007fff9571276f ReceiveNextEventCommon + 432
	29  HIToolbox                           0x00007fff957125af _BlockUntilNextEventMatchingListInModeWithFilter + 71
	30  AppKit                              0x00007fff97cc4df6 _DPSNextEvent + 1067
	31  AppKit                              0x00007fff97cc4226 -[NSApplication _nextEventMatchingEventMask:untilDate:inMode:dequeue:] + 454
	32  AppKit                              0x00007fff97cb8d80 -[NSApplication run] + 682
	33  AppKit                              0x00007fff97c82368 NSApplicationMain + 1176
	34  libdyld.dylib                       0x00007fff8253a5ad start + 1
)
```

latenitefilms:
Another possible bug?

```
Fatal Exception: NSInvalidArgumentException
0  CoreFoundation                 0x7fffc4bc4563 (Missing)
1  libobjc.A.dylib                0x7fffd9fa31da (Missing)
2  CoreFoundation                 0x7fffc4ae051d (Missing)
3  internal.so                    0x11037a7d8 -[HSMIDIDeviceManager observeValueForKeyPath:ofObject:change:context:] (internal.m:159)
4  Foundation                     0x7fffc658d35d (Missing)
5  Foundation                     0x7fffc658cbf4 (Missing)
6  Foundation                     0x7fffc65a0339 (Missing)
7  CommandPost                    0x106675345 -[MIKMIDIDeviceManager addInternalDevicesObject:] (MIKMIDIDeviceManager.m:388)
8  CommandPost                    0x106674fdd -[MIKMIDIDeviceManager handleMIDIObjectAddNotification:] (MIKMIDIDeviceManager.m:333)
9  CommandPost                    0x106674221 MIKMIDIDeviceManagerNotifyCallback (MIKMIDIDeviceManager.m:375)
10 CoreMIDI                       0x1068e60de (Missing)
11 CoreMIDI                       0x1068c36f4 (Missing)
12 CoreMIDI                       0x1068df6dd (Missing)
13 CoreFoundation                 0x7fffc4b44289 (Missing)
14 CoreFoundation                 0x7fffc4b44201 (Missing)
15 CoreFoundation                 0x7fffc4b3c105 (Missing)
16 CoreFoundation                 0x7fffc4b3b544 (Missing)
17 HIToolbox                      0x7fffc409aebc (Missing)
18 HIToolbox                      0x7fffc409acf1 (Missing)
19 HIToolbox                      0x7fffc409ab26 (Missing)
20 AppKit                         0x7fffc2631a54 (Missing)
21 AppKit                         0x7fffc2dad7ee (Missing)
22 AppKit                         0x7fffc26263db (Missing)
23 AppKit                         0x7fffc25f0e0e (Missing)
24 libdyld.dylib                  0x7fffda884235 (Missing)
```

asmagill:
The most recent crash you show is a different one related to KVO... I'll have to look closer at your code tonight/tomorrow and see if something is maybe being collected out of order or if a watcher isn't getting cleared soon enough. 

(I remember having a weird issue with KVO a while back where I had to put special checks in the object destructor because the watcher wasn't always being released before ARC collected the object even though it was being cleared during lua garbage collection... It was weird because the exact same code template has worked everywhere else I've used KVO, it was just this one case I had to take special steps... of course I can't recall which one it was or even if it was in core or in one of my third-party modules, so I'll have to do some searching...)

The previous crash (2 messages up) was likely caused by the same thing that prompted the last fix... an array which *shouldn't* be empty apparently can be... make sure everywhere you're using an NSArray or NSMutableArray, you check that it's count is > 0 (or the index you're interested in, but I think both of these were trying to access element 0) before getting the object and at least you can control how it errors out instead of generating the crash dump.

cmsj:
@latenitefilms FYI, this PR appears to have picked up too many commits, or the wrong branch or something - it's a full merge of CommandPost at the moment :)

latenitefilms:
Opps. Will try and fix.

latenitefilms:
Had a quick play, but no idea how to undo whatever i

latenitefilms:
So... it looks in the last commit I actually merged the main CommandPost repo into this branch. Oops! I have no idea how to undo that, so I might have to submit a new PR. Will try fix later tonight. If you want to merge this more urgently though, you could always just grab the two files within the “extensions/midi” folder, scrap the rest and reconfigure the Xcode project. Sorry!

latenitefilms:
@cmsj - I THINK this is now fixed in https://github.com/Hammerspoon/hammerspoon/pull/1629

Any issues let me know! Apologies - I'm still pretty rubbish at Git!

PR #1616, opened 2017-12-08
Extended hs.osascript with functions to execute as/js from a file

rkalis:
This allows people to import an AppleScript file to execute, instead of writing it inline in a Lua file. This also makes it easier to debug these scripts before integrating it with Hammerspoon as the files can be run separately first.

### DIFF
--- a/extensions/osascript/LICENSE
+++ b/extensions/osascript/LICENSE
@@ -1,7 +1,8 @@
 The MIT License (MIT)
 
 Original work Copyright (c) 2014 Aaron Magill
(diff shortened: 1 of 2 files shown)


cmsj:
@rkalis please ignore the stickler-ci comments for now - we've only just enabled it and I'm still bedding it down as to what is reasonable :)

rkalis:
@cmsj Aah good to know, I was confused already, as some of the new lines of documentation (like the returns) were exactly the same as some existing ones 😅😅

cmsj:
Thanks! 👍

PR #1622, opened 2017-12-16
Update .stickler.yml

latenitefilms:
@cmsj - I THINK this is what's missing?


cmsj:
Thanks. I hope nobody noticed the irony of me failing to configure the feature I added to Stickler ;)

PR #1625, opened 2017-12-20
Add a default toolbar to hs.console. Closes #1475

### LINKED ISSUE #1475: Can't get back to hammerspoon preferences
Hello there, 

I disabled both the dock and the menu bar icons. However, when I launch hammerspoon from spotlight or the terminal (`open -a Hammerspoon`), I'm greeted with only the console, not the preference window. Am I missing something?

Thanks!

cmsj:
*CL* *Added*: Default toolbar is now present in the Console window

Closes #1475

### DIFF
--- a/extensions/_coresetup/init.lua
+++ b/extensions/_coresetup/init.lua
@@ -568,6 +568,25 @@ hs.fileDroppedToDockIconCallback = nil
     end)
     return result
   end
+
+  -- Set up the default Console toolbar
+--- hs.console.defaultToolbar
+--- Constant
+--- Default toolbar for the Console window
+---
+--- Notes:
+---  * This is an `hs.toolbar` object that is shown by default in the Hammerspoon Console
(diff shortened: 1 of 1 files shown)


cmsj:
@asmagill quick five minute first-draft of what this could be. One thing I noticed, when I click on the Help button I get a few exceptions logged from WKWebView that bits of it are being called on background threads that shouldn't be, which I think is in `extensions/webview/internal.m`'s `webview_url()`. Is that `dispatch_async()` necessary?

asmagill:
I'm not seeing any type of logging from attempts to use the background thread when I paste the code into my setup... I'll try a stock clean install later tonight and see if it's something in my settings or custom built Hammerspoon.

The async call was added with https://github.com/Hammerspoon/hammerspoon/pull/1153 and does no actual lua work itself; it just delays loading a new url if the current one hasn't finished rendering (which was causing errors if you changed the URL too quickly with some dynamically generated pages)

cmsj:
@asmagill try running it from Xcode, the exception was from https://developer.apple.com/documentation/code_diagnostics/main_thread_checker

PR #1628, opened 2017-12-23
Try switching NSData conversion options to basically not happening at all, in hs.socket

cmsj:


latenitefilms:
@cmsj & @asmagill - Is it worth updating CocoaAsyncSocket from 7.4.3 to 7.6.2 whilst you're working on this extension, or does that have the potential to break things?

latenitefilms:
@cmsj & @asmagill - Completely unimportant, but just curious... Why does the use of `hs.socket` create this message in the Console:

`2017-12-24 16:38:28: -- Loading extension: socket`

...even when you use `socket = require("hs.socket")`? I can't see anything in `socket/init.lua`, so I'm assuming the cause must be somewhere within the Objective-C code?

asmagill:
@cmsj, changing lines 177-179 of internal.m to

~~~objc
    [skin requireModule:"hs.socket"] ;
    lua_getfield(skin.L, -1, "timeout") ;
~~~

should get rid of the notification @latenitefilms is talking about.

cmsj:
Since we just had a release, I'm merging this for now - please shout anyone if you run into issues with it :)

latenitefilms:
Awesome, thank you!

PR #1632, opened 2017-12-26
move webview load checks back to main thread

asmagill:
@cmsj, per #1625, I can confirm running Hammerspoon in XCode shows "Main Thread Checker: UI API called on a background thread" for the WKWebView loading property because running within XCode includes the wrapper library described here: https://developer.apple.com/documentation/code_diagnostics/main_thread_checker

In this particular case, I think the checker is being a little overprotective, but that doesn't mean it won't change in the future, so... as suggested, I've moved the checks back into the main queue, though still in an async block to hopefully continue to prevent the issue described in #1151.

@latenitefilms, can you test this pull and see if this modification of #1153 still prevents the -999 errors you were seeing?

### DIFF
--- a/extensions/webview/internal.m
+++ b/extensions/webview/internal.m
@@ -879,7 +879,7 @@ static int webview_url(lua_State *L) {
         NSURLRequest *theNSURL = [skin luaObjectAtIndex:2 toClass:"NSURLRequest"] ;
         if (theNSURL) {
             if (theView.loading) [theView stopLoading] ;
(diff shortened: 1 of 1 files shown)


latenitefilms:
@asmagill - Will do!

cmsj:
@asmagill thanks. this seems like the most sensible option, so let's see what @latenitefilms finds :)

latenitefilms:
@asmagill  & @cmsj - Sorry for the delayed reply. Looks like it's working fine for me! Merge away!

PR #1643, opened 2018-01-08
hs.keycodes - Fix input source switch failure in case of input method has the same bundle ID with parent

waynezhang:
`TISInputSourceRef` passed to `TISSelectInputSource` should always be selectable


cmsj:
@asmagill I think this makes sense - any thoughts?

cmsj:
Thanks!

Issue #1407, opened 2017-05-07
Detecting User Inactivity?


latenitefilms:
Is there any API already built into Hammerspoon to detect user inactivity?

If, not it might be a useful callback for `hs.host`?

Thoughts @asmagill & @cmsj ?

This might be relevant? http://www.xs-labs.com/en/archives/articles/iokit-idle-time/

asmagill:
Looks pretty straightforward to query the idle time... I don't know much about IOKit, but I don't think we'd be able to leverage the OS to generate a callback when it exceeded a specific value (if someone knows otherwise, please correct me if I'm wrong!).

Like pasteboard changes, you'd need to write your own timer to check at whatever granularity you cared about and then perform whatever action.

latenitefilms:
Awesome! Well, I'd be very happy to just use a `hs.timer` if someone smarter than me is able to add the idle time to `hs.host`. Thanks @asmagill !

PR #1666, opened 2018-01-30
Fix up some static analysis issues found by Xcode

cmsj:


cmsj:
@asmagill I'd like to get your input on these before committing them. These don't clean up all of the static analysis issues found, partly because I don't understand some of the others, and partly because I think at least two of them are wrong/unhelpful.

extensions/host/internal.m
```diff
@@ -557,13 +557,11 @@ static int hs_idleTime(lua_State *L) {
         CFNumberGetValue((CFNumberRef)idle, kCFNumberSInt64Type, &time) ;
     } else {
         IOObjectRelease(ioIterator) ;
-        CFRelease(idle) ;
```
cmsj: I don't think we should be calling `CFRelease()` on `idle` - it was retrieved from an array we own, with a `Get` call, which means we are not responsible for freeing it, just the overall array.

extensions/host/internal.m
```diff
@@ -557,13 +557,11 @@ static int hs_idleTime(lua_State *L) {
...
         CFRelease(properties) ;
         return luaL_error(L, "Unsupported type %d for HIDIdleTime", type) ;
     }
 
     IOObjectRelease(ioIterator) ;
-    CFRelease(idle) ;
```
cmsj: ditto.

extensions/host/internal.m
```diff
@@ -692,6 +690,7 @@ static int hs_vramSize(lua_State *L) {
                         if (ValueInBytes) Size >>= 20;
 
                         lua_pushnumber(L, Size) ;
+                        CFRelease(VRAMSize);
```
cmsj: Per https://developer.apple.com/documentation/iokit/1514537-ioregistryentrysearchcfproperty?language=objc - we are responsible for freeing the result of `IORegistryEntrySearchCFProperty()`

extensions/httpserver/internal.m
```diff
@@ -37,7 +37,7 @@ @interface HSHTTPServer : HTTPServer
 
 @interface HSHTTPDataResponse : HTTPDataResponse
 @property int hsStatus;
-@property (nonatomic, copy) NSMutableDictionary *hsHeaders;
+@property (nonatomic, copy) NSDictionary *hsHeaders;
```
cmsj: The `copy` property implies an immutable copy will be made, so marking the property as `NSMutableDictionary` makes no sense.

extensions/ipc/internal.m
```diff
@@ -129,7 +129,11 @@ static int ipc_remotePort(lua_State *L) {
     NSString *portName = [skin toNSObjectAtIndex:1] ;
 
     HSIPCMessagePort *port = [[HSIPCMessagePort alloc] init] ;
-    if (port) port.messagePort = CFMessagePortCreateRemote(NULL, (__bridge CFStringRef)portName) ;
+    if (port) {
+        CFMessagePortRef portRef = CFMessagePortCreateRemote(NULL, (__bridge CFStringRef)portName) ;
+        port.messagePort = portRef;
+        CFRelease(portRef);
+    }
```
cmsj: `HSIPCMessagePort`'s `messagePort` property is implicitly `strong`, so it will retain a reference. If we don't `CFRelease()` here we're leaking.

extensions/streamdeck/HSStreamDeckManager.m
```diff
@@ -46,7 +46,9 @@ - (id)init {
         self.discoveryCallbackRef = LUA_NOREF;
 
         // Create a HID device manager
-        self.ioHIDManager = IOHIDManagerCreate(kCFAllocatorDefault, kIOHIDManagerOptionNone);
+        IOHIDManagerRef managerRef = IOHIDManagerCreate(kCFAllocatorDefault, kIOHIDManagerOptionNone);
+        self.ioHIDManager = managerRef;
+        CFRelease(managerRef);
```
cmsj: Ditto.

extensions/window/internal.m
```diff
@@ -188,7 +188,6 @@ static BOOL window_presstab(AXUIElementRef win, CFIndex tabIndex) {
 
     worked = YES;
 cleanup:
-    if (tab) CFRelease(tab);
```
cmsj: Per the first item in this review `tab` was obtained from an array with a `Get` function, so we don't own its memory.

cmsj:
@asmagill There are three other analysis issues in the code, two of which relate to callbacks leaking memory, but the docs for those callbacks explicitly state that the caller will `CFRelease()` the result, so I'm not too worried about those. There is one other on `markdown.c` in `hs.docs` which suggests a possible codepath where `calloc()` is called with an argument of `0`. Perhaps you could take a look at that? The code is a bit complex for me to just go wading in with my boots on, and ruin everything ;)

extensions/host/internal.m
asmagill: That was blindly copied and pasted from the source referenced in #1407; since that line's only reached if an error occurs, I don't think it has actually ever been invoked.  I agree with its removal and suspect it was just an oversight on my part.

extensions/host/internal.m
asmagill: Same source as above, but I'm a little surprised it never caused an issue during my testing since this would get called everytime a number is returned... however, I agree that one would expect it to follow the Get rule.

 `IORegistryEntryCreateCFProperties` creates the dictionary, and the docs don't say whether or not that uses the default value callback functions when doing so... I guess try we try it and see.

extensions/host/internal.m
asmagill: Likely also a copy-paste mistake.

extensions/httpserver/internal.m
asmagill: no objections

extensions/ipc/internal.m
asmagill: Are core foundation type properties default strong? I always thought strong was limited to Objective-C objects and CF objects were treated like any other pointer.

Explicitly making it strong and compiling results in:

~~~
hammerspoon/extensions/ipc/internal.m:14:1: error: property with 'retain (or strong)' attribute must be of object type
@property (strong) CFMessagePortRef   messagePort ;
~~~

cmsj: Huh, that’s a really good point. I’m not sure what to do about this one now.

asmagill: I did come across a discussion in the clang forums that talks about casting CF types when defining them as properties to NSObject so that ARC could handle memory management within the auto-synthesized getter and setter functions -- they were clear, though, that it *only* affected the code within the synthesized functions and use of the ivar within the class (i.e. accessing it as `_var` instead of as `self.var`) still required us to handle proper memory management... I'll see if I can find links to the official documentation.

Probably the first thing I need to do is get my setup caught up to yours, pull out a few of my more experimental modules that I'm not really working on at the moment, and run with the same analyzers and debug options within XCode so I can see the same reports you're getting in real time and in context.  Tonight is pretty busy, but I'll shoot for being in a position to compare apples to apples tomorrow.

cmsj: Thinking about this some more, I think Xcode's analysis error may well be simply that it doesn't understand what's happening with the memory - it sees us assign it to a property, probably knows that the property won't retain, and then the end of the function arrives. It can't track it across to the `__gc` functions where the memory is released, so there probably isn't actually a leak here, just an implicit behaviour. I'm not super worried about that, but it would be nice if we could find a way to hint to the compiler that we know what we're doing :)

PR #1668, opened 2018-02-01
Switch LuaSkin initialisation to allow for setting the delegate extremely early - before Lua is bootstrapped, so we can always have a logging delegate available.

cmsj:


latenitefilms:
@cmsj - I still can't get it to work, but that's maybe because of my poor Git skills.

cmsj:
@latenitefilms what's not working for you?

latenitefilms:
@cmsj - I still couldn't see any messages in Fabric when I tried merging this code into CommandPost.

cmsj:
@latenitefilms odd. I just did a release build, ran it outside of Xcode, did `hs.crash.crash()` in the console, re-launched the build and it uploaded: http://crashes.to/s/76e3d506b85 - click on "View all sessions" and it has all the `require` logs.

latenitefilms:
@cmsj - Found it! Thanks!!
screen shot 2018-02-04 at 12 28 39 am

PR #1685, opened 2018-02-14
Updated MIKMIDI & Documentation Tweaks

latenitefilms:
- Updated MIKMIDI to use master branch
- Updated in-line channel documentation


latenitefilms:
@cmsj & @asmagill - I updated the `podfile` to use the MIKMIDI master branch, so that we can take advantage of a few bug fixes in there. I also corrected a mistake in the in-line documentation.

Merge away if happy!

cmsj:
My concern there is that every time I do a `pod update` in the future, we're just going to unquestioningly grab the latest master of MIKMIDI, which might cause issues if they change their API.

Is it possible to pin the pod to a specific revision?

latenitefilms:
That’s what I was originally planning to do, but then we need to wait for the developers to push out a “release” before we can take advantage of the changes. I’d prefer to merge in the bug fixes as soon as we can. The master branch always seems to be “stable”, and if they do change the API, I’ll jump onto updating hs.midi ASAP if that works for you?

latenitefilms:
@cmsj - Alternatively, I could just do the latest and great MIKMIDI changes in CommandPost, however, will that cause sync issues when I merge from Hammerspoon to CommandPost (which I basically do straight away, as soon as you make a change)?

cmsj:
@latenitefilms perfect, thanks!

Issue #1600, opened 2017-11-13
Configurable fade speed for alerts


pbdeuchler:
Is it possible to change the function signature of `hs.alert.show` to control the speed of fade out/fade in for alerts? I'm pretty proficient with lua but I couldn't figure out where to make the change, if someone points me in the right direction I could probably whip up a PR in the next couple of days.

cmsj:
Modifying `hs.alert.show()` will be tricky because its final parameter is an optional number, so if we add another optional number it will be impossible to tell which one the user meant.

If you take a look at `extensions/alert/init.lua` you'll see that `0.15` is used throughout as a constant value for the fade time. My suggestion would be to make that a module level variable that users can override. See `hs.alert.defaultStyle` for a reasonable indication of how we declare/document module variables (although obviously for the fade time it would be a number not a table :)

I'm going to close this because I think I've covered what you want, but please feel free to re-open it if you want to discuss this further 🙂

PR #1692, opened 2018-02-18
Add Fade in/out duration to hs.alert

trofimovilya:
Add fadeInDuration and fadeOutDuration property to style parameter

There is such issue:
[Issue 1600](https://github.com/Hammerspoon/hammerspoon/issues/1600)

Particularly for me it is useful in order to have an ability to instantly recreate an alert without FadeIn effect.
I tried to use `hs.alert` to alert about Spotify/iTunes volume changes. So when I press for example a volume down button, I expected to have a single alert showing current volume state of a player. I can prevent multiple alert by closing the alert using `closeSpecific` function without fade out effect, but new alert is created above the old one with fade in effect and it looks like a sort of blink effect.


cmsj:
Thanks!

PR #1709, opened 2018-03-16
[READY] Issue #1708 - Fixed callback bug in hs.tangent

### LINKED ISSUE #1708: Bug in hs.tangent potentially throwing away useful data
Andy from Tangent reports...

> **Multiple controls**
> When moving the trackerballs in the X axis only they behave OK however when moving vertically (together with some small unintentional horizontal movement) they are very steppy. It’s as if the horizontal movement takes priority and the vertical movement is only processed occasionally. I can achieve the same effect by moving Temperature and Hue at the same time – Temp takes priority and Hue only responds if you stop moving Temp. I’m pretty sure that I know what’s going on here:
>  
> If multiple controls are moved at the same time on the panel (e.g. a single trackerball which is effectively 2 encoders) then the message which you receive from the Hub will messages encapsulate a number of commands. Each message starts with a 32 bit integer value which indicates the number of bytes that will follow to complete the message. Commands then follow one by one until the message is complete.
>  
> **Format:**            `<numBytes>, <command1>, <command2>, …`
>  
> I strongly suspect that you’re reading the first command and then discarding any subsequent commands within the message.
> I’ve been able to confirm this by swapping the Temp and Hue parameters over on the panel. Now Hue takes priority over Temp i.e. if more than one command is initiated within the poll period (40mS) then the command caused by the lowest ID encoder is acted upon while others are discarded.

@cmsj or @asmagill - Any chance you could help me solve this?

latenitefilms:
- Fixed bug in `hs.tangent` where only the first command in a series of
commands was being processed.
- Closes #1708


latenitefilms:
@cmsj - This seems to be working now, so I believe this is all good to merge!

Thanks so much for all your help!!

PR #1733, opened 2018-04-07
Add Lua stack guarding functions to all C->Lua entrypoints, since they must have zero side-effects on the stack

cmsj:


cmsj:
So this is the next step on my path to causing everything to crash until we figure out why things are crashing.

The tl;dr here is that when we go from a C callback into Lua, we must take great care to leave the Lua stack as we found it. I think we all knew that at some point, and I would note that @asmagill seems to be very good at it, but I thought it might be sensible to formalise things with a couple of macros that kill Hammerspoon if it's not the case.

Turning on the API checker has revealed a few `stack overflow` situations, and I think some of them are happening because C->Lua codepaths are slowly leaking things onto the stack until it fills up. So, in the same spirit of enabling the API checker, I want to enable these stack guard macros in release builds, so we can figure out quickly, which paths are leaking, and fix them (and also prevent new ones from creeping in by always using these macros).

Any objections? :)

asmagill:
I whole heartedly agree, though my ability to help address some of the crashes as we get reports will remain spotty for the next month or so; that said, it's good to see such progress and attention to the project and I will help as and when I can!

cmsj:
One detail that I think I want to get cleared up is about stack balancing when Lua is calling a C function - my current understanding is that we don't have to take care to pop everything in that situation because we specify how many values we're returning, so Lua is able to remove everything else from the stack for us.

@asmagill does that match your understanding? If not, we're going to need a whole heap more stack guarding ;)

asmagill:
This ended up being longer then I expected, so **tl;dr** -- I think you are correct -- within a function invoked through the Lua interpreter, the function's context is "isolated" so we can be as messy as we want within the function and be assured that only the specified number of arguments will be on the stack when it completes, or 1 in the case of a lua error being thrown, e.g. `luaL_error` or `luaL_argerror`.

-- more detail of my observations, and some things I've tried and thought about on this topic --

As I understand it, if something is ultimately invoked with `lua_call` (which is invoked by `lua_pcall` which is used by LuaSkin's `protectedCallAndTraceback:nresults:`) what appears to happen (I'm not sure about the actual implementation details, but this is what I've observed) is that the function and arguments passed in are popped from the stack, a "pseudo stack" is created for use within the c-function, and then upon completion, only the specified number of return arguments are pulled from the "pseudo stack" and put onto the main stack (explicit nil's are added if the pseudo stack is smaller then the specified number of return arguments). The pseudo stack is also cleared, so anything extra created within the function itself is automatically taken care of for us. The one caveat to keep in mind is that even if we specify 0 return arguments, if the function results in a lua error, it will always return 1 result on the main stack -- the error string, which we need to pop in the case of an error to achieve a net zero effect on the stack.

What I have done in the past to verify I haven't missed anything for an Objective-C callback function (e.g delegate method, timer callback, etc.) is something like this (this is a contrived example, but it should get the idea across):

~~~objc
- (void)delegateCallbackExample:(NSObject *)object {
  LuaSkin *skin = [LuaSkin shared] ;
  int stackTop = lua_absindex(skin.L, lua_gettop(skin.L)) ;

  ... prep work if any for the Objective-C part of the callback ...

  ... push function and argument ...

  if (![skin protectedCallAndTraceback:argCount nresults:0]) {
    // an error occurred, so stack pos -1 will have an error string... do what we want with it then...
    lua_pop(L, 1) ;
  } // else {
    // in this case we didn't have return arguments to worry about, but if we did, act on them then
    // pop them as well.
  // }

  ... follow-up work if any for the Objective-C part of the callback ...

  int stackTop2 =  lua_absindex(skin.L, lua_gettop(skin.L)) ;
  if (stackTop != stackTop2) {
    [skin logWarn:[NSString stringWithFormat:@"*** Stack Mismatch: %ld upon entry, %ld on exit", stackTop, stackTop2]] ;
  }
}
~~~

I've thought about ways to integrate this automatically into LuaSkin but all I've been able to come up with is:
1. check if `lua_pcall` is leaking within `protectedCallAndTraceback:nresults:` (which I've never observed and would indicate a Lua problem and not specifically a Hammerspoon one)... it wouldn't be able to tell us anything about the Objective-C code  before or after it invokes `protectedCallAndTraceback:nresults:`.
2. Create something like a stack checkpoint method and then a second method to verify that the stack size has only changed by the amount specified (most likely 0 most of the time) -- this would require us to add lines to the beginning and end of each Objective-C callback/delegate block, though... something like:

~~~objc
- (void)delegateCallbackExample:(NSObject *)object {
  [LuaSkin setStackCheckPoint] ;

  LuaSkin *skin = [LuaSkin shared] ;

  ... prep work if any for the Objective-C part of the callback ...

  ... push function and argument ...

  if (![skin protectedCallAndTraceback:argCount nresults:0]) {
    // an error occurred, so stack pos -1 will have an error string... do what we want with it then...
    lua_pop(L, 1) ;
  } // else {
    // in this case we didn't have return arguments to worry about, but if we did, act on them then
    // pop them as well.
  // }

  ... follow-up work if any for the Objective-C part of the callback ...

  [LuaSkin clearStackCheckPointWithDelta:0] ; // in this case we want a net 0 change
}
~~~

And then `clearStackCheckPointWithDelta:` would log a message when the stack change was other then what was expected.

Even in this case, though, we'd need to enforce it through code review as it requires us determining where we want to look for leaks -- I can't think of any way to automate it more.

If you've got other ideas, though, I'd love to hear them.

cmsj:
@asmagill thanks for the detailed reply :)

I checked the Lua docs and they concur - Lua->C transitions do get a temporary stack, so that's good, we can ignore those (within reason - the stack only comes with 20 slots, so we should pay attention for places where we're pushing things in a loop. I added `[LuaSkin growStack]` for that recently, and have deployed it in a few places, although probably some of them are unnecessary and were reactions to stack overflow crash reports).

The error-message-on-stack-after-pcall I've attempted to mitigate by adding `[LuaSkin protectedCallAndError]` which handles the error condition by calling `[LuaSkin logError]` and popping the error, and I've moved almost all of our calls to `protectedCallAndTraceback` to the new one. It's not appropriate 100% of the time, but I think it is a safer method to default to using.

The stack checkpoint idea is interesting - IIUC you'd forcibly clear the stack down to the expected size? I think I might prefer it if we just balance the stack properly in the C callbacks, which is what this PR is aimed at forcing us to do. Also FWIW I did it with `#define`s so they're easily changed/NOOPed, and lighter weight than a LuaSkin method call. (I'm intending on extending them a little so in debug builds, they'll actually log the entry/exit stack sizes).

As much as I'd like to just land this PR as-is, that feels a bit unfair on the poor users who will have to deal with `assert()` aborting their instances when an untidy callback gets triggered, so I'll try and at least contrive a simple test of each affected callback locally, before I commit :)

cmsj:
Ok, I tested almost all of these properly, but the last few I did by verifying the code by hand. Hopefully I didn't miss anything.

One slightly concerning thing was the changes I've made in hs.location - the documentation said all of the callback entrypoints should be passing `locationObject`, but many were not, and one pushed 4 arguments, but claimed to `protectedCallAndError` that it had only pushed 3, so that was likely a stack leak.

PR #1750, opened 2018-04-17
Add hs.chooser.globalCallback and refactor hs.chooser to be able to support it

cmsj:


cmsj:
I'm going to merge this now so I can get some daily testing on it myself. @latenitefilms - you use hs.chooser more than I do, could you let me know if you spot any problems?

latenitefilms:
@cmsj - Will do! I'll merge into CommandPost shortly, so we'll see what explodes.

PR #1791, opened 2018-05-12
Better docs.

wkoszek:
About Wiki I learned from IRC. Instead of doing this, put Wiki reference here.

There should be simple reference to Hello-World example, which I'm putting. This is to validate whether everything is fine.


cmsj:
Thanks for this - I slightly reworked the wording/layout and pushed the changes :)

Issue #1680, opened 2018-02-11
Request: API to control LEDs on the keyboard


prsquee:
Hello! 

I have my capslock key remapped for hyper, left the LED unused. It would be nice to have an API to control/watcher for the LEDs on the keyboard. Could be used to indicate new notifications while the screen is off. 

Reference:
http://osxbook.com/book/bonus/chapter10/kbdleds/
http://osxbook.com/book/bonus/chapter10/kbdleds/download/keyboard_leds.c

thanks!

blahgeek:
I'd also love to see this feature implemented.

Currently `hs.hid.capslock` API can change the caps lock state (and of course the caps lock LED would change with it as well). But it would be great to have another API to control caps lock LED only without affecting caps lock state (say, `hs.hid.capslock_led`?)

The reference code can be found [here](https://developer.apple.com/library/content/samplecode/HID_LED_test_tool/Listings/main_c.html), it should be trivial to add it to hammerspoon if anyone has some time :)

PR #1802, opened 2018-05-26
Add hs.hid.led to control LEDs on the keyboard

blahgeek:
#1680 

```
-- turn on caps lock LED
hs.hid.led.set("caps", true)
-- turn off num lock LED
hs.hid.led.set("num", false)
```

The key difference between `hs.hid.led.set("caps", true)` and `hs.hid.capslock.on()` (existing API) is that, `hs.hid.led` toggles the LED state directly without changing the caps lock.

The API is a little bit inconsistent with `hs.hid.capslock` but I didn't come out a better way.

### DIFF
--- a/extensions/hid/led.m
+++ b/extensions/hid/led.m
@@ -1,0 +1,128 @@
+#import <Foundation/Foundation.h>
+#include <mach/mach_error.h>
+#include <IOKit/hid/IOHIDUsageTables.h>
+
+static NSMutableDictionary* _CreateMatchingDict(Boolean isDeviceNotElement,
+                                                uint32_t inUsagePage,
+                                                uint32_t inUsage);
+
+static NSMutableDictionary* _CreateMatchingDict(Boolean isDeviceNotElement,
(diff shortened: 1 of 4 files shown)


extensions/hid/led.m
```diff
@@ -0,0 +1,128 @@
...
+                    //IOHIDDeviceClose(refs[deviceIndex], 0);
+                }
+                break;
+            }
+        }
+        CFRelease(elements);
+    }
+    CFRelease(mgr);
+    
+Oops:
```
cmsj: `Oops` also needs to check if `mgr` is non-NULL and `CFRelease()` it, since there are a couple of places that could jump to `Oops` with `mgr` set.

blahgeek: Thanks for the review! New commit pushed.

cornfeedhobo:
@cmsj Could you take another review of this? This would be a really nice feature for those that customize the touchbar.

cmsj:
Thanks!👍

PR #1814, opened 2018-06-07
Cleanup cleanup-canvas

cmsj:


cmsj:
I'm going to merge this because it seems reasonable. If the `local currentTransform` part breaks something, @asmagill can yell at me 😁

asmagill:
I'll run some tests tonight or tomorrow, but I think you're correct -- I think I intended to build off of that transform when append was true before remembering that I could just chain off of the existing transform (since it's just a table with metamethods) as done a couple of lines later with `obj[index].transformation:translate(point.x, point.y)...`

PR #1819, opened 2018-06-07
Cleanup cleanup-doc

cmsj:


cmsj:
This one needs a closer look before merging, there are some odd looking changes.

extensions/doc/makeLuaDocs.lua
```diff
@@ -195,7 +195,8 @@ print("++ Parsing manual...")
 
 -- Get manual version first -- first pass we ignore keys that aren't numbers
 
-    local a, b, k = 1, 0, ""
+    local a = 1
+    local b, k
```
cmsj: This seems potentially concerning.

extensions/doc/makeLuaDocs.lua
```diff
@@ -230,8 +231,8 @@ print("++ Parsing manual...")
 -- ignorable, as they're likely links within the manual itself to lists already included in a manual section,
 -- but post them just in case I need to look for them later if the format changes.
 
-    local a, b, k = 1, 0, ""
-    local text, posTable = nil, nil
+    a = 1
```
cmsj: as does this

extensions/doc/makeLuaDocs.lua
latenitefilms: `b` and `k` are redefined straight away [here](https://github.com/cmsj/hammerspoon/blob/d5c32b2799b534ae355cfeb3615831f052cc1f42/extensions/doc/makeLuaDocs.lua#L203):

```lua
a, b, k = LuaManual:find("[\r\n][^\r\n]*<[aA][^\r\n]*%s+[nN][aA][mM][eE]%s*=%s*\"([%d%.]+)\"[^\r\n]*>", a)
```

Therefore, their original value is never used.

extensions/doc/makeLuaDocs.lua
latenitefilms: `b` and `k` are already defined as `local` [earlier in the code](https://github.com/cmsj/hammerspoon/blob/d5c32b2799b534ae355cfeb3615831f052cc1f42/extensions/doc/makeLuaDocs.lua#L199). Their initial values are also overwritten [here](https://github.com/cmsj/hammerspoon/blob/d5c32b2799b534ae355cfeb3615831f052cc1f42/extensions/doc/makeLuaDocs.lua#L203).

`text` and `posTable` are also already defined as `local` [earlier in the code](https://github.com/cmsj/hammerspoon/blob/d5c32b2799b534ae355cfeb3615831f052cc1f42/extensions/doc/makeLuaDocs.lua#L200).

PR #1822, opened 2018-06-07
Cleanup cleanup-expose

cmsj:


extensions/expose/init.lua
```diff
@@ -318,11 +321,11 @@ local function getHints(self,windows)
     end
   end
   local function normalize(t,n) --change in place
-    local _
     while #t>0 and tlen(t.apps)>0 do
       if n>self.ui.maxHintLetters or (tlen(t.apps)==1 and n>1 and not hasSubHints(t))  then
         -- last app remaining for this hint; give it digits
-        local app=next(t.apps)
```
cmsj: I haven't checked this in detail yet, but I'm wondering if the `next()` call is important here even if `app` isn't used?

latenitefilms: @cmsj - Good point - that makes sense. It should probably just be `next(t.apps)` without the `local app`?

cmsj: That would be my expectation, yes.

latenitefilms: Fixed in my branch.

cmsj:
Could you re-post your branch as a PR?

PR #1833, opened 2018-06-07
Cleanup cleanup-ipc

cmsj:


extensions/ipc/init.lua
```diff
@@ -460,21 +460,21 @@ module.__defaultHandler = function(self, msgID, msg)
...
+        local fn = function(ss)
+            local fn, err = load("return " .. ss)
+            if not fn then fn, err = load(ss) end
             if fn then return fn() else return err end
         end
 
         local results = table.pack(pcall(function() return fn(str) end))
 
-        local str = ""
+        str = ""
```
cmsj: Why did this go from a local to a global?

latenitefilms: `str` is previously defined as `local` [here](https://github.com/cmsj/hammerspoon/blob/768cedb4e8b738164a4dd3cdef44eef53f2b1dbf/extensions/ipc/init.lua#L435).

PR #1837, opened 2018-06-07
Updated hs.tangent

latenitefilms:
- Improved `hs.tangent` to make it faster and easier to use


latenitefilms:
@cmsj & @asmagill - This is good to go.

PR #1866, opened 2018-06-13
#1865 - Create Objects Above Fullscreen Windows

### LINKED ISSUE #1865: hs.chooser should have a higher window level by default, and make the window level configurable
- it probably does something as the fullscreen app looses focus, but there is no visible ui...
- contrary to Alfred which achieve this feat

That would be cool to have that!

latenitefilms:
- Updated in-line documentation in regard to hiding the Hammerspoon
Dock Icon if you want Hammerspoon objects to appear ABOVE fullscreen
windows.
- Closes #1865


cmsj:
I think there is something to be said for documenting this, but we should not close #1865 until its level is configurable in the same way `hs.drawing` and `hs.canvas` objects are.

PR #1878, opened 2018-06-23
Added hs.notify:alwaysShowAdditionalActions()

### LINKED ISSUE #1877: hs.notify with additionalActions glitches
I've been playing with the `additionalActions` options in `hs.notify`, which are cool. A couple of little things that would make them work a bit better, IMHO:

1. Currently, there is no visual hint that extra options exist. With Apple's notifications (for Reminder, for example), by default you just see the `action` button, like so:
banners_and_alerts

But when the mouse hovers over it a 'downward chevron' appears:
banners_and_alerts

This would be super helpful.

2. When I do click-n-hold on the action, the drop-down box of options appears as expected. If I just release the mouse click without selecting an option, the notification stays visible as expected, but if I then just do a click on the action button without holding, it takes two clicks for it to actually activate.

Anyway, just a couple of minor things.

latenitefilms:
- Added `hs.notify:alwaysShowAdditionalActions()`
- Closes #1877


latenitefilms:
@cmsj - I believe this is good to go!

PR #1879, opened 2018-06-23
Changed Hammerspoon's default notification style to an alert

latenitefilms:


cmsj:
Is there some background on why this is necessary/better?

latenitefilms:
@cmsj - By default Hammerspoon only displays banner notifications - not alert notifications. If you want to display buttons on your `hs.notify` notification, you need to manually change the Hammerspoon notification alert style to "alerts". This can be slightly confusing if you have your alert style set to "banners", and you're wondering why your `hs.notify` buttons aren't appearing. I think it would be better if Hammerspoon defaulted to alerts out-of-the-box, and the user can always turn back to banners if needed.

screen shot 2018-07-07 at 10 49 07 pm

cmsj:
I think we need some further work on this - by defaulting to alerts, even a simple `hs.notify.show()` is leaving a permanent alert notification. I'm fine with people being able to generate alerts, but I don't think that our API should be creating them by default.

latenitefilms:
@cmsj - Completely up to you. I'm happy for you to just revert this commit if you want. Alternatively we could modify `hs.notify.show()` so that it automatically triggers `hs.notify:withdraw()` after a certain amount of time? Personally, I think having buttons in macOS notifications is more important than not - but again, happy either way as for my purposes I can just enable this in CommandPost anyway.

cmsj:
I've made a fairly significant change - notifications will now autowithdraw after 5 seconds. It can be overridden with a new `:withdrawAfter()` method, or an attribute to `.new()`, but cannot be overridden for notifications created with `.show()` (I don't think the latter is a problem because the default behaviour therefore isn't changing, and there would be no way to add action items to such a notification anyway.

cmsj:
@asmagill may or may not object to my changes, but if mine come out, this PR will need to be reverted too.

latenitefilms:
Looks good to me! Thanks @cmsj !

asmagill:
There is a lot about the notification module that I would do differently now, but I don't want to rewrite it from scratch any time soon (it was one of my first real forays into Objective-C specific programming)... if this works and meet everyones's needs and expectations, then I have no objections.

PR #1880, opened 2018-06-23
Fixed bug in hs.chooser:refreshChoicesCallback()

### LINKED ISSUE #1876: hs.chooser:refreshChoicesCallback() doesn't take into account query string?
When you trigger `hs.chooser:refreshChoicesCallback()` it doesn't seem to take into account the current query string?

For example, if you have the following choices:

```
a1
a2
a3
b1
b2
b3
```

...and your query string is "a", you'll only see the choices starting with "a".

However, if you trigger `:refreshChoicesCallback()` via a contextual menu, it reloads all the choices, but doesn't take into account the existing query string value, so "b1", "b2" and "b3" are now visible, even though the query string still says "a".

Here's some test code:

```lua
test = hs.chooser.new(function(result)
	print(string.format("Result: %s", hs.inspect(result)))
end)

local function getChoices()
	return {
		{ ["text"] = "a1" },
		{ ["text"] = "a2" }, 
		{ ["text"] = "a3" }, 
		{ ["text"] = "b1" }, 
		{ ["text"] = "b2" }, 
		{ ["text"] = "b3" }, 				
	}
end

test:choices(getChoices)

test:rightClickCallback(function()
	test:refreshChoicesCallback()
end)

test:show()
```

I've tried capturing the `:query()` then re-applying it after `:refreshChoicesCallback()`, but that has no impact.

Any ideas @cmsj or @asmagill ?

latenitefilms:
- `hs.chooser:refreshChoicesCallback()` now takes into account the
current query string
- Closes #1876


latenitefilms:
@cmsj - This is good to go.

cmsj:
I've just reverted this via adbbbb69 because it broke scrolling through choices with cursor keys.

latenitefilms:
Ah bugger! Sorry about that!

Back to the drawing board on #1876 then I guess. Any ideas?

Issue #1804, opened 2018-05-26
Add hs.styledtext.validFont(name)


latenitefilms:
If you type `hs.styledtext.fontInfo("asdf")` into the Console, it will generate a LuaSkin warning, and return the default system font. This isn't overly helpful if you're trying to determine if a font exists or not. I feel like it would be more useful if it just returned `nil` if a font doesn't exist?

```lua
> hs.styledtext.fontInfo("asdf")
2018-05-27 01:09:01: 01:09:01 ** Warning:   LuaSkin: invalid font specified: asdf
ascender           12.568359375
boundingRect       h=17.6337890625 w=21.099609375 x=-3.9736328125 y=-3.63720703125 
capHeight          9.15966796875
descender          -2.7421875
displayName        System Font Regular
familyName         .AppleSystemUIFont
fixedPitch         false
fontName           .AppleSystemUIFont
italicAngle        0.0
leading            0.0
maximumAdvancement h=0.0 w=16.8974609375 
numberOfGlyphs     2596
pointSize          13.0
underlinePosition  -1.9677734375
underlineThickness 0.76171875
xHeight            6.8427734375
```

asmagill:
I believe the thinking was that when requesting a font, at least in some cases (especially when rendering something), it would be better to default to the system font rather than no font at all... and IIRC the LuaSkin conversion code which creates the NSFont object is where it determines if the font details provided are valid or not -- since I don't think we want to change that behavior where an NSFont object is required (e.g. when rendering text for drawing or canvas), it would probably be easier to make a list of the specific functions/methods which should return nil or an empty table and modify those rather than try to make the conversion code more intelligent.

latenitefilms:
No worries! Maybe something like `hs.styledtext.validFont(name)`?

PR #1882, opened 2018-06-25
*CL Addition* adds functions to get fonts by family name

asmagill:
Adds

~~~
Function: hs.styledtext.fontFamilies() -> table

Returns the names of all font families installed for the system.

Parameters:
 * None

Returns:
 * a table containing the names of every font family installed for the system.
~~~

and

~~~
Function: hs.styledtext.fontsForFamily(familyName) -> table

Returns an array containing fonts available for the specified font family or nil if no fonts for the specified family are present.

Parameters:
 * `familyName` - a string specifying the font family to return available fonts for. The strings should be one of the values returned by the `hs.styledtext.fontFamiles` function.

Returns:
 * a table containing an array of available fonts for the specified family. Each array entry will be a table, also as an array, in the following order:
   * a string specifying the font name which can be used in the `hs.drawing:setTextFont(fontname)` method.
   * a string specifying the basic style of the font (e.g. Bold, Italic, Roman, etc.)
   * a table containing one or more strings specifying common names for the weight of the font. ISO equivalent names are preceded with "ISO:". Possible values are:
            `{ "ultralight" }`
            `{ "thin", "ISO:ultralight" }`
            `{ "light", "extralight", "ISO:extralight" }`
            `{ "book", "ISO:light" }`
            `{ "regular", "plain", "display", "roman", "ISO:semilight" }`
            `{ "medium", "ISO:medium" }`
            `{ "demi", "demibold" }`
            `{ "semi", "semibold", "ISO:semibold" }`
            `{ "bold", "ISO:bold" }`
            `{ "extra", "extrabold", "ISO:extrabold" }`
            `{ "heavy", "heavyface" }`
            `{ "black", "super", "ISO:ultrabold" }`
            `{ "ultra", "ultrablack", "fat" }`
            `{ "extrablack", "obese", "nord" }`
   * a table specifying zero or more traits for the font as defined in the `hs.styledtext.fontTraits` table. A field with the key `_numeric` is also set which specified the numeric value corresponding to the traits for easy use with the `hs.styledtext.convertFont` function.
~~~

These are first steps for a planned font browser spoon I hope to work on later in July

### DIFF
--- a/extensions/styledtext/init.lua
+++ b/extensions/styledtext/init.lua
@@ -239,7 +239,7 @@ module.fontsForFamily = function(...)
             { "ultra", "ultrablack", "fat" },
             { "extrablack", "obese", "nord" }
         }
-        for i,v in ipairs(results) do
+        for _,v in ipairs(results) do
             v[3] = fontWeights[v[3]] or string.format("** unrecognized font weight: %d", v[3])
             local style, styleTable = v[4], { _numeric = v[4] }
             for k, v2 in pairs(module.fontTraits) do
--- a/extensions/styledtext/init.lua
+++ b/extensions/styledtext/init.lua
@@ -179,11 +179,84 @@ module.lineStyles    = ls.makeConstantsTable(module.lineStyles)
 module.lineAppliesTo = ls.makeConstantsTable(module.lineAppliesTo)
 
 module.fontNames = function(...)
-    return ls.makeConstantsTable(module._fontNames(...))
+    local results = module._fontNames(...)
+    return results and ls.makeConstantsTable(results) or nil
+end
+
+module.fontFamilies = function(...)
+    local results = module._fontFamilies(...)
+    return results and ls.makeConstantsTable(results) or nil
 end
 
 module.fontNamesWithTraits = function(...)
-    return ls.makeConstantsTable(module._fontNamesWithTraits(...))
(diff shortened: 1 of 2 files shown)


asmagill:
also cleans up some XCode build warnings, but apparently not lua ones... what are we doing for unused variables now? preceding them with an underscore or do I need to declare them as unused somewhere?

cmsj:
For the Lua one, just name the variable `_`

latenitefilms:
FYI - the reason I added hs.styledtext.validFont() was so that I could determine if a font is valid without generaring a LuaSkin error. Final Cut Pro has access to a whole bunch of fonts that don’t appear in Font Book (they’re contained internally within the Final Cut Pro application bundle as well as in the bundles of 3rd party effects), so I’m currently using a Terminal command to get a list of open files when Final Cut Pro is running, building a list of fonts from that, then comparing that to hs.styledtext’s results to remove duplicates. Without hs.styledtext.validFont() I get a heap of LuaSkin errors which fills my Console, which I want to avoid.

latenitefilms:
Original issue here: https://github.com/Hammerspoon/hammerspoon/issues/1804

asmagill:
Not sure how validFont got removed... I must have accidentally used an earlier copy of `internal.m`... I'll add it back when I fix the stickler issue.

PR #1885, opened 2018-06-26
*CL Addition* adds hs.styledtext.loadFont

asmagill:
@latenitefilms see if this helps -- by registering the font at the specified path, you can use it as if it were installed in the system -- the `hs.styledtext` and `hs.canvas` functions/methods will see it as a valid font after being registered.


latenitefilms:
You're an absolute legend. THANK YOU! Will test out ASAP.

latenitefilms:
This is perfect, THANK YOU!!

PR #1910, opened 2018-07-19
#1909 hs.inspect table bug

### LINKED ISSUE #1909: Bug in hs.inspect when dealing with tables with complex `__init` event functions.
In the case where a `table` has a metatable with an `__init` function that will always return a new table instance for a key value, `hs.inspect` will throw an error upon encountering the new table instance because it wasn't encountered in the first pass. Essentially, it just needs to handle the case where it has not logged it in `tableAppearances` before.

I've put a test case in the pull request which will accompany this issue shortly.

randomeizer:
Fixes #1909.


cmsj:
Good work for adding a test, but I think `hs.inspect` doesn't currently have any tests, so it would need an `HSinspect.m` in `Hammerspoon Tests` too I think?

randomeizer:
That would probably help :P I've pushed a fix.

cmsj:
Merged with some changes (note that the name of the Lua test function has to match the name in ObjC if `RUN_LUA_TEST()` is used.

latenitefilms:
@cmsj - I think maybe something went wrong when you merged this? It seems you merged the test (then commented it out), but didn't actually merge the fix?

cmsj:
@latenitefilms it looks to me like the actual fix is in `inspect/init.lua`: https://github.com/Hammerspoon/hammerspoon/blob/master/extensions/inspect/init.lua#L239

latenitefilms:
Opps, apologies, I must have been looking in a wrong branch.

PR #1919, opened 2018-08-11
Added links to @asmagill's templates

latenitefilms:


asmagill:
I have no specific plans to move that or the internal.m or Makefile templates I usually use, though I do modify them from time to time as XCode, needs, or conventions change...

Might be better for me to put together a couple of wiki pages and add it to the Hammerspoon wiki... I keep meaning to do a full tutorial, but there are just so many variances.  A couple of templates with comments and (at least my) common practices might be easier to maintain and in a place less likely to disappear or get reorganized in a clean up or shuffle.

cmsj:
I'm going to close this because I think we should have all our docs on the project's own sites, but it would be good to have these included.
We should really go back to having a template extension that we promise to always use for our own new extensions and that way it'll stay up to date.

PR #1928, opened 2018-09-01
Issue #1927 - hs.fs.dir() Documentation

### LINKED ISSUE #1927: Better Documentation for hs.fs.dir()
Based on the API documentation it's not clear that `hs.fs.dir()` will return a `nil` value for the iterator function if something doesn't work. For example:

```
> iterFn, dirObj = hs.fs.dir("/Users/Guest/Documents")
print(string.format("iterFn: %s", iterFn))
print(string.format("dirObj: %s", dirObj))
2018-08-28 12:11:11: iterFn: nil
2018-08-28 12:11:11: dirObj: cannot open /Users/Guest/Documents: Permission denied
```

This becomes problematic when you're using a `for` loop, for example:

```
for file in hs.fs.dir("/Users/Guest/Documents") do 
	print(file)
end
```

Maybe either the documentation or the function itself needs to be re-thought @cmsj or @asmagill ?

latenitefilms:
- Fixed `hs.fs.dir()` documentation
- Closes #1927


cmsj:
👍

PR #1932, opened 2018-09-05
hs.socket removed debugging

latenitefilms:
- Removed LuaSkin debug messages when writing and reading socket data,
as this was causing `hs.socket` to lock up when reading and writing a
lot of data quickly (i.e. when used in `hs.tangent`).


cmsj:
Good catch!

PR #1933, opened 2018-09-05
Tweaks to hs.tangent

latenitefilms:
- Removed `hs.tangent.interval`
- Added better debugging messages to socket callback

### DIFF
--- a/extensions/tangent/init.lua
+++ b/extensions/tangent/init.lua
@@ -1056,7 +1056,7 @@ end
 --- Notes:
 ---  * Full documentation for the Tangent API can be downloaded [here](http://www.tangentwave.co.uk/download/developer-support-pack/).
 ---  * The callback function should expect 1 argument and should not return anything.
----   * The 1 argument will be a table, which can contain one or many commands. Each command is it's own table with the following contents:
+---  * The 1 argument will be a table, which can contain one or many commands. Each command is it's own table with the following contents:
 ---    * id - the message ID of the incoming message
 ---    * metadata - A table of data for the Tangent command (see below).
 ---  * The metadata table will return the following, depending on the `id` for the callback:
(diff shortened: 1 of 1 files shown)


extensions/tangent/init.lua
```diff
@@ -1265,7 +1260,7 @@ end
 --- Function
 --- Updates the Hub with a menu value.
 --- The Hub then updates the displays of any panels which are currently
----   showing the menu.
+--    showing the menu.
```
cmsj: I don’t think this is intentional?

extensions/tangent/init.lua
```diff
@@ -1061,7 +1056,7 @@ end
 --- Notes:
 ---  * Full documentation for the Tangent API can be downloaded [here](http://www.tangentwave.co.uk/download/developer-support-pack/).
 ---  * The callback function should expect 1 argument and should not return anything.
----  * The 1 argument will be a table, which can contain one or many commands. Each command is it's own table with the following contents:
+--   * The 1 argument will be a table, which can contain one or many commands. Each command is it's own table with the following contents:
```
cmsj: I don’t think changing the `—` was intentional?

extensions/tangent/init.lua
```diff
@@ -1695,6 +1690,47 @@ local MESSAGE_SIZE = 1
 -- Message Body.
 local MESSAGE_BODY = 2
 
+-- socketCallback(data, tag) -> none
```
cmsj: `hs.tangent.socketCallback` here. Also I’m not sure what the reasoning is for exposing the callback - is it for people to use as their own Hs.socket callback?

cmsj:
Added some comments in the code

extensions/tangent/init.lua
latenitefilms: Ah, no - not sure what happened there!

extensions/tangent/init.lua
latenitefilms: Ah, no - not sure what happened there!

extensions/tangent/init.lua
latenitefilms: It's not exposed - it's just in a local function.

latenitefilms:
Fixed! Thanks @cmsj !

Issue #1944, opened 2018-09-25
LuaSkin Nullability Issues in Xcode 10.0


latenitefilms:
Any ideas @cmsj or @asmagill ?

<img width="2487" alt="screen shot 2018-09-25 at 3 58 17 pm" src="https://user-images.githubusercontent.com/22286696/45995558-dba20780-c0db-11e8-9d0a-9411e49a54d5.png">

latenitefilms:
Maybe related? https://oleb.net/2018/@keypath/

latenitefilms:
FYI: this is outside my skillset, so I'm going to leave @asmagill or @cmsj to address this issue.

asmagill:
@cmsj since LuaSkin was your creation, I defer to your preferences.

Basically the compiler is complaining because keys to dictionaries can't be nil, but the shortcut `@(cstring)` to create an NSString object is now marked nullable.

The fixes as I see it are as @latenitefilms suggest and to change the warnings, at least for this part of the project, or to create place holder vars, e.g. `NSString *placholder = @(cstring)`, verify that it's not nil, and then use placeholder instead of the shortcut.

I'm leaning towards the second approach because while some of the cases do already verify that the cstring isn't NULL (not all do, though), it is theoretically possible that an invalid UTF8 sequence beginning the cstring could also cause the NSString version to be nil, so the proper approach is to do the conversion and verification ourselves.

I'm fine doing the work, but as I said, LuaSkin was your creation and I don't know if/what plans you might have for it outside of Hammerspoon, so I'll defer to your preference.

latenitefilms:
Creating place holder vars sounds like a much better idea to me! Thanks for reviewing @asmagill !

latenitefilms:
@asmagill - Given we might not get a reply from @cmsj for a little while, what are the chances of you doing a pull request in the meantime as a temporary workaround?

asmagill:
Most of the cases were in the LuaSkin registration of helper functions, so to trigger them, you'd have to be building an external module... a couple, though, could be triggered by user data at run time.

In any case, see #1953 and let me know how it works for you. You'll still need #1943 though, figured we can keep issues separate for now unless they involve overlapping files.

latenitefilms:
Woohoo! That all works now - thanks heaps @asmagill !

[...]

Issue #1941, opened 2018-09-22
HS won't build with Xcode 10 on 10.13.6


chdiza:
Normally I just do `make release`.  After updating to Xcode 10, this bombs out quickly.  Below are the entire contents of `Release-build.log`.

```
note: Using new build system

** CLEAN SUCCEEDED **

note: Using new build system
note: Planning build
note: Constructing build description
Build system information
error: unexpected duplicate task: CompileAssetCatalog /Users/chdiza/Library/Developer/Xcode/DerivedData/Hammerspoon-adlzscacpnutjgcpsrpwvdewubqc/Build/Products/Release/Hammerspoon.app/Contents/Resources /opt/bld/hammerspoon/Hammerspoon/Images.xcassets (in target 'Hammerspoon')

Build system information
error: unexpected duplicate task: CompileAssetCatalog /Users/chdiza/Library/Developer/Xcode/DerivedData/Hammerspoon-adlzscacpnutjgcpsrpwvdewubqc/Build/Products/Release/Hammerspoon.app/Contents/Resources /opt/bld/hammerspoon/Hammerspoon/Images.xcassets (in target 'Hammerspoon')
```

Just for fun I tried plain old `make` but the result is the same.

latenitefilms:
I think what you're now seeing is the same as issue #1944. I believe you can get around this by diabling the `-Wnullable-to-nonnull-conversion` warning - but I haven't done a pull request, because I'm not sure this is the best way to tackle it - I'll await @asmagill or @cmsj's guidance on this one.

chdiza:
I just tried building (with `make release`) again on High Sierra with Xcode 10, hoping that 5787754c7a64707070f1 will have fixed things.  It still fails, but the error is different:
```
The following build commands failed:
	PhaseScriptExecution [CP]\ Embed\ Pods\ Frameworks /Users/chdiza/Library/Developer/Xcode/DerivedData/Hammerspoon-adlzscacpnutjgcpsrpwvdewubqc/Build/Intermediates.noindex/Hammerspoon.build/Release/Hammerspoon.build/Script-E849AEC367EA9502A25B9A7C.sh
(1 failure)
make: *** [build/Hammerspoon.app] Error 65
```

When I look in `Release-build.log`, I find this at the end:
```
[...]

PR #1953, opened 2018-10-02
wrap cstring->NSString shortcuts to prevent compile errors in Xcode 10

asmagill:
The use of `@(cstring)` as a shortcut to converting a c style string to an objective-c one is considered nullable in Xcode 10 which prevents its use where non-nullable is required (e.g. as NSDictionary keys). This corrects that.

To compile under Xcode 10, you will also need to apply #1941 to the master branch.

I have not really tested this -- my setup loads and I haven't noticed any new warnings or loss of functionality, but that's not really a formal testing methodology 😜

Other Xcode 10/Mojave tweaks may follow, but I wanted to get this out to address #1944 so others can continue testing and working.

### DIFF
--- a/LuaSkin/LuaSkin/Skin.m
+++ b/LuaSkin/LuaSkin/Skin.m
@@ -594,18 +594,24 @@ nextarg:
     if (level == 0) level = 3 ;
 
     if (className && helperFN) {
-        if (self.registeredNSHelperFunctions[@(className)]) {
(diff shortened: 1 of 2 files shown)


latenitefilms:
Legend, thanks @asmagill !!

cmsj:
Apologies, I fixed this without realising you'd already pushed a PR :/

PR #1981, opened 2018-11-13
Fix small issues/typos in docs

malob:
`hs.deezer.rw()` and `hs.spotify.rw()` are both missing the parens in their function definitions in their documentation.


malob:
Added another commit so that `desc` is populated in docs json file.

cmsj:
Thanks!

PR #1998, opened 2018-12-13
Fix focusWindowWest documentation typo

dnlggr:
Fixes a tiny typo in the `focusWindowWest` documentation.
It should compare west to "left" instead of "right". 🗺


cmsj:
Thanks!

Issue #1572, opened 2017-10-07
NSURL BookmarkData objects?


latenitefilms:
Howdy everyone!

There's some data in a plist file I'm trying to access - specifically:

```
	<key>FFRecentLibraries</key>
	<array>
		<data>
		Ym9va2wFAAAAAAQQMAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
		AAAAOAQAAAUAAAABAQAAVXNlcnMAAAAMAAAAAQEAAGNocmlzaG9ja2luZwcA
		AAABAQAARGVza3RvcAAZAAAAAQEAAExORi1KMTAyOTMgLSBCb25kcyAtIFJl
		ZWwAAAASAAAAAQEAADAyIC0gRmluYWwgQ3V0IFBybwAADgAAAAEBAAAwMSAt
		IExpYnJhcmllcwAAIwAAAAEBAABMTkYtSjEwMjkzIC0gQm9uZHMgLSBSZWVs
		LmZjcGJ1bmRsZQAcAAAAAQYAAAQAAAAUAAAAKAAAADgAAABcAAAAeAAAAJAA
		AAAIAAAABAMAANxpBgAAAAAACAAAAAQDAABvqwkAAAAAAAgAAAAEAwAAgKsJ
		AAAAAAAIAAAABAMAAOKCPgAAAAAACAAAAAQDAAAugz4AAAAAAAgAAAAEAwAA
		MoM+AAAAAAAIAAAABAMAALqEPgAAAAAAHAAAAAEGAADgAAAA8AAAAAABAAAQ
		AQAAIAEAADABAABAAQAACAAAAAAEAABBv06cUgAAABgAAAABAgAAAgEAAAAA
		AAAPAQAAAAAAAAABAAAAAAAACAAAAAQDAAAFAAAAAAAAAAQAAAADAwAA9QEA
		AAgAAAABCQAAZmlsZTovLy8MAAAAAQEAAE1hY2ludG9zaCBIRAgAAAAEAwAA
		AHDE2NEBAAAIAAAAAAQAAEG+t9/xAAAAJAAAAAEBAAA5NkI5NUM1NS1DQjk0
		LTM3RkMtOEY3RC1GMzFCREYyNDAwMTkYAAAAAQIAAIEAAAABAAAA7xMAAAEA
[...]

PR #2003, opened 2018-12-21
[READY] Issue #1572 - hs.fs.getPathFromBookmark

latenitefilms:
- Initial commit for `hs.fs.getPathFromBookmark`. Currently it’s not
working properly.

### DIFF
--- a/extensions/fs/internal.m
+++ b/extensions/fs/internal.m
@@ -1017,6 +1017,9 @@ static int fs_displayName(lua_State *L) {
 ///
 /// Returns:
 ///  * Bookmark data.
+///
+/// Notes:
+///  * This is primarily just intended for testing `hs.fs.getPathFromBookmark`.
 static int fs_getHomeDirectoryAsBookmark(lua_State *L) {
     LuaSkin *skin = [LuaSkin shared] ;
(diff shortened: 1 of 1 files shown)


latenitefilms:
Currently this doesn't work as expected, and I'm not really sure why. For example:

```lua
> hs.fs.getPathFromBookmark(hs.fs.getHomeDirectoryAsBookmark())
2018-12-21 14:51:38: ********
2018-12-21 14:51:38: 14:51:38 ERROR:   LuaSkin: Error resolving URL from bookmark: Error Domain=NSCocoaErrorDomain Code=259 "The file couldn’t be opened because it isn’t in the correct format."
2018-12-21 14:51:38: ********
nil
```

However, if I move the code from within `hs.fs.getHomeDirectoryAsBookmark` inside `hs.fs.getPathFromBookmark`, it works as expected, so I can only assume something is going funky when being translated by LuaSkin.

Any ideas @asmagill or @cmsj ?

latenitefilms:
Never mind, I had another look at how `hs.socket` works, and made some adjustments that seem to now work - yay!

Feel free to tweak and change as needed, and if happy, merge away!

PR #2004, opened 2018-12-22
Add ForceToGray and InvertedPolarity to hs.screen

zuckschwerdt:
Adds setForceToGray, getForceToGray, getInvertedPolarity, setInvertedPolarity to hs.screen

These functions seem to fit in hs.screen with minimal code. They operate on CoreGraphics and thus are Functions, not Methods on a screen.

The function names follow the API call names, but could perhaps simply be called `Grayscale` and `Inverted`?

Uses perhaps to to signify chorded hotkeys or certain modes or states.

### DIFF
--- a/extensions/screen/internal.m
+++ b/extensions/screen/internal.m
@@ -645,6 +645,88 @@ static int screen_setBrightness(lua_State *L) {
     return 1;
 }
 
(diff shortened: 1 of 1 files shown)


cmsj:
Thanks!

PR #2014, opened 2019-01-05
[READY] Tweak to hs.chooser

latenitefilms:
- Explicitly require `hs.window` so we don’t get the “Loading
Extension” message in the Console.


cmsj:
Good catch, thanks!

PR #2027, opened 2019-01-26
[READY] Add secure text prompt

skrypka:
Add `secureField` option to `hs.dialog.textPrompt` if true PasswordField field is going to be using instead of TextField.

It looks like this:
screen shot 2019-01-26 at 8 21 47 pm
 
/reviewme @cmsj


cmsj:
Thanks!

PR #2035, opened 2019-02-06
configurable hs.hints icon alpha

szymonkaliski:
```lua
hs.hints.iconAlpha = 0.0 -- invisible hints icons (text still visible)
hs.hints.iconAlpha = 1.0 -- fully visible hints
```


cmsj:
Thanks!

PR #2050, opened 2019-03-16
[READY] Issue #2049 - hs.webview freezing

### LINKED ISSUE #2049: hs.webview freezing on while loop?
Howdy!

I have a `hs.webview` that has a `hs.webview.toolbar`. The toolbar buttons change the contents of the webview, which all works great.

However, when I click the toolbar buttons really quickly, it can cause Hammerspoon/CommandPost to freeze.

If I'm reading the crash logs correctly, the problem seems to lie here:

[`while (theView.loading) {}`](https://github.com/Hammerspoon/hammerspoon/blob/master/extensions/webview/internal.m#L1396)

It seems in my case, `theView.loading` never goes `false`.

I wonder if a timeout value should be added to this `while` loop?

Thoughts @asmagill & @cmsj ?

Here's the crash log: https://gist.github.com/latenitefilms/021eb13351f83c726c63420da33e0669

latenitefilms:
- Fixed a bug which could cause a `hs.webview` to freeze
- Closes #2049

### DIFF
--- a/extensions/webview/internal.m
+++ b/extensions/webview/internal.m
@@ -1176,8 +1176,8 @@ static int webview_reload(lua_State *L) {
     HSWebViewView   *theView = theWindow.contentView ;
 
     if (theView.loading) [theView stopLoading] ;
+    while (theView.loading) {}
     dispatch_async(dispatch_get_main_queue(), ^{
-        while (theView.loading) {}
         WKNavigation *navID ;
         if (lua_type(L, 2) == LUA_TBOOLEAN && lua_toboolean(L, 2))
             navID = [theView reload] ;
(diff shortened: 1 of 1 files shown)


latenitefilms:
@cmsj & @asmagill - This seems to fix #2049.

asmagill:
I think I see why this helps... if the toolbar button action is to change the URL, then rapidly pressing it (or not so rapidly if the pages take a while to load) could in theory end up with multiple blocks queued all of which want to load another page as soon as the flag changes... instead, you hold off on adding to the queue *until* the flag is cleared so hopefully there is only ever one request pending.

You should do the same for [reload](https://github.com/Hammerspoon/hammerspoon/blob/f56d0722f725e680fee152058c0ec513b76d719f/extensions/webview/internal.m#L1180) and [html](https://github.com/Hammerspoon/hammerspoon/blob/f56d0722f725e680fee152058c0ec513b76d719f/extensions/webview/internal.m#L1396)

latenitefilms:
Awesome, thanks @asmagill ! Branch updated, feel free to merge if happy.

PR #2060, opened 2019-03-28
[READY] Issue #2055 - Fixes hs.socket crash

### LINKED ISSUE #2055: Crash - hs.socket - LuaSkin::pushLuaRef was passed a NOREF/REFNIL ref
See: http://crashes.to/s/a5440b5090d

The problem seems to be somewhere in here:
https://github.com/Hammerspoon/hammerspoon/blob/master/extensions/socket/internal.m#L26

latenitefilms:
- Added in some logic to make sure the `hs.socket` callbacks actually
exist before triggering to avoid a crash.
- Closes #2055


latenitefilms:
@asmagill - Feel free to merge away if happy. Thank you!

Issue #1681, opened 2018-02-12
Feature Request - Tabs for Chooser


latenitefilms:
It would be amazing if you could add a customisable toolbar to `hs.chooser`. 

For example:

<img width="1387" alt="console_revamp_idea" src="https://user-images.githubusercontent.com/22286696/36083446-62fd19cc-1006-11e8-90e4-cb47a92530a6.png">

@asmagill - Maybe this is something that could be tackled with `hs.guitk` at some stage in the future?

latenitefilms:
Also see: https://colorsnapper.com

latenitefilms:
It would actually be really cool if we could "attach" a `hs.webview.toolbar` to a `hs.chooser`.

I've tried playing around but my Xcode skills aren't really up to it.

Any hints or tricks @asmagill or @cmsj ?

latenitefilms:
Given you're both online at the moment @asmagill & @cmsj - any ideas on how best I can try and get this working? Basically I want to try and attach a `NSToolbar` to a `hs.chooser` window.

Related: https://dikant.de/2007/09/05/nstoolbar-basics/

asmagill:
The challenge will probably be that @cmsj wrote the chooser using the layout tools within XCode while I wrote the toolbar module completely in Objective-C eschewing any of the helpers or placeholder values XCode offers... I've always meant to learn how to use the layout tools in XCode, but haven't had the need or job requirements to do so...

That said, assuming that the window/panel object that comprises the chooser has a place for the toolbar, unless I'm missing something, it shouldn't be much harder than adding it to the console window was (some of the window styles available don't show or have space for some of the add-on objects (toolbars, rulers, etc.) even when they're attached).

I'll take a closer look and see if I think it's something I can do relatively soon or if I need to plan it for later (or give you pointers on where to look if you want to tackle it yourself in the meantime)

PR #1974, opened 2018-10-30
[READY] Issue #1876 - Updated refreshChoicesCallback() to take into account current query string

latenitefilms:
- Updated `hs.chooser:refreshChoicesCallback()` to take into account
the current query string
- Was originally submitted in #1903 however @cmsj needed to revert.
- Closes #1876

PR #2062, opened 2019-03-29
[READY] add toolbar support to hs.chooser

asmagill:
Probably address #1681

This is not ready for prime time, but for those who want to give it a whirl, here is the preliminary support... still needs testing, but this sample should get you started:

~~~lua
local chooser = require("hs.chooser")
local toolbar = require("hs.webview.toolbar")
local canvas  = require("hs.canvas")
local inspect = require("hs.inspect")
local stext   = require("hs.styledtext")

local module = {}

local list1, list2, list3 = {}, {}, {}

for i = 1, 10, 1 do
    table.insert(list1, { text = string.char(96 + i) } )
    table.insert(list2, { text = tostring(i) })
    table.insert(list3, { text = string.char(64 + i) })
end

local changeChooserEntries = function(list, bar, parent, item)
    parent:choices(list)
    parent:query("")
    bar:selectedItem(item)
end

local toolbarItems = {
    {
        id         = "choice1",
        selectable = true,
        label      = "Lower Case",
        image      = canvas.new{ h = 50, w = 50 }:appendElements{
                            {
                                type = "rectangle",
                                strokeColor = { white = 1 },
                                fillColor   = { red   = .5 },
                            }, {
-- vertical centering is hard without putting the text into a local variable, finding its bounding box
-- with hs.canvas:minimumTextSize, then calculating the proper frame based on canvas dimensions... this
-- is "close enough" for this example...
                                frame = { h = 50, w = 50, x = 0, y = -6 },
                                text = stext.new("a", {
                                    font = { name = ".AppleSystemUIFont", size = 50 },
                                    paragraphStyle = { alignment = "center" }
                                }),
                                type = "text",
                            }
                    }:imageFromCanvas(),
        fn         = function(...) changeChooserEntries(list1, ...) end,
    },
    {
        id         = "choice2",
        selectable = true,
        label      = "Numeric",
        image      = canvas.new{ h = 50, w = 50 }:appendElements{
                            {
                                type = "rectangle",
                                strokeColor = { white = 1 },
                                fillColor   = { green = .5 },
                            }, {
                                frame = { h = 50, w = 50, x = 0, y = -6 },
                                text = stext.new("1", {
                                    font = { name = ".AppleSystemUIFont", size = 50 },
                                    paragraphStyle = { alignment = "center" }
                                }),
                                type = "text",
                            }
                    }:imageFromCanvas(),
        fn         = function(...) changeChooserEntries(list2, ...) end,
    },
    {
        id         = "choice3",
        selectable = true,
        label      = "Upper Case",
        image      = canvas.new{ h = 50, w = 50 }:appendElements{
                            {
                                type = "rectangle",
                                strokeColor = { white = 1 },
                                fillColor   = { blue  = .5 },
                            }, {
                                frame = { h = 50, w = 50, x = 0, y = -6 },
                                text = stext.new("A", {
                                    font = { name = ".AppleSystemUIFont", size = 50 },
                                    paragraphStyle = { alignment = "center" }
                                }),
                                type = "text",
                            }
                    }:imageFromCanvas(),
        fn         = function(...) changeChooserEntries(list3, ...) end,
    },
}

local _toolbar = toolbar.new("chooserToolbarTest")
                        :addItems(toolbarItems)
                        :canCustomize(true)
                        :setCallback(function(...)
                                          print("+++ Oops! You better assign me something to do!")
                                     end)

local _chooser = chooser.new(function(...)
    print("Chooser results = " .. inspect(table.pack(...), { newline = " ", indent = "" }))
end):attachedToolbar(_toolbar)

changeChooserEntries(list1, _toolbar, _chooser, "choice1")

module._toolbar = _toolbar
module._chooser = _chooser

module.show = function() _chooser:show() end

return module
~~~

Save that somewhere, then do:

~~~
> test = dofile("_scratch/chooserToolbar.lua")

> test.show()
~~~

If you want to make changes for testing and avoid reloading Hammersoon, make sure to do:

~~~
> test._chooser:delete()

> test._toolbar:delete()
~~~

before re-entering the `dofile` line.

### DIFF
--- a/extensions/_coresetup/init.lua
+++ b/extensions/_coresetup/init.lua
@@ -583,7 +583,7 @@ hs.fileDroppedToDockIconCallback = nil
 ---
 --- Notes:
 ---  * This is an `hs.toolbar` object that is shown by default in the Hammerspoon Console
----  * You can remove this toolbar by adding `hs.console.toolbar(nil)` to your config, or you can replace it with your own `hs.toolbar` object
+---  * You can remove this toolbar by adding `hs.console.toolbar(nil)` to your config, or you can replace it with your own `hs.webview.toolbar` object
   local toolbar = require("hs.webview.toolbar")
   local console = require("hs.console")
   local image = require("hs.image")
--- a/extensions/chooser/init.lua
+++ b/extensions/chooser/init.lua
@@ -10,6 +10,22 @@ require("hs.drawing.color")
 local chooser = require("hs.chooser.internal")
 local window = require("hs.window")
 
+--- hs.chooser:attachedToolbar([toolbar | nil]) -> hs.chooser object | currentValue
+--- Method
+--- Get or attach/detach a toolbar to/from the chooser.
+---
+--- Parameters:
+---  * `toolbar` - if an `hs.webview.toolbar` object is specified, it will be attached to the chooser.  If an explicit nil is specified, the current toolbar will be removed from the chooser.
+---
+--- Returns:
+---  * if a toolbarObject or explicit nil is specified, returns the hs.chooser object; otherwise returns the current toolbarObject or nil, if no toolbar is attached to the chooser.
+---
+--- Notes:
+---  * this method is a convenience wrapper for the `hs.webview.toolbar.attachToolbar` function.
+---
+---  * If the toolbarObject is currently attached to another window when this method is called, it will be detached from the original window and attached to the chooser.  If you wish to attach the same toolbar to multiple chooser objects, see `hs.webview.toolbar:copy`.
+hs.getObjectMetatable("hs.chooser").attachedToolbar = require"hs.webview.toolbar".attachToolbar
+
 --- hs.chooser.globalCallback
 --- Variable
 --- A global callback function used for various hs.chooser events
--- a/extensions/chooser/internal.m
+++ b/extensions/chooser/internal.m
@@ -780,6 +780,13 @@ static int userdata_gc(lua_State* L) {
             chooser.completionCallbackRef = [skin luaUnref:refTable ref:chooser.completionCallbackRef];
             chooser.rightClickCallbackRef = [skin luaUnref:refTable ref:chooser.rightClickCallbackRef];
             chooser.isObservingThemeChanges = NO;  // Stop observing for interface theme changes.
+
+            NSWindow *theWindow = chooser.window ;
+            if (theWindow.toolbar) {
+                theWindow.toolbar.visible = NO ;
+                theWindow.toolbar = nil ;
+            }
+
             chooser = nil;
         }
     }
--- a/extensions/console/internal.m
+++ b/extensions/console/internal.m
@@ -545,7 +545,9 @@ static int console_behavior(lua_State *L) {
 ///  * a string of "visible" or "hidden" specifying the current (possibly changed) state of the window title's visibility.
 ///
 /// Notes:
-///  * When a toolbar is attached to the Hammerspoon console (see the `hs.webview.toolbar` module documentation), this function can be used to specify whether the Toolbar appears underneath the console window's title ("visible") or in the window's title bar itself, as seen in applications like Safari ("hidden").
+///  * When a toolbar is attached to the Hammerspoon console (see the `hs.webview.toolbar` module documentation), this function can be used to specify whether the Toolbar appears underneath the console window's title ("visible") or in the window's title bar itself, as seen in applications like Safari ("hidden"). When the title is hidden, the toolbar will only display the buttons as small icons without labels, regardless of the toolbar settings.
+///
+///  * If a toolbar is attached to the console, you can achieve the same effect as this function with `hs.console.toolbar():inTitleBar(boolean)`
 static int console_titleVisibility(lua_State *L) {
     LuaSkin *skin = [LuaSkin shared] ;
     [skin checkArgs:LS_TSTRING | LS_TOPTIONAL, LS_TBREAK] ;
--- a/extensions/console/internal.m
+++ b/extensions/console/internal.m
@@ -545,7 +545,7 @@ static int console_behavior(lua_State *L) {
 ///  * a string of "visible" or "hidden" specifying the current (possibly changed) state of the window title's visibility.
 ///
 /// Notes:
-///  * When a toolbar is attached to the Hammerspoon console (see the `hs.webview.toolbar` module documentation), this function can be used to specify whether the Toolbar appears underneath the console window's title ("visible") or in the window's title bar itself, as seen in applications like Safari ("hidden"). When the title is hidden, the toolbar will only display the buttons as small icons without labels, regardless of the toolbar settings.
+///  * When a toolbar is attached to the Hammerspoon console (see the `hs.webview.toolbar` module documentation), this function can be used to specify whether the Toolbar appears underneath the console window's title ("visible") or in the window's title bar itself, as seen in applications like Safari ("hidden"). When the title is hidden, the toolbar will only display the toolbar items as icons without labels, and ignores changes made with `hs.webview.toolbar:displayMode`.
 ///
 ///  * If a toolbar is attached to the console, you can achieve the same effect as this function with `hs.console.toolbar():inTitleBar(boolean)`
 static int console_titleVisibility(lua_State *L) {
--- a/extensions/webview/internal.m
+++ b/extensions/webview/internal.m
@@ -2940,6 +2940,12 @@ static int userdata_gc(lua_State* L) {
         theWindow.windowCallback   = [skin luaUnref:refTable ref:theWindow.windowCallback] ;
         theView.navigationCallback = [skin luaUnref:refTable ref:theView.navigationCallback] ;
         theView.policyCallback     = [skin luaUnref:refTable ref:theView.policyCallback] ;
+
+        if (theWindow.toolbar) {
+            theWindow.toolbar.visible = NO ;
+            theWindow.toolbar = nil ;
+        }
+
         [theWindow close] ; // ensure a proper close when gc invoked during reload; nop if hs.webview:delete() is used
 
         // emancipate us from our parent
--- a/extensions/webview/internal.m
+++ b/extensions/webview/internal.m
@@ -2123,7 +2123,7 @@ static int webview_windowTitle(lua_State *L) {
 /// Notes:
 ///  * See also [hs.webview:windowStyle](#windowStyle) and [hs.webview.windowMasks](#windowMasks).
 ///
-///  * When a toolbar is attached to the webview, this function can be used to specify whether the Toolbar appears underneath the webview window's title ("visible") or in the window's title bar itself, as seen in applications like Safari ("hidden"). When the title is hidden, the toolbar will only display the buttons as small icons without labels, regardless of the toolbar settings.
+///  * When a toolbar is attached to the webview, this function can be used to specify whether the Toolbar appears underneath the webview window's title ("visible") or in the window's title bar itself, as seen in applications like Safari ("hidden"). When the title is hidden, the toolbar will only display the toolbar items as icons without labels, and ignores changes made with `hs.webview.toolbar:displayMode`.
 ///
 ///  * If a toolbar is attached to the webview, you can achieve the same effect as this method with `hs.webview:attachedToolbar():inTitleBar(boolean)`
 static int webview_titleVisibility(lua_State *L) {
--- a/extensions/webview/internal.m
+++ b/extensions/webview/internal.m
@@ -2110,6 +2110,53 @@ static int webview_windowTitle(lua_State *L) {
     return 1 ;
 }
 
+/// hs.webview:titleVisibility([state]) -> webviewObject | string
+/// Function
+/// Get or set whether or not the title text appears in the webview window.
+///
+/// Parameters:
+///  * `state` - an optional string containing the text "visible" or "hidden", specifying whether or not the webview's title text appears when webview's window style includes "titled".
+///
+/// Returns:
+///  * if a value is provided, returns the webview object; otherwise returns the current value.
+///
+/// Notes:
+///  * See also [hs.webview:windowStyle](#windowStyle) and [hs.webview.windowMasks](#windowMasks).
+///
+///  * When a toolbar is attached to the webview, this function can be used to specify whether the Toolbar appears underneath the webview window's title ("visible") or in the window's title bar itself, as seen in applications like Safari ("hidden"). When the title is hidden, the toolbar will only display the buttons as small icons without labels, regardless of the toolbar settings.
+///
+///  * If a toolbar is attached to the webview, you can achieve the same effect as this method with `hs.webview:attachedToolbar():inTitleBar(boolean)`
+static int webview_titleVisibility(lua_State *L) {
+    LuaSkin *skin = [LuaSkin shared] ;
+    [skin checkArgs:LS_TUSERDATA, USERDATA_TAG, LS_TSTRING | LS_TOPTIONAL, LS_TBREAK] ;
+    HSWebViewWindow *theWindow = get_objectFromUserdata(__bridge HSWebViewWindow, L, 1, USERDATA_TAG) ;
+
+    NSDictionary *mapping = @{
+        @"visible" : @(NSWindowTitleVisible),
+        @"hidden"  : @(NSWindowTitleHidden),
+    } ;
+
+    if (lua_gettop(L) == 1) {
+        NSNumber *titleVisibility = @(theWindow.titleVisibility) ;
+        NSString *value = [[mapping allKeysForObject:titleVisibility] firstObject] ;
+        if (value) {
+            [skin pushNSObject:value] ;
+        } else {
+            [skin logWarn:[NSString stringWithFormat:@"unrecognized titleVisibility %@ -- notify developers", titleVisibility]] ;
+            lua_pushnil(L) ;
+        }
+    } else {
+        NSNumber *value = mapping[[skin toNSObjectAtIndex:2]] ;
+        if (value) {
+            theWindow.titleVisibility = [value intValue] ;
+            lua_pushvalue(L, 1) ;
+        } else {
+            return luaL_argerror(L, 2, [[NSString stringWithFormat:@"must be one of '%@'", [[mapping allKeys] componentsJoinedByString:@"', '"]] UTF8String]) ;
+        }
+    }
+    return 1 ;
+}
+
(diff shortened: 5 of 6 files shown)


asmagill:
@latenitefilms not sure if you've seen this yet... give it a try when you get a chance and see if it gives you what you're looking for.  Since we're tweaking the chooser window and directly, I'd also be interested in any reports positive or negative about the chooser window's behavior or appearance when you change things like the toolbar item icon size, label on/off, etc.

latenitefilms:
I’m away from my laptop currently but I’ll definitely test this out ASAP! Thanks heaps, as always, for all your help and support! HUGELY appreciated!

latenitefilms:
@asmagill - This is AWESOME, thanks heaps!!

The biggest issue at the moment is that the toolbar hides the `hs.chooser` search box.

Basically, the toolbar seems to be sitting over the top of the original chooser interface, rather than sitting on the top of it (i.e. making the chooser height bigger).

For example:

Screen Shot 2019-03-31 at 9 06 38 pm

This is using the below code:

```lua
local chooser = require("hs.chooser")
local console = require("hs.console")
local inspect = require("hs.inspect")
local styledtext = require("hs.styledtext")
local toolbar = require("hs.webview.toolbar")
local image = require("hs.image")

local completionFn = function(result)
	print(string.format("Result: %s", result and inspect(result)))
end

local fancyText = styledtext.new("First Choice", {
        color = {hex = "#FF0000", alpha = 1},
        font = { name = "Futura", size = 18 },
        
    })

local fancySubText = styledtext.new("This is the subtext of the first choice", {
        color = {hex = "#FF0000", alpha = 1},
        font = { name = "Futura", size = 12 },
    })

console.printStyledtext(fancyText)
console.printStyledtext(fancySubText)

local function getChoices()
    return {
        {
            ["text"] = fancyText,
            ["subText"] = fancySubText,
            ["uuid"] = "0001"
        },
        { 
            ["text"] = fancyText,
            ["uuid"] = "Bbbb"
        },
        { 
            ["text"] = "Third Possibility",
            ["subText"] = "What a lot of choosing there is going on here!",
            ["uuid"] = "III3"
        },
        {
            ["text"] = fancyText,
            ["subText"] = fancySubText,
            ["uuid"] = "0001"
        },
        { 
            ["text"] = fancyText,
            ["uuid"] = "Bbbb"
        },
        { 
            ["text"] = "Third Possibility",
            ["subText"] = "What a lot of choosing there is going on here!",
            ["uuid"] = "III3"
        },
        {
            ["text"] = fancyText,
            ["subText"] = fancySubText,
            ["uuid"] = "0001"
        },
        { 
            ["text"] = fancyText,
            ["uuid"] = "Bbbb"
        },
        { 
            ["text"] = "Third Possibility",
            ["subText"] = "What a lot of choosing there is going on here!",
            ["uuid"] = "III3"
        },
        {
            ["text"] = fancyText,
            ["subText"] = fancySubText,
            ["uuid"] = "0001"
        },
        { 
            ["text"] = fancyText,
            ["uuid"] = "Bbbb"
        },
        { 
            ["text"] = "Third Possibility",
            ["subText"] = "What a lot of choosing there is going on here!",
            ["uuid"] = "III3"
        },
        {
            ["text"] = fancyText,
            ["subText"] = fancySubText,
            ["uuid"] = "0001"
        },
        { 
            ["text"] = fancyText,
            ["uuid"] = "Bbbb"
        },
        { 
            ["text"] = "Third Possibility",
            ["subText"] = "What a lot of choosing there is going on here!",
            ["uuid"] = "III3"
        },				
    }
end


local toolbarItems = {
    {
        id         = "choice1",
        label      = "Button 1",
        fn         = function() print("hello") end,
        image      = image.imageFromName(image.systemImageNames["Computer"]),
    },
    {
        id         = "choice2",
        label      = "Button 2",
        fn         = function() print("hello") end,
        image      = image.imageFromName(image.systemImageNames["Caution"]),
    },
    {
        id         = "choice3",
        label      = "Button 3",
        fn         = function() print("hello") end,
        image      = image.imageFromName(image.systemImageNames["Advanced"]),
    },
}

local chooserToolbar = toolbar.new("chooserToolbarTest")
                        :addItems(toolbarItems)
                        :canCustomize(true)
                        :setCallback(function(...)
                                          print("+++ Oops! You better assign me something to do!")
                                     end)

theChooser = chooser.new(completionFn)
	:choices(getChoices)
	:rightClickCallback(function() 
	    theChooser:refreshChoicesCallback() 
	    print("refresh!")
	end)
	:attachedToolbar(chooserToolbar)
	:bgDark(true)	
	:show()	
```

Basically, what I THINK should be happening is that if the toolbar is added, then the height of the `hs.chooser` `NSWindow` should be increased by the height of the toolbar.

However, it seems getting the height of a toolbar is not that simple. This MIGHT be relevant?

http://mirror.informatimago.com/next/developer.apple.com/documentation/Cocoa/Conceptual/Toolbars/Tasks/ToolbarHeight.html

latenitefilms:
Playing around with this a bit more... This might require some Interface Builder magic.

Any ideas @cmsj how we can "push" the main interface content underneath the `NSToolbar`?

asmagill:
Did my example code cover the search box? I distinctly remember seeing it and even typing into it a few times during my tests... Can't test your code right now myself to see if I get the same behavior you show here, but if we can determine that it's related to some of the differences, it might point to a solution that doesn't require the interface builder... then again, it might, so...

Either way, let me know if you get a chance, and I'll try your example later today as well.

asmagill:
Had a thought... I was poking around while developing this a bit with one of my modules that works as a lua-objc bridge (very unstable which is why I've never really pushed it, but useful sometimes to see whats going on internally and sometimes tweak them without recompiling)... one of the things I tried was to set the resizable flag on the window as well while working on it...

Try changing lines 906 and 912 of `toolbar_internal.m` as follows:

~~~objc
    if (isChooser) theWindow.styleMask &= ~(NSWindowStyleMaskTitled | NSWindowStyleMaskResizable)  ; // chooser isn't normally titled or resizable

...

    if (isChooser) theWindow.styleMask |= (NSWindowStyleMaskTitled | NSWindowStyleMaskResizable) ; // only titled windows can have toolbars
~~~

Let me know if that makes a difference.

asmagill:
OK see if this pull works any better...

Known issues/differences/limitations:
1. adding a toolbar to a chooser changes it visually in the following ways:
  A. adds an empty space at the top where the window title would usually go
  B. reduces choices list size by a couple of items (to make room for the toolbar)
  C. corners of chooser become rounded as opposed to square
2. if you hide the toolbar (`hs.webview.toolbar:visible(false)`) while the chooser is visible, the window shrinks by the size of the toolbar, but the title bar remains. When the chooser is shown again, it returns to its normal size, but if the toolbar is editable, you can bring it back by right clicking on the titlebar and select "Customize Toolbar", then click on Done -- this will increase the chooser window size to fit the toolbar (until shown again later, when it again returns to its normal size).
3. if you remove the toolbar entirely (`hs.chooser:attachedToolbar(nil)`) while the chooser is visible, the entire chooser window shrinks by the size of the toolbar and title and goes back to squared corners; when the chooser is shown again (even without the toolbar) it returns to its normal size.

From what I can determine, all of the visual changes of (1) are caused by making the windows mask titled, but I have been unable to get the toolbar to attach without making the window titled (though I haven't specifically read anywhere that its impossible, I just haven't found a way around it) -- @cmsj do you have any thoughts or sources on this?

And as for (2) and (3) in general it shouldn't be a problem unless you add toolbar buttons to do so like I did in my tests -- see https://github.com/asmagill/hammerspoon-config-take2/blob/master/_scratch/chooserToolbar.lua

latenitefilms:
Woohoo! This works great now @asmagill  - thanks heaps!!

I don't think there's many/any situations where you'd want to hide a toolbar whilst the chooser is already visible, so I don't think the known issues/differences/limitations are a problem.

Somewhat related, in Google I can across this website, which is a pretty handy visual reference:

https://github.com/lukakerr/NSWindowStyles

I tried to add this:

```objc
            if (isChooser) {
                theWindow.styleMask = NSWindowStyleMaskTitled ; // only titled windows can have toolbars
                theWindow.titleVisibility = NSWindowTitleHidden;
            }
```

But it suffers this problem (i.e. you loose the text description under the toolbar items):

https://stackoverflow.com/questions/42915982/macos-any-way-to-hide-window-title-and-not-toolbar-item-labels
https://stackoverflow.com/questions/35087150/unified-nstoolbar-titlebar-with-icons-and-labels

I also tried ticking "Hide Title" in the Window section of the Interface Builder, but it has the same outcome.

I think if we want to allow users to have the option of having text labels on their toolbar, then having a title bar seems to be the only option.

However, I personally don't have an issue with not allowing text labels on the chooser toolbar, as it looks pretty cool:

Screen Shot 2019-04-01 at 8 53 48 am

You can also use the below if we don't want the chooser to be draggable with a toolbar:

```objc
[theWindow setMovable: NO];
```

I'd love to see this merged into Hammerspoon sooner rather than later, so happy to trust whatever you decide is the best route!

Whilst you're working on `hs.chooser` if you have any thoughts on #1974, let me know!

Thanks again for all your help!! HUGELY appreciated!

asmagill:
That is actually a behavior I have been curious about for a while now -- making the toolbar more like the Safari toolbar where the buttons are in the title space... I definitely want to add that as an option (with the documentation note that doing so suppresses text labels)

I'll see what I can do tonight/tomorrow to add this. Thanks!

cmsj:
I can't really offer any useful guidance here, I am *really* bad at using Interface Builder and it's frankly a miracle that hs.chooser's UI works at all, but anything else I've tried to get it to do (like hiding the results pane until it's needed) has never worked properly :/

asmagill:
Hmm... apparently I found out about `titleVisibility` at some point before because it's already possible with `hs.console` for the main console... guess I need to add it to `webview` and `chooser` now.

latenitefilms:
@asmagill - Oh yeah, I forgot that `hs.console` already has this ability!

Whilst I've got you... is it possible to make the Console not the "key window" (which may or may not be the correct term)? 

For example:

- I have a `hs.webview` and console open. The webview is frontmost.
- I CMD+TAB to Safari, which "covers up" both the webview and console.
- I CMD+TAB back to Hammerspoon. The console is activated - even though it wasn't the last used window.

latenitefilms:
Maybe related: https://stackoverflow.com/questions/11936915/cocoa-prevent-window-activation

latenitefilms:
Actually... ignore my last two messages. I need to do some further testing.

asmagill:
Ok, added `hs.webview:titleVisibility`, `hs.webview.toolbar:inTitleBar`, removed the ability to move the chooser window when it's in titled (toolbar friendly) mode, and cleaned up some documentation mistakes.

Modified my example to do basic test, so check the link I give above if you want to see it. Haven't tested it much beyond that yet, so... do your worst and I'll try not to cry too much 😜

asmagill:
Re the webview not being the frontmost window when you come back to Hammerspoon even when it was when you left it, I've noticed that too with the built-in documentation and have been giving it some thought, but don't have a good solution yet... `canBecomeKeyWindow` won't do it -- that affects whether or not the window can accept user input and we already have a way to set that for webviews.

I don't have time to play with it right now, but my first thought would be to try adding something like this to the webview window object and see what happens/breaks:

~~~objc
-(BOOL)canBecomeMainWindow { return YES; }
~~~

If that works, then we'll need to come up with a good method name/context so that it can be programmatically controlled like `canBecomeKeyWindow` is in webview's `internal.m`.

latenitefilms:
I wonder if draggable should be optional when a toolbar is added?

latenitefilms:
Looking great @asmagill - thanks heaps!!

Screen Shot 2019-04-03 at 8 37 34 pm

latenitefilms:
All seems to be working great! Thanks again @asmagill !

Merge away if you're happy!

Screen Shot 2019-04-03 at 10 40 10 pm

asmagill:
@cmsj, I still can't get into the Better Code Hub details, and I don't know how seriously you're concerned with what it monitors/reports; otherwise, what are your thoughts on the compromises made re visual aspects when a toolbar is added? Other thoughts on this?

Pending bug reports or change suggestions, I'm not expecting to do any more work on this particular pull, so whenever you're good with it, let me know.

latenitefilms:
FWIW, I merged this code into CommandPost yesterday, and pushed out a release, and I haven't had any screaming users yet, so all looks good to me!

latenitefilms:
@asmagill - Is there any reason why only one `hs.webview.toolbar` item can be `selectable`? I was kinda hoping that I could use multiple buttons as toggles, but this currently isn't possible, even if each button is in its own group.

asmagill:
Only one toolbar item can be "selected" (have a highlighted background) at a time, but multiple items can be selectable; you need to add the `selectable` attribute and set it to `true` (it's false if not specified) to each item you want to be able to have the background when it is the last one clicked on... if you look at my example, you'll see that the first three items i define have this property.

latenitefilms:
Yeah, I was hoping that multiple items could be "selected" (have a highlighted background) at once - imagine bold, italic and underline in TextEdit. All good - not a major issue.

latenitefilms:
@asmagill - Next question... would it be possible to add a `nonactivating` option to a `hs.chooser`, like you can do with `hs.webview`?

For example:
```lua
:windowStyle({"titled", "nonactivating", "closable", "HUD", "utility"})
```

I'd love it if when I triggered the chooser, that it kept my current open app active?

latenitefilms:
Actually... maybe it's not `nonactivating` that's causing this, maybe it's:

```lua
:level(drawing.windowLevels.floating)
```

asmagill:
I think what you're wanting is in fact non activating... setting the level to floating just makes the window hover above other things when its visible -- if it takes input (mouse clicks, keyboard entry, etc.) then it will still activate Hammerspoon as the current app.

And, before you ask, adding the `NSNonactivatingPanelMask` would require significant work at this time as it only works for `NSPanel` objects, not `NSWindow` objects. If you look closely at the code for `hs.canvas` and `hs.webview`, they are subclasses of `NSPanel` (which is itself a subclass of `NSWindow`) for this very reason.

I'm not saying it couldn't be done, but it would require some thought and probably significant coding and testing... since `NSPanel` is a subclass of `NSWindow`, in theory, it *should* be doable, but since the window and controller are defined with the XCode Interface Builder, we'd have to start there and I know next to nothing about that... I don't know if it would be as simple as toggling a few options or require creating a completely new object with the builder... once in place, 90+% of the existing `hs.chooser` code would probably remain the same as long we made sure to turn on a few of the things that are on by default for `NSWindow` but off by default for `NSPanel`.

So, maybe add it as a "nice feature for the future", but it's not something that could be easily added to this pull.

latenitefilms:
Ah, ok - thanks heaps for the explanation @asmagill.

Have you ever played around with this undocumented method?

```
@interface NSWindow (Private)
- (void )_setPreventsActivation:(bool)preventsActivation;
@end

[myWindow _setPreventsActivation:true];
```

asmagill:
I have not... it might do what you want; then again I've seen more than one private API method hang around long after it no longer did anything, so... YMMV.

Probably worth a try but I won't have time myself until the weekend... If you want to give it a shot, I'd probably add it to the `hs.chooser` module directly since, if it works, it has benefits beyond the addition of a toolbar.

latenitefilms:
Yeah, I haven't had much luck with the private API - not sure if it just doesn't do anything or if I'm using it wrong.

Changing `@interface HSChooserWindow : NSWindow` to `@interface HSChooserWindow : NSPanel` doesn't seem to break anything, which is interesting. This automatically updates the options in the Interface Builder.

However, changing the Panel Style to "HUD Panel" and ticking "Non Activating" in the interface building still doesn't achieve what I want. It does now open the chooser without making Hammerspoon active, however the chooser itself is not active, so as soon as I click on the search text box, Hammerspoon gets focus again (which is not what I expected with "Non Activating" ticked).

Screen Shot 2019-04-04 at 6 03 49 pm

latenitefilms:
Actually commenting out these two lines:

```objc
//-(BOOL)canBecomeMainWindow { return YES; }
//-(BOOL)canBecomeKeyWindow { return YES; }
```

...has solved the focus issue. Now I can click on the Chooser without it changing apps! Woohoo!

Next problem to solve is that when the Chooser initially loads, it doesn't actually get focus so I still need to click on the chooser initially.

latenitefilms:
Well, I feel like I've tried every combination of things, but I can't get the chooser `NSPanel` to get focus without actually clicking on it. Any words of wisdom @cmsj ?

latenitefilms:
I've tried everything I can think of to try and get `hs.chooser` working as a non-activating `NSPanel` but have had no luck. I've added a seperate issue here: #2067

latenitefilms:
@asmagill - Looks like you are actually using that undocumented API call elsewhere:

https://github.com/asmagill/hammerspoon_asm.undocumented/blob/master/touchbar/internal.m#L185

Maybe this might give you some clues?

latenitefilms:
@asmagill - I just realised that I can only get "nonactivating" to work on a `hs.webview` if I have a toolbar. If I don't have a toolbar, it will always activate Hammerspoon when I click on the `hs.webview`.

asmagill:
@latenitefilms do you have your code posted somewhere I can review? My first thought is to ask exactly *where* you're setting the nonactivating flag; some of the changes my code implements only occur when a toolbar is being assigned, so if you just lumped it in the same code block, that could be why.

latenitefilms:
@asmagill - Here's the change, where I added back in a toolbar to make non-activating work:

https://github.com/CommandPost/CommandPost/commit/1bd7c385329a85e23b42c2f290cbc9e0d31a5678#diff-5fd28b56fda6c9c87887b68aff438027L352

Hopefully it makes SOME sense.

latenitefilms:
FYI: I also tried to give using `[myWindow _setPreventsActivation:true];` another bash, but I still can't get the right combination for it to work. `preventsActivation` definitely works, as your virtual Touch Bar works as expected.

asmagill:
My bad - I missed that you said on a webview, I assumed because of the thread it was still your attempts on the non-activating chooser...

I'll have to give it some thought, not sure... my understanding is that the flag should be sufficient, but I'll try and take a closer look later and see what I can determine.

latenitefilms:
Sorry for the confusion... the "non-activating" issue is with a `hs.webview`. Basically, I have to have a toolbar visible for `nonactivating` to work.

However, the `preventsActivation` undocumented preference, I've been trying to apply to a `hs.chooser`.

Basically, I want to get "non-activating" work on both webview's and choosers, with and without toolbars. Currently I can only get them working on webviews WITH toolbars.

PR #2063, opened 2019-03-29
Fix some static analysis issues relating to hs.webview blocks not explicitly retaining self. # Requesting a pull to hammerspoon:master from cmsj:cleanup-application # # Write a message for this pull request. The first block # of text is the title and the rest is the description.

cmsj:


cmsj:
@asmagill if you have some time (no rush, this isn't really a bug), could you take a look at this? Relates to the static analyser in Xcode - it says that previously we were implicitly retaining self and this change makes it explicit.

asmagill:
lgtm... I remember starting to see this warning when I was last doing regular work on Hammerspoon -- I think its one of those "it's better to be explicit then make assumptions" type of things that the clang developers have decided we should be watching for.

Don't know that it's worth making a concentrated effort to track them all down, but as we work on a module here or there and notice the warnings, I'm all for adding the explicit references.

cmsj:
@asmagill I've already chased almost all of them down, we're almost at a place now where the only ones left are the analyser not understanding what we're doing (e.g. hs.crash) or things we legitimately can't fix (like the claimed IPC leaks).

PR #2070, opened 2019-04-08
Rigid following of stickler "suggestions" broke metatables for canvas elements

asmagill:
By rigidly following stickler's suggestions without considering the surrounding code, a case where `_` was being used as a real parameter and not just as a place holder broke and caused metatables for canvas elements to break.

To see what I mean, take a canvas object you've created for something and type `canvasObject[1]` (where `canvasObject` is whatever variable you're storing the canvas in, of course) into the console before and after applying these changes...

### DIFF
--- a/extensions/canvas/init.lua
+++ b/extensions/canvas/init.lua
@@ -530,8 +530,8 @@ elementMT.__newindex = function(_, k, v)
     end
 end
 
-elementMT.__pairs = function(_)
-    local obj = elementMT.__e[_]
+elementMT.__pairs = function(s)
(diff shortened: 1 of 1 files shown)


latenitefilms:
Opps! My bad probably - apologies!

asmagill:
Well, I'm the first to notice, so... evidently it's not a widely used or known feature

It's not a required feature for the module to work (obviously), but it is a useful debugging tool to be able to see each element's attributes in the console (at least to me...)

cmsj:
Apologies for missing that in review.

PR #2082, opened 2019-04-25
[READY] Issue #2077 - nil error in hs.spoons

### LINKED ISSUE #2077: hs.spoons.use: attempt to index a nil value (global 'spoon')
`hs.spoons.use`, which looks like a great way to load spoons, calls: 

```lua
      local spn=hs.loadSpoon(name)
```
 This fails to set the global variable `spoon`, which results in the error above.  This occurs because `hs.spoons.list`, which is called on use of `use` (via `isInstalled`), references this global variable.  Perhaps the global should just be included by default if the spoons module is going to use it throughout.  Or at minimum `use` should enable the global by calling `loadSpoon` with a 2nd argument of `true`.

latenitefilms:
- Fixed a `nil` error in `hs.spoons`
- Closes #2077


cmsj:
Before we land something like this, I'd like to understand why `spoon` is nil.

cmsj:
Closing in favour of 9866d934

PR #2083, opened 2019-04-25
[READY] Issue #2067 - Made hs.chooser a non-activating NSPanel

### LINKED ISSUE #2067: Convert hs.chooser into a non-activating NSPanel
Currently `hs.chooser` uses a `NSWindow`, which doesn't support "non activating". It would be great if we could convert `hs.chooser` from a `NSWindow` to a `NSPanel` so that we can customise the `windowStyle`, like you can with a `hs.webview`. Specifically, I want to be able to activate a `hs.chooser` from another application (i.e. Safari), and when the chooser appears, Safari (not Hammerspoon) still has focus.

Discussed here: 

https://github.com/Hammerspoon/hammerspoon/pull/2062#issuecomment-479764102

latenitefilms:
- Changed `hs.chooser` from using a `NSWindow` to a non-activating
`NSPanel`.
- Closes #2067


latenitefilms:
@asmagill - FYI: when this code is used in conjunction with #2062, it seems adding a toolbar makes the `hs.chooser` non-activating (i.e. you can't actually type text into the console). I think this is due to the way you're currently applying the `theWindow.styleMask` in `hs.webview.toolbar.attachToolbar`, but I haven't found the right combination to make it work yet.

latenitefilms:
Given all the changes made to `hs.chooser`, I'm going to close this pull request and try again.

PR #2086, opened 2019-05-03
Target 10.12

cmsj:


cmsj:
I'd be interested to hear objections to us finally moving on from targeting 10.10. This PR gets us up to 10.12.

latenitefilms:
Fine by me!

asmagill:
No objections if it doesn't break anything.

cmsj:
I can't promise it doesn't break anything, but I'll at least do a full test run before I merge this.

cmsj:
Also this isn't working at all right now because of some inscrutable cocoapods build failure :/

cmsj:
I've been able to run through the test suite locally and things seem generally in good shape.

asmagill:
To a lesser degree, I guess the only outstanding question then would be places where we detect the 10.10 version either through compiler directives or at run time with weak bindings...

While neither should break anything, and may make someone's efforts to "back port" Hammerspoon easier should they want to, it does make the code more complex in a handful of places... is it worth it to make a pull at some point in the near future where we remove this legacy code just to clean up the sources? We probably still have some tests for 10.9 and lower lurking around in the sources as well that could be cleaned out.

PR #2090, opened 2019-05-03
[READY] Rework chooser ui

cmsj:


cmsj:
Note that this depends on the 10.12 target PR

cmsj:
Tested as working with the new hs.chooser toolbar attach feature. Hopefully people prefer this new tweaked design - all I really did was change the autolayout constraints a bit, since I now understand them a lot better than when I first made the chooser UI.

latenitefilms:
@cmsj - Curious... have you tested this in High Sierra? I wonder if these changes are causing what I'm seeing in #2108.

Issue #2098, opened 2019-05-07
The road to 2.0


cmsj:
I have thoughts :)

Over the last year or so, I've been isolating more and more of my config into Spoons, and one problem I keep running into is that most of our modules assume there will just be one callback function.

One way out of this would be to allow multiple callbacks to be registered, but I think that would likely end up feeling messy.

I think a much more usable setup would be if we redesign the modules around a scheme like hs.watchable. I really like the idea of random parts of Spoons/configs being able to just register themselves to receive events of a certain nature, and the underlying modules would take care of creating the relevant OS watchers and feeding events into the watchable-like system.

My thoughts would be that we'd just arbitrarily declare one of the upcoming 0.9.x releases as 1.0.0 and then get to work remaking all the callback modules, the first release of which would be 2.0.0, to signify the giant API breakage.

@asmagill and @latenitefilms - as our most active contributors, what are your thoughts on this?

cmsj:
(other thoughts are certainly welcome, in case there is any doubt)

cmsj:
In a more general sense, I'd see such a 2.0 as an opportunity to rethink a lot of our APIs, and bring them up to date with how things tend to work in more modern modules (e.g. getting rid of separate get/set methods)

latenitefilms:
We have an extension in CommandPost called [`cp.prop`](https://github.com/CommandPost/CommandPost/blob/develop/src/extensions/cp/prop/init.lua) that might be of interest?

To quote the manual:

```lua
--- === cp.prop ===
---
--- This is a utility library for helping keep track of single-value property states. Each property provides access to a single value. Must be readable, but may be read-only. It works by creating a table which has a `get` and (optionally) a `set` function which are called when changing the state.
---
--- --
---
--- ## Features:
--- ### 1. Callable
--- A `prop` can be called like a function once created. Eg:
---
--- ```lua
--- local value = true
--- local propValue = prop.new(function() return value end, function(newValue) value = newValue end)
--- propValue() == true     -- `value` is still true
--- propValue(false) == false   -- now `value` is false
--- ```
---
--- ### 2. Togglable
--- A `prop` comes with toggling built in - as long as the it has a `set` function. Continuing from the last example:
---
--- ```lua
--- propValue:toggle()  -- `value` went from `false` to `true`.
--- ```
---
---  **Note:** Toggling a non-boolean value will flip it to `nil` and a subsequent toggle will make it `true`. See the [toggle method](#toggle) for more details.
---
[...]

Issue #1567, opened 2017-10-03
hs.menubar.new(false) appearing in system menubar?


latenitefilms:
I have a `hs.menubar` "attached" to a `hs.chooser`, so that when I right click on the `hs.chooser` it pops up the menubar. I've noticed though, that ever now and again, even though I'm using `hs.menubar.new(false)`, the menubar will appear in the system menubar. It doesn't have any text value or icon, so you'd only notice if you click up there, or if another application has something in the menubar and you notice the weird space between the icons - but it's a little... odd.

Possibly related to #1566?

latenitefilms:
I can reproduce this with:

```
-- hs.chooser Example

local inspect = require("hs.inspect")

local logger = require("hs.logger")
logger.defaultLogLevel = 'debug'
log = logger.new("chooser")

local chooser = require("hs.chooser")
local menubar = require("hs.menubar")
local mouse = require("hs.mouse")

showAChoice = true
numberOfChoices = 100000

function choicesA()
	local result = {}

	for i=1, numberOfChoices, 1 do
		table.insert(result, {
			["text"] = "Choice " .. tostring(i),
			["subText"] = "This is the subtext of choice " .. tostring(i),
			["uuid"] = tostring(i)
		})
	end

	return result
end

function choicesB()
	local result = {}

	for i=1, numberOfChoices, 1 do
		table.insert(result, {
			["text"] = "New Choice " .. tostring(i),
			["subText"] = "This is the subtext of choice " .. tostring(i),
			["uuid"] = tostring(i)
		})
	end

	return result
end

myMenubar = menubar.new(false)
	:setMenu({
       {
       	title = "Refresh",
       	fn = function()
       		log.df("Refreshing Choices.")
       		if showAChoice then
       			choices = choicesB()
       		else
       			choices = choicesA()
       		end
       		showAChoice = not showAChoice
			myChooser:refreshChoicesCallback()
       	end
       },
   })

function completionFn(value)
	log.df("completionFn value: %s", inspect(value))
end

function choicesFn(value)
	if not choices then
		choices = choicesA()
	end
	return choices
end

function rightClickFn()
	local point = mouse.getAbsolutePosition()
	myMenubar:popupMenu(point)
end

function queryChangedFn(value)
	log.df("queryChangedFn value: %s", inspect(value))
end

function showFn()
	log.df("showFn triggered.")
end

myChooser = chooser.new(completionFn)
	:choices(choicesFn)
[...]

PR #2133, opened 2019-08-01
Fix detached menus so they don't leave traces in menubar

asmagill:
address #1567 and menubar notes in #2098
also clears compiler warnings during build

### DIFF
--- a/extensions/menubar/init.lua
+++ b/extensions/menubar/init.lua
@@ -41,7 +41,7 @@ end
 ---  * None
 ---
 --- Returns:
----  * an hs.geometry rect describing the menubar item's frame
+---  * an hs.geometry rect describing the menubar item's frame or nil if the menubar item is not currently in the menubar.
 ---
 --- Notes:
 ---  * This will return a frame even if no icon or title is set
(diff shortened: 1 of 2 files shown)


asmagill:
Details, for the interested:

Cleaned up compiler warnings, mostly converting nested objective-c calls into multiple lines which verify object existence before invoking next method.

Converted assignments of `nil` assigned to `void *` pointers to `NULL` since `nil` is suppose to be reserved for objective-c objects and `NULL` is used used for C type pointers; the compiler is supposed to be smart enough to correct for this, but this is more explicit.

Fix for persisting pop-up menus: when originally written, because the `NSStatusItem` class does not list an initializer, all status items were created through the `NSStatusBar` class (which are attached to the menubar) and pop-up menus were removed from the menubar with the `removeStatusItem:` method.

This worked until sometime around macOS 10.11 or so (not sure exactly); after that, because the status item originated in the menu bar, macOS would occasionally re-attach the status item, even after a removal, if the `NSStatusItem` object wasn't collected via ARC quickly enough (we kept it around in our own wrapper because the `NSMenu` used as the pop-up was a property of the `NSStatusItem` object.)

This change uses the generic initializer that all descendants of `NSObject` inherit (e.g. `[[NSStatusItem alloc] init]`) and then copies the properties of the status item in the menubar into this new status item, and then allows the status item which was created in the menubar to be collected immediately, preventing traces of it from being re-added to the menubar.

asmagill:
@cmsj, can't see BCH details for this either:
Screen Shot 2019-08-01 at 3 42 42 PM

cmsj:
I can’t even see BCH these days 😕 I’ll remove it at some point.

latenitefilms:
As discussed in the other issue, I've tested this code with CommandPost, and all working great - so feel free to merge away when ready @cmsj !

PR #2138, opened 2019-08-16
[READY] Convert hs.chooser into a non-activating NSPanel

### LINKED ISSUE #2067: Convert hs.chooser into a non-activating NSPanel

latenitefilms:
- Closes #2067


cmsj:
Looks good to me, thanks very much! This is a pretty spectacular improvement :D

cmsj:
Argh, sorry, I had to revert again. It's the arrow key navigation that breaks.

Issue #1782, opened 2018-05-01
Crash - Skin.m line 406 - -[LuaSkin checkArgs:]


latenitefilms:
See: http://crashes.to/s/f477e2cc3d9

Code in question: [`[_skin pushNSObject:stdOutArg];`](https://github.com/Hammerspoon/hammerspoon/blob/master/extensions/task/internal.m#L930)

**Logs:**
```
1443 | \| | 08:00:43:467 (UTC) | \| | hs.task terminationHandler block encountered an exception: *** -[NSConcreteFileHandle readDataOfLength:]: Resource temporarily unavailable
-- | -- | -- | -- | --
1444 | \| | 08:00:43:467 (UTC) | \| | (WARNING) hs.task terminationHandler block encountered an exception: *** -[NSConcreteFileHandle readDataOfLength:]: Resource temporarily unavailable
1445 | \| | 08:00:43:470 (UTC) | \| | hs.task terminationHandler block encountered an exception: *** -[NSConcreteFileHandle readDataOfLength:]: Resource temporarily unavailable
1446 | \| | 08:00:43:470 (UTC) | \| | (WARNING) hs.task terminationHandler block encountered an exception: *** -[NSConcreteFileHandle readDataOfLength:]: Resource temporarily unavailable
```

latenitefilms:
Maybe related? 

https://stackoverflow.com/questions/7871061/how-to-use-nsfilehandles-writeabilityhandler

cmsj:
I disagree with your "Code in question" - if you look at "Keys" section in "View all Sessions" on the crash report, there's:

`Assertion failed: (ref != LUA_NOREF && ref != LUA_REFNIL), function -[LuaSkin pushLuaRef:ref:], file /Users/chrishocking/Github/CommandPost-App/LuaSkin/LuaSkin/Skin.m, line 389.`

which is more likely to be one of the two lines preceding the `pushNSObject`. It's unlikely to be the `userData->luaStreamCallback` ref push, since that was validated on line 908, which leaves `userData->selfRef`. My reading of the code is that this can only be `LUA_NOREF` or `LUA_REFNIL` in three situations:

1. The task object hasn't been created yet
2. The task object has been GC'd
3. The task object's `terminationHandler` has been called

It seems impossible that 1. could be relevant here. So, that leaves the question of whether the file handles can call the observer block (as defined between lines 891 and  958, after the Task has either terminated on its own, or because we GC'd it.

[...]

PR #2139, opened 2019-08-17
hs.task - check selfRef validity in stream callback

asmagill:
Ok, my morning appointment got cancelled, so here's a potential fix based on my comments in #1782

### DIFF
--- a/extensions/task/internal.m
+++ b/extensions/task/internal.m
@@ -179,7 +179,7 @@ void create_task(task_userdata_t *userData) {
 ///   * stdOut - A string containing the standard output of the process
 ///   * stdErr - A string containing the standard error output of the process
 ///  * streamCallbackFn - A optional callback function to be called whenever the task outputs data to stdout or stderr. The function must return a boolean value - true to continue calling the streaming callback, false to stop calling it. The function should accept three arguments:
-///   * task - The hs.task object
+///   * task - The hs.task object or nil if this is the final output of the completed task.
 ///   * stdOut - A string containing the standard output received since the last call to this callback
 ///   * stdErr - A string containing the standard error output received since the last call to this callback
 ///  * arguments - An optional table of command line argument strings for the executable
@@ -189,6 +189,7 @@ void create_task(task_userdata_t *userData) {
 ///
 /// Notes:
 ///  * The arguments are not processed via a shell, so you do not need to do any quoting or escaping. They are passed to the executable exactly as provided.
+///  * When using a stream callback, the callback may be invoked one last time after the termination callback has already been invoked. In this case, the `task` argument to the stream callback will be `nil` rather than the task userdata object and the return value of the stream callback will be ignored.
 static int task_new(lua_State *L) {
     // Check our arguments
     LuaSkin *skin = [LuaSkin shared];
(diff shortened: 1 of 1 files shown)


joemiller:
@asmagill If the `task` arg to the callback func is `nil` will there (possibly) be any data in the `stdOut` or `stdErr` args or should the callback always return immediately when `task == nil` ?

cmsj:
@joemiller I suspect the answer right now would be we don’t know. I’ll try and test it shortly and report back.

asmagill:
For a slight variant to the code presented in #1782, I get this in the console:

~~~


> task = hs.task.new("_scratch/foo.sh", function(exitCode, stdOut, stdErr)
	print(string.format("termination rc    : %s", exitCode))
	print(string.format("termination stdOut: %s", stdOut))
	print(string.format("termination stdErr: %s", stdErr))		
end, function(ud, stdOut, stdErr)
	print(string.format("stream ud    : %s", ud))
	print(string.format("stream stdOut: %s", stdOut))
	print(string.format("stream stdErr: %s", stdErr))		
	return true
end):start()

stream ud    : hs.task: _scratch/foo.sh  (0x6080003b3548)
stream stdOut: hello args

stream stdErr: 
stream ud    : hs.task: _scratch/foo.sh  (0x6080003b3548)
stream stdOut: 
stream stdErr: + echo hello args
+ sleep 2

stream ud    : hs.task: _scratch/foo.sh  (0x6080003b3548)
stream stdOut: i am still here 18261

stream stdErr: 
stream ud    : hs.task: _scratch/foo.sh  (0x6080003b3548)
stream stdOut: 
stream stdErr: + echo i am still here 18261
+ sleep 2

stream ud    : hs.task: _scratch/foo.sh  (0x6080003b3548)
stream stdOut: ok i am done now 12438

stream stdErr: 
termination rc    : 0
termination stdOut: 
termination stdErr: 
stream ud    : nil
stream stdOut: 
stream stdErr: + echo ok i am done now 12438
+ exit 0

~~~

So, it looks like the final stream callback (which you'll note is after the termination callback) did contain output... in this case it was the results of the `set -x` command in the script to output each line as it's executed to `stdErr`.

So, my take is that you shouldn't assume that `stdOut` and `stdErr` will be empty, even when the userdata object is `nil`.

asmagill:
It probably depends upon when the task flushes the `stdOut` and `stdErr` buffers... for the sample script I've been using:

~~~sh
#!/bin/bash

set -x

echo hello args $@
sleep 2
echo i am still here $RANDOM
sleep 2
echo ok i am done now $RANDOM
#ls -lR
#exit 1
exit 0
~~~

I commented out the `set -x` line so that the last "output" is the last `echo` command, and running the same test in the console does not invoke the stream callback after the termination callback is invoked.

A few other tests (changing the last `echo` to `echo -n`, redirecting it to `stdErr`, clearing all of the lines after the last echo, including the `exit`) don't change this behavior, so I'm going to make the broad and poorly minimally claim that *most* scripts probably won't ever have this issue unless the shell itself is generating some of the output (like `bash` does when `set -x` is invoked), but that with any compiled task invoked, it's going to depend entirely upon when it's flushing the `stdOut` and `stdErr` buffers -- and if there is anything in one of the buffers when the task terminates that has not been flushed, *that* is when you'll get a stream callback *after* the termination callback.

cmsj:
I haven’t had a moment to test, but I agree with @asmagill’s take. Any process that shares the file descriptors (so, possibly any parent or child process, plus the OS if it feels like it), could write to the file descriptors at basically any time. Fun!

joemiller:
awesome. I hope to try this branch myself tomorrow, or master if it's merged by then. thanks all

PR #2140, opened 2019-08-21
fix error when triggering mouse event with modifier

wwwjfy:
modifer is the value in the table, so the index to convert should be -1
instead of -2

### DIFF
--- a/extensions/eventtap/event.m
+++ b/extensions/eventtap/event.m
@@ -793,7 +793,7 @@ static int eventtap_event_newMouseEvent(lua_State* L) {
     if (!lua_isnoneornil(L, 4) && (lua_type(L, 4) == LUA_TTABLE)) {
         lua_pushnil(L);
         while (lua_next(L, 4) != 0) {
-            modifier = lua_tostring(L, -2);
+            modifier = lua_tostring(L, -1);
             if (!modifier) {
                 [skin logBreadcrumb:[NSString stringWithFormat:@"hs.eventtap.event.newMouseEvent() unexpected entry in modifiers table: %d", lua_type(L, -1)]];
                 lua_pop(L, 1);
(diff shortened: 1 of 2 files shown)


wwwjfy:
This also fixes the error:
```
hs.eventtap.event.newMouseEvent(hs.eventtap.event.types["leftMouseDown"], hs.geometry.point(0,0), {"alt"}):post()
invalid key to 'next'
stack traceback:
	[C]: in function 'hs.eventtap.event._newMouseEvent'
	...n.app/Contents/Resources/extensions/hs/eventtap/init.lua:167: in function 'hs.eventtap.event.newMouseEvent'
	[string "return hs.eventtap.event.newMouseEvent(hs.eve..."]:1: in main chunk
	[C]: in function 'xpcall'
	...app/Contents/Resources/extensions/hs/_coresetup/init.lua:422: in function <...app/Contents/Resources/extensions/hs/_coresetup/init.lua:402>
```

asmagill:
There is a little more going on here...

If you review the documentation, this constructor for `hs.eventtap.event` reads "modifiers - An optional table containing zero or more of the following keys:" which means it's expecting a table like `{ cmd=true, shift=true }`.

And the code without this commit, does that, except for two things:
1. it never checks that the value of the `key` is actually true (i.e. not `nil` or `false`), so `{ shift = false }` would be as valid as `{ shift = true }` which is inconsistent with most every lua program that uses key-value pairs for arguments... and definitely bad style.
2. more importantly, it's inconsistent with every other constructor in this module which just wants a list of modifiers.

So, since I did most of the coding for this module, let me just say "oops, my bad!"

If you could tweak the documentation for `hs.eventtap.event.newMouseEvent` (it should be in the module's `init.lua` file since this is wrapped) to match the other constructors for this module, I'll land this.

@cmsj, what is the status of the master branch? Is it safe for me to merge things, once this is documented properly, or are you still in the middle of tweaking for the new macOS/XCode?

cmsj:
Master is open for business 🙂

wwwjfy:
ah I see, `{alt=true}` works for me in current version.
But yes, it's better to be consistent with others.

I updated the doc (just gave an example). Feel free to add more changes where you see suitable. Thanks!

asmagill:
Thanks!

PR #2144, opened 2019-08-22
Allow hs.eventtap.event.newMouseEvent() to accept mouseMoved

cmsj:
@asmagill someone on IRC brought up a question about how to send mouse movements with `hs.eventtap` and in digging around with the `mouseEventDeltaX` property, I can't get it to work.

I've tried both:

```lua
foo= hs.eventtap.event.newEvent()
foo:setType(hs.eventtap.event.types.mouseMoved)
foo:setProperty(hs.eventtap.event.properties.mouseEventDeltaY, 100)
foo:location(foo:location())
foo:post()
```

and (with this PR applied):

```lua
foo= hs.eventtap.event.newMouseEvent(hs.eventtap.event.types.mouseMoved, hs.mouse.getAbsolutePosition())
foo:setProperty(hs.eventtap.event.properties.mouseEventDeltaY, 100)
foo:post()
```

but neither seems to cause the mouse pointer to actually move.

Am I doing something wrong, or is this broken somehow?

### DIFF
--- a/extensions/eventtap/event.m
+++ b/extensions/eventtap/event.m
@@ -860,6 +860,8 @@ static int eventtap_event_newMouseEvent(lua_State* L) {
         button = kCGMouseButtonRight;
     else if (strcmp(buttonString, "other") == 0)
         button = kCGMouseButtonCenter;
+    else if (strcmp(buttonString, "none") == 0)
+        button = 0;
 
     if (!lua_isnoneornil(L, 4) && (lua_type(L, 4) == LUA_TTABLE)) {
         lua_pushnil(L);
--- a/extensions/eventtap/init.lua
+++ b/extensions/eventtap/init.lua
@@ -140,7 +140,7 @@ end
 --- Creates a new mouse event
 ---
 --- Parameters:
----  * eventtype - One of the values from `hs.eventtap.event.types`
+---  * eventtype - One of the mouse related values from `hs.eventtap.event.types`
 ---  * point - An hs.geometry point table (i.e. of the form `{x=123, y=456}`) indicating the location where the mouse event should occur
 ---  * modifiers - An optional table containing zero or more of the following keys:
 ---   * cmd
@@ -160,6 +160,8 @@ function module.event.newMouseEvent(eventtype, point, modifiers)
         button = "right"
     elseif eventtype == types["otherMouseDown"] or eventtype == types["otherMouseUp"] or eventtype == types["otherMouseDragged"] then
         button = "other"
+    elseif eventtype == types["mouseMoved"] then
+        button = "none"
     else
         print("Error: unrecognised mouse button eventtype: " .. tostring(eventtype))
         return nil


asmagill:
Hmm... can confirm that it's not working...

But this does:

~~~lua
delta = 100
foo= hs.eventtap.event.newEvent()
foo:setType(hs.eventtap.event.types.mouseMoved)
mouseAt = hs.mouse.getAbsolutePosition()
mouseAt.y = mouseAt.y + delta
foo:location(mouseAt)
foo:post()
~~~

as does this (after the pull, didn't try it before applying the pull):

~~~lua
delta = 100
mouseAt = hs.mouse.getAbsolutePosition()
mouseAt.y = mouseAt.y + delta
hs.eventtap.event.newMouseEvent(hs.eventtap.event.types.mouseMoved, mouseAt):post()
~~~

Looks like the important part is setting the location to the destination rather than adjusting the delta properties. I suppose that the deltas might be nice if we think something else might be looking at the events we're posting.

Not sure if you want to code a specific helper function for moving the mouse based on these new findings or land this as is, since it is a better implementation when creating a move event with no button press associated.

shayanjm:
So I'm the one who brought this up in IRC. I'm trying to use Hammerspoon to automate mouse movements within a game built on Unity, which seems to center the mouse on the screen. I tried both examples (I think the second one only works with the patch), and neither worked in the game. That being said, I was able to get Karabiner's mouse movement to work fine in-game. I suspect we need to be able to manipulate the mouse delta directly to get results here (as opposed to selecting a mouseAt position), but I could be wrong. Any thoughts?

asmagill:
You could try throwing the delta in anyways, even if traditional mouse movement doesn't require it, and see if it help with your specific situation. Otherwise I suspect we're just not injecting the event early enough into the stream and KE is probably the way to go.

Something like this:

~~~lua
delta = 100
foo= hs.eventtap.event.newEvent()
foo:setType(hs.eventtap.event.types.mouseMoved)
mouseAt = hs.mouse.getAbsolutePosition()
foo:setProperty(hs.eventtap.event.properties.mouseEventDeltaY, delta)
mouseAt.y = mouseAt.y + delta
foo:location(mouseAt)
foo:post()
~~~

shayanjm:
@asmagill - that worked! Thanks for looking at this.

cmsj:
@asmagill thoughts on landing this PR anyway, so people can create a `newMouseEvent()` with `mouseMoved` already set?

asmagill:
It might have been lost in the dialog, but an earlier question I posed is should we land this as is, or do you also want to add a helper function for mouse movement from the current position that also takes a table with x and y specified and preset the new location and delta's?

With this pull as it is, @shayanjm would have to do the following:

~~~lua
delta = 100
mouseAt = hs.mouse.getAbsolutePosition()
mouseAt.y = mouseAt.y + delta
myEvent = hs.eventtap.event.newMouseEvent(hs.eventtap.event.types.mouseMoved, mouseAt)
myEvent:setProperty(hs.eventtap.event.properties.mouseEventDeltaY, delta):post()
~~~

It might be nice to have a constructor specifically for mouse movement, something like:

~~~
hs.eventtap.event.newMouseMoveBy(delta, [from]) -> event
~~~

Where `delta` is a point table specifying the x and y delta's and `from` is optionally the point the mouse moves "from"... internally the function would take `from` or `hs.mouse.getAbsolutePosition()`, apply the deltas to it, use that new point as the target for the event, and also set the `mouseEventDeltaX` and `mouseEventDeltaY` properties.

Of course, this then suggests a companion `hs.eventtap.event.newMouseMoveTo(dest, [from])` where it calculates the deltas as `dest.[xy]` - `from.[xy]` where `from` defaults to `hs.mouse.getAbsolutePosition` if not specified...

Your call; as is, the new pull is still easier then creating the event completely from scratch, but we should probably document somewhere in here that creating mouse movement that other apps can see is a two step process -- `target` needs to be the destination, not the position before the movement, and `mouseEventDelta[XY]` need to reflect the magnitude of movement that other apps will see (the event doesn't apply them to the target location for us; they're the movement that got us *to* the target location) -- otherwise the movement is simply a re-positioning of the mouse pointer.

cmsj:
I'm going to merge this as-is, and maybe we can circle back to add an `hs.eventtap.event.newMoveMouseBy()` as @asmagill suggests (anyone fancy filing an Issue, or writing the function? 😁)

PR #2147, opened 2019-08-23
[READY] Added hs.webview.toolbar.uniqueName()

### LINKED ISSUE #2109: Add hs.webview.toolbar.get?
Currently when you use `hs.webview.toolbar.new` you need to assign each toolbar object with a unique name. However, currently there's no way to check that name isn't already in use.

latenitefilms:
- Closes #2109


latenitefilms:
@asmagill - Very good point! Fixed!

asmagill:
lgtm

PR #2148, opened 2019-08-23
[READY] Added hs.webview:isVisible()

### LINKED ISSUE #2072: Add isVisible() to hs.webview?
Currently you can use `hs.webview:hswindow() == nil` to detect if a webview is visible or not, but a native `:isVisible()` method would be handy?

latenitefilms:
- Closes #2072


asmagill:
lgtm

Issue #1793, opened 2018-05-15
Better hs.uielement Documentation?


latenitefilms:
Should `hs.uielement:newWatcher` actually be `hs.uielement.newWatcher`?

I was under the impression that all constructors should be functions not methods?

latenitefilms:
I've also just noticed the following bug:

```lua
2018-05-16 00:03:37: Welcome to the Hammerspoon Console!
You can run any Lua code in here.

2018-05-16 00:03:39: -- Lazy extension loading enabled
2018-05-16 00:03:39: -- Loading extension: notify
2018-05-16 00:03:39: -- Loading extension: fnutils
2018-05-16 00:03:39: -- Can't find ~/.hammerspoon/init.lua; create it and reload your config.
2018-05-16 00:03:41: ********
2018-05-16 00:03:41: 00:03:41 ERROR:   LuaSkin: hs.dockIconClickCallback: attempt to call a nil value
stack traceback:
2018-05-16 00:03:41: ********

> hs.uielement.focusedElement()
2018-05-16 00:03:43: -- Loading extension: uielement
userdata: 0x60000064ca18
2018-05-16 00:03:43: ********
2018-05-16 00:03:43: 00:03:43 ERROR:   LuaSkin: hs.application.watcher callback: ....app/Contents/Resources/extensions/hs/uielement/init.lua:78: attempt to index a userdata value (local 'app')
stack traceback:
	....app/Contents/Resources/extensions/hs/uielement/init.lua:78: in function <....app/Contents/Resources/extensions/hs/uielement/init.lua:77>
2018-05-16 00:03:43: ********
2018-05-16 00:03:43: ********
2018-05-16 00:03:43: 00:03:43 ERROR:   LuaSkin: hs.application.watcher callback: ....app/Contents/Resources/extensions/hs/uielement/init.lua:78: attempt to index a userdata value (local 'app')
stack traceback:
	....app/Contents/Resources/extensions/hs/uielement/init.lua:78: in function <....app/Contents/Resources/extensions/hs/uielement/init.lua:77>
2018-05-16 00:03:43: ********
2018-05-16 00:03:46: ********
2018-05-16 00:03:46: 00:03:46 ERROR:   LuaSkin: hs.application.watcher callback: ....app/Contents/Resources/extensions/hs/uielement/init.lua:78: attempt to index a userdata value (local 'app')
stack traceback:
[...]

PR #2149, opened 2019-08-23
[READY] Fixed various bugs in hs.uielement

### LINKED ISSUE #1104: hs.application.watcher error using hs.uielement
I set up a hotkey to grab the current text selection like this:

```
local text = hs.uielement.focusedElement():selectedText()
```

It seems to be firing this error every time I switch application context. Doesn't seem to affect what I'm doing, but it fills up the log:

```
LuaSkin: hs.application.watcher callback error: ....app/Contents/Resources/extensions/hs/uielement/init.lua:72: attempt to index a userdata value (local 'app')
```

Build: 0.9.50
OSX Version: 10.12.1

latenitefilms:
- Closes #1104

### DIFF
--- a/extensions/uielement/init.lua
+++ b/extensions/uielement/init.lua
@@ -75,7 +75,7 @@ uielement.watcher.titleChanged     = "AXTitleChanged"
 local appWatchers = {}
 
 local function appCallback(_, event, app)
-    if appWatchers[app:pid()] and event == application.watcher.terminated then
+    if app and (appWatchers[app:pid()] and event == application.watcher.terminated) then
         fnutils.each(appWatchers[app:pid()], function(watcher) watcher:_stop() end)
         appWatchers[app:pid()] = nil
     end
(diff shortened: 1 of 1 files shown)


latenitefilms:
@cmsj & @asmagill - You might want to review the discussion [here](https://github.com/Hammerspoon/hammerspoon/issues/1793#issuecomment-389339927) before merging this, as there may still be some other work that needs to be done.

asmagill:
The changes look good as far as catching and preventing lua errors when assumed parameters aren't present (or are `nil`). Do we know if, for the specific cases where errors were being thrown, if the parameter was *supposed* to be `nil` or does it being `nil` suggest a lower level problem?

latenitefilms:
I've never really had much luck with `hs.uielement` in general to be honest, so I haven't look into it too deeply.

latenitefilms:
FWIW: Unless @asmagill or @cmsj has time to review `hs.uielement` in more depth (which I'm guessing neither of you probably do?), I'd suggest we just merge this, as the original issue (#1104) said they were just having log spam, rather than something not working.

PR #2151, opened 2019-08-23
[READY] Added hs.mouse.count()

### LINKED ISSUE #1587: Detect external mouse?
Is there any easy way to detect whether or not an external mouse is connected?

I thought about using `hs.usb.attachedDevices()`, but that won't work as a lot of external wireless mice have a USB receiver.

Thoughts @asmagill ?

latenitefilms:
- Closes #1587


latenitefilms:
@cmsj & @asmagill - This seems to work. It uses codes from an [external repository](http://icculus.org/manymouse/) using a [zlib license](http://hg.icculus.org/icculus/manymouse/raw-file/tip/LICENSE.txt). It contains code only relevant for Windows/Linux that can probably be removed, but that's above my skillset.

If there's a better/easier way to do this without ManyMouse, feel free to implement!

asmagill:
I'm a little leery of bringing in a lot of code for which we only care about a tiny section... it would probably be better (though maybe a little more work at first) to isolate the specific mac os code in their project and replicate just that.

I've only skimmed ManyMouse, so I'm not entirely sure what all it offers, but do you envision a use/need for more than the count of mouse devices? I'm sure with a little time I can isolate that section... I've done it before with some of the other additions we've made over the years (wow, it really is that long now!)

latenitefilms:
As far as I can tell ManyMouse is basically just a wrapper for IOKit's HID Manager API - so I actually think the amount of macOS-related code needed is pretty small. The reason this pull request has so many file changes is because currently I've left in all the Windows & Linux code.

I'll see if I can remove all the non-macOS code without breaking things.

latenitefilms:
@asmagill - I've successfully removed a lot of the other platform code, so hopefully it now looks a little less scary. I'm still completely open to removing ManyMouse all together - I'm just not sure how.

asmagill:
Was just looking at the mac specific code... it works by hijacking the run-loop temporarily, so that the HIDManager will callback as it enumerates the devices identified matching the specified criteria (i.e. "mouse"), then once they've all been enumerated, allows program flow to continue on...

And I thought using the iterator functions of IOKit was annoying... Oof!

But after half an hour of googling, and another reading through the HIDManager header,I can't find an easier way to do it.

In their case it makes a kind of sense because they want to attach a callback to queue events from each mouse separately... for our purposes here, it just seems... inelegant... and annoying that there doesn't seem to be an obvious easier way, especially since we're just releasing the event callback almost immediately with the call to `quit`.

Ok, add a Notes section to the docs for `hs.mouse.count` and add a line that says something along the lines of "This function leverages code from ManyMouse which can be found at http://...the-appropriate-url..." and I'm happy with it.

@cmsj, your thoughts?

asmagill:
It does give me an idea for something that could act as a watcher and tell when keyboards or mice are added or removed (really any HID device -- is "device" really necessary here? just sounds weird without it, though...), even when they aren't USB devices, but... that's for another day, probably in v2.

cmsj:
I'm fine with this as-is. Couple of thoughts:

 1. Is the "Apple Internal Keyboard / Trackpad" a fairly stable string?
 2. Would there be any value in also being able to return a list of the attached mice names?

PR #2153, opened 2019-08-23
[READY] Changed CFBundleVersion instances to CFBundleShortVersionString

### LINKED ISSUE #2152: Confusion now that CFBundleVersion is an incrementing build number
In the current public version of Hammerspoon `CFBundleVersion` is the same as the `CFBundleShortVersionString`, however in a [recent commit](https://github.com/Hammerspoon/hammerspoon/commit/4447448a406107e4a97f3e67aa0769ae7a45e185) this behaviour has now changed, and the `CFBundleVersion` is an automatically increasing build number.

This poses two interesting issues:

1) Using `hs.processInfo.version` returns the `CFBundleVersion`, so you get the build number rather than the expected version information. This can be easily fixed by modifying the `push_hammerAppInfo` function in `MJLua.m`. However maybe this build information should also be exposed?

2) The release scripts used to require the version as a parameter for protection, however now they require the build number, which is a bit confusing. This can again be fixed by modifying the release script.

However, when investigating this - I can across this:

https://stackoverflow.com/questions/19726988/what-values-should-i-use-for-cfbundleversion-and-cfbundleshortversionstring

...which made me wonder if rather than just having `CFBundleVersion` as a single number that increments each time, should it actually also contain the version information? For example, rather than `5011`, should it be `0.9.76.5011`? Thoughts @cmsj ?

latenitefilms:
- `hs.processInfo.version` now refers to `CFBundleShortVersionString`
- `hs.processInfo.build` now refers to `CFBundleVersion `
- Updated release script
- Closes #2152

### DIFF
--- a/Hammerspoon/MJLua.m
+++ b/Hammerspoon/MJLua.m
@@ -186,7 +186,8 @@ static int core_reload(lua_State* L) {
 static int push_hammerAppInfo(lua_State* L) {
     LuaSkin *skin = [LuaSkin shared];
     NSDictionary *appInfo = @{
-                              @"version": [[[NSBundle mainBundle] infoDictionary] objectForKey:@"CFBundleVersion"],
+                              @"version": [[[NSBundle mainBundle] infoDictionary] objectForKey:@"CFBundleShortVersionString"],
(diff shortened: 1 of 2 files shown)


cmsj:
Thanks!

PR #2155, opened 2019-08-25
[READY] skip unknown warnings during xcode10 build

asmagill:
@cmsj, with your recent changes to better prepare for XCode 11, compiling under XCode 10 fails with an error about an unknown warning, specifically `-Wsizeof-pointer-div`...

This fix allows it to compile without error under XCode 10... since you're listing `-Wsizeof-pointer-div` as something to ignore anyways, I don't think these additions should be an issue with the master branch and the future with XCode 11, but I defer to your judgement.

### DIFF
--- a/LuaSkin/LuaSkin/Skin.m
+++ b/LuaSkin/LuaSkin/Skin.m
@@ -277,6 +277,7 @@ NSString *specMaskToString(int spec) {
 
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wconstant-conversion"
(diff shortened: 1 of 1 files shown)


asmagill:
@cmsj, objections? I'd like to land this if possible so I don't have to keep making changes to `Skin.m` that I have to make sure don't end up in other pull requests!

latenitefilms:
FWIW: Looks good to me! With these changes, everything still builds correctly on XCode 10 & 11 (Beta).

cmsj:
Thanks @asmagill - I had no idea that option existed :)

PR #2156, opened 2019-08-25
remove check for docs.json file for IPC

asmagill:
The latest incarnation of IPC was originally a third party module which tried to load its own documentation, if the `docs.json` file existed in its folder. This check is no longer necessary since the module is now part of core and its documentation is part of the built in provided documentation.


asmagill:
conflicts with #2159; will redo pull at later date

PR #2159, opened 2019-08-29
[READY] hs.doc speedify

### LINKED ISSUE #2145: Performance of loading docstrings
I've noticed for a while that reloading HS causes a noticeable beachball and prompted by someone else asking on IRC, I finally profiled it.

I'm pretty sure that it's entirely down to hs.json.decode time, which, without a config, can only mean the docstrings being loaded (5.3MB of JSON).

@asmagill fancy spitballing some ideas of how we could optimise this?

asmagill:
Speeds up `hs.doc` and Hammerspoon (re)load time by moving heavy lifting of reading and parsing json doc files to Objective-C.

* No longer preloads documentation files when they are registered -- they are loaded the first time a help function is invoked after the file is registered.
* No longer preloads all installed spoon documentation -- onlyactually  loaded spoon docs will be registered unless you invoke `hs.doc.preloadSpoonDocs` first.
* No longer checks to see if spoon doc exists, attempting to generate it if it doesn't.
* Removes some unnecessary functions (primarily `hs.doc.spoonsupport`)
* Creates `objectWrapper` userdata so `hs.doc.hsdocs` can access raw json data faster

Needs testing, but I think this is probably pretty close to being ready.

Should close #2145


asmagill:
@cmsj not sure how to tell stickler to ignore specific unused variables... I suppose I could change the watchable callback to use `...` instead and I think the `module.help` function's `__tostring` method could have self replaced by `_` since I think I remember that being ignored by default, but... personally I prefer seeing that I've prepared the correct number of arguments and like the explicit `self` to make it obvious this is expected to be a method.

Thoughts or suggestions? I'll change them if you prefer.

latenitefilms:
@asmagill You can suffix with `-- luacheck: ignore` to tell @stickler-ci to ignore a line.

latenitefilms:
FWIW, everything seems to build and work fine here! Woohoo!

Let me know if there's anything specific you'd like me to test.

asmagill:
@latenitefilms nothing specific, I just want the people who do custom builds to hammer on it a little (use `help...` and the web docs (click on the `i` icon in the default toolbar for the console and browse a little) even if you don't usually...

I'm pretty confident in it, but it is a complete departure from the previous approach and want to make sure I haven't done anything which has unexpected results in weird situations.

And if you're willing and have a little time, poke around the internals a little bit... for example, to see the objectWrapper up close (I do have some ideas on how to integrate it into LuaSkin so it can easily be used elsewhere that I may try to get to next week or so), you can get at the two main dictionaries used in `internal.m` with something like the following (this example shows that the objects do in fact change after issuing the `help` command):

Try this immediately after a reload:

~~~lua
docTree = getmetatable(hs.doc)._documentationTreeObject()
registeredFiles = getmetatable(hs.doc)._registeredFilesObject()

print("pairs(docTree)")
for k,v in pairs(docTree) do print(k,v) end

print("\npairs(docTree.spoon)")
for k,v in pairs(docTree.spoon) do print(k,v) end

print("\npairs(registeredFiles)")
for k,v in pairs(registeredFiles) do print(string.format("%s %s %s", k,v.spoon, v.json)) end
~~~

Then type `help` in the console

Try the above again and see the differences... `inspect` is of limited use since it sees userdata as  userdata, rather then as a table, but you can use `pairs` on anything that shows up as a dictionary type, and `ipairs` on anything that is an array type (look closely at the userdata types printed the second time you run the code above. Key names can be used as you'd use them for tables (see `docTree.spoon` above, which could just as easily have been `docTree["spoon"]`) and the array types can be accessed like `object[<integer>]`.

asmagill:
And pending any bugs discovered, I think I'm done, at least with the `hs.doc` portion... I found a situation where the web browser toolbar wasn't updating the back arrow every time, but I'm having problems replicating it in a controlled way, so I'll poke some more this weekend, but as that code hasn't changed in this pull, landing this if I don't figure it out quickly doesn't change anything that wasn't already there in the previous releases.

asmagill:
@cmsj, need some XCode help...

I'm trying to move the objectWrapper stuff from `hs.doc` into `LuaSkin`.  You can see a proof of concept that this works at https://github.com/asmagill/hammerspoon/tree/doc_LSow which creates `hs.luaskin`, which early in `hs/_coresetup/init.lua` is used to overwrite `ls` (which is created in `luaskin.lua`, loaded in LuaSkin's `createLuaState`) and loads the lua side methods to accompany the changes to `Skin.m` and `Skin.h`.

This is a horrible approach because it means LuaSkin now *requires* Hammerspoon for full functionality.


What I would *prefer* to do can be found at https://github.com/asmagill/hammerspoon/tree/doc_LSowInLS, which tries to add an `objectWrapper` target to the LuaSkin project... if you look closely my changes to `luaskin.lua` and the luaopen function of `luaskin.m` you'll see that it's location agnostic, as long as the the library name is corrected in `luaskin.lua` and the `so` library loaded is in the same directory as `luaskin.lua`... but, when I build this second branch, I get the following in `build/Debug.log`:

~~~
Build system information
error: Cycle in dependencies between targets 'Hammerspoon' and 'LuaSkin'; building could produce unreliable results.
Cycle path: Hammerspoon → LuaSkin → Hammerspoon
Cycle details:
→ That command depends on command in Target 'Hammerspoon': script phase “[CP] Embed Pods Frameworks”
→ Target 'Hammerspoon' has target dependency on Target 'LuaSkin'
→ Target 'LuaSkin' has target dependency on Target 'objectWrapper'
~~~

If I take `objectWrapper` out of LuaSkin's dependancies, it's never built and Hammerspoon won't load.

Thoughts? Ideas? Suggestions? I'm about the end of my familiarity with XCode...

latenitefilms:
This may or may not be helpful, but I've seen this error before, and it fixed itself after I did a "clean".

cmsj:
Agree with @latenitefilms - something about the new build system generates a weird cyclical dependency after the first build. I don’t understand what it’s doing, but cleaning the build folder usually helps 😕

asmagill:
"Clean Build Folder" under the Project menu? Didn't seem to help.

While I think the error message itself is more unclear then it should be, I can see why there might be a cycle... the objectWrapper library target *does* require the LuaSkin framework (it uses LuaSkin in the methods being added to lua with the shared library), but the LuaSkin target requires the objectWrapper library in its final release bundle, though not to *build* the framework (I'm probably misusing at least one or more words in that statement)

I'm not sure what the best approach here is... if I were doing this old-school with Makefiles and the command line, I'd probably arrange things so that the objectWrapper shared library isn't built until after the LuaSkin framework was compiled and then probably use a script to move the objectWrapper library into its final location inside the framework. I'll try and work on this some more tonight or tomorrow, just figured it was worth putting it out there so multiple eyes might have a chance to look at it.

asmagill:
@cmsj, @latenitefilms -- For clarification, what I'm trying to do is have `luaskin.m`, which is currently in `LuaSkin/LuaSkin/` compiled as a dynamic library so `luaskin.lua` can load it. This objective-c code needs to be linked against LuaSkin because its methods require this, but also needs to be contained within the LuaSkin framework bundle like `luaskin.lua` is.

Not sure of the best way to do this; https://github.com/asmagill/hammerspoon/tree/doc_LSowInLS is just my first attempt patterned closely on the way we add module libraries to Hammerspoon, but I don't think that's going to work.

As always, any ideas are appreciated... I'm going to let this ruminate in my mind for a while and see what other possibilities I can come up with...

asmagill:
@cmsj, had an idea that panned out, but not sure if it fits with your ideas of proper separation of functionality or not...

Instead of compiling `luaskin.m` separately, I added it to the LuaSkin framework as a dependency so it's compiled into LuaSkin itself (its no longer a separate `luaskin.so` or `libObjectWrapper.so` library to be loaded by `luaskin.lua`). `luaopen_luaskin_internal` is then invoked by LuaSkin's `createLuaState` *before* it tries to load the `luaskin.lua` so that the metatable and constructor are already part of lua before `luaskin.lua` extends them.

The only public function of `luaskin.m` is `luaopen_luaskin_internal` itself, and this is declared in `Skin.h`, not `LuaSkin.h`, so it shouldn't be accidentally visible elsewhere.

You can see the changes to `Skin.m`, `Skin.h`, `luaskin.lua`, and `luaskin.m` at https://github.com/asmagill/hammerspoon/tree/doc_LSownLS2/LuaSkin/LuaSkin to review and see if you have any objections.

I'm not 100% happy with this -- it makes changes to `luaskin.m` require a complete rebuild of Hammerspoon (or at least LuaSkin, if it can be built independently and copied into an existing `Hammerspoon` app bundle? never tried before...) and it makes it a little less obvious, in terms of object files, that they are separate... but in a sense they aren't since this *is* an extension to LuaSkin...

If you don't mind the approach I've taken, I'll roll this pull (#2159) forward to match the branch I link to above and we can go ahead and land this whenever -- while the wrapper doesn't currently support modifying mutable objects, the infrastructure is in place and isn't needed for `hs.doc`; and since nothing else uses the wrapper yet, this lack isn't a show stopper. (And I may have time tonight or tomorrow to add it anyways, in any case, just not sure yet.)

latenitefilms:
FWIW: This is all definitely above my pay-grade, so will leave with you smarter people to solve.

latenitefilms:
FYI: This is the build error I constantly see - which "cleaning" fixes:

Screen Shot 2019-09-07 at 10 55 52 am

cmsj:
@latenitefilms that seems to be a weird side effect of xcode's new build system. I get it too, have done since we switched.

@asmagill I'm fine with luaskin.m being compiled into LuaSkin, and I'm happy for you to merge this whenever you're ready (which the current subject suggests you are, but I thought I'd check before clicking merge :)

asmagill:
Yes, this is ready. I can merge it in a few hours myself if you want.

cmsj:
Thanks @asmagill. I've hand-merged it since GitHub didn't think it could do it.

PR #2168, opened 2019-09-05
Update init.lua

roeybiran:


cmsj:
The docs aren’t actually wrong, this is a subtle pointer to a feature of Lua where you don’t need parentheses to call a function.

I’m not personally a fan, but I also can’t pretend it doesn’t exist 🙂

roeybiran:
Wow, I had no idea such a feature exists, thanks for the enlightenment!

Keep up the with the amazing work on HS! (Can’t imagine my life without it)

PR #2182, opened 2019-09-17
[READY] Added hs.json.write() & hs.json.read()

### LINKED ISSUE #2146: Add ability to read/write JSON files to/from file at a specific path
As discussed in #2145

latenitefilms:
- Added `hs.json.write()` & `hs.json.read()`
- Closes #2146

### DIFF
--- a/extensions/json/internal.m
+++ b/extensions/json/internal.m
@@ -85,10 +85,121 @@ static int json_decode(lua_State* L) {
     }
 }
 
+/// hs.json.write(data, path, [prettyprint], [replace]) -> boolean
+/// Function
+/// Encodes a table as JSON to a file
+///
+/// Parameters:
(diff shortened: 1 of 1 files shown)


latenitefilms:
This is ready for your review @cmsj & @asmagill .

cmsj:
I hate to be nit-picky, but I always think that code of the form:

```
if (something) {
  // Long section of code
} else {
  // Handle error and return
}
```

looks better (mainly because of indenting) if you do:

```
if (!something) {
  // Handle error and return
}
// Long section of code
```

latenitefilms:
@cmsj - No problemo! Fixed.

It would be great to put these kind of code formatting recommendations in a document somewhere. I still get confused as to the best way to format documentation, as there's so much inconsistency in the Hammerspoon extensions.

PR #2207, opened 2019-10-11
Updated to support Catalina Music.app

cuyler:
This a simple update that duplicates the iTunes extension with a Music extension. The app control functionality, via AppleScript works exactly the same.

I considered NOT using the generic 'Music' namespace, however that is technically what the app is called.

* For folks looking to just add music support: you can just grab this extension and manually add it to an existing install:

Go to: /Applications/Hammerspoon.app/Contents/Resources/extensions/hs
create a new folder/directory called "music" and drop this in and restart Hammerspoon.

You'll need to update iTunes references from 'hs.itunes' to 'hs.music' but it should otherwise work identically.

### DIFF
--- a/extensions/music/LICENSE
+++ b/extensions/music/LICENSE
@@ -1,0 +1,21 @@
+Released under MIT license.
+
+Copyright (c) 2014 Linell Bonnette
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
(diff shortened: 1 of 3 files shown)


cmsj:
Hmm, yeah, the generic `hs.music` is very unfortunate. I almost wonder if we should make `hs.itunes` detect which OS version it's on and change the name of the app it sends events to?

cuyler:
That's a good take, honestly. I use HS for keyboard controlling my music so I was desperate for a quick fix. Let me take another look and I'll remove and resubmit this PR if I can make it reasonable.

rsefer:
https://github.com/Hammerspoon/hammerspoon/blob/3fe9fb425505c76a1fb98e7a047b7f0983494862/extensions/itunes/init.lua#L26-L35

```lua
local applicationName = 'iTunes'
if hs.host.operatingSystemVersion().major >= 10 and hs.host.operatingSystemVersion().minor >=15 then
  applicationName = 'Music'
end
if local _cmd = 'tell application "' .. applicationName .. '" to ' .. cmd
```

cmsj:
I'm going to close this because I just merged a PR that detects the macOS version and uses `iTunes` or `Music` appropriately. Thanks very much though!

PR #2212, opened 2019-10-20
[NEED HELP] Disabled allowsVibrancy in hs.chooser

### LINKED ISSUE #2180: Chooser choice images look too dark or too dark
With the below sample code I'm getting the darked/lightened icons in the screenshots. I'm using Hammerspoon 0.9.75 on macOS 10.14.6 (18G87). It looks like a blending mode is causing this because with multiple results moving the selection onto an item will emphasise the effect.

```lua
ch = hs.chooser.new(function(x) end)
ch:choices(function()
    local app = hs.application.find('Google Chrome')
    return {{text = 'Test', image = hs.image.imageFromAppBundle(app:bundleID())}}
end)
--ch:bgDark(true)
ch:show()
```

image

image

I also tried saving the resulting image out to a file, in case the image from the app bundle was behaving strangely but that looks fine.

```lua
local app = hs.application.find('Google Chrome')
hs.image.imageFromAppBundle(app:bundleID()):saveToFile('test.png')
```

image

latenitefilms:
- Set all the `allowsVibrancy` values to `NO`, to avoid the issue
described in issue #2180.
- Closes #2180


cmsj:
@latenitefilms presumably this means the chooser window won't have the nice blurry transparency at all? I wonder if we can just disable vibrancy for the image.

latenitefilms:
@cmsj - To be honest, I didn't notice any major difference, but maybe that's because I'm using dark mode? I feel like the "nice blurry transparency" disappeared in Mojave, but I could be mistaken? Will try and have another look at this over the next day or so.

latenitefilms:
Sorry, I tried playing around with this, but I have absolutely no idea how to disable `allowsVibrancy` for only the `NSImageView`.

dsdshcym:
I have the same issue as #2180.

I'm not familiar with mac app development (cannot build hammerspoon on my local). But I think maybe setting `allowVibrancy` to `NO` only in `HSChooserCell.m` can fix the issue and keep the window blurry at the same time?

PR #2227, opened 2019-10-31
[READY] Added an optional callback function to hs.image.imageFromURL

### LINKED ISSUE #2224: Add a non-blocking variant of hs.image.imageFromURL
My use case in particular falls within the context of `hs.chooser` choices, each choice item needing an icon image from an external URL as shown in the gif below.

test

Due to the blocking nature of `hs.image.imageFromURL`, using it blocks not only the chooser modal itself but the entire application in general, which isn't a great experience. For example attempting to load http://deelay.me/20000/https://i.imgur.com/2lrs7zF.jpg (simulated delay of 20s) blocks the whole app for the full 20 seconds.

It'd be awesome if there were a way to fetch images from URLs without blocking the main application thread.

latenitefilms:
- Closes #2224


cmsj:
So, this is technically still blocking, because although it's using `dispatch_async()`, it's specifying the main queue. I'll rewrite it shortly to do the image fetching on a background thread so it doesn't block the main thread at all.

latenitefilms:
@cmsj - Ah, interesting! I copied the code from [`hs.application:getMenuItems()`](https://github.com/Hammerspoon/hammerspoon/blob/master/extensions/application/internal.m#L1099), does that mean that code is also blocking, or am I misunderstanding something?

cmsj:
@latenitefilms nope, that code is also blocking, but that code *has* to run on the main thread because it touches the UI 🙂

latenitefilms:
Awesome - thanks for the explanation @cmsj !

cmsj:
Closing in favour of 74ddcea

PR #2225, opened 2019-10-31
[NEED HELP] Fixed potential crash in hs.application:getMenuItems()

latenitefilms:
- Closes #2220

PR #2233, opened 2019-11-08
[REVIEW] Fixed crash in hs.application:getMenus()

### LINKED ISSUE #2220: Crash - hs.application:getMenus()
Any idea why this would happen @cmsj ?

```
2019-10-26 10:57:24: ********
2019-10-26 10:57:24: 10:57:24 ERROR:   LuaSkin: hs.application:getMenus(): attempt to call a number value
stack traceback:
2019-10-26 10:57:24: ********
```

latenitefilms:
- Added protection in the callback for the unlikely event that `app` is
no longer valid
- Closes #2220


latenitefilms:
This is very similar to my failed #2225 attempt, and to be honest, I'm not really sure if `(!app)` is the same as `(app != nil)` in this context. I've pushed this code into CommandPost and made some tweaks to some Lua code using the function, so we'll see if that makes any difference.

latenitefilms:
Nope, this fix didn't work either - still seeing crashes in Fabric.io. Doh!

PR #2237, opened 2019-11-14
[REVIEW] adds hs.screen:setOrigin() method

treffynnon:
This adds a new method to the hs.screen API to allow a displays origin to be set  from lua.

An example of calling the code might be:

```lua
hs.screen.mainScreen():setOrigin(100,100)
```

Relates to the conversation https://freenode.logbot.info/hammerspoon/20191112#c2857975

### DIFF
--- a/Hammerspoon Tests/HSscreen.m
+++ b/Hammerspoon Tests/HSscreen.m
@@ -58,6 +58,10 @@
     RUN_LUA_TEST()
 }
 
+- (void)testSetOrigin {
+    RUN_LUA_TEST()
+}
+
 - (void)testFrames {
     RUN_LUA_TEST()
 }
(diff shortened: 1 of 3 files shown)


cmsj:
@treffynnon apologies, please ignore stickler for now, it's getting some stuff wrong here.

treffynnon:
@cmsj is this the kind of feature that will be merged into the project though? I am running my own local build of Hammerspoon at the moment so I can set the origin.

As a side note - my script uses `hs.screen.rotation()` too so I think it would be nice to create a buffer of screen updates and flush them all at once like the underlying Apple API. I don't know how to write this though so I just update the screen twice - once for rotation and once for origin.

cmsj:
@treffynnon yep, this is exactly the sort of thing we want in Hammerspoon. As for having a transactional API - it's something I'm trying to avoid, because we risk users leaving things in a weird, inconsistent state. It's uglier to have multiple stages of display configuration, but at least we are more in control of the outcome. Thanks for submitting this PR!

PR #2264, opened 2019-12-23
[REVIEW] Added hs.random()

### LINKED ISSUE #2260: Add more cryptographically secure random number generator
As discussed in https://github.com/Hammerspoon/Spoons/issues/145 - @psifertex writes:

> The current PasswordGenerator spoon uses Lua's math.rand seeded with os.time which is very cryptographically insecure. TL;DR randomness is hard and a plugin with the goal of generating passwords should either be very, very well vetted or (better!) just use an existing library that's known to have solid cryptographic properties.

One possible solution?

```objc
#include <stdlib.h>
// ...
// ...
int r = arc4random_uniform(74);
```

Another solution might be [SecRandomCopyBytes](https://developer.apple.com/documentation/security/1399291-secrandomcopybytes?language=objc)?

Although I do note that Lua 5.4 will have a "new implementation for `math.random`", but there's no release date yet - it's also not clear from the WIP documentation as to whether or not it'll be cryptographically secure.

latenitefilms:
- Closes #2260


cmsj:
oh, darn, I had been working on this a bit and just pushed my version, without realising you were also working on it!

latenitefilms:
No worries at all! I'll close this, however see my comment [here](https://github.com/Hammerspoon/hammerspoon/issues/2260#issuecomment-568578015).

PR #2265, opened 2019-12-25
[READY] Added hs.screenRecordingState()

### LINKED ISSUE #2208: Add ability to check the Screen Recording permissions
I haven't personally updated to Catalina yet - but it seems even if you're not "screen recording", you still might Screen Recording permissions to do certain things. It's probably worth adding something similar to [`hs.accessibilityState(shouldPrompt)`](http://www.hammerspoon.org/docs/hs.html#accessibilityState) to check for permission.

**Code Example:**
https://stackoverflow.com/questions/57957198/how-to-trigger-screen-recording-permission-system-modal-dialog-on-macos-catalina/58142253#58142253

**Examples of other applications documentation:**
https://www.macbartender.com/Screen-Recording-Permission/
https://matthewpalmer.net/vanilla/screen-recording-permission.html
https://support.displaylink.com/knowledgebase/articles/1916602-macos-catalina-10-15-screen-recording-permission

latenitefilms:
- Closes #2208


latenitefilms:

asmagill:
Confirmed works for me.

@cmsj, objections to landing?

asmagill:
Didn't someone post a link recently here or in the google groups to a library that had checks for a variety of "permissions" Hammerspoon could request? I'm wondering if there are any others in there we should consider while we're adding this... or does this close the loop on things we can currently do with Hammerspoon that requires special permissions?

Other than that thought, I think this should be ready.

cmsj:
I think we might have all of them after this 😁
+1 from me

asmagill:
K, I'm working on a build right now with this and a couple of others to test myself; assuming they all work as advertised (and I expect them to), I'll merge them later today.

latenitefilms:
This one?

https://github.com/MacPaw/PermissionsKit

Would be good to have a check for Full Disk Access.

asmagill:
Yes, that was it. I agree, full disk access is also useful to have a check for.

The others... well, I've dabbled in a module that tried to interpret the calendar info but never went anywhere with it (and as I recall, I was prompted about access when I did work on it). I'll bookmark the link in case it comes up again in the future.

But I do agree full disk access should be added. Will you have time? I can probably do it tonight if you prefer.

latenitefilms:
I think Full Disk Access should probably be done in a separate pull request.

I’m also not sure we can use same methods as PermissionsKit, as I don’t think it works correctly with a hardened runtime.

I’ll try and have a play later today.

asmagill:
I was going to dig into their source and see exactly what they're doing for the full disk check -- no reason to import a whole framework/library if we only need one part. But, I'll defer to you.

latenitefilms:
Agreed, sorry was just saying we’ll probably need to slightly revise the technique they’re using. Given how many radar’s exist because of dodgy permissions APIs I’d imagine we’re gonna have to be constantly playing catch up with each macOS update.

asmagill:
I'm ready to merge this if you're planning the full disk permissions to be a separate pull...

latenitefilms:
Merge away!

PR #2267, opened 2019-12-26
[READY] Added hs.window.list()

### LINKED ISSUE #2266: Add ability to get another applications window title/position/size without Accessibility Permissions
See: https://stackoverflow.com/questions/3039717/how-to-get-another-application-windows-title-position-and-size-in-mac-os-witho

latenitefilms:
- Closes #2266


cmsj:
What's the advantage of this over hs.window:allWindows() ?

asmagill:
@cmsj, this will allow getting a lot of information about potentially *all* windows (ok, the code does filter a few of the least useful out), even those `hs.window.allWindows` leaves out and does it without using the Accessibility interface so it's really fast.  Downside, it's strictly key-value data, not an `hs.window` object.  Running this and then finding the Hammerspoon console window, I see:

~~~
{
    kCGWindowAlpha = 0.80000001192093,
    kCGWindowBounds = {
      Height = 399.0,
      Width = 1280.0,
      X = 0.0,
      Y = 608.0
    },
    kCGWindowIsOnscreen = true,
    kCGWindowLayer = 0,
    kCGWindowMemoryUsage = 1248,
    kCGWindowName = "Hammerspoon Console",
    kCGWindowNumber = 3197,
    kCGWindowOwnerName = "Hammerspoon",
    kCGWindowOwnerPID = 47164,
    kCGWindowSharingState = 1,
    kCGWindowStoreType = 1
  }
~~~

I experimented with a similar function earlier but kind of dropped it when I found a different way to do what I wanted (I think it was when I was trying to find a way to differentiate between Hammerspoon drawings/canvases and Hammerspoon webviews before I figured out I could change their accessibility identifiers, but it was a while back, so I'm not positive).

Aside from maybe being a little more explicit in the details that this doesn't get you the window object, that you may or may *not* be able to match a given entry's window id to an `hs.window` object returned by `hs.allWindows`, and that some of the fields are used internally and may not have an obvious use or meaning for us (i.e. I don't want to field questions like "what does field XYZZY mean?") I don't have a problem with this being added.

Maybe also lay out in the details what the most obvious useful field names are (e.g. `kCGWindowBounds`, `kCGWindowName`, etc.)

asmagill:
Oh the other thing I think will need to be addressed is that IIRC bounds are given in the macOS mapping, not Hammerspoon’s inverted one, so either a helper needs to be written, the results need to be iterated through and adjusted, or details need to be made explicit in the docs to this effect. IMO preferably one of the former.

asmagill:
Maybe I spoke too soon on the inverted `y` value in bounds... somebody run some of their own tests to verify, but moving my console window around and rechecking gives results I'd expect from within Hammerspoon, so... I may be wrong on that particular recollection...

latenitefilms:
The main reason I added this was to get window position and title information without Accessibility permissions so that I could create Accessibility Permission overlays like "Remote Buddy":

Screen Shot 2019-12-26 at 11 53 15 am

latenitefilms:

cmsj:
Hand-merged into the new hs.window and updated `[LuaSkin shared]`. Thanks!

Issue #1245, opened 2017-03-01
question about multiple watchers


szymonkaliski:
My HS config is by now quite big, and I'm using the same watchers in multiple places, for example battery watcher in one file is displaying notifications about percentage, in another file I have another battery watcher instance which is automatically unmount usb drives, etc... 

My question is could that possibly impact my old macbook performance, and would it be better to just have one watcher per each watcher type that I'm using, or is having multiple instances fine?

asmagill:
The simple answer is yes, they do add up.  Which ones you should worry about most... that's a somewhat tougher question.

tl;dr - this is on my mind as well and I am working out some ideas that should hopefully help reduce the impact of this in the future.  At a minimum, I recommend taking a look at `hs.watchable` and see if that can help you with setting up single watchers that can trigger multiple callbacks.

### My thoughts and plans, FWIW

Consider what is happening when a watcher/notification/eventtap is triggered (this is very simplified and generic and I'm being loose with some of the terminology, so it may not be exact for any particular given watcher):
* first there is the "event" which causes the macOS to generate the "notification"
* Hammerspoon is notified/detects the change which causes a context shift to the Hammerspoon application (i.e. notable overhead)
* If the logic in the compiled portion of Hammerspoon believes that the callback should occur, the Lua state is invoked (more notable overhead)
* The callback returns, Hammerspoon shifts context back to the OS
* If multiple watchers for the same thing are registered, go back to step 2 for the next one

Some random thoughts of my own regarding this and some of the things I am working on related to it:

* This was part of the rationale behind `hs.watchable` which allows you to create a path that other code can subscribe to.  When the watcher/code which "owns" the path changes its value, it triggers updates to the subscribers through Lua metamethods.  Yes, we stay in Lua longer, which is arguably the slowest stage, but it means we only need one watcher (and thus context shift) for however many essentially identical watchers we want

* Push as much of the logic to the operating system or compiled code as possible:  For example, when using `hs.distributednotifications.new`, it is significantly more efficient to specify a `name` (and when possible `object`) parameter rather then use a series of if-then comparisons in the callback to determine if we actually care about the notification.

[...]

Issue #1116, opened 2016-11-30
KVO lite for Lua?


asmagill:
I've started using https://github.com/asmagill/hammerspoon_asm/tree/master/watchable so I can "communicate" between files in my setup and so that I can reduce the number of watchers that are really doing exactly the same thing (e.g. using `hs.caffeinate.watcher` to detect when the system is about to go to sleep).

Would this be useful in core or just confusing to people not familiar with KVO?

junkblocker:
This is a must! :+1:

PR #2272, opened 2019-12-30
[READY] Added hs.pasteboard.watcher

### LINKED ISSUE #2254: Add hs.pasteboard.watcher
@cmsj & @asmagill - I realise there's sadly no notification API for when the Pasteboard is updated, so you need to poll - however, I wonder rather than doing it in Lua-land via `hs.timer` would it be better if this polling was done in Objective-C land on a background thread? Is this even possible? Thoughts?

latenitefilms:
- ~~This is an initial attempt for `hs.pasteboard.watcher` - currently
each watcher uses it’s own unique `NSTimer`, which is probably not
ideal. We should probably just have a single global `NSTimer`, but this
probably complicates things if you have multiple watchers?~~
- Closes #2254


latenitefilms:
**TODO:** Investigate using Grand Central Dispatch as an alternative to using an NSTimer.

@asmagill & @cmsj - Feel free to jump in at any time with any comments, thoughts, ideas or suggestions!

latenitefilms:
I probably should also allow you to specify a specific pasteboard name to watch for.

latenitefilms:
@asmagill & @cmsj - This is now working. You can use this Lua code to test:

```lua
generalPBWatcher = hs.pasteboard.watcher.new(function(v) print(string.format("General Pasteboard Contents: %s", v)) end)
specialPBWatcher = hs.pasteboard.watcher.new(function(v) print(string.format("Special Pasteboard Contents: %s", v)) end, "special")
hs.pasteboard.writeObjects("This is on the general pasteboard.")
hs.pasteboard.writeObjects("This is on the special pasteboard.", "special")
```

I have no idea if using an `NSTimer` is better or worse than GCD, or other methods, so I refer to your genius on that one.

I'm also not really sure the best way to code it so that if a user creates multiple watchers for the same thing (i.e. multiple "general pasteboard" watchers or multiple watchers using the same "name"), that internally only a single `NSTimer` is used (as it seems a little crazy to have multiple `NSTimer`'s doing the exact same thing).

cmsj:
How about we use this as a testbed for creating hs.watchable? Then we’ll have one timer object that can be fanned out to multiple Lua callbacks in a generic way that we can then re-use on the road to 2.0? 😁

latenitefilms:
I assume you mean an Objective-C-based watchable framework (as discussed in #2098), as opposed to the existing Lua-based [`hs.watchable`](http://www.hammerspoon.org/docs/hs.watchable.html) (as discussed in #1245 and implemented in #1116)?

The only reason I wrote this in Objective-C and not Lua, is because I assume it would perform better as there would be less jumping back and forth between Obj-C & Lua-land?

Either way, I'm all for it! However, if building the "watchable framework" (for lack of a better term) is going to take a really long time (i.e. months), then I'd suggest pushing out this simpler version, and replacing it's internals later down the line when it's ready. As you say, lots of people, including spoon developers want a Pasteboard Watcher, so would be good to add in sooner rather than later.

I'm not sure how much help I'll be with this, but happy to help out in any way I can!

asmagill:
I won't have time to review this until later, but re multiple callbacks, my initial thought is to take a look at `hs.location`, if I recall correctly... in this module, we were faced with the fact that the macOS limits each application to one object for location queries, so the one object is stored and managed within the `init.lua` file -- assigning and removing callback functions is handled in lua by storing a unique identifier for each in a lua table, then the objective-c callback invokes a function in `init.lua` which checks these internal tables to see which callbacks to invoke. It sounds more complex then it really is...

I'll comment more when I have a chance to review your actual code later tonight.

latenitefilms:
@asmagill - I've had a look at `hs.location` - thanks for the tip! It would be fairly easy to add a pasteboard watcher if we're staying completely in Lua-land using a `hs.timer`, but I wanted to see if there's any benefit of keeping things all in Objective-C land, given how often we want to check the Pasteboard for changes, and also considering users might have multiple spoons all watching the Pasteboard.

I've just had another play and pushed some more tweaks and changes.

It's currently just using a single `NSTimer` which is posting a `NSNotificationCenter` notification, which then triggers all the individual watchers.

It seems to work pretty well, but I'm not really sure if this is a good way of doing things - it's more an experiment than anything else.

asmagill:
I’ll take a look at it tomorrow (with daughter tonight) but that *is* one of the purposes of the notification manager — for one section of code in an app to notify another of changes... sounds like its probably a more elegant solution then I was considering which involved basically replicating the hs.location idea in Objective-C with an NSDictionary and metamethod shenanigans...

latenitefilms:

extensions/pasteboard/watcher.m
```diff
@@ -0,0 +1,316 @@
...
+        [sharedPasteboardTimer invalidate];
+        sharedPasteboardTimer = nil;
+    }
+    
+    // Watcher is no longer running:
+    self.isRunning = NO;
+}
+@end
+
+HSPasteboardTimer *createHSPasteboardTimer(int callbackRef, NSString *pbName) {
```
asmagill, line 136: Point of style: most often a subclass will have an initiallizer that does this kind of initial setup, rather than a helper function like this... something like the following up in the HSPasteboardTimer implementation:

~~~objc
+ (instancetype)initForName:(NSString *)pbName withCallback:(int)callbackRef {
    self = [super init] ;
    if (self) {
        _fnRef = callbackRef ;
        _pbname = pbName ;
        // or, assuming you haven't made the properties readonly, and prefer
        // the style, you could instead use:
        //     self.fnRef = callbackRef ;
        //     self.pbname = pbName ;
    }
    return self ;
}
~~~

and then in `hs.pasteboard.watcher.new`, use:

~~~objc
HSPasteboardTimer *timer = [[HSPasteboardTimer alloc] initForName:pbName withCallback:callbackRef] ;
~~~

but this works as well, so... your call.

extensions/pasteboard/watcher.m
```diff
@@ -0,0 +1,316 @@
...
+    }
+    
+    // Remove the Metatable so future use of the variable in Lua won't think its valid
+    lua_pushnil(L);
+    lua_setmetatable(L, 1);
+
+    return 0;
+}
+
+static int meta_gc(lua_State* __unused L) {
```
asmagill, line 269: By not invalidating `sharedPasteboardTimer` and setting it to nil here in the `meta_gc` function, you do realize that your very first watcher is never released by ARC (its notification watcher and callback function ref *are* cleared by it's lua __gc function, though, so it isn't dangling within `lua`) even after a reload of Hammerspoon/CommandPost and will still be running every half second, even if no watchers are created after the reload.

Minor impact, but I'd be more comfortable with something like this here (untested, but I think this should be sufficient):

~~~objc
    if (sharedPasteboardTimer) {
        [sharedPasteboardTimer invalidate] ;
        sharedPasteboardTimer = nil ;
    }
~~~

asmagill:
I'm wondering if you should also add a module level function to adjust the period of the timer... I can imagine a case where something that automated a lot of copy/cut and pasting might be fast enough that a pasteboard watcher might miss some changes...

extensions/pasteboard/watcher.m
latenitefilms: Good to know! I'm happy to try and keep things in-line with the rest of the Hammerspoon code, so more than happy to go with whatever you feel fits best.

latenitefilms:
@asmagill - Unless I'm missing something, I don't think this should matter, as we're using `changeCount`, so even if a million pasteboard actions are triggered within that half a second, they should all be handled by the watcher?

extensions/pasteboard/watcher.m
latenitefilms: Good point, thanks!

latenitefilms:
...but yes, I do agree, having the watcher interval adjustable makes sense.

asmagill:
My primary use for the pasteboard is a clipboard recorder -- it keeps a record of the last 50 textual things I've copied to the clipboard so I can "reselect" the from a menu...

Yes, the callback for the watcher would get fired, but in theory more than one change could have happened to the clipboard before the next timer firing -- and only the latest content would be available to the callback function. Granted, you can say this about any interval, so it is a trade off between how quickly you think the clipboard will change versus how often you want the timer to fire and check.

Now .5 sec is reasonable (and maybe even a little fast) if you're only concerned with human speeds for selecting and copying (which, to be honest, I mostly am). But I can imagine cases where people might be wanting to capture what other applications that might use the clipboard (or a named pasteboard) when passing stuff back and forth, and that could definitely be more often than half a second.

I'd recommend making it set-able at the module level, but probably leave the default at .5. If someone finds that they are missing stuff, then they can change it.

latenitefilms:
Ah, I see what you mean. I forgot that I was actually returning anything in the callback - my original thought was that I'd just use it to trigger my own code that processes the pasteboard contents.

Maybe rather than returning a string of the last pasteboard item, it should return a table of pasteboard items?

asmagill:
Note quite...

Your timer fires a notification every .5 seconds.

Each notification watcher that is currently active is then triggered and checks to see if changecount for its specific pasteboard has changed... if so, it calls the lua callback with the contents of the pasteboard; otherwise it just exits the method.

But within that .5 seconds, the pasteboard could have theoretically changed multiple times -- changecount might not be just 1 more, but any number of times more. We only have access to the contents at the time the notification is acted on, not what it might have contained .1 or .2 (etc.) seconds ago.

latenitefilms:
Ahhhhhh... now I understand. Apologies - was trying to multitask, and clearly not very good at it! Is there any downside of checking the pasteboard `changeCount` more regularly? If it's done in a timer it won't block any Lua code, right?

asmagill:
It does run on the main application loop, so when the NSTimer fires, there is an event queued for the main application loop; when the event is handled, `sharedPasteboardNotification` is posted to the notification center, which then queues events on the main application loop for every registered observer for that notification, so... there is *some* overhead that could impact other queued tasks, including Lua code.

But unlike handling the polling and dispatch in `lua` itself (as was discussed earlier), the impact is very slight, especially when there is actually no change to report.

Like all things, it's a trade off... and NSTimers are also bad in terms of energy efficiency, so... I'd probably leave it at .5, or maybe move it to 0.25 (but no lower) for the default and then add a function allowing the user to change it if the find a need to.

latenitefilms:
Great, thanks! Will try and make these changes later tonight for your review - however, if you want, feel free to make the changes yourself and merge in.

latenitefilms:
@asmagill - Your requested changes have now been made. I believe this is now good to go.

asmagill:
Doc header still says poll interval defaults to .5; docs for actual interval getter/setter says .25. Other than making the docs consistent, I'd say this one is ready.

asmagill:
FWIW, `hs.pasteboard.watcher.interval()` reports .25, so fix the header and this is ready as well.

latenitefilms:
Fixed! Thanks @asmagill !

PR #2279, opened 2020-01-05
[READY] Added hs.relaunch()

### LINKED ISSUE #2278: Add function to relaunch Hammerspoon - hs.relaunch()
I propose that `hs.restart()` should restart the Hammerspoon app.

I tried doing this with `hs.task()`, based on what I read [here](https://vgable.com/blog/2008/10/05/restarting-your-cocoa-application/), but couldn't get it to work, so I believe to achieve this, we'll need a little helper app as discussed [here](https://stackoverflow.com/a/27497031/6925202).

latenitefilms:
- Closes #2278


cmsj:
Does this work reliably? I think I've mostly seen people do this with helper binaries.

latenitefilms:
Yep, seems to work fine on Mojave & Catalina. @asmagill suggested this method in #2278.

asmagill:
I've used this function since the Mjolnir days when I decided doing the double `dlclose` that macOS apps require to unload dynamic libraries wasn't a safe way for me to quickly reload modules I was developing and testing... I've bound it to `cmd-alt-shift-ctrl r` so a complete reload of Hammerspoon is always only a keystroke away.

The first version didn't use the application path, and that was problematic because it sometimes loaded an alternate binary if one existed in the XCode temp folders... once switching to `hs.processInfo.bundlePath`, it's worked flawlessly for me.

cmsj:
That's good enough for me, thanks!

PR #2287, opened 2020-01-11
Show images from chooser choices with their original colors

### LINKED ISSUE #2180: Chooser choice images look too dark or too dark

dsdshcym:
- fixes #2180
- inspired by #2212
- by only disabling vibrancy for Cell and TableView, we can keep the blurry effect

after:
image

(notice the blurry effect for the `Create pull request` button behind the chooser)

### DIFF
--- a/extensions/chooser/HSChooserCell.m
+++ b/extensions/chooser/HSChooserCell.m
@@ -11,7 +11,7 @@
 @implementation HSChooserCell
 
 - (BOOL)allowsVibrancy {
-    return YES;
+    return NO;
 }
 
 @end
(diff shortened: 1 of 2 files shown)


cmsj:
Awesome, thanks!

PR #2293, opened 2020-01-16
[READY] Added hs.websocket

### LINKED ISSUE #2056: How to use websocket schemes ws:// or wss://
I have an application that can be controlled via websocket protocol. How can I connect to it and  exchange messages? Are there are any examples?

### LINKED ISSUE #2228: Add documentation for hs.http.websocket
There is ample documentation for `hs.httpserver.websocket`, but there is no documentation for `hs.http.websocket` even though it exists as a perfectly usable websocket client implementation.

### LINKED ISSUE #2229: hs.http.websocket needs onopen and onclose callbacks
hs.http.websocket has a callback for received messages, but it needs callbacks for open and close events too.

### LINKED ISSUE #2230: hs.http.websocket crashes hammerspoon if send is called too soon
the following code will cause hammerspoon to crash

```lua
ws = hs.http.websocket('ws://server', callback)
ws:send('hello')
```

Workaround:

```lua
ws = hs.http.websocket('ws://server', callback)
hs.timer.doAfter(2, function() ws:send('hello') end)
```

latenitefilms:
- `hs.http.websocket` has been removed and replaced with `hs.websocket` which uses SocketRocket instead of PocketSocket.
- Closes #2229
- Closes #2056
- Closes #2230
- Closes #2228


latenitefilms:
@cmsj & @asmagill - Do you want to do a quick idiot-check of this before merging?

latenitefilms:
@asmagill - When you get a chance, I'd love your review and advice on this one. I've basically just moved the undocumented `hs.http.websocket` into `hs.websocket`, and switched from using PocketSocket to SocketRocket.

latenitefilms:
Here's some test code that seems to work:

```lua
wsserver = hs.httpserver.new()
	:setCallback(function(requestType, path, headers, raw)
		print(string.format("requestType: %s", requestType))
		print(string.format("path: %s", path))
		print(string.format("headers: %s", hs.inspect(headers)))
		print(string.format("raw: %s", raw))
	end)
	:setPort(1234)
	:websocket("/ws", function(message)
		print(string.format("server message: %s", message))
	end)
	:start()

ws = hs.websocket.new("ws://localhost:1234/ws", function(message)
	print(string.format("client message: %s", message))
end)

hs.timer.doAfter(1, function()
	ws:send("sending from client to server")
	wsserver:send("sending from server to client")
end)
```

latenitefilms:
@cmsj & @asmagill - I'd love to merge this into Hammerspoon sooner rather than later if that's possible?

asmagill:
Not sure if I have anything to test this against, but I'll give it some thought today (FWIW, I once tried to get `hs.http.websocket` to work with Kodi, but never got it to work -- at the time, I assumed that it was an issue with Kodi because people were reporting issues with their implementation on systems with both IPv4 and IPv6 enabled -- which I was using -- but maybe not; however I've since replaced that machine, so I need to think about what else I have that might use websockets).

On a more organizational/backwards-compatibility level of things, at least until we formally move to 2.0:

1) does `hs.http.websocket` point to this? If anyone is using the existing module, we don't want that to suddenly stop.
2) in terms of what Hammerspoon/Lua sees, is this function-wise and method-wise identical to `hs.http.websocket`? Or were methods/functions changed? Additions don't matter as much, it's breaking existing expectations that I'm concerned about.
3) If 2 is *No*, how hard would a wrapper be to write?

I'll let you know my thoughts after I look at this closer tonight

cmsj:
Would https://www.websocket.org/echo.html be useful for testing?

latenitefilms:
> 1. does hs.http.websocket point to this? If anyone is using the existing module, we don't want that to suddenly stop.

I've just restored `hs.http.websocket` to maintain backwards compatibility. I've also added in documentation for this module, as it never existed before.

`hs.websocket` has a slightly different callback format than `hs.http.websocket` (it returns a table instead of just a string), which could easily be fixed with a wrapper function, however, internally they seem to do things slightly differently, so for full backwards compatibility, it probably just make sense to leave the original code alone. Although, given that `hs.http.websocket` was undocumented, I doubt that many people are actually using it.

> Would https://www.websocket.org/echo.html be useful for testing?

Definitely. I'll try and write some proper tests for the extension later today.

latenitefilms:
In the meantime... you can use this test code to setup Hammerspoon as a websocket server to test `hs.websocket` as a client:

```lua
wsserver = hs.httpserver.new()
	:setCallback(function(requestType, path, headers, raw)
		print(string.format("requestType: %s", requestType))
		print(string.format("path: %s", path))
		print(string.format("headers: %s", hs.inspect(headers)))
		print(string.format("raw: %s", raw))
	end)
	:setPort(1234)
	:websocket("/ws", function(message)
		print(string.format("server message: %s", message))
	end)
	:start()

ws = hs.websocket.new("ws://localhost:1234/ws", function(event, message)
	print(string.format("client event: %s (%s)", event, message))
end)

hs.timer.doAfter(1, function()
	ws:send("sending from client to server")
	wsserver:send("sending from server to client")
end)
```

To test the echo you can use:

```lua
wssEcho = hs.websocket.new("wss://echo.websocket.org/", function(event, message)
	print(string.format("client event: %s (%s)", event, message))
end)
hs.timer.doAfter(3, function()
	wssEcho:send("sending from client to server")
end)
```

latenitefilms:
@cmsj - To be honest, I'm not exactly sure how I should do the tests for this, as they're asynchronous? Thoughts?

Also, any ideas why I'm seeing this crash during a reload?

```objc
Crashed Thread:        0  Dispatch queue: com.apple.main-thread

Exception Type:        EXC_BAD_INSTRUCTION (SIGILL)
Exception Codes:       0x0000000000000001, 0x0000000000000000
Exception Note:        EXC_CORPSE_NOTIFY

Termination Signal:    Illegal instruction: 4
Termination Reason:    Namespace SIGNAL, Code 0x4
Terminating Process:   exc handler [17717]

Application Specific Information:
Crashing on exception: ERROR: LuaSkin::pushLuaRef was passed a NOREF/REFNIL ref

Application Specific Backtrace 1:
0   CoreFoundation                      0x00007fff36975acd __exceptionPreprocess + 256
1   libobjc.A.dylib                     0x00007fff6107ba17 objc_exception_throw + 48
2   CoreFoundation                      0x00007fff369907e6 +[NSException raise:format:arguments:] + 98
3   Foundation                          0x00007fff38c31791 -[NSAssertionHandler handleFailureInMethod:object:file:lineNumber:description:] + 194
4   LuaSkin                             0x0000000102b25af8 -[LuaSkin pushLuaRef:ref:] + 563
5   internal.so                         0x0000000106cde8c8 __40-[HSWebSocketDelegate webSocketDidOpen:]_block_invoke + 116
6   libdispatch.dylib                   0x00007fff627fc5f8 _dispatch_call_block_and_release + 12
7   libdispatch.dylib                   0x00007fff627fd63d _dispatch_client_callout + 8
8   libdispatch.dylib                   0x00007fff6280868d _dispatch_main_queue_callback_4CF + 1135
9   CoreFoundation                      0x00007fff368bfa87 __CFRUNLOOP_IS_SERVICING_THE_MAIN_DISPATCH_QUEUE__ + 9
10  CoreFoundation                      0x00007fff368bf1b1 __CFRunLoopRun + 2289
11  CoreFoundation                      0x00007fff368be66e CFRunLoopRunSpecific + 455
12  HIToolbox                           0x00007fff35b1d1ab RunCurrentEventLoopInMode + 292
13  HIToolbox                           0x00007fff35b1cee5 ReceiveNextEventCommon + 603
14  HIToolbox                           0x00007fff35b1cc76 _BlockUntilNextEventMatchingListInModeWithFilter + 64
15  AppKit                              0x00007fff33eb577d _DPSNextEvent + 1135
16  AppKit                              0x00007fff33eb446b -[NSApplication(NSEvent) _nextEventMatchingEventMask:untilDate:inMode:dequeue:] + 1361
17  AppKit                              0x00007fff33eae588 -[NSApplication run] + 699
18  AppKit                              0x00007fff33e9dac8 NSApplicationMain + 777
19  libdyld.dylib                       0x00007fff6284a3d5 start + 1
20  ???                                 0x0000000000000001 0x0 + 1

Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   com.apple.AppKit              	0x00007fff342b9867 -[NSApplication _crashOnException:] + 109
1   com.apple.AppKit              	0x00007fff342b9748 -[NSApplication reportException:] + 916
2   org.latenitefilms.CommandPost 	0x0000000102895bf0 CLSNSApplicationReportException(objc_object*, objc_selector*, NSException*) + 69
3   com.apple.AppKit              	0x00007fff3437a0d1 uncaughtErrorProc + 145
4   com.apple.CoreFoundation      	0x00007fff369e9ab9 __handleUncaughtException + 775
5   libobjc.A.dylib               	0x00007fff6107dedd _objc_terminate() + 91
6   org.latenitefilms.CommandPost 	0x0000000102895954 CLSTerminateHandler() + 355
7   libc++abi.dylib               	0x00007fff5fad619e std::__terminate(void (*)()) + 8
8   libc++abi.dylib               	0x00007fff5fad6213 std::terminate() + 51
9   libdispatch.dylib             	0x00007fff627fd651 _dispatch_client_callout + 28
10  libdispatch.dylib             	0x00007fff6280868d _dispatch_main_queue_callback_4CF + 1135
11  com.apple.CoreFoundation      	0x00007fff368bfa87 __CFRUNLOOP_IS_SERVICING_THE_MAIN_DISPATCH_QUEUE__ + 9
12  com.apple.CoreFoundation      	0x00007fff368bf1b1 __CFRunLoopRun + 2289
13  com.apple.CoreFoundation      	0x00007fff368be66e CFRunLoopRunSpecific + 455
14  com.apple.HIToolbox           	0x00007fff35b1d1ab RunCurrentEventLoopInMode + 292
15  com.apple.HIToolbox           	0x00007fff35b1cee5 ReceiveNextEventCommon + 603
16  com.apple.HIToolbox           	0x00007fff35b1cc76 _BlockUntilNextEventMatchingListInModeWithFilter + 64
17  com.apple.AppKit              	0x00007fff33eb577d _DPSNextEvent + 1135
18  com.apple.AppKit              	0x00007fff33eb446b -[NSApplication(NSEvent) _nextEventMatchingEventMask:untilDate:inMode:dequeue:] + 1361
19  com.apple.AppKit              	0x00007fff33eae588 -[NSApplication run] + 699
20  com.apple.AppKit              	0x00007fff33e9dac8 NSApplicationMain + 777
21  libdyld.dylib                 	0x00007fff6284a3d5 start + 1
```

cmsj:
We can do async tests - take a look at how hs.socket does it, or hs.http :)

As for the crash, the important line is:

```Crashing on exception: ERROR: LuaSkin::pushLuaRef was passed a NOREF/REFNIL ref```

Something somewhere is pushing an invalid reference into Lua. Perhaps a callback ref from a deallocated object?

**Comment by latenitefilms** [comment]
@cmsj - Awesome, thanks! Will have a proper look at the `hs.socket` tests and see what I can come up with.

In regards to the crash, I THOUGHT the `if (self.fn == LUA_NOREF) {` line would protect it, but maybe I need to put it WITHIN the `dispatch_async(dispatch_get_main_queue(), ^{`?

```objc
- (void)webSocketDidOpen:(SRWebSocket *)webSocket;
 {
     if (self.fn == LUA_NOREF) {
         return;
     }
         _lua_stackguard_entry(skin.L);

         [skin pushLuaRef:refTable ref:self.fn];
         [skin pushNSObject:@"open"];

         [skin protectedCallAndError:@"hs.websocket callback" nargs:1 nresults:0];
         _lua_stackguard_exit(skin.L);
     });
 }
```

latenitefilms:
Any ideas @asmagill ?

asmagill:
You’re checking when you *dispatch* the block, not when the block is actually being *executed*... there is a small, bit not quite 0% chance of it being collected before the block is actually executed. Move the check into the block itself.

latenitefilms:
Legend, thanks heaps! It seems to crash fairly often when I do an automatic reload, so I'll see if this fixes it.

latenitefilms:
For anyone playing along at home - after using this code for a while now, I think it's pretty stable and reliable. I still need to write some proper tests for it, which I'm hoping to do in the next week or so, then I think we're all good to review and merge.

latenitefilms:
I'm still planning to come back and write some tests for this extension, but if you get bored @cmsj and want to get this extension merged sooner rather than later, feel free to use [the examples above](https://github.com/Hammerspoon/hammerspoon/pull/2293#issuecomment-577400063) as a starting point for the tests. No stress if not - I'll get to it ASAP.

FWIW - I've been using this extension daily in CommandPost (controlling a Loupedeck CT), and it's been working great.

Podfile
```diff
@@ -17,6 +17,7 @@ pod 'Crashlytics', '3.14.0'
 pod 'Fabric', '1.10.2'
 pod 'Sparkle', '1.22.0', :configurations => ['Release']
 pod 'MIKMIDI', '1.7.0'
+pod 'SocketRocket'
```
cmsj, line 20: Can we pin this pod to a specific version please? I don't mind what the version number is - whatever the latest version that works for you, but it'll save us pain in the future if we have a pin.

extensions/websocket/internal.m
```diff
@@ -0,0 +1,294 @@
...
+        _webSocket.delegate = self;
+    }
+    return self;
+}
+- (void)webSocket:(SRWebSocket *)webSocket didReceiveMessage:(id)message {
+    dispatch_async(dispatch_get_main_queue(), ^{
+        if (self.fn == LUA_NOREF) {
+            return;
+        }
+        LuaSkin *skin = [LuaSkin shared];
```
cmsj, line 38: All of the `[LuaSkin shared]` should be updated to the new `[LuaSkin sharedWithState:NULL]`.

Podfile
latenitefilms: Done.

extensions/websocket/internal.m
latenitefilms: Done

asmagill:
I think I'd prefer just one library for websockets... and then maybe a link or wrappers to maintain `hs.http.websocket` compatibility... Otherwise, we're duplicating functionality and replicating two Pods instead of just one. Unless we know for certain their is a fundamental reason to maintain both...

But other than that concern, I defer to you guys as to when this is ready for primetime...

latenitefilms:
I think the problem is that they both behave differently. However, as explained above, given that `hs.http.websocket` was undocumented, I doubt that many people are actually using it.

I'm happy for @asmagill and/or @cmsj - to adjust this pull request as they see fit and merge whenever they think it's ready.

FWIW - I've already merged `hs.websocket` into CommandPost, and it's working great communicating with the Loupedeck CT.

jeffmikels:
I don't mean to bump a dead thread, but has this been committed in any builds yet?

latenitefilms:
No sorry, not yet - however it's included in the [CommandPost](https://commandpost.io) betas if you want to grab just that extension from there.

asmagill:
@latenitefilms I know you mention a possible wrapper above to provide backwards compatibility, but you also mention some internals happening differently... how differently?

I guess my biggest ... annoyance, for lack of a better word ... would be the inclusion of multiple Pods that exist solely for one specific purpose each that (externally at least) do the same thing... that seems like a lot of mostly unused duplicated code that just creates bloat.

latenitefilms:
I think when I first looked into this I was (wrongly) assuming that PocketSocket was used for other things in `hs.http`, not just the undocumented `hs.http.websocket` functionality. Looking at the [original pull request](https://github.com/Hammerspoon/hammerspoon/commit/f55812384ebb83153580423248fab50e545f91f4) adding the feature, it seems that PocketSocket is only used for `hs.http.websocket`.

Given this, I think the best way forward is remove PocketSocket and `hs.http.websocket`, then make an undocumented Lua wrapper for `hs.http.websocket` for the off-chance that people are actually using the feature successfully. I don't think there's any point "exposing" the documentation as it hasn't been there for 4 years anyway, and as far as I can tell, the few people who have tried to use it, have been unsuccessful.

I'll have a play tonight, unless yourself or @cmsj just want to action. I'm not exactly sure how easy it is to remove a Pod - so will probably need @cmsj's help anyway.

latenitefilms:
@asmagill - As per your suggestion, I've removed `PocketSocket`, and created an undocumented wrapper for the legacy `hs.http.websocket` function.

@cmsj - I'm not sure how the best way to handle these changes in terms of updating all the Xcode projects pod references - so I might have to leave that to you to fix up? This branch loads and tests fine - but I can't merge the `master` back into this branch without conflicts.

asmagill:
I agree that the actual merging will probably require @cmsj to do so manually...

As to maintaining backwards compatibility with the existing version, I cobbled the following together from the previous issues/pulls regarding the original websocket additions... if you can get the wrapper to mimic it, I'd consider the wrapper tested enough... as you mentioned before, it wasn't even documented, so I sincerely doubt it was used by more than a handful of people, if that...

At any rate, my sample:

~~~lua
> server = hs.httpserver.new()
server:setCallback(function() return "hello from server", 200, {} end)
server:setPort(8888)
server:websocket("/ws", function(msg)
    return "server saw '" .. msg .. "' and sends it back!"
end):start()

sRepeater = hs.timer.doEvery(5,function() server:send(tostring(math.random())) end)

client = hs.http.websocket("ws://localhost:8888/ws", function(msg)
    print("client received '" .. msg .. "'")
end)

client received '0.087619876954705'
client received '0.33632046543062'

-- this has to be sent at least a second or so after `client` is created or it crashes
-- Hammerspoon... another reason to swap this version out completely as otherwise we'd
-- need to track the crash down...
> client:send("client says boo!")
hs.http.websocket: localhost (0x6060002e29a8)

client received 'server saw 'client says boo!' and sends it back!'
client received '0.40862569445744'
client received '0.46942672831938'
~~~

latenitefilms:
Ummm... it kinda works.

For some reason `client:send("client says boo!")` doesn't seem to be working as intended.

Screen Shot 2020-05-28 at 7 14 36 pm

latenitefilms:
Ummm, when I try and send a message from the client, it closes the server websocket connection.

```lua
> client:send("test")
hs.websocket: connected (0x606000a80228)
2020-05-28 19:44:10: 19:44:10    LuaSkin: (secondary thread): Closed websocket connection
```

asmagill:
I'm not at a point where I can try to merge this into my build, so I've just skimmed the files in this pull request, but the only `send` implementation I can find talks about sending a message *to* the client... (and as we see above, that does appear to be working fine in the wrapper as you get the random numbers)... with this new Pod, is sending *from* the client a different method that I've missed (or needs to be implemented)? Or is this a case where the terms "client" and "server" are interchangeable depending upon which side you're looking at it from?

Can you implement the *intention* of my bi-directional example directly without using the wrapper?

latenitefilms:
The “server” in your test code example is using `hs.httpserver:websocket(path, callback) -> object` which remains untouched by this pull request.

In terms of the “client” - I’ve tried replacing `hs.http.websocket` with `hs.websocket` so there’s no wrapper involved, but it has the same issue.

What’s strange is I’m sure this USED to work.

Maybe `hs.httpserver:websocket(path, callback) -> object` was also using the code in `hs.http.websocket` somehow?

Yes, merging this code into your own Hammerspoon build could be tricky, but if you wanted to have a test locally you could probably just manually download this branch from the GitHub website and build locally for testing purposes? Obviously no pressure of course - only if you want to and have time.

I’ll do some more digging tomorrow too.

latenitefilms:
I wonder if something broke in `hs.httpserver:websocket` outside of this pull request when all the co-routine stuff was implemented? I can’t spot anything obvious, but maybe worth having a quick check?

latenitefilms:
Looking on my phone, is this code triggering the right callback? Should it be wsCallback instead of just callback?

https://github.com/Hammerspoon/hammerspoon/pull/2307/files#diff-441d5de73cc02527c7d1f07b41c72282R81

latenitefilms:
Ok, something is broken in this branch, because if I run the following code in `master` it works:

Screen Shot 2020-05-30 at 11 57 40 am

asmagill:
when you say "between this branch and master", what changes or commits are you specifically referring to? Obviously something has changed in there that needs to be looked at closer. I'm hoping to try integrating these changes (or this pull itself, if I'm lucky) into my local build this weekend to see if I can add anything more.

latenitefilms:
To be honest, I'm not sure at this point. 

Thinking out loud:

- My test code on [23rd Jan](https://github.com/Hammerspoon/hammerspoon/pull/2293#issuecomment-577400063) was working fine.
- [5692bbe](https://github.com/Hammerspoon/hammerspoon/pull/2293/commits/5692bbe5a8d5d5f80c39fc36f9bc47c3a5ff2756): 12th of Feb I merged `master` back into this branch.
- [7e5eb26](https://github.com/Hammerspoon/hammerspoon/pull/2293/commits/7e5eb2620d18d0e31d32356da8c5e3d9c0ca0702): 12th of Feb I fixed potential crash during reloads, which shouldn't affect anything.
- [89659c8](https://github.com/Hammerspoon/hammerspoon/pull/2293/commits/89659c85e38fb5de9578cbf45ac03adc05d976b5): 23rd April I merged `master` back into this branch.
- [91ba7ef](https://github.com/Hammerspoon/hammerspoon/pull/2293/commits/91ba7efc3d66311202ac9201f4bb00e7c02db125): Added specific SocketRocket version to `Podfile` and made `hs.websocket` coroutine friendly, which again shouldn't break anything.
- [b3e3bdf](https://github.com/Hammerspoon/hammerspoon/pull/2293/commits/b3e3bdfc3d56f8b1265eef7397e7d0cd3427e4b3):  Removed PocketSocket Pod and `hs.http.websocket`. Added wrapper for `hs.http.websocket`, which now uses `hs.websocket`. I'm ASSUMING this is where things started to break.

Let me try and run the test again, without that last commit.

latenitefilms:
I ignored the last commit ([b3e3bdf](https://github.com/Hammerspoon/hammerspoon/pull/2293/commits/b3e3bdfc3d56f8b1265eef7397e7d0cd3427e4b3) - where I removed the `PocketSocket` Pod and made `hs.http.websocket` just a wrapper to `hs.websocket`) - but @asmagill's test code still doesn't work:

Screen Shot 2020-05-30 at 2 53 06 pm

It seems that whenever `hs.websocket` tries to send a message to `hs.httpserver`, the `hs.httpserver` instances closes its connection, whereas, if I use the legacy `hs.http.websocket` it works fine.

I wonder if internally `PocketSocket` and `SocketRocket` are using similar code, and there's some kind of conflict when they're communicating with one another?

I can only assume that that's the case, because this test works fine:

```lua
wssEcho = hs.websocket.new("wss://echo.websocket.org/", function(event, message)
	print(string.format("client event: %s (%s)", event, message))
end)
hs.timer.doAfter(3, function()
	wssEcho:send("sending from client to server")
end)
```

Maybe it's because `hs.websocket` is both acting as a client AND server (for lack of better terms)?

Long story short... as far as I can tell `hs.websocket` is working correctly. It's just that `hs.httpserver` doesn't like getting messages from it (whereas `hs.http.websocket` had no problems with this).

My thoughts are given `hs.websocket` is working fine with `wss://echo.websocket.org/` and also communicating successfully with my Loupedeck CT hardware - it should be fine to merge in, but maybe it just needs a note that you can't send messages to `hs.http.websocket` (which a user would never need to do anyway)?

latenitefilms:
@cmsj - I've added some tests, which all pass on my local machine in Xcode.

Any questions let me know!

asmagill:
Sorry, got distracted this weekend and never had a chance to test this myself... did you ever figure out the hs client sending to the hs server issue? Or are we going to leave it since its in the wrapper only?

latenitefilms:
No, `hs.httpserver` still disconnects when you try and send messages to it using `hs.websocket` - I have no idea why. However, `hs.websocket` seems to work correctly with every other websocket implementation I’ve tried connecting it to (echo test, web services, hardware devices like the Loupedeck CT, etc).

asmagill:
alright, I'll give it a shot after checking the other pulls I'm reviewing right now... this one, like axuielement, makes changes to the Hammerspoon project files so I have to merge them by hand... I'm waiting to do it last. (come to think of it, pasteboard watcher does too, but for some reason that merged fine without conflict... maybe because it's a single file added to an existing subproject as opposed to adding in a whole new subproject?)

Unless there is something I'm missing, I can't fathom why it *shouldn't* be possible for a local client to talk to a local server, so before we throw our hands up and call it a "know limitation", I want to take a shot at it myself.

latenitefilms:
Agreed - would definitely be great to solve the mystery before merging in. I’ll have another play myself today as well.

asmagill:
I added a `[LuaSkin logInfo...` line in the `hs.httpserver` `internal.m` file within the `- (void)didReceiveMessage:(NSString *)msg` method of `HSWebSocket`... it's never invoked. `didOpen` is invoked when the client is created (I bypassed the wrapper and created it with `hs.websocket` so I could see all of the messages it received, and it *does* receive the "open" message right after `didOpen` logs its message).

And `didClose` is invoked when I try to send something from the client to the server... looking at the core for `WebSocket.m` in CocoaHTTPServer, it looks like the the only way `didClose` could be invoked without `didReceiveMessage:` first is if CocoaHTTPServer considered the frame invalid... are we running with the latest version of the CocoaHTTPServer pod? I've never worked directly with Pods, how would I check and possibly update it in my setup?

I'm going to sleep on it and see if I have any brainstorms in the morning.

cmsj:
The top level Podfile encodes the versions we’re using of each pod. If you have Cocoapods installed, you can modify the Podfile and run “pod update”.

latenitefilms:
Yes, CocoaHTTPServer 2.3 seems to be the latest release, but it hasn't been updated since 7 Mar 2013, so that could very well be the issue (as the repo has a lot of issues in it).

latenitefilms:
[CocoaHTTPServer2](https://github.com/huangjimmy/CocoaHTTPServer2) seems to fix some websocket issues (continuation frame & 64bit length data support), however isn't compatible with the CocoaLumberjack we have installed, so not really sure of the best solve.

latenitefilms:
FWIW - I tried updating to CocoaLumberjack 3.6.1 and CocoaHTTPServer2 2.4.2, and whilst I no longer see the "closed" message, it still doesn't work for whatever reason:

Screen Shot 2020-06-03 at 9 48 17 pm

latenitefilms:
I finally worked out why my original tests were working fine, but no longer worked! Woohoo!

Originally I had:

```objc
NSString *message = [skin toNSObjectAtIndex:2];
```

...but we changed it to:

```objc
NSData *message = [skin toNSObjectAtIndex:2 withOptions: LS_NSLuaStringAsDataOnly];
```

If I change it back, the tests work correctly. Woohoo!

I've now changed it to:

```objc
id message = [skin toNSObjectAtIndex:2 withOptions:LS_NSPreserveLuaStringExactly];
```

...which also now works, and I believe solves the same issue that we were originally trying to solve anyway (i.e. if we send a string it'll pass an `NSString`, whereas if we send data it'll pass `NSData`)?

FWIW, I tried doing something like this:

```objc
id message = [skin toNSObjectAtIndex:2 withOptions:LS_NSPreserveLuaStringExactly];
NSData *data;

if ([message isKindOfClass:[NSString class]]) {
	data = [message dataUsingEncoding:NSUTF8StringEncoding];
} else {
	data = (NSData *) message;
}
```

...but it had the same issue, so it seems for whatever reason, `hs.httpserver` doesn't mind receiving a `NSString`, but it breaks if you send it `NSData` - whereas, all other websocket implementations don't seem to have a problem - for example, this code works fine in both cases:

```lua
wssEcho = hs.websocket.new("wss://echo.websocket.org/", function(event, message)
	print(string.format("client event: %s (%s)", event, message))
end)
hs.timer.doAfter(3, function()
	wssEcho:send("sending from client to server")
end)
```

Thoughts @asmagill and/or @cmsj ?

latenitefilms:
Ok, so it looks like [CocoaHTTPServer2](https://github.com/huangjimmy/CocoaHTTPServer2) added:

```objc
- (void)webSocket:(WebSocket *)ws didReceiveData:(NSData *)data;
```

...so maybe `hs.httpserver` should be updated to use CocoaHTTPServer2 and receive `NSData` instead of just `NSString`'s?

Either way, hopefully this now gives you the clues you need @asmagill to finally wrap up this pull request.

cmsj:
Thanks @latenitefilms and @asmagill for looking at this, I haven't had time to really dig into it at all and I'm trying to get the hs.uielement stuff back on track (which is currently breaking my brain quite significantly!)

latenitefilms:
@cmsj - No worries at all. I think this now all finally makes sense, and it's actually a limitation of `hs.httpserver`, where it was only originally designed to work with `NSString`'s, simply because that was a limitation of the `CocoaHTTPServer` framework.

Based on [this commit](https://github.com/huangjimmy/CocoaHTTPServer2/commit/55d4c7a0e8b6fb701cdb08b0bec011fbda9ad97e), it looks like `WS_OP_BINARY_FRAME` was simply being ignored by `CocoaHTTPServer` - so if you sent it binary data, it would just fail.

This has been solved in [CocoaHTTPServer2](https://github.com/huangjimmy/CocoaHTTPServer2).

So now the question is... do we want to actually update `hs.httpserver` with this new-found knowledge in this pull request, or just leave it for another issue?

I'll leave this up to @asmagill  & @cmsj to decide.

@cmsj - Let me know if there's anything I can do to help with `hs.uielement`!

cmsj:
I had a quick look at CocoaHTTPServer2 and it seems like it's a fork, rather than a newer version, so we might well be swapping one piece of abandonware for another.

All of these pods worry me a lot. Sooner or later one of them is going to break hard on a new macOS and we'll be stuck.

latenitefilms:
Correct, `CocoaHTTPServer2` is just a fork. There's [hundreds](https://github.com/robbiehanson/CocoaHTTPServer/network/members) of different `CocoaHTTPServer` forks it seems - but `CocoaHTTPServer2` seems to be the only one I can find that adds these websocket `NSData` additions. It hasn't been updated since **27 Jul 2016**.

The most active/recent `CocoaHTTPServer` fork seems to be [here](https://github.com/TechSmith/CocoaHTTPServer) (updated 6 days ago), but doesn't solve the websocket issue, so is not really helpful in this particular case.

I guess one option is we could fork `CocoaHTTPServer2` into the Hammerspoon organisation, and maintain it based on changes to other forks as required? Obviously less than ideal though as it means far more work.

Or we just ignore this particular problem, as clearly no one has ever run into it before, as `hs.websocket` will never really be used with `hs.httpserver` anyway?

asmagill:
With these new findings, I'm wondering if it doesn't make the most sense to just remove websocket support from `hs.httpserver` all together... I know it breaks backwards compatibility, but it seems it was never a fully compliant/stable/working solution to begin with.

I know I argued for backwards compatibility, and it's still a desirable goal, but it appears in this case, not a reasonable one.

@cmsj, are you concerned about the CocoaHTTPServer pod breaking for basic web server support as well at some point? Do we need to consider (in a different thread/pull) a replacement for `hs.httpserver` at some point as well?

latenitefilms:
I’d imagine there’s quite a few people using the HTTP Server, so I’d be tempted to just leave it as is for now, or attempt to patch it using some of these active forks?

Either way, I feel like this is a much bigger/different discussion than this pull request tries to solve, and the main objective at this point is trying to get `hs.websocket` stable and safe to merge into the `master`. Are you happy with the last pull request I did as a workaround, or should I undo it, and we should always just send `NSData` regardless of the fact that Hammerspoon’s HTTP Server can’t receive it?

I’m also happy to try and fix the underlying HTTP Server websocket issue if that’s something you’d like to try and solve before we merge this?

Thoughts?

asmagill:
When did we originally change it in an attempt to get it to support NSData directly? Was that the first step along your path to this pull? Or was it for another issue from someone else?

I'm inclined to agree and let this land as is, but if you were the reason we originally tried the change, I'm even more inclined 😜

My experience with forcing raw-data that isn't utf8 compliant into NSStrings is that is *mostly* works, unless there are embedded nulls (0 byte values), or you're doing actual string manipulation that relies on utf-8/16 compliance, and then all bets are off. But Apple doesn't *guarantee* that it will work, which is why I prefer to use NSData when possible if we're allowing binary data. But this is a reasonable work around, given the state of the Pods in use.

latenitefilms:
It was changed in [7bb9bbe](https://github.com/Hammerspoon/hammerspoon/pull/2293/commits/7bb9bbef12ea3d8d2fba2ffeb6b5aae99dfaf7ce). The note says:

> - `hs.websocket:send()` now uses `NSData`. Thanks @asmagill !

...so I think you suggested this in another pull request (I think maybe when we changed this in  `hs.sockets:write()`?).

Either way, I'm happy to leave it as:

```objc
NSData *message = [skin toNSObjectAtIndex:2 withOptions: LS_NSLuaStringAsDataOnly];
```

...and I can add a note saying it doesn't work with `hs.httpserver` because `hs.httpserver` doesn't support `NSData` messages, or we can keep it as:

```objc
id message = [skin toNSObjectAtIndex:2 withOptions:LS_NSPreserveLuaStringExactly];
```

...which I believe will mean it will send a `NSString` unless you supply a Lua string containing character byte sequences which cannot be converted to a proper UTF8 Unicode character.

Let me know which you'd prefer?

asmagill:
Just to confirm, this change is to the `hs.websocket:send` method, correct? I'd say if the websocket pod you're using can properly handle *either* an NSData or an NSString being sent to its `send` method, then use the 
~~~objc
id message = [skin toNSObjectAtIndex:2 withOptions:LS_NSPreserveLuaStringExactly];
~~~
 form as that will return NSData only when the data can't be interpreted as valid UTF8. We'll maintain a minimum of backwards compatibility and can decide if we need to rethink `hs.httpserver` entirely at another point if it becomes necessary.

latenitefilms:
Correct, this is for the `hs.websocket:send()` method, and yes, SocketRocket's send method supports both `NSString` and `NSData`.

Given this, I believe everything is now ready to merge! Woohoo!

cmsj:
@asmagill I’m pretty conflicted about this whole situation. My concern for CocoaHTTPServer is mostly that at some point the SDK will change and render it unusable, but that’s something we can worry about in the future (and probably lots of other people will suddenly care too).

It would be lovely if we could have a single HTTP client/server module that can do text serving and websocket serving and text clienting and websocket clienting, but that doesn’t seem to be reasonable at the moment, so I’m fine with splitting things down.

@latenitefilms does your hs.websocket just do client operations? Or can it do server stuff too?

cmsj:
(Also as a note to whoever ends up merging this, it will need a “pod update” with a newer version of Cocoapods than @latenitefilms has installed)

asmagill:
I've never worked on the pods before, so I'll leave merging this to you; I believe it's as ready as its going to get without further considering the CocoaHTTPServer issue.

It's a shame that CocoaHTTPServer isn't properly up to the task and isn't being maintained... it should be noted, though I don't think it's a show stopper, that implementing the websocket server in a different Pod means that the same port number can't be used for both a web site *and* have a websocket served by Hammerspoon. CocoaHTTPServer is making the decision based on the path requested as to which of its subclasses actually handles the incoming request; another Pod providing the server won't be able to attach to an already in use port number. But as Hammerspoon's web serving was never meant to be a full blown server solution, I don't see this as a problem, just something to document if/when we do implement the websocket server elsewhere and/or replace CocoaHTTPServer.

latenitefilms:
> @latenitefilms does your hs.websocket just do client operations? Or can it do server stuff too?

No, SocketRocket only acts as a client.

The SocketRocket documentation [mentions](https://github.com/facebookarchive/SocketRocket#websocket-server-implementation-recommendations):

> **WebSocket Server Implementation Recommendations**
> SocketRocket has been used with the following libraries:
> - [Tornado](https://github.com/tornadoweb/tornado)
> - Go's [WebSocket package](https://godoc.org/golang.org/x/net/websocket) or Gorilla's [version](http://www.gorillatoolkit.org/pkg/websocket).
> - [Autobahn](http://autobahn.ws/testsuite/) (using its fuzzing client).
> 
> The Tornado one is dirt simple and works like a charm. ([IPython notebook](http://ipython.org/ipython-doc/dev/interactive/htmlnotebook.html) uses it too). It's much easier to configure handlers and routes than in Autobahn/twisted.

However, I'm not sure if any of these are suitable for Hammerspoon/Objective-C.

> (Also as a note to whoever ends up merging this, it will need a “pod update” with a newer version of Cocoapods than @latenitefilms has installed)

Sorry - it seems updating cocoapods on Mojave is a bit of a hassle! I finally got the secret combination of Terminal commands working to update it successfully. I've updated this pull request.

In regards to `CocoaHTTPServer`, I'll have a play to see if I can "patch" it using the various forks that exist in another branch.

latenitefilms:
@cmsj - FYI: I've updated this branch based on your recent commit where you added `RUN_TWO_PART_LUA_TEST_WITH_TIMEOUT` into the test suite itself.

Unless anyone has any objections, I believe this is good to go.

latenitefilms:
@asmagill - I've just updated this branch as well, so fingers and toes crossed it makes merging easier for you.

PR #2302, opened 2020-02-09
Fixed doc comment line break

rsefer:
[hs.grid](http://www.hammerspoon.org/docs/hs.grid.html) has an erroneous code/line-break in the documentation for the hs.grid.set function: "...the window's current screen will be used"

image

I've never contributed to the documentation before, so it's likely I haven't done this right - my apologies!


rsefer:
Did run `make doc` and checked the generated html markup - looks good. I believe the failed travis build is unrelated.

image

cmsj:
Thanks!

PR #2307, opened 2020-02-19
[REVIEW] Fix (potentially) buggy dispatches in various modules

asmagill:
May change title later, as I suspect I'll add more as I do the changes for proper coroutine support; however, as that's going onto a separate branch for the purposes of testing before formally releasing it into the wild, I figured some of other bugs I'm finding along the way should probably be applied to master in the mean time...

Currently:

* `hs.webview.toolbar` checked for valid fn ref before dispatch, but not within, so could crash if callback disappeared (e.g. __gc)
* `hs.webview.reload` checked lua stack within dispatch for parameter *after* stack purged, so `validate` parameter was being ignored; in addition, the logic was inverted from the docs

### DIFF
--- a/extensions/application/internal.m
+++ b/extensions/application/internal.m
@@ -7,6 +7,8 @@
 #define get_app(L, idx) *((AXUIElementRef*)luaL_checkudata(L, idx, "hs.application"))
 #define nsobject_for_app(L, idx) [NSRunningApplication runningApplicationWithProcessIdentifier: pid_for_app(L, idx)]
 
+static NSMutableSet *backgroundCallbacks ;
+
 static pid_t pid_for_app(lua_State* L, int idx) {
     get_app(L, idx); // type-checking
     lua_getuservalue(L, idx);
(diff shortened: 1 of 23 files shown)


asmagill:
The failing tests appear to be in HSTestSocket, and that passes when I run it in XCode on my machine, so I'm going to ignore Travis (I usually do) until/unless someone can prove it's right and I'm just getting false positives.

asmagill:
#### Notes to self

Keeping a list here of things I'll go back and check/fix after first pass through for #2308 -- dispatches are relatively easy to fix, which is why I've been doing them in tandem.  What I'm listing here will take a little more thought...
 - - -
* The following are functions/methods that use a callback mechanism but don't store `fnRef` in a userdata, so no `__gc` to flag if `fnRef` becomes invalid before actual dispatch is executed. If the callback doesn't fire till after reload, causes crash that's hard to diagnose... need fix, maybe module level array that module level _gc clears out. 

~~~
hs.webview.datastore
    fetchRecords
    removeRecordsFor
    removeRecordsAfter
    
hs.application
	getMenuItems

hs.image
    imageFromURL
~~~

* Check all modules that use `performSelectorOnMainThread:` -- if `waitUntilDone:` is NO, they *all* make assumptions about `fnRef` remaining valid

latenitefilms:
`getMenuItems` is still very crashy for me, so might be worthwhile reviewing those crash logs at same time to see if there’s any other improvements you can do?

latenitefilms:
@cmsj @asmagill - I believe this code has already been merged now, so this pull request can be closed?

asmagill:
Yes, these fixes were also a part of #2308; I wasn't sure when we'd be applying the coroutine fixes, so I wanted this in case we decided to phase them in separately.

PR #2320, opened 2020-03-12
Use key name as alert message when binding hotkeys for Spoons

k-obrien:
I wanted to remove the need to rewrite the body of `hs.spoons.bindHotKeysToSpec` for each new Spoon if I wanted to include the hotkey name as an alert message.

I thought about adding a new optional name-to-message map to the function but decided making the smaller change introduced less risk and would also encourage Spoon authors to use thoughtful and descriptive naming for their hotkey names.

If a user does not want to show the alert for hotkeys as per current behaviour when using  `hs.spoons.bindHotKeysToSpec` then they can simply set the `hs.hotkey.alertDuration` to 0.

### DIFF
--- a/SPOONS.md
+++ b/SPOONS.md
@@ -120,13 +120,13 @@ If your Spoon provides some kind of background activity, e.g. timers, watchers, 
 #### Hotkeys
 
 If your Spoon provides actions that a user can map to hotkeys, you should expose a `:bindHotKeys()` method. The method should accept a single parameter, which is a table.
-The keys of the table should be strings that describe the action performed by the hotkeys, and the values of the table should be tables containing modifiers and keynames/keycodes.
+The keys of the table should be strings that describe the action performed by the hotkeys, and the values of the table should be tables containing modifiers and keynames/keycodes and, optionally, a message to be displayed via `hs.alert()` when the hotkey has been triggered.
 
 For example, if the user wants to map two of your actions, `show` and `hide`, they would pass in:
 
 ```lua
   {
-    show={{"cmd", "alt"}, "s"},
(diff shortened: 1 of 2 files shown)


asmagill:
I'm generally not a fan of changing long standing behavior, but then I don't use hotkey alerts at all -- if I want a visual confirmation of one of my hotkeys, I've always embedded something in the callback itself.

I'm wondering if it might be better to change the map passed into `hs.spoons.bindHotKeysToSpec` so that the name can be optionally set on a per-hotkey basis (I can imagine someone wanting *some* hotkeys to show an alert, but not *all* of them... plus some common names will get reused... a lot).

Example (modified from the sample in the Spoon documentation):

~~~lua
  {
    show={{"cmd", "alt"}, "s", label = "Show this spoons output"},
    hide={{"cmd", "alt"}, "h"}
  }
~~~

And then in `hs.spoons.bindHotKeysToSpec` check to see if `label` is set for a each entry.

Or if you're really wedded to using the function name as the label, then add an optional parameter to `hs.spoons.bindHotKeysToSpec` so that it sets the labels when `true` is passed in.

I don't use the feature, so I'll differ to the consensus, but those are my thoughts...

k-obrien:
I like the idea of your first alternative; I actually considered it when I first wrote the PR but decided to go for the path of least resistance. Upon reflection I think that decision was wrong. I'll make the suggested change. Thanks.

k-obrien:
@asmagill Implemented change as suggested.

asmagill:
@cmsj, objections?

cmsj:
@asmagill none :)

PR #2333, opened 2020-04-01
fix xcode 11.4 compile error

deftsp:
See also `https://developer.apple.com/documentation/xcode_release_notes/xcode_11_4_release_notes?language=objc`


cmsj:
Thanks! @asmagill fixed this in master by just reverting to `#import` so we should be good now.

Issue #2339, opened 2020-04-08
Problems after coroutine & modernise hs.application/window/uielement merge


latenitefilms:
@cmsj & @asmagill - I have now merged the latest Hammerspoon master into `CommandPost-App`. The first issue I've come across is:

```lua
2020-04-08 17:21:54: *** ERROR: ...ost.app/Contents/Resources/extensions/hs/window/init.lua:947: attempt to index a nil value (global 'uielement')
stack traceback:
	...ost.app/Contents/Resources/extensions/hs/window/init.lua:947: in metamethod '__index'
	...rishocking/Github/CommandPost/src/extensions/cp/init.lua:40: in main chunk
	[C]: in function 'xpcall'
	...Github/CommandPost/src/extensions/hs/_coresetup/init.lua:794: in function 'hs._coresetup.setup'
	(...tail calls...)
```

latenitefilms:
We've also seemed to have lost [`hs.window.timeout()`](https://github.com/Hammerspoon/hammerspoon/pull/2252).

latenitefilms:
Another one:

```lua
2020-04-08 17:33:18: ********
2020-04-08 17:33:18: 17:33:18 ERROR:   LuaSkin: hs.timer callback error: ...ost.app/Contents/Resources/extensions/hs/window/init.lua:515: ERROR: incorrect type 'no value' for argument 2 (expected boolean)
stack traceback:
	[C]: in method '_bringtofront'
	...ost.app/Contents/Resources/extensions/hs/window/init.lua:515: in method 'focus'
	...mandPost/src/plugins/finalcutpro/fullscreen/dockicon.lua:53: in function <...mandPost/src/plugins/finalcutpro/fullscreen/dockicon.lua:52>
2020-04-08 17:33:18: ********
```

asmagill:
@cmsj, the current branch includes changes to the underlying userdata for `hs.window` and `hs.application`, doesn't it?

Assuming this is so, since I'm seeing breakage with some of my own stuff that uses `hs._asm.axuieleemnt`, @latenitefilms I'm going to need to update axuielement, since it makes assumptions about the structures behind the userdata for these elements... it was easier than requiring `hs._asm.axuielement` to be recompiled everytime Hammerspoon was rebuilt.

I'll take a closer look at it later today.

latenitefilms:
@asmagill - Yes, it does - although I haven't spotted any `hs._asm.axuielement` errors yet.

[...]

PR #2340, opened 2020-04-09
make HSuicore coroutine friendly

asmagill:
Because HSuicore methods are called from module functions, they have to be state aware to be coroutine safe.

Should fix at least some of #2339


cmsj:
Thanks!

Issue #2334, opened 2020-04-01
hs.webview freezes?


latenitefilms:
Every now and again if I have a `hs.webview` that is being updated through code, whilst something else is happening in Hammerspoon, I get a freeze. I've never really looked into it, or worked out why it does it.

A quick look at the `hs.webview` code, and I've noticed there's a couple of:

```objc
if (theView.loading) [theView stopLoading] ;
while (theView.loading) {}
```

@asmagill - I wonder if it's a possibility of `theView.loading` never returning false, hence the freeze?

latenitefilms:
Does this give you any clues?

```lua
Date/Time:       2020-04-03 23:56:31 +1100
End time:        2020-04-03 23:56:32 +1100
OS Version:      Mac OS X 10.14.6 (Build 18G3020)
Architecture:    x86_64h
Report Version:  28

Data Source:     Stackshots
Shared Cache:    0x4aa6000 7F556C80-8A46-36EC-A2EF-6592B7F1F89D
Shared Cache:    0 3BF98159-9727-306D-B435-377EFF981064

Command:         CommandPost
Path:            /Users/USER/*/CommandPost.app/Contents/MacOS/CommandPost
Identifier:      org.latenitefilms.CommandPost
Version:         1.0.7-beta.6 (5878)
Parent:          launchd [1]
PID:             94383

Event:           hang
Duration:        1.20s (process was unresponsive for 15 seconds before sampling)
Steps:           12 (100ms sampling interval)

Hardware model:  MacBookPro14,3
Active cpus:     8

Time Awake Since Boot: 340000s
Time Since Wake: 300000s

Fan speed:       2537 rpm

--------------------------------------------------
Timeline format: stacks are sorted chronologically
Use -i and -heavy to re-report with count sorting
--------------------------------------------------


Heaviest stack for the main thread of the target process:
  12  start + 1 (libdyld.dylib + 91093) [0x7fff5c3e33d5]
  12  NSApplicationMain + 777 (AppKit + 10952) [0x7fff2da35ac8]
  12  -[NSApplication run] + 699 (AppKit + 79240) [0x7fff2da46588]
  12  -[NSApplication(NSEvent) _nextEventMatchingEventMask:untilDate:inMode:dequeue:] + 1361 (AppKit + 103531) [0x7fff2da4c46b]
  12  _DPSNextEvent + 1135 (AppKit + 108413) [0x7fff2da4d77d]
[...]

PR #2348, opened 2020-04-29
[READY] Change empty delay loops in webview to timers.

asmagill:
may address #2334

@latenitefilms let me know if this makes a difference.

### DIFF
--- a/extensions/webview/internal.m
+++ b/extensions/webview/internal.m
@@ -11,6 +11,8 @@
 static int           refTable ;
 static WKProcessPool *HSWebViewProcessPool ;
 
+static NSMapTable    *delayTimers ;
+
 static inline NSRect RectWithFlippedYCoordinate(NSRect theRect) {
     return NSMakeRect(theRect.origin.x,
                       [[NSScreen screens][0] frame].size.height - theRect.origin.y - theRect.size.height,
@@ -24,6 +26,29 @@ static inline NSRect RectWithFlippedYCoordinate(NSRect theRect) {
 static int userdata_gc(lua_State* L) ;
 static int NSError_toLua(lua_State *L, id obj) ;
 static int SecCertificateRef_toLua(lua_State *L, SecCertificateRef certRef) ;
+
+void delayUntilViewStopsLoading(HSWebViewView *theView, dispatch_block_t block) {
+    if (!delayTimers) delayTimers = [NSMapTable strongToWeakObjectsMapTable] ;
+
+    NSTimer *existingTimer = [delayTimers objectForKey:theView] ;
+    if (existingTimer) {
+        [existingTimer invalidate] ;
+        [delayTimers removeObjectForKey:theView] ;
+        existingTimer = nil ;
+    }
+
+    NSTimer *newDelay = [NSTimer scheduledTimerWithTimeInterval:0.01
+                                                        repeats:YES
+                                                          block:^(NSTimer *timer) {
+        if (!theView.loading) {
+            [delayTimers removeObjectForKey:theView] ;
+            [timer invalidate] ;
+            dispatch_async(dispatch_get_main_queue(), block) ;
+        }
+    }] ;
+
+    [delayTimers setObject:newDelay forKey:theView] ;
+}
 
 #pragma mark - our window object
 
@@ -897,11 +922,12 @@ static int webview_url(lua_State *L) {
         NSURLRequest *theNSURL = [skin luaObjectAtIndex:2 toClass:"NSURLRequest"] ;
         if (theNSURL) {
             if (theView.loading) [theView stopLoading] ;
-            while (theView.loading) {}
-            dispatch_async(dispatch_get_main_queue(), ^{
+
+            delayUntilViewStopsLoading(theView, ^{
                 WKNavigation *navID = [theView loadRequest:theNSURL] ;
                 theView.trackingID = navID ;
             }) ;
+
             lua_pushvalue(L, 1) ;
             return 1 ;
         } else {
@@ -1174,10 +1200,9 @@ static int webview_reload(lua_State *L) {
     HSWebViewView   *theView = theWindow.contentView ;
 
     if (theView.loading) [theView stopLoading] ;
-    while (theView.loading) {}
     BOOL validate = (lua_type(L, 2) == LUA_TBOOLEAN) ? (BOOL)lua_toboolean(L, 2) : NO ;
 
-    dispatch_async(dispatch_get_main_queue(), ^{
+    delayUntilViewStopsLoading(theView, ^{
         WKNavigation *navID ;
         if (validate)
             navID = [theView reloadFromOrigin] ;
@@ -1185,6 +1210,7 @@ static int webview_reload(lua_State *L) {
             navID = [theView reload] ;
         theView.trackingID = navID ;
     }) ;
+
     lua_pushvalue(L, 1) ;
     return 1 ;
 }
@@ -1392,11 +1418,12 @@ static int webview_html(lua_State *L) {
     NSString *theBaseURL = (lua_type(L, 3) == LUA_TSTRING) ? [skin toNSObjectAtIndex:3] : nil ;
 
     if (theView.loading) [theView stopLoading] ;
-    while (theView.loading) {}
-    dispatch_async(dispatch_get_main_queue(), ^{
+
+    delayUntilViewStopsLoading(theView, ^{
         WKNavigation *navID = [theView loadHTMLString:theHTML baseURL:[NSURL URLWithString:theBaseURL]] ;
         theView.trackingID = navID ;
     }) ;
+
     lua_pushvalue(L, 1) ;
     return 1 ;
 }
@@ -3027,6 +3054,13 @@ static int userdata_gc(lua_State* L) {
             child.parent = nil ;
         }
 
+        NSTimer *reloadTimer = [delayTimers objectForKey:theView] ;
+        if (reloadTimer) {
+            [reloadTimer invalidate] ;
+            [delayTimers removeObjectForKey:theView] ;
+            reloadTimer = nil ;
+        }
+
         theView.navigationDelegate = nil ;
         theView.UIDelegate         = nil ;
         theWindow.contentView      = nil ;
@@ -3043,6 +3077,15 @@ static int meta_gc(lua_State* __unused L) {
     if (HSWebViewProcessPool) {
         HSWebViewProcessPool = nil ;
     }
+
+    if (delayTimers) {
+        NSEnumerator *enumerator = [delayTimers objectEnumerator];
+        NSTimer *timer ;
+        while ((timer = [enumerator nextObject])) [timer invalidate] ;
+        [delayTimers removeAllObjects] ;
+    }
+    delayTimers = nil ;
+
     return 0 ;
 }
 
--- a/extensions/webview/internal.m
+++ b/extensions/webview/internal.m
@@ -30,6 +30,8 @@ static int SecCertificateRef_toLua(lua_State *L, SecCertificateRef certRef) ;
 void delayUntilViewStopsLoading(HSWebViewView *theView, dispatch_block_t block) {
     if (!delayTimers) delayTimers = [NSMapTable strongToWeakObjectsMapTable] ;
 
+//     if (theView.loading) [theView stopLoading] ;
+
     NSTimer *existingTimer = [delayTimers objectForKey:theView] ;
     if (existingTimer) {
         [existingTimer invalidate] ;
@@ -43,6 +45,7 @@ void delayUntilViewStopsLoading(HSWebViewView *theView, dispatch_block_t block) 
         // make sure were wenen't queued in the runloop before the timer was invalidated by another "load" event
         if (timer.valid) {
             if (!theView.loading) {
+                [theView stopLoading] ; // stop loading other resources
                 [delayTimers removeObjectForKey:theView] ;
                 [timer invalidate] ;
                 block() ;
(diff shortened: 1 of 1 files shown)


latenitefilms:
Awesome, thank you! Will attempt to merge into CommandPost and see.

latenitefilms:
Yep - now when I click my `hs.webview.toolbar` buttons REALLY quickly, I can't get CommandPost to freeze. Woohoo! Thanks @asmagill !

Although, when first loading the `hs.webview`, it now seems dramatically slower than before?

latenitefilms:
You can see how I'm actually using the code here if it's helpful:

https://github.com/CommandPost/CommandPost/blob/develop/src/plugins/core/toolbox/manager/init.lua#L295

asmagill:
Ok, let me know if this is any better... I changed the timer delay to 1/1000 sec, removed the unnecessary dispatch when the view is ready for updates (we're already on the main thread, so it wasn't really necessary), change the valid run modes for the timer to be a little more inclusive (like `hs.timer`), and set the initial fire time to immediately.

I'll take a closer look at your usage later (probably tomorrow) and see if any other tweaks come to mind, either in my code or in your approach.

latenitefilms:
Legend, thank you! Building now...

latenitefilms:
Sorry for the delayed reply - we added a new addition to the family.

I need to do some more scientific testing - but it still seems noticeably slower to open new `hs.webview`'s for some reason?

latenitefilms:
Ok, so ignore my last post - either I was using the wrong build, or restarting my computer fixed the delay.

However, I am still noticing some weird things. If I click my toolbar items REALLY fast, eventually `hs.webview:html()` stops working (i.e. it doesn't update the webview content).

Obviously, in this use case, I can't imagine users clicking toolbar items this quickly, but hey, you never know.

If it makes any difference, the data that I'm supplying to `hs.webview:html()` could take a little while to generate (as I'm building a UI in HTML on the fly).

asmagill:
@latenitefilms -- assuming its safe to still ignore your slowdown issue which, to be honest, I have no idea about -- this shouldn't be *any* slower, especially for the first load of a webview, see if this helps with your incessant predilection for repeated button mashing 😜

It was theoretically possible that if a timer had fired and pushed its action block onto the application queue, but some other event which would start a new "load" event was already on the queue, you could get a situation where the old one would delete the timer for the new one. This now checks that the timer for a given queued event is still valid before running its "load" block which *should* prevent such a race condition from being possible.

latenitefilms:
Ummm, the latest changes didn't seem to fix the issue.

I have a `hs.webview`, which has a toolbar that activate different content using `hs.webview:html()`. Some of the "panels" take a while to load, as we're pulling in data and generating the HTML on the fly. If I click back and forth between some of these slower panels, then the panels stop updating because `hs.webview:html()` stops working.

Any ideas?

Screen Shot 2020-05-16 at 2 33 06 am

latenitefilms:
Not sure if it's relevant or not, but I get these messages in Console, when the HTML contents isn't updating correctly:

Screen Shot 2020-05-16 at 11 03 27 pm

latenitefilms:
Maybe related?

https://bugs.webkit.org/show_bug.cgi?id=173384

latenitefilms:
Ok, so `webview:loading()` (which is `[theView isLoading]` in Obj-C land) returns `true` when it fails to restart. It seems to be stuck loading?

asmagill:
Ran some tests of my own last night:

Initial webview:
~~~lua
w = hs.webview.new({ x = 100, y = 100, h = 500, w = 400 }):show()
~~~

Then I tried the following:
~~~lua
for i = 1, 100, 1 do w:html(tostring(i)) end
~~~

Which works -- the last entry appears in the webview. Then it struck me, of course it does because none of the timers ever actually fire before getting invalidated.

So then I tried:

~~~lua
coroutine.wrap(function() for i = 1, 99, 1 do w:html(tostring(i)) ; coroutine.applicationYield() ; print(i) end end)()
~~~

And the webview now hangs after the first or second call. Typing `w:loading()` returns true and `w:stopLoading()` does absolutely nothing -- `w:loading()` remains true. Doing `w:reload()` also does nothing -- but right clicking in the webview and selecting "Reload" causes the final version to finally display.

A little digging and I find a bunch of back and forth about what the "loading" property of WKWebView is supposed to mean, but the consensus seems to be that it refers to the "getting" of the primary resource for the page (the content of the actual URL/html) and *not* any subsequent resources it may then require (images, external javascript/css, etc.). It is also completely unrelated to rendering of the actual content, just the acquisition of it. But in my experiments it *looks like* it is during the rendering where the "loading" property gets changed... and the `stopLoading` method *doesn't* toggle this... which leads me to "WKWebView is a big fat giant PITA and sucks Donkey #$@%$#!!!"

So in a fit of madness, I commented out lines 931, 1209, and 1427 of webview's `internal.m` file (basically the three places where the line reads `if (theView.loading) [theView stopLoading] ;` and then recompiled.

Now, the loop in the wrapped coroutine works... it appears to cycle through most if not all of the actual pages, but it doesn't hang. Not sure what the implication will be if the HTML takes a long time to render or with remote URLs, rather than the rather simplistic one line of text in my tests...

I'm not ready to push these changes because if true, it kind of makes me wonder why the "stopLoading" method exists at all, if its so broken, so I want to see if I can dig deeper into whats actually happening; however, if you don't mind trying these changes yourself, I'd be interested in how it affects real HTML and not my simplistic example. I'm specifically curious about (1) does it actually interrupt HTML that takes a "long time" to render (not load), and (2) are *any* of the page changes being prevented, or do *all* of them end up getting performed and then immediately replaced by the "next" request in the queue?

I'm going to dig into the WebKit2 sources again and see if I can find out more about the `loading` and `stopLoading` code... while Apple's implementation differs some, doing so in the past has led to some insights that helped with writing this module, so it seems like the best place to start...

latenitefilms:
Maybe related?

https://stackoverflow.com/questions/11635854/what-does-stoploading-really-do

latenitefilms:
FYI: Removing the three `if (theView.loading) [theView stopLoading] ;` lines fixed all the issues. It's now working great, even with really slow/complex "panels".

asmagill:
Any significant slowdowns, flashing of content, or side effects?

latenitefilms:
Not that I can see. It seems to work fine switching panels quickly. If I click on a slower  -to-load panel, it just keeps the previous panels content until the new panel’s content is ready. Seems to all work as intended, and fixes the original issue of Hammerspoon freezing when changing panels quickly.

I wonder if you could just get rid of the `if (theView.loading)` instead, so that it always stops loading first within timer? Or will it crash if `theView.loading` is `false`? Or is it the `[theView stopLoading]` that’s causing the problem in the first place?

asmagill:
I think the problem is with `[theView stopLoading]` specifically... the code to trigger the stop and then wait for `webview.loading` to become false was added for #1153 -- if you try to load new content before `loading` becomes false, you run into -999 errors.

I think, if `stopLoading` is triggered at the wrong point, it prevents whatever usually sets `loading` to false from being executed. Of course, without looking at the source for WKWebview, that's a guess... I am loading the github repo for webkit now (it's got like a bajillion deltas to process and has now been running for about 45 minutes with about 20% left), so I can look at what the public version is doing, but then Apple has always tweaked even the stuff they release to the OS community for themselves, so...

I'm leaning towards pushing my changes (with the lines commented out) and calling this ready for inclusion, but I want to run a couple of more tests and dig into the webkit source a little first.

asmagill:
Digging through and understanding WebKit is going to take a while, but what little I do understand so far does seem to confirm my theory that if `stopLoading` is called at the wrong time, the state never reaches `State::Finished`, and if it doesn't reach that state, then no new data will be processed and `loading` will continue to return `true`... Doing a forced reload from within the context menu seems to execute a "reset" method that isn't published (and isn't in the WKWebView but one of it's proxies) that even the `reload` *method* doesn't include...

And, as a test, I tried modifying my above coroutine example to the following and it showed the earlier behavior of seeming to lock up:

~~~lua
coroutine.wrap(function() for i = 1, 99, 1 do w:stopLoading() ; coroutine.applicationYield() ; w:html(tostring(i)) ; print(i) end end)()
~~~

Which makes me wonder if the `stopLoading` method should be removed completely...

If I can figure out the specific case where `stopLoading` is useful and a way to detect if we're *at* or *in* that state, it should remain -- it is defined for the class after all -- but so far the code and docs aren't helping much...

asmagill:
Yeah... I want to remove `stopLoading` entirely... @cmsj, @latenitefilms I'm going to need your thoughts on that.

I just tried the following with the same results we've been getting...

~~~lua
a = { "https://www.google.com", "https://www.yahoo.com", "https://www.amazon.com", "https://www.microsoft.com", "https://www.apple.com" }
w = hs.webview.new({ x = 100, y = 100, h = 500, w = 400 }, { developerExtrasEnabled = true }):show()
coroutine.wrap(function() for i,v in ipairs(a) do w:url(v) ; coroutine.applicationYield() ; print(i) end end)()
~~~

works with the google page flashing up before the apple one does at the end (so the timers for the interim ones are successfully invalidated)

but if I introduce a `stopLoading` into the mix, e.g.

~~~lua
coroutine.wrap(function() for i,v in ipairs(a) do w:stopLoading() ; coroutine.applicationYield() ; w:url(v) ; print(i) end end)()
~~~

or even

~~~lua
coroutine.wrap(function() for i,v in ipairs(a) do if w:loading() then w:stopLoading() end ; coroutine.applicationYield() ; w:url(v) ; print(i) end end)()
~~~

it stops at some point (usually the first URL, but once in a great while, the second) and is stuck in the "loading" state requiring a reload from the context menu to free it (`w:reload()` doesn't cut it...)

latenitefilms:
Well, it definitely doesn’t actually stop loading as the name would suggest, so I’m all for removing it, or just making the documentation private so we don’t break anyone that is using it for some reason.

asmagill:
FWIW https://bugs.webkit.org/show_bug.cgi?id=202543 basically confirms at least part of what we're observing, but no real action appears to have been taken...

asmagill:
Another option... modify `stopLoading` (and its docs to make this clear) so that if the page itself is still loading it does nothing... the docs are a little unclear. Consensus on the internet is that the `loading` property is about the base page, not any background ajax or resource files referenced by the page itself; but `stopLoading` seems to only "break things" when `loading` is true...

I'm not enough a HTML guru to craft a good test, but I can confirm that adding https://github.com/Hammerspoon/hammerspoon/pull/2348/files#diff-8464e0071249c699ef07682f5856adaaR48 doesn't break the desired behavior. Of course it may just be a NOP... as I said, I don't trust my HTML/Web Programming enough to really tell...

asmagill:
@latenitefilms please try this latest pull if you get a chance. If it works, I'll still need to modify the `stopLoading` method and docs... and if it doesn't, I'll still need to *remove* the `stopLoading` method and docs...

So expect at least one more test :-)

cmsj:
@asmagill im going to completely defer to you on this one. I’m not at all familiar with hs.webview 🙂

latenitefilms:
Woohoo! All looks good to me! Merge away!

Thanks heaps for all your help and testing @asmagill !

asmagill:
@cmsj, the only nice thing I can say about `hs.webview` is that I don't hate it as much as I hate `hs.eventtap`. And I don't hate it enough to rewrite it given what I know now. So I guess that's two things...

Ok, if you're ok with the verbiage in the `hs.webview:stopLoading()` method documentation, then this is ready.
~~~
> help.hs.webview.stopLoading
Method: hs.webview:stopLoading() -> webviewObject

Stop loading additional content for the webview.

Parameters:
 * None

Returns:
 * The webview object

Notes:
 * this method does not stop the loading of the primary content for the page at the
   specified URL
 * if [hs.webview:loading](#loading) would return true, this method does nothing 
   -- see notes:
   * The documentation from Apple is unclear and experimentation has shown that if
     this method is applied before the content of the specified URL has loaded, it
     can cause the webview to lock up; however it appears to stop the loading of
     additional resources specified for the content (external script files, external
     style files, AJAX queries, etc.) and should be used in this context.
~~~

If you don't get to it before I do, I'll merge it in the next day or so when I start formally preparing axuielement for inclusion in core (I've stripped it down to the minimum necessary for @latenitefilms  but yet with enough to start replacing some of our blocking and/or slower application and window methods with more responsive ones -- the `getMenuItems` version using axuielement doesn't bring HS to a halt when used on something like Safari). Eventually I may add some of the other filters and search methods back, but I want to finally be able to take a look at `hs.uielement` and see if we can't maybe simplify the whole mix a bit first.

latenitefilms:
Awesome! I'm going to merge this into CommandPost now.

I've never been able to get `hs.uielement` to work properly - so I'd probably just depreciate it.

asmagill:
As I understand it, `hs.uielement` is closely integrated with the watcher functionality for `hs.application` and `hs.window`. I'm not sure if it has specific utility beyond that or not -- I honestly haven't looked that closely in a long time.

It's my suspicion that everything `hs.uielement` does, `hs._asm.axuielement[.observer]` can do (or can be made to do with very minor additions), but at the moment, that is just a guess.

Every time I've tried to modify uielement in the past, I've broken something... since then, @cmsj has done a refactoring of application, window, and uielement, so perhaps its cleaner/clearer now? In any case, there *is* overlap, and while I personally feel axuielement is clearer and easier to use, rather than deprecation, it's likely that there will be some merging or mixing of uielement and axuielement to reduce code duplication while maintaining backwards compatibility and to prevent breaking anything that anyone may be currently doing with it.

cmsj:
Agreed. Also, we don’t need to make any decisions about uielement before landing axuielement.

PR #2353, opened 2020-05-05
Update CocoaAsyncSocket from 7.6.3 to 7.6.4

latenitefilms:
Release notes can be found [here](https://github.com/robbiehanson/CocoaAsyncSocket/releases/tag/7.6.4).


cmsj:
Thanks for not running "pod install" or "pod update" I hate those diffs and I'd much rather run the command myself :)

PR #2359, opened 2020-05-09
make MJConfigFile resolve symlinks

### LINKED ISSUE #2357: Double-Clicking an Installed Spoon Deletes It
Steps to reproduce:
1. Install or put a spoon in ./Spoons.
2. Double click or open it.
3. You’ll see this message:
Screen Shot 2020-05-06 at 23 07 02
4.  The .spoon is removed immediately — it doesn’t go to the trash.

And, a question:
Is it possible to make Hammerspoon/macOS/Finder stop treat .spoons as packages, but as the folders they are? It would make their management much easier for me, for example opening them directly as folders with launchers such as LaunchBar.

Edit: my ./hammerspoon directory is **symlinked** to ~.

Thanks!

luckman212:
My attempt to fix #2357 
Very limited testing — "it works for me" 🙂
Thanks @latenitefilms

### DIFF
--- a/Hammerspoon/MJConfigUtils.m
+++ b/Hammerspoon/MJConfigUtils.m
@@ -6,5 +6,5 @@ NSString* MJConfigDir(void) {
 }
 
 NSString* MJConfigFileFullPath(void) {
-    return [MJConfigFile stringByStandardizingPath];
+    return [MJConfigFile stringByResolvingSymlinksInPath];
 }


cmsj:
I had a look at the docs and it seems like `stringByStandardizingPath` will resolve symlinks only if the supplied path is absolute, for relative ones it doesn't. I think we still want the other things it's doing though, so I would suggest we go with:

`[[MJConfigFile stringByStandardizingPath] stringByResolvingSymlinksInPath];`

PR #2368, opened 2020-05-15
Release Script Bug Fix

latenitefilms:
- The `SENTRY_AUTH_TOKEN` was not being exported for use with
`sentry-cli`


latenitefilms:
I think this is right @cmsj - but feel free to close if I'm missing something.

cmsj:
Yep, that makes sense. Good spot!

Issue #2361, opened 2020-05-11
[Suggestion] Add hs.math.isNan


asmagill:
Putting this out there for comments...

It doesn't come up often, and there is no rush on this, but since we now have a place where this actually fits, it would be nice to have a way to quickly check if a number is actually the `nan` value.

~~~lua
hs.math.isNan = function(x)
    return x ~= x
end
~~~

Example:

~~~lua
> hs.math.isNan = function(x)
    return x ~= x
end

> return math.huge, hs.math.isNan(math.huge)
inf	false

-- yeah, I know they'll be different; this is just an example
> return math.random(), hs.math.isNan(math.random())
0.29472790984437	false

> return 0/0, hs.math.isNan(0/0)
nan	true
~~~

I'm indifferent on the name... `isnan`, `isNan`, `isNAN`, `notANumber`...

cmsj:
How about isNaN? I think so long as the first N is capitalised, I don’t have strong feelings about the rest, but I do think this is a good idea to add.

asmagill:
Ok, I like `hs.math.isNaN`. The docs should be explicit that it is testing for "equaling" the mathematical concept of a numerical value being "not-a-number" -- technically a string value is also "not a number", but this test will return false in this case.

What about also adding the other related C library queries?

e.g.
~~~
int isinf(double value);   /* gives -1 for -inf, 1 for inf, 0 otherwise */
int finite(double value);  /* gives 1 for not NaN and not inf */
~~~

I can see a use case for both, but I wonder if `hs.math.isInfinite` should just return `true` for both `math.huge` and `-math.huge` rather then a signed response... or maybe go ahead and return -1 and 1 (since they will both evaluate to `true` in a conditional) but `false` when the number *isn't* one of the infinities?

And as to naming, `hs.math.isFinite` and `hs.math.isInfinite`? Or the shorter version like the C library counterparts?

asmagill:
And while I'm avoiding going out to get groceries and run errands, another thought...

Should `hs.math` also mirror the contents of the lua `math` library? I can see myself doing something stupid in one of my scripts like this and wondering why `math.sin` no longer exists:

~~~lua
local math = require("hs.math")
... other lua code...
... until something like:
    print(math.sin(90 * math.pi / 180))
~~~

asmagill:
should have closed this when #2371 landed

PR #2371, opened 2020-05-18
[DISCUSSION] add hs.math.isNaN, isFinite, isInfinite

asmagill:
Addresses #2361

If the docstrings and names are good, then I think this is ready to go.

I also added comments about why `hs.math.minFloat` is accessed through metamethods rather than just assigning it directly to the module... I had to stare at it for a few minutes myself to remember why I made that decision the first time!

### DIFF
--- a/extensions/math/init.lua
+++ b/extensions/math/init.lua
@@ -2,11 +2,42 @@
 ---
 --- Various helpful mathematical functions
 ---
---- This module will also mirror the contents of the built-in Lua `math` library so it is safe to do something like the following in your own code and still have access to both libraries:
+--- This module includes, and is a superset of the built-in Lua `math` library so it is safe to do something like the following in your own code and still have access to both libraries:
 ---
 ---     local math = require("hs.math")
 ---     local n = math.sin(math.minFloat) -- works even though they're both from different libraries
-
+---
+--- The documentation for the math library can be found at http://www.lua.org/manual/5.3/ or from the Hammerspoon console via the help command: `help.lua.math`. This includes the following functions and variables:
+---
+---   * hs.math.abs        - help available via `help.lua.math.abs`
+---   * hs.math.acos       - help available via `help.lua.math.acos`
+---   * hs.math.asin       - help available via `help.lua.math.asin`
+---   * hs.math.atan       - help available via `help.lua.math.atan`
+---   * hs.math.ceil       - help available via `help.lua.math.ceil`
+---   * hs.math.cos        - help available via `help.lua.math.cos`
+---   * hs.math.deg        - help available via `help.lua.math.deg`
+---   * hs.math.exp        - help available via `help.lua.math.exp`
+---   * hs.math.floor      - help available via `help.lua.math.floor`
+---   * hs.math.fmod       - help available via `help.lua.math.fmod`
+---   * hs.math.huge       - help available via `help.lua.math.huge`
+---   * hs.math.log        - help available via `help.lua.math.log`
+---   * hs.math.max        - help available via `help.lua.math.max`
+---   * hs.math.maxinteger - help available via `help.lua.math.maxinteger`
+---   * hs.math.min        - help available via `help.lua.math.min`
+---   * hs.math.mininteger - help available via `help.lua.math.mininteger`
+---   * hs.math.modf       - help available via `help.lua.math.modf`
+---   * hs.math.pi         - help available via `help.lua.math.pi`
+---   * hs.math.rad        - help available via `help.lua.math.rad`
+---   * hs.math.random     - help available via `help.lua.math.random`
+---   * hs.math.randomseed - help available via `help.lua.math.randomseed`
+---   * hs.math.sin        - help available via `help.lua.math.sin`
+---   * hs.math.sqrt       - help available via `help.lua.math.sqrt`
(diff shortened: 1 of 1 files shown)


latenitefilms:
I think it could also be handy to add documentation for all the built-in `math` functions too?

asmagill:
These are already documented. If you type `help.lua.math` into the console you should see it. And with 5.4 on the horizon I suppose I should dig out the scripts I used to make the necessary json file and see if it works against the 5.4 documentation pages...

latenitefilms:
Understood, but I still think it would be handy if the `math` functions were mentioned in the [API docs](http://www.hammerspoon.org/docs/index.html) to help out new users, who might also be using Lua for the first time.

cmsj:
I would like to hope that anyone diving into Hammerspoon is probably ok with looking for Lua’s math library :)

extensions/math/init.lua
```diff
@@ -1,8 +1,66 @@
 --- === hs.math ===
 ---
 --- Various helpful mathematical functions
+---
+--- This module will also mirror the contents of the built-in Lua `math` library so it is safe to do something like the following in your own code and still have access to both libraries:
```
cmsj, line 5: How about we change "will also mirror" to "includes, and is a superset of"? It moves it into a more active voice and is explicit that this is Lua math + more stuff.

asmagill:
Re the active vs passive voice, I don't have a strong opinion, so that's fine... but re documenting the built-in functions, I looked at the other module we have that does something similar, `hs.utf8` and this is from its documentation:

~~~
> help.hs.utf8
Functions providing basic support for UTF-8 encodings

Prior to upgrading Hammerspoon's Lua interpreter to 5.3, UTF8 support was provided by including the then beta version of Lua 5.3's utf8 library as a Hammerspoon module.  This is no longer necessary, but to maintain compatibility, the Lua utf8 library can still be accessed through `hs.utf8`.  The documentation for the utf8 library can be found at http://www.lua.org/manual/5.3/ or from the Hammerspoon console via the help command: `help.lua.utf8`. This affects the following functions and variables:

  * hs.utf8.char          - help available via `help.lua.utf8.char`
  * hs.utf8.charPattern   - help available via `help.lua.utf8.charpattern`
  * hs.utf8.codepoint     - help available via `help.lua.utf8.codepoint`
  * hs.utf8.codes         - help available via `help.lua.utf8.codes`
  * hs.utf8.len           - help available via `help.lua.utf8.len`
  * hs.utf8.offset        - help available via `help.lua.utf8.offset`

Additional functions that are specific to Hammerspoon which provide expanded support for UTF8 are documented here.
~~~

Would adding verbiage like this to the head of the `hs.math` documentation that mentions the other functions and how to get at their documentation be a reasonable compromise?

cmsj:
+1 from me. Thanks!

asmagill:
Something like:

~~~
=== hs.math ===

Various helpful mathematical functions

This module includes, and is a superset of the built-in Lua `math` library so it is
safe to do something like the following in your own code and still have access to
both libraries:

    local math = require("hs.math")
    local n = math.sin(math.minFloat) -- works even though they're both from different libraries

The documentation for the math library can be found at http://www.lua.org/manual/5.3/
or from the Hammerspoon console via the help command: `help.lua.math`. This
includes the following functions and variables:

  * hs.math.abs        - help available via `help.lua.math.abs`
  * hs.math.acos       - help available via `help.lua.math.acos`
  * hs.math.asin       - help available via `help.lua.math.asin`
  * hs.math.atan       - help available via `help.lua.math.atan`
  * hs.math.ceil       - help available via `help.lua.math.ceil`
  * hs.math.cos        - help available via `help.lua.math.cos`
  * hs.math.deg        - help available via `help.lua.math.deg`
  * hs.math.exp        - help available via `help.lua.math.exp`
  * hs.math.floor      - help available via `help.lua.math.floor`
  * hs.math.fmod       - help available via `help.lua.math.fmod`
  * hs.math.huge       - help available via `help.lua.math.huge`
  * hs.math.log        - help available via `help.lua.math.log`
  * hs.math.max        - help available via `help.lua.math.max`
  * hs.math.maxinteger - help available via `help.lua.math.maxinteger`
  * hs.math.min        - help available via `help.lua.math.min`
  * hs.math.mininteger - help available via `help.lua.math.mininteger`
  * hs.math.modf       - help available via `help.lua.math.modf`
  * hs.math.pi         - help available via `help.lua.math.pi`
  * hs.math.rad        - help available via `help.lua.math.rad`
  * hs.math.random     - help available via `help.lua.math.random`
  * hs.math.randomseed - help available via `help.lua.math.randomseed`
  * hs.math.sin        - help available via `help.lua.math.sin`
  * hs.math.sqrt       - help available via `help.lua.math.sqrt`
  * hs.math.tan        - help available via `help.lua.math.tan`
  * hs.math.tointeger  - help available via `help.lua.math.tointeger`
  * hs.math.type       - help available via `help.lua.math.type`
  * hs.math.ult        - help available via `help.lua.math.ult`

Additional functions and values that are specific to Hammerspoon which provide
expanded math support are documented here.
~~~

latenitefilms:
Perfect! Great idea!

cmsj:
@asmagill feel free to merge this whenever you're happy with it :)

PR #2393, opened 2020-06-11
create invalidCallback for hs.chooser

SebastienWae:
Create an `invalidCallback` method for `hs.chooser`.

Similar to the the valid property of Alfred's script filter items. [alfred.com#script-filter](https://www.alfredapp.com/help/workflows/inputs/script-filter/json/)

This add an optional `valid` property to `hs.chooser` choices that default to `true`, if set to false when the choice is selected the `invalidCallback` will be invoked instead of the `completionCallback` and the chooser interface will stay instead of hiding.

This allows for example multi-step workflows with the Seal Spoon and can be used as an alternative to Alfred.

Example use case for an App manager:
* Fuzzy search the name of a running application
  - All choices have `valid` propriety set to `false`
* The user press enter
  - The `invalidCallback` method is invoked
  - Autocomplete the name of the selected app
  - Change to choices related to the app
    * All choices have `valid` propriety set to `true`
    * All choices have a custom function
* The user press enter
  - The `completionCallback` is invoked
  - The choice custom function is invoked

### DIFF
--- a/extensions/chooser/HSChooser.h
+++ b/extensions/chooser/HSChooser.h
@@ -53,6 +53,7 @@
 @property(nonatomic) int queryChangedCallbackRef;
 @property(nonatomic) int completionCallbackRef;
 @property(nonatomic) int rightClickCallbackRef;
+@property(nonatomic) int invalidCallbackRef;
 
 // A pointer to the hs.chooser module's references table
 @property(nonatomic) int *refTable;
--- a/extensions/chooser/HSChooser.m
+++ b/extensions/chooser/HSChooser.m
@@ -44,6 +44,7 @@
         self.choicesCallbackRef = LUA_NOREF;
         self.queryChangedCallbackRef = LUA_NOREF;
         self.rightClickCallbackRef = LUA_NOREF;
+        self.invalidCallbackRef = LUA_NOREF;
         self.completionCallbackRef = completionCallbackRef;
 
         self.hasChosen = NO;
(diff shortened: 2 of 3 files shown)


asmagill:
After reviewing the code, there is a potential point of confusion that should be documented -- it should be made clear that even if the `valid` parameter for an entry is set to `false`, if no `invalidCallback` has been defined, the default completion handler will be invoked instead.

asmagill:
In reviewing a usage of `hs.chooser` of mine that visually implements something similar to what this pull will provide by invoking the `show` method from within the `chooserCallback` when the choice should be treated like the entry into a sub-menu, it occurs to me that a slightly different approach might be cleaner, at least in some cases, so I want to put it out for discussion before landing this...

What if instead of adding a new callback, we change the behavior of the `chooserCallback` so that if can return an optional boolean value indicating whether or not the chooser should be closed?

* keep the `valid` attribute and it's default assumed value of true
* if the `chooserCallback` does not return a value (what it currently does) or returns nil, then the logic is determined by the `valid` attribute -- if `valid` == true, hide the chooser display, if `valid` == false, keep the chooser visible.
* if the `chooserCallback` returns a boolean value, then the `valid` attribute is ignored and the return value determines if the chooser display is hidden (returns true) or remains on screen (returns false).

The pro is that you can set a default when determining the chooser items, but programmatically override it within the `chooserCallback` based on factors that might have changed (in my case, it's the holding of a specific modifier key when selecting some items); the con is that it makes the `chooserCallback` a potentially more complex function.

Thoughts?

I'll go with the consensus, so if people think this pull is the cleaner approach, I'm ok with landing this pull.

PR #2397, opened 2020-06-14
[READY] Added hs.serial

### LINKED ISSUE #2391: [New Extension] hs.serial - communicate with external devices through a serial port
I'm thinking basically a wrapper for [ORSSerialPort](https://github.com/armadsen/ORSSerialPort).

The reason being, I'm hoping to add support for the [Tour Box](https://www.tourboxtech.com), which seems to use a virtual COM port for communication.

latenitefilms:
- Added `ORSSerialPort` pod
- Added `hs.serial`
- Closes #2391


latenitefilms:
@asmagill & @cmsj - This all seems to be working, however, it definitely needs a review to make sure I haven't done anything silly in terms of garbage collection, etc.

asmagill:
I'll try and take a look at this sometime this week.

I tried my hand at a serial module a while back that worked for me, but I was uncomfortable at the time including it in the core modules because it used some techniques that were new to me (at the time)... and then I stopped using it so never really got around to testing it thoroughly enough to think about suggesting it again. This is not to say I want to offer it up as an alternative -- I prefer using a library that has wider support and testing then mine ever got. But I do have some questions about functionality, in case I do go back to wanting to use serial support from within Hammerspoon...

* does this library (and module as implemented for Hammerspoon) allow you to change the baud rate on the fly or only when the port is opened?
* enabling/disabling both hardware *and* software flow control? 
* can (at least) DTR be toggled under software control?

If these features are available but not currently broken out into module functions/methods, I'd ask that they be added.

This may all be moot and already present -- I haven't had a chance to look closely at it yet, but they are the features I'd want in a serial port module.

asmagill:
(oh, btw my version can be viewed at https://github.com/asmagill/hammerspoon_serial/tree/master if you want to look at the functionality it offered)

latenitefilms:
Garnsnabbit! I should have known that @asmagill would have already played with this at some point! My bad for not checking before getting stuck into this.

`hs._asm.serial` already looks more feature-rich than what I threw together, and you write that it's based partly on `ORSSerialPort` anyway - so it might be easier to throw away this pull request anyway, and retrofit `hs._asm.serial` to make use of the `ORSSerialPort` pod if that's the direction you'd prefer?

Looking at the `ORSSerialPort` [manual](http://cocoadocs.org/docsets/ORSSerialPort/2.0.2/Classes/ORSSerialPort.html), it does offer the following properties to configure a serial port:

- `baudRate` property
- `allowsNonStandardBaudRates` property
- `numberOfStopBits` property
- `shouldEchoReceivedData` property
- `parity` property
- `usesRTSCTSFlowControl` property
- `usesDTRDSRFlowControl` property
- `usesDCDOutputFlowControl` property

...so I believe `ORSSerialPort` can do everything you want - it just needs to be exposed.

I'll have a play with `hs._asm.serial` myself and see if it works with what I was doing.

latenitefilms:
FYI: I'm going to expose the rest of the `ORSSerialPort` properties in this `hs.serial` implementation anyway, then we can properly evaluate whether it's worth continuing down this route, or scrap this pull request and go with `hs._asm.serial`.

One potential benefit of using the `ORSSerialPort` pod is that it has a [packet parsing API](https://github.com/armadsen/ORSSerialPort/wiki/Packet-Parsing-API), which sounds pretty cool, and means you can avoid manually maintaining a buffer - but I have no idea how you would best implement this in Lua, so regardless, if we do decide this pull request is worth continuing, I'll definitely need your help/guidance with that side of things anyway.

cmsj:
I'm saying all of this without having really reviewed either of your serial port modules, or even having used them, so I'm well aware that it might not make any sense, but....

Honestly I'm having very mixed feelings about external dependencies at the moment. One the one hand, Pods are a risk to our ability to even compile HS at any random point in the future, on the other hand, integrating third party code into HS (I'm thinking of LuaFileSystem) is a risk to our ability to keep third party dependencies up to date.
I *think* I favour the latter option because at least the code is just a few files in our tree and not something that Cocoapods controls and rewrites at will, but I'm not really sure.

latenitefilms:
I think it kinda has to be reviewed on a case-by-case basis depending on the dependancy.

In this case, like [MIKMIDI](https://github.com/mixedinkey-opensource/MIKMIDI), [ORSSerialPort](https://github.com/armadsen/ORSSerialPort) is being actively developed by a really awesome, responsive and proactive developer - so I think it's worthwhile using the pod.

In the case of `CocoaHTTPServer`, in which the original repository has essentially stopped any updates - the benefits of open source projects like this is that others can always fork them and continue them on in forked repositories. My suggestion and plan with that is to fork it myself, and bring across the necessary websocket fixes that other people have done in various forks.

Whilst I agree, it's kinda nice to have an extension "self contained" (like something like `hs.streamdeck` is with no external dependancies) - you kinda run into the similar issues, because for example with `hs._asm.serial`, in it's current form it has problems (i.e. it crashes Hammerspoon if you fail to close serial port before reloading), and really relies on @asmagill to fix them - although the benefit of keeping things "self contained" is that anyone can make a fix and a simple pull request directly within Hammerspoon, rather than having to fix an external repository then do a `pod update`.

Either way, in regards to this particular issue, I'm going to keep working on my own implementation of `hs.serial` for now using [ORSSerialPort](https://github.com/armadsen/ORSSerialPort) because it's a good learning experience, I need it sooner rather than later for CommandPost, and because I already realised that there's so many things I've poorly thought about in the current implementation that will need a completely re-think and re-design. 

I'm happy for this not to be merged into Hammerspoon urgently, and I can just use it in CommandPost once I get it working properly, and we can wait and see what @asmagill thinks about his own `hs._asm.serial` implementation.

asmagill:
Meh, I don't have strong feelings one way or the other.

Re external dependancies, in theory they will be more thoroughly tested than anything we throw together, but with that comes an update schedule (or lack of one) that's outside our control.

Basically we have two kinds: ones that we implement more-or-less "as is" from their original source (e.g. `inspect` and `lfs`), and ones that provide libraries that we then write our own modules/interfaces for (I think every CocoaPod currently included).

I'm not sure that we *should* have included `inspect` and `lfs` the way we did (not that I'm advocating removing them now -- I use them too much), -- LuaRocks works perfectly well with a Hammerspoon installation, once you get a couple of paths corrected; it just doesn't have a shiny GUI or provide quite the "install once and run" model that I think Hammerspoon sought in its early days and still does, to some extent. But it puts keeping them up to date completely on us since our copy is separate from the codes "regular" installation environment.

And as to the CocoaPods... well, that's the point of external libraries -- to encapsulate what you don't want to have to write yourself, and then write what you *do* want on top of it. And hopefully you choose a library that's well maintained and current, so you don't have to do so yourself. And (correct me if I'm wrong, as I've never added a Pod myself), I *think* keeping them up to date is relatively simple with a single command, though it has to be run before re-building Hammerspoon.

- - -

Re the serial module, I'd be willing to do a complete rewrite of my module at some point, knowing what I know now... And I'm not above looking at the `ORSSerialPort` source and see if their way of doing things sparks any specific ideas (in fact, I think I did just that as one of my sources when I originally wrote it)... I've often reviewed other implementations when figuring out how I want to program something, and as far as I can recall, I've usually (always? I at least strive for always, but it's hard to remember for certain) made mention of such, at least in comments, and usually in the documentation as well.

cmsj:
@latenitefilms yeah it's the HTTPServer situation that has me nervous. The various dependencies and the main pod are basically all unmaintained as far as I can see, and sooner or later that *is* going to bite us, and we have no obvious alternative pod.

I'm very glad to hear that ORSSerialPort is being actively developed. I think that ought to be our minimum requirement for pulling in an additional pod, so on that basis I'd be ok with merging this module when it's ready.

latenitefilms:
> And (correct me if I'm wrong, as I've never added a Pod myself), I think keeping them up to date is relatively simple with a single command, though it has to be run before re-building Hammerspoon.

Correct, you just need to execute `pod update` from Terminal.

> And I'm not above looking at the ORSSerialPort source and see if their way of doing things sparks any specific ideas (in fact, I think I did just that as one of my sources when I originally wrote it)...

Correct, you did. The module is largely based on code found at http://playground.arduino.cc/Interfacing/Cocoa and https://github.com/armadsen/ORSSerialPort.

> yeah it's the HTTPServer situation that has me nervous. The various dependencies and the main pod are basically all unmaintained as far as I can see, and sooner or later that is going to bite us, and we have no obvious alternative pod.

I'll fork `CocoaHTTPServer` myself and try take on the role of actively maintaining it for the purposes of Hammerspoon & CommandPost - that way you can always make pull requests yourself if things break.

> I'm very glad to hear that ORSSerialPort is being actively developed. I think that ought to be our minimum requirement for pulling in an additional pod, so on that basis I'd be ok with merging this module when it's ready.

I'll try finish this pull request, so it's functional, then we can get @asmagill to review and see if he think he can do a better job without the need for `ORSSerialPort`. Might take me a while anyway.

latenitefilms:
@asmagill - I've just updated this branch as well, so fingers and toes crossed it makes merging easier for you.

asmagill:
forgot to add "closes ..." to commit message instead of accepting default

PR #2426, opened 2020-08-01
Updated www.hammerspoon.org URLs to use HTTPS

garrettreid:
While navigating Hammerspoon documentation, I noticed I often ended up on non-TLS web pages. This is a bummer, since www.hammerspoon.org is set up just fine for HTTPS. This pull request tries to address that, updating Hammerspoon links where it seemed safe to do so.

A couple things to note:
- Hammerspoon's SSL cert is for [www.hammerspoon.org](https://www.hammerspoon.org), not [hammerspoon.org](https://hammerspoon.org). CertBot will happily issue certificates with Subject Alternative Names that work for multiple domains at once; this is recommended.
- I didn't update the URL in the rtf, due to principle of minimum possible changes and the existing link being to http://hammerspoon.org (the HTTPS version is not covered by the current LetsEncrypt certificate).
- HTTP links can also be mitigated by server config; redirect from HTTP to HTTPS and [add a HSTS header](https://en.wikipedia.org/wiki/HTTP_Strict_Transport_Security) to teach clients to always visit HTTPS.

### DIFF
--- a/Hammerspoon/Hammerspoon-Info.plist
+++ b/Hammerspoon/Hammerspoon-Info.plist
@@ -278,7 +278,7 @@
 			<key>UTTypeIdentifier</key>
 			<string>org.hammerspoon.Hammerspoon.Spoon</string>
 			<key>UTTypeReferenceURL</key>
-			<string>http://www.hammerspoon.org/</string>
+			<string>https://www.hammerspoon.org/</string>
 			<key>UTTypeTagSpecification</key>
 			<dict>
 				<key>public.filename-extension</key>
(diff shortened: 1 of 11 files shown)


asmagill:
@cmsj, I'm going to let you weigh in on this... I don't feel strongly one way or the other.

latenitefilms:
I vote for removing the "www" all together, so that `www.hammerspoon.org` redirects to `hammerspoon.org`, and suggest forcing https.

cmsj:
We need to loop in @Habbie since he looks after the domain/DNS :)

Habbie:
Both www and the bare domain point to github IPs, so I don't think I need to make any change?

Habbie:
I suspect we can remove 'www' from https://github.com/Hammerspoon/hammerspoon.github.io/blob/master/CNAME to fix it.

cmsj:
Hmm, yeah I just remembered that GitHub is a bit picky about SSL certs, it generates one for precisely the name its given. I don't think I'm in any rush to change this right now :)

cmsj:
Either way, I think this PR is fine, so I'll merge it now.

PR #2427, opened 2020-08-07
Use standard subrole for canvas window

alin23:
I looked through the codebase but I could not find a good reason for having a non-standard subrole for canvas windows.

This would help yabai in [filtering out AXUnknown windows](https://github.com/koekeishiya/yabai/pull/636) and in [drawing a window stack indicator](https://github.com/AdamWagner/stackline/pull/14#issue-462606696) that only reacts to non-canvas windows

### DIFF
--- a/extensions/canvas/internal.m
+++ b/extensions/canvas/internal.m
@@ -901,7 +901,7 @@ static int userdata_gc(lua_State* L) ;
             return _subroleOverride ;
         }
     } else {
-        return [[super accessibilitySubrole] stringByAppendingString:@".Hammerspoon"] ;
+        return [super accessibilitySubrole] ;
     }
 }
 


asmagill:
There is an undocumented `hs.canvas` method -- `canvasObject:_accessibilitySubrole(x)` which allows changing this on a per canvas basis... if `x` is an explicit empty string (i.e. `""`) then it will return the default value; if you specify a specific string, then it will use that instead, and if you specify an explicit `nil`, then it goes back to the current behavior.

This was originally done so that `hs.window.filter` didn't potentially see some of these as "regular" windows that it needed to take notice of. I don't use `hs.window.filter` myself, so I'm not sure what a good test for it would be, but I'll see if I can come up with something this weekend to check before and after trying your fix; or if you do use `hs.window.filter`, what has your experience been specifically when applied to the Hammerspoon application itself?

I originally wanted the method for setting the subrole arbitrarily (`canvasObject:_accessibilitySubrole(x)`) for some additions I wanted to add with a `gui toolkit` set of modules, but that's kind of on long term hold at the moment, so if we can confirm that this doesn't break `hs.window.filter`, then it would be cleaner to remove all of the subrole change code entirely and allow it to inherit the `NSPanel/NSWindow` default.

And if it does break `hs.window.filter`, then I can fairly quickly add a function that you can stick in your Hammerspoon `init.lua` config file that tells canvas to invert the behavior for all canvas objects (i.e. i if the override is `nil`, then return the regular result as opposed to our modified one) for people who use things like `yabai` and don't use `hs.window.filter`.

asmagill:
Without having a good test or user of `hs.window.filter` to test this with, I don't want to make any major changes to the way things work by default -- should I go ahead with the global canvas function to allow users of Yabai to use the standard subrole by default instead of our doctored one? That way we're not changing anything that others may unknowingly be counting on and users of Yabai only need to add one line to their Hammerspoon `init.lua` file.

I ask because I'm trying to get a release ready, and I can easily add a module level toggle for this sometime tomorrow as my final (hopefully!) contribution to the 0.9.79 release push.

dominiklohmann:
It is generally a bad idea to use completely custom window roles or subroles, as it makes filtering windows for both actual accessibility tools and applications like Yabai harder.

Please use the window roles for just what they are: roles. Putting additional information in there makes it more than a role, and the field should not be abused for that purpose.

asmagill:
@cmsj, feel free to override my thinking on this, and I'll go with your decision

The change to the subrole for the original `hs.drawing` objects was made for a reason -- the plethora of drawing objects that were considered "filterable" windows in Hammerspoon was causing issues with `hs.window.filter` and the decision was made because it required the least changes to *our* codebase. `hs.canvas` has since replaced `hs.drawing`, but it's basic approach of putting graphical elements into transparent `NSPanel` objects is still the same and inherited the same override.

Until someone can show me that `hs.window.filter` no longer has an issue with the default subrole (or someone has the time to rewrite it to get around the issue) I am not inclined to introduce a change that will break existing assumptions of our own modules, especially since (1) I don't use `hs.window.filter` so I don't have a good idea of how to test it, and (2) I don't know how many people the change would effect -- it has been in place for multiple years (almost as long as Hammerspoon itself has) and this is the first time anyone has brought up an issue with it.

It's a valid issue, I admit, but one that has a simple workaround that doesn't break things for anyone else and doesn't require a potentially significant rewrite of one of our own modules.

I have almost completed testing of an update to `hs.canvas` that will allow you and others who want to use an external window manager tool to manipulate Hammerspoon and Hammerspoon canvas objects to add one line of code to your `init.lua` file and have the subroles reverted for all canvas objects.

@cmsj, as I stated above, if you disagree, I'll go with your decision.

latenitefilms:
FWIW - We don't actually use `hs.window.filter` in CommandPost, but the few times I've looked into it's code, it's pretty complicated and messy. It also has some performance issues, being all written in Lua. I think long term, `hs.window.filter` would probably be something that could be rewritten in Objective-C/Swift, or depreciated, and replaced with some newer functions that provided similar functionality.

In regards to this specific issue, it should be easy enough to filter out `hs.canvas` objects using `hs.window.filter` if you did revert back to standard sub-roles.

alin23:
@asmagill I understand your concerns here, I wouldn't merge a breaking change either. If there are ways to satisfy this case and keep backwards compatibility I would do that instead.

But the reason I opened this PR in the first place is because I could not find a single instance of `AXUnknown.Hammerspoon` being used in the Hammerspoon codebase. Maybe I'm misunderstanding here, are you saying that users of `hs.window.filter` use this role manually to filter out canvas windows? Or maybe you can point me to a place in the codebase that uses this role.

I could only find these lines:
https://github.com/Hammerspoon/hammerspoon/blob/d93695c83e19e78fbd7fec16206ce642a6862bc3/extensions/window/init.lua#L739
https://github.com/Hammerspoon/hammerspoon/blob/d47d16d80064ab0f26f3b0626d248e446b375262/extensions/window/filter.lua#L1377

I tried to test this change using Stackline which heavily uses `hs.window.filter` but I can't get past this code:
```lua
stackline.wf = wf.new():setOverrideFilter{ -- {{{
    visible = true, -- (i.e. not hidden and not minimized)
    fullscreen = false,
    currentSpace = true,
    allowRoles = 'AXStandardWindow',
} -- }}}
```

I get the following traceback:
```lua
2020-09-09 11:12:17: *** ERROR: ...n.app/Contents/Resources/extensions/hs/window/filter.lua:1431: attempt to index a nil value (field '_element')
stack traceback:
	...n.app/Contents/Resources/extensions/hs/window/filter.lua:1431: in upvalue 'startAppWatcher'
	...n.app/Contents/Resources/extensions/hs/window/filter.lua:1559: in upvalue 'startGlobalWatcher'
	...n.app/Contents/Resources/extensions/hs/window/filter.lua:1649: in upvalue 'checkTrackSpacesFilters'
	...n.app/Contents/Resources/extensions/hs/window/filter.lua:592: in function <...n.app/Contents/Resources/extensions/hs/window/filter.lua:518>
	(...tail calls...)
	/Users/alin/.hammerspoon/stackline/stackline/stackline.lua:12: in main chunk
	[C]: in function 'rawrequire'
	...app/Contents/Resources/extensions/hs/_coresetup/init.lua:642: in function 'require'
	/Users/alin/.hammerspoon/init.lua:13: in main chunk
	[C]: in function 'xpcall'
	...app/Contents/Resources/extensions/hs/_coresetup/init.lua:693: in function 'hs._coresetup.setup'
	(...tail calls...)
```

asmagill:
The string ".Hammerspoon" was added so that `hs.window.filter` automatically *ignored* Hammerspoon canvas objects, rather than detected them... because the subrole wasn't *just* "AXUnknown" (or "AXSystemDialog", if `hs.canvas:clickActivating` is set to false for the object), `hs.window.filter` never even considers the object.

Without adding more complex logic or tracking canvas's as they are created, it was the simplest way to "hide" the drawing/canvas objects from the filter's consideration.

Long term, maybe the best answer *is* to rewrite `hs.window.filter` and stop playing games with the subrole. However, if we want to get a new release out soon, the best short term solution is something like #2453 because anything else is going to take time and testing to avoid breaking something for someone and I wouldn't even know where to begin with it.

alin23:
Yes, https://github.com/Hammerspoon/hammerspoon/pull/2453 seems like a good solution. 
It will allow yabai users to fix the issue with just one line in init.lua

PR #2448, opened 2020-09-04
Remove int to BOOL casts

saagarjha:
This is an error by default with newer compilers.

### DIFF
--- a/LuaSkin/LuaSkin/luaskin.m
+++ b/LuaSkin/LuaSkin/luaskin.m
@@ -184,7 +184,7 @@ static int obj_ud_newindex(lua_State *L) {
     lua_getfield(L, -1, "mutable") ;
     isMutable = isMutable && lua_toboolean(L, -1) ;
     lua_getfield(L, -2, "arrayAutoConversion") ;
-    arrayAutoConversion = (BOOL)lua_toboolean(L, -1) ;
+    arrayAutoConversion = lua_toboolean(L, -1) ;
     lua_pop(L, 3) ;
 
     if (isMutable) {
--- a/extensions/doc/internal.m
+++ b/extensions/doc/internal.m
@@ -278,7 +278,7 @@ static int doc_registerJSONFile(lua_State *L) {
     LuaSkin *skin = [LuaSkin sharedWithState:L] ;
     [skin checkArgs:LS_TSTRING, LS_TBOOLEAN | LS_TOPTIONAL, LS_TBREAK] ;
     NSString *path   = [skin toNSObjectAtIndex:1] ;
-    BOOL     isSpoon = (lua_gettop(L) > 1) ? (BOOL)lua_toboolean(L, 2) : NO ;
+    BOOL     isSpoon = (lua_gettop(L) > 1) ? lua_toboolean(L, 2) : NO ;
 
     // some tricks used to figure out if the docs.json file exists duplicate final "/" before "docs.json"
     // so rather then track them all down, just adjust it here; otherwise we have two "different" paths
--- a/extensions/hid/internal.m
+++ b/extensions/hid/internal.m
@@ -125,7 +125,7 @@ static int hid_led_set(lua_State* L) {
     [skin checkArgs:LS_TSTRING, LS_TBOOLEAN, LS_TBREAK];
 
     NSString *name = [skin toNSObjectAtIndex:1];
-    long target_value = (BOOL)lua_toboolean(L, 2) ? 1 : 0;
+    long target_value = lua_toboolean(L, 2);
     bool ret = false;
 
     if ([name isEqualToString:@"caps"]) {
--- a/extensions/socket/internal.m
+++ b/extensions/socket/internal.m
@@ -588,7 +588,7 @@ static int socket_connected(lua_State *L) {
     [[LuaSkin sharedWithState:L] checkArgs:LS_TUSERDATA, USERDATA_TAG, LS_TBREAK];
     HSAsyncTcpSocket* asyncSocket = getUserData(L, 1);
 
-    lua_pushboolean(L, (BOOL)get_socket_connections(asyncSocket));
+    lua_pushboolean(L, !!get_socket_connections(asyncSocket));
     return 1;
 }
 
--- a/extensions/timer/internal.m
+++ b/extensions/timer/internal.m
@@ -145,7 +145,7 @@ static int timer_new(lua_State* L) {
 
     BOOL continueOnError;
     if (lua_isboolean(L, 3))
-        continueOnError = (BOOL)lua_toboolean(L, 3) ;
+        continueOnError = lua_toboolean(L, 3) ;
     else
         continueOnError = NO ;
 


asmagill:
I'm not sure I like the "double not" approach to fixing these... is there a new warning class for clang that we can possibly disable to resolve this instead? I don't want something as broad as ignoring all implicit casts, but maybe there is something more specific?

What specific error message are you receiving during compilation?

saagarjha:
```
/Users/saagarjha/Git/hammerspoon/LuaSkin/LuaSkin/luaskin.m:187:33: error: cast from function call of type 'int' to non-matching type 'BOOL' (aka 'bool') [-Werror,-Wbad-function-cast]
    arrayAutoConversion = (BOOL)lua_toboolean(L, -1) ;
                                ^~~~~~~~~~~~~~~~~~~~
```

which is a strange flag to put it under but I guess, but I guess they really don't want you casting an integral type to a `BOOL`. However, you can just assign to it directly and it does the same thing, which means that this patch would work too:

```diff
diff --git a/LuaSkin/LuaSkin/luaskin.m b/LuaSkin/LuaSkin/luaskin.m
index a363dc39..1b84d933 100644
--- a/LuaSkin/LuaSkin/luaskin.m
+++ b/LuaSkin/LuaSkin/luaskin.m
@@ -184,7 +184,7 @@ static int obj_ud_newindex(lua_State *L) {
     lua_getfield(L, -1, "mutable") ;
     isMutable = isMutable && lua_toboolean(L, -1) ;
     lua_getfield(L, -2, "arrayAutoConversion") ;
-    arrayAutoConversion = (BOOL)lua_toboolean(L, -1) ;
+    arrayAutoConversion = lua_toboolean(L, -1) ;
     lua_pop(L, 3) ;
 
     if (isMutable) {
diff --git a/extensions/doc/internal.m b/extensions/doc/internal.m
index 0a743674..2424c55e 100644
--- a/extensions/doc/internal.m
+++ b/extensions/doc/internal.m
@@ -278,7 +278,7 @@ static int doc_registerJSONFile(lua_State *L) {
     LuaSkin *skin = [LuaSkin sharedWithState:L] ;
     [skin checkArgs:LS_TSTRING, LS_TBOOLEAN | LS_TOPTIONAL, LS_TBREAK] ;
     NSString *path   = [skin toNSObjectAtIndex:1] ;
-    BOOL     isSpoon = (lua_gettop(L) > 1) ? (BOOL)lua_toboolean(L, 2) : NO ;
+    BOOL     isSpoon = (lua_gettop(L) > 1) ? lua_toboolean(L, 2) : NO ;
 
     // some tricks used to figure out if the docs.json file exists duplicate final "/" before "docs.json"
     // so rather then track them all down, just adjust it here; otherwise we have two "different" paths
diff --git a/extensions/hid/internal.m b/extensions/hid/internal.m
index 6fd634e5..82b10eac 100644
--- a/extensions/hid/internal.m
+++ b/extensions/hid/internal.m
@@ -125,7 +125,7 @@ static int hid_led_set(lua_State* L) {
     [skin checkArgs:LS_TSTRING, LS_TBOOLEAN, LS_TBREAK];
 
     NSString *name = [skin toNSObjectAtIndex:1];
-    long target_value = (BOOL)lua_toboolean(L, 2) ? 1 : 0;
+    long target_value = lua_toboolean(L, 2) ? 1 : 0;
     bool ret = false;
 
     if ([name isEqualToString:@"caps"]) {
diff --git a/extensions/socket/internal.m b/extensions/socket/internal.m
index 7a9aaf2d..6f528e28 100644
--- a/extensions/socket/internal.m
+++ b/extensions/socket/internal.m
@@ -588,7 +588,7 @@ static int socket_connected(lua_State *L) {
     [[LuaSkin sharedWithState:L] checkArgs:LS_TUSERDATA, USERDATA_TAG, LS_TBREAK];
     HSAsyncTcpSocket* asyncSocket = getUserData(L, 1);
 
-    lua_pushboolean(L, (BOOL)get_socket_connections(asyncSocket));
+    lua_pushboolean(L, !!get_socket_connections(asyncSocket));
     return 1;
 }
 
diff --git a/extensions/timer/internal.m b/extensions/timer/internal.m
index 6328194b..9b4f0446 100644
--- a/extensions/timer/internal.m
+++ b/extensions/timer/internal.m
@@ -145,7 +145,7 @@ static int timer_new(lua_State* L) {
 
     BOOL continueOnError;
     if (lua_isboolean(L, 3))
-        continueOnError = (BOOL)lua_toboolean(L, 3) ;
+        continueOnError = lua_toboolean(L, 3) ;
     else
         continueOnError = NO ; 
```

What do you think?

asmagill:
<sigh> I remember putting some of those casts in there because warnings popped up about implicit casting *to* BOOL from int... Wish the clang developers would make up their minds... If I'm feeling generous, I chalk it up to yet one more of the changes prepping for Apple's moving to an ARM architecture and some of the behind the scenes changes that will require... when I'm not feeling generous, I just curse at them in my head...

Ok, I guess the best approach is to just drop the cast altogether... it's certainly easier to read than the double not and doesn't rely on hoping compiler optimization removes the unnecessary code.

I'm hoping to start landing some of the outstanding pulls tomorrow afternoon so we can get 0.9.79 ready for @cmsj to release. If you get a chance to clean this up before then, it would save me some work... otherwise I will when I merge this specific pull.

saagarjha:
I believe the rationale behind it was that on Apple silicon `BOOL` is a one-bit value…wait, that still doesn't really make sense. I really have no idea why the cast is considered illegal. I pushed the version where I got rid of most of the double nots, let me know if there's anything else that I can help with. I'd guess that @cmsj already has some sort of fix for this locally so perhaps he can weigh in on what he'd like to do here.

cmsj:
Yeah this is a fairly silly compiler warning and is only showing up because we have -Weverything (it wouldn't be included in -Wall). I do have some kind of local hack for this, since it shows up on Apple Silicon builds, but I can't remember what I did.

I'll take a look and figure out the least worst option here. What's annoying about it is that I can't find a decent rationale for why it's a problem, beyond some vague posts about maybe casting function returns is just a bad idea in general, but they're not well supported arguments.

asmagill:
I've started merging some of the pull requests and building between each major one on my machine (macOS 10.15.16, Xcode 11.7) but haven't run into this issue yet... out of curiosity, which version of XCode are you using? I think I heard recently about macOS 10.16 going into public beta, and there is usually an XCode beta along side the new macOS releases... are you using that?

I'm planning to hold this specific request until the last merge (a couple of the remaining pulls require conflict resolution, so I hope to get to them tonight or tomorrow morning at the latest)... I'll then check to see if there is in fact a new beta for Xcode and install it so I can make sure that the new additions haven't added any additional places that need to have the BOOL casts removed as well.

saagarjha:
Yep, this is Xcode Version 12.0 beta 6 (12A8189n).

asmagill:
Ok, I was only able to get the errors compiling, even with Xcode 12 beta 6, when compiling for "Any Mac (Apple Silicon, Intel)" (i.e. not "My Mac"), and even then, the proposed fix to `luaskin.m` didn't work -- I had to change it to `arrayAutoConversion = (BOOL)(lua_toboolean(L, -1)) ;`... surprisingly this was the only place where it seemed to matter, the others seemed ok with just removing the `(BOOL)` cast.

Also, as I mentioned above, I'm not a fan of the double nots, but removing them from the `extensions/socket/internal.m` file didn't work either (casting NSInteger to int error, IIRC)... I ended up changing it to `lua_pushboolean(L, get_socket_connections(asyncSocket) ? true : false);` which it did like.

Any objections to landing this pull with the changes outlined in this comment, or do you want to do some more digging on your end?

latenitefilms:
FWIW - `lua_pushboolean(L, get_socket_connections(asyncSocket) ? true : false);` looks and sounds good to me!

asmagill:
Also just confirmed that these changes do not interfere with compiling with Xcode 11.7.

latenitefilms:
I'm still on Xcode 11.3.1 on Mojave - will that still work?

asmagill:
I've uploaded the current master with this pull and my tweaks to a branch on my personal Hammerspoon fork -- you can find it at https://github.com/asmagill/hammerspoon/tree/compileTest if you want to try it with your compiler version.

latenitefilms:
I just tested 22d6863 and all working great on Mojave - thanks @asmagill !

PR #2453, opened 2020-09-09
allow inverting custom subrole logic for canvas windows

asmagill:
Alternative to #2427 

With this pull, adding `hs.canvas.useCustomAccessibilitySubrole(false)` to your Hammerspoon `init.lua` file will ensure that all canvas objects use default Apple standard subroles by default. This should conform to what third-party accessibility enabled tools expect but may cause issues with `hs.window.filter` and other modules that can optionally use it (`hs.window.layout`, `hs.window.switcher`, etc.)

### DIFF
--- a/extensions/canvas/internal.m
+++ b/extensions/canvas/internal.m
@@ -4,6 +4,7 @@
 
 static const char *USERDATA_TAG = "hs.canvas" ;
 static int refTable = LUA_NOREF;
+static BOOL defaultCustomSubRole = YES ;
 
 // Can't have "static" or "constant" dynamic NSObjects like NSArray, so define in lua_open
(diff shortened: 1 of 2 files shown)


asmagill:
@cmsj, before proceeding with the final few pulls, I want your input on a few things:

1. Should I land this pull, a variant of #2427 (rather than change the string returned it would be cleaner to just remove the override method altogether) or leave things as they are.
2. Should we land #2426?

I'm downloading the latest Xcode beta now to see if #2448 is sufficient or requires additional tweaks, expect to have that landed later tonight... after that it's up to you concerning the above and whatever else you might have -- I'll be good for 0.9.79 to land whenever.

PR #2457, opened 2020-09-12
[READY] Added hs.fs.urlFromPath

### LINKED ISSUE #2456: Add hs.fs.urlFromPath()

latenitefilms:
- Fixed a potential memory leak in `hs.fs.pathToBookmark()`
- Closes #2456


extensions/fs/internal.m
```diff
@@ -1090,6 +1091,34 @@ static int fs_pathFromBookmark(lua_State *L) {
...
+///  * path - The path
+///
+/// Returns:
+///  * A string or `nil` if path is invalid.
+static int fs_urlFromPath(lua_State *L) {
+    LuaSkin *skin = [LuaSkin sharedWithState:L] ;
+    [skin checkArgs:LS_TSTRING, LS_TBREAK] ;
+
+    NSString *filePath = [skin toNSObjectAtIndex:1];
+    char *absolutePath = realpath([filePath stringByExpandingTildeInPath].UTF8String, NULL);
```
asmagill, line 1108: More curiosity then anything else... does this differ from NSString's stringByStandardizingPath (https://developer.apple.com/documentation/foundation/nsstring/1407194-stringbystandardizingpath?language=occ)? I've used the NSString method in a couple of places myself, so if there is a difference, we may need to standardize...

extensions/fs/internal.m
```diff
@@ -1090,6 +1091,34 @@ static int fs_pathFromBookmark(lua_State *L) {
...
+
+    if (!absolutePath) {
+        lua_pushnil(L);
+        return 1;
+    }
+    
+    NSURL *fileURL = [[NSURL alloc] initFileURLWithPath:[NSString stringWithFormat:@"%s", absolutePath]];
```
asmagill, line 1115: Probably "faster" to use `[NSString stringWithUTF8String:absolutePath]`

extensions/fs/internal.m
```diff
@@ -1035,6 +1035,7 @@ static int fs_pathToBookmark(lua_State *L) {
                     relativeToURL:nil
                     error:nil];
     [skin pushNSObject:bookmarkData] ;
+    free(absolutePath);
```
asmagill, line 1038: good catch

extensions/fs/internal.m
latenitefilms: I don't think [`stringByStandardizingPath`](https://developer.apple.com/documentation/foundation/nsstring/1407194-stringbystandardizingpath) will convert `~/Desktop` to `/Users/*USERNAME*/Desktop`?

extensions/fs/internal.m
latenitefilms: I tried that but got:

`Implicit conversion from nullable pointer 'NSString * _Nullable' to non-nullable pointer type 'NSString * _Nonnull'`

latenitefilms:
@asmagill - Feel free to make whatever changes are required and merge whenever happy. Thanks heaps!

latenitefilms:
@asmagill - Any ideas why I'm getting this in CommandPost:

```lua
> hs.inspect(hs.fs.urlFromPath("/Users/chrishocking/Desktop/temp/this is a; test 0005.aif"))
{
  __luaSkinType = "NSURL",
  filePath = "/Users/chrishocking/Desktop/temp/this is a; test 0005.aif",
  url = "file:///Users/chrishocking/Desktop/temp/this%20is%20a%3B%20test%200005.aif"
}
```

But this in Hammerspoon:

```lua
> hs.inspect(hs.fs.urlFromPath("/Users/chrishocking/Desktop/temp/this is a; test 0005.aif"))
"file:///Users/chrishocking/Desktop/temp/this%20is%20a%3B%20test%200005.aif"
```

extensions/fs/internal.m
asmagill: Pretty sure it does, but can't test it right now (according to the docs, the first step is `Expanding an initial tilde expression using stringByExpandingTildeInPath.`)

asmagill:
The `hs.sharing` module installs conversion functions for `NSURL` -- try loading `hs.sharing` in Hammerspoon and then try your function. The original LuaSkin fallback for NSURL was to return the string, but conversion functions, if they exist, take precedence.

extensions/fs/internal.m
asmagill: Ok, Apple did get a little more anal about Nullable warnings a few Xcode versions back and (in my opinion) dropped the ball when it came to updating some of their method definitions. Meh, it was just a thought.

asmagill:
I should say, if they exist and are loaded... mostly it isn't a problem because except for NSDate and NSURL, the built in LuaSkin conversions cover only lua equivalent types, so unrecognized classes would either return `nil` or a string containing something along the lines of "unrecognized type", depending upon the options passed to LuaSkin's `pushNSObject:` method. It was only when we started doing more complex things with url's that a need for a specific type became necessary which overrides the built in one.

extensions/fs/internal.m
latenitefilms: As usual, you're absolutely right. I've updated - thanks!

latenitefilms:
@asmagill - I've update some other functions based on your suggestion too. I believe this is now good to go, subject to your review and approval.

latenitefilms:
@asmagill - Ahhhhhhhh, I know now why we were using `realpath([filePath stringByExpandingTildeInPath].UTF8String, NULL);` instead of `[[filePath stringByStandardizingPath] stringByResolvingSymlinksInPath]` - there was a reason for it.

`realpath([filePath stringByExpandingTildeInPath].UTF8String, NULL);` only returns a value if the file/path is valid, whereas `[[filePath stringByStandardizingPath] stringByResolvingSymlinksInPath]` will always return a string.

I think we need to revert back to the old code. Sorry!

asmagill:
IIRC, pathwatcher can detect the creation of paths that didn't exist previously, so if I'm following your logic, it should probably continue to use the `[[filePath stringByStandardizingPath] stringByResolvingSymlinksInPath]` approach; by that same logic, `hs.fs` and `hs.doc` should probably use `realpath` instead.

If you agree with my reasoning, I can throw together a new pull in the next couple of hours to reflect this.

asmagill:
or there is #2463

PR #2463, opened 2020-09-15
Fixed an issue that was introduced in #2457

latenitefilms:
Discussed here: https://github.com/Hammerspoon/hammerspoon/pull/2457#issuecomment-692836030


asmagill:
Think I accidentally hit merge and not comment-and-merge...

Recap -- landing this because my comment in #2457 re `hs.doc` only impacts developers when updating core module outside of main source (to reduce compile time during testing) and not main users... fixing it might prevent an annoyance (duplicate doc messages to console) but only during development, not once changes merged back to core.

Will re-assess if issue becomes more common, but for now, this should be sufficient.

PR #2464, opened 2020-09-16
[READY] Fixed bug in hs.fs.urlFromPath()

latenitefilms:
Some paths with non-alphanumeric characters (such as single quotation marks) could break this function - it would return a URL string, but the result wouldn’t be valid for `hs.pasteboard.writeObjects({url=path})` for example. I assume it’s because of the `UTF8String`.

### DIFF
--- a/extensions/fs/internal.m
+++ b/extensions/fs/internal.m
@@ -1112,7 +1112,8 @@ static int fs_urlFromPath(lua_State *L) {
         return 1;
     }
     
-    NSURL *fileURL = [[NSURL alloc] initFileURLWithPath:[NSString stringWithFormat:@"%s", absolutePath]];
+    NSString *urlPath = [[filePath stringByStandardizingPath] stringByResolvingSymlinksInPath];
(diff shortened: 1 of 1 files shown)


asmagill:
How certain/close are you the you've caught the edge cases you care about with urls and paths now? Should I wait a bit before testing and merging this?

latenitefilms:
I've now tested this with a whole bunch of random strings, and I can't break it, so I THINK it's all good to merge. I pushed out a public beta of CommandPost last night, and got feedback back pretty quickly that it was broken, and this fix seems to solve those complaints.

We now just use `realpath([filePath stringByExpandingTildeInPath].UTF8String, NULL)` to check the file exists - so there's no longer any conversion happening - it's all `NSString`.

asmagill:
I'm going to wait until morning (my time) just to be certain... but it lgtm.

asmagill:
Ok, so not morning, but it's in there now 😜

PR #2471, opened 2020-09-19
updated internal lua docs for 5.4

asmagill:
Ok, easier than I had hoped.

In case someone else needs to do this in the future and I'm not available, I actually included the necessary file at https://github.com/Hammerspoon/hammerspoon/blob/master/extensions/doc/makeLuaDocs.lua when I did this last time.

It's not included in the distributed application, so you'd need to clone the repository, then from the console do `doFile("/..path.to.repository/extensions/doc/makeLuaDocs.lua")` which will create `lua.json` in your config dir. Move this to `/..path.to.repository/extensions/doc/lua.json` and submit a new pull request.

I'll do a more thorough spot-check tomorrow, but it looks good so far.


asmagill:
Stickler you bogon -- it's not a lua file! it's a json file! ... and probably is too large for you anyways...

Issue #2474, opened 2020-09-20
Window filter subscribe is broken on 0.9.79


folke:
Running the test script below gives an error.

```lua
filter = hs.window.filter
filter.setLogLevel(10)
filter.default:subscribe(filter.windowFocused, function() print("foo") end)
```

Output:
```log
2020-09-20 20:42:24: 20:42:24 wf-default: hs.window.filter.default (0x6000028e4680) instance deleted
2020-09-20 20:42:24:             wfilter: unregistered 0 apps
2020-09-20 20:42:24: -- Lazy extension loading enabled
2020-09-20 20:42:24: -- Loading ~/.hammerspoon/init.lua
2020-09-20 20:42:24: -- Loading extension: window
2020-09-20 20:42:24: -- Loading extensions: window.filter
2020-09-20 20:42:24: 20:42:24 wf-default: new hs.window.filter: wf-default (0x6000028f7840) - empty
2020-09-20 20:42:24:          wf-default: setting filter for Flux: reject
2020-09-20 20:42:24:          wf-default: setting filter for Music Manager: reject
2020-09-20 20:42:24:          wf-default: setting filter for CheatSheet: reject
2020-09-20 20:42:24:          wf-default: setting filter for loginwindow: reject
2020-09-20 20:42:24:          wf-default: setting filter for MenuMetersApp: reject
2020-09-20 20:42:24:          wf-default: setting filter for CursorSense Manager: reject
2020-09-20 20:42:24:          wf-default: setting filter for Google Drive: reject
2020-09-20 20:42:24:          wf-default: setting filter for Alfred 2: reject
2020-09-20 20:42:24:          wf-default: setting filter for CrashPlan menu bar: reject
2020-09-20 20:42:24:          wf-default: setting filter for PopClip: reject
2020-09-20 20:42:24:          wf-default: setting filter for Grandview: reject
2020-09-20 20:42:24:          wf-default: setting filter for Bartender: reject
2020-09-20 20:42:24:          wf-default: setting filter for Notification Center: reject
2020-09-20 20:42:24:          wf-default: setting filter for 1Password mini: reject
2020-09-20 20:42:24:          wf-default: setting filter for SystemPal: reject
2020-09-20 20:42:24:          wf-default: setting filter for Isolator: reject
2020-09-20 20:42:24:          wf-default: setting filter for BetterSnapTool: reject
2020-09-20 20:42:24:          wf-default: setting filter for Radium: reject
[...]

PR #2483, opened 2020-09-21
hs.window.filter fixes

asmagill:
May address #2474 and linked issues.

1. Previous `hs.uielement.watcher` was a table and `hs.window.filter` checked for pid by accessing table directly instead of through methods. New `hs.uielement.watcher` is a userdata, so you methods required.

2. `hs.uielement.watcher` objects were not having metatable removed during garbage collection, resulting in Hammerspoon crash during reload if `hs.window.filter` used anywhere. Metatable is now removed during garbage collection.

@cmsj, I'm not entirely happy about #2 because all it really does is push back the error into lua space (which a reload can survive). Somehow we've got a userdata (perhaps in a queued callback) that still "exists" even though its ref count has reached 0, which is the real issue. Thoughts on how to try and narrow down the search to a better solution?

### DIFF
--- a/extensions/uielement/watcher.m
+++ b/extensions/uielement/watcher.m
@@ -158,6 +158,8 @@ static int userdata_gc(lua_State* L) {
             watcher = nil;
         }
     }
+    lua_pushnil(L) ;
+    lua_setmetatable(L, 1) ;
     return 0;
 }
 
--- a/extensions/window/filter.lua
+++ b/extensions/window/filter.lua
@@ -1387,7 +1387,7 @@ appWindowEvent=function(win,event,_,appname,retry)
     end
     if apps[appname].windows[id] then return log.df('%s (%d) already registered',appname,id) end
     local watcher=win:newWatcher(windowEvent,appname)
-    if not watcher._element.pid then
+    if not watcher:pid() then
       log.wf('%s: %s has no watcher pid',appname,role or (win.role and win:role()))
       if retry>MAX_RETRIES then log.df('%s: %s has no watcher pid',appname,win.subrole and win:subrole() or (win.role and win:role()) or 'window')
       else
@@ -1412,7 +1412,7 @@ local function startAppWatcher(app,appname)
   local watcher = app:newWatcher(appWindowEvent,appname)
   watcher:start({uiwatcher.windowCreated,uiwatcher.focusedWindowChanged})
   App.new(app,appname,watcher)
-  if not watcher._element.pid then
+  if not watcher:pid() then
     log.wf('No accessibility access to app %s (no watcher pid)',(appname or '[???]'))
   end
 end
@@ -1428,7 +1428,7 @@ local function startAppWatcher(app,appname,retry,nologging)
   if retry>1 and not pendingApps[appname] then return end --given up before anything could even happen
 
   local watcher = app:newWatcher(appWindowEvent,appname)
-  if watcher._element.pid then
+  if watcher:pid() then
     pendingApps[appname]=nil --done
     watcher:start({uiwatcher.windowCreated,uiwatcher.focusedWindowChanged})
     App.new(app,appname,watcher)


latenitefilms:
Crazy idea that might not be relevant... could the usage of `hs.uielement.watcher` be replaced with `hs.axuielement` now that's merged in for the purposes of `hs.window.filter`?

asmagill:
Probably but that's a bigger rewrite then I want to tackle at the moment... maybe in a couple of weeks, but right now, after we put out the fires that are to be expected with a new release (especially one as broad as this one was), I'd like to concentrate on a couple of modules that actually interest me 😁

latenitefilms:
Understood and fair enough. I've tried looking at `hs.window.filter` in the past, but it's pretty confusing and complicated. The one time I tried to fix the @sticker-ci errors in `hs.window.filter` I broke it, haha. 

I feel like something like this would be better implemented in Objective-C-land anyway for performance reasons. I would imagine there's also probably faster/easier methods in Apple's own APIs for filtering windows?

Regarding point 2, I wonder if this commented out code is relevant?

https://github.com/Hammerspoon/hammerspoon/commit/08b1213f459fac653b73ecfcd9a679e69023b56e#diff-2eca8514adcbcbcd0642d8320e0f6b31L19

latenitefilms:
Also, looking at `hs.uielement` again now, I still think it's pretty confusing?

For example:

```lua
> hs.uielement.focusedElement()
hs.uielement: 0x6000028ff7b8
```

It returns a `hs.uielement` object - in this case, a window:

```lua
> hs.uielement.focusedElement():isWindow()
true

> hs.uielement.focusedElement():role()
AXWindow
```

...however, I can't actually access it as a `hs.window` object, so I'm not sure how useful it is?

Also, if `hs.uielement` only ever returns a `hs.uielement` object that's a window or an application, how exactly do you use `hs.uielement:selectedText()`?

**Related:** 
- Better hs.uielement Documentation? #1793

cmsj:
Thanks @asmagill this is some nasty fallout from the modernisation of hs.application/hs.window/hs.uielement.

I agree we should focus right now on getting things working again, but long term I love the idea of being able to remove hs.uielement - it was always a painful integration point that I should probably never have allowed to merge in the form it took.

I don't have any thoughts currently for solving #2 better, but this seems patch seems like a sensible precaution for now. I'll merge this shortly and push out a new release.

asmagill:
I *suspect* it's some sort of mis-match between `selfRefCount` and ARC's internal retain count for the `HSUIElementWatcher` object itself, but trying to resolve that conflict by either artificially adjusting `selfRefCount` in the module's `start` and `stop` methods, or by removing `watcherRef` altogether (you already have the `HSUIElementWatcher` object in the callback through the observer's context, so you don't really *need* `watcherRef` except as a hedge against the fact that storing the `HSUIElementWatcher` as a `void*` in the context doesn't guarantee that ARC hasn't collected the object)... but none of this helped. And gives the same error -- `:pid()` getting called during collection which means that *something* is still trying to execute a callback even after it's collected (and `selfRefCount` reaches 0), but I can't tell what...

I'm on board with the "this stops the crashing" for now... we may be leaking some memory, but I've banged my head against this enough for the moment.

- - -

So I can let the back of my brain churn for a while on the requirements of a rewrite using `hs.axuielement` (which, incidentally, doesn't use the observer context, so there is less worry about ARC's retain count getting out of sync through mechanisms outside of our control), I want to check -- `hs.uielement` is never used directly by itself, correct? It's providing a common set of functionality for watchers for `hs.window` and `hs.application` but `hs.uielement` itself could be removed entirely as long as those "common functions" were implemented another way, correct?

A quick grep through the extensions folder shows that `hs.uielement` is used primarily by `hs.window.filter` and indirectly by `hs.tabs` to dereference event types (so maybe `hs.uielement` will still "exist" but as a mostly empty wrapper so that others using the same dereferencing won't break)?

I guess I'm going to have to become a lot more familiar with `hs.window.filter` before considering the rewrite since that seems to be the part that relies on it the most. And that gets immediate issues filed when it breaks 😜

PR #2488, opened 2020-09-22
hs.window fix: move moveScreen commands back into metatable

### LINKED ISSUE #2476: `hs.window` method broken on 0.9.79
Calling this function will raise error after upgraded to `v0.9.79`.

```lua
function send_window_next_screen()
  hs.notify.show("Next Screen", "", "")
  local win = hs.window.focusedWindow()
  win:moveOneScreenEast(0)
end
```

Error in console:
```
2020-09-21 07:29:22: 07:29:22 ERROR:   LuaSkin: hs.hotkey callback: /Users/myuser/.hammerspoon/init.lua:24: attempt to call a nil value (method 'moveOneScreenEast')
stack traceback:
	/Users/myuser/.hammerspoon/init.lua:24: in function 'send_window_next_screen'
	/Users/myuser/.hammerspoon/init.lua:13: in function </Users/myuser/.hammerspoon/init.lua:13>
```

asmagill:
Fixes #2476

### DIFF
--- a/extensions/window/init.lua
+++ b/extensions/window/init.lua
@@ -943,15 +943,16 @@ do
     return window[k]
   end
   local mt=getmetatable(window)
-  if mt.__index==uielement then
-    --inject "lazy loading" for submodules
-    mt.__index=function(t,k)
-      if t==window and submodules[k] then return loadSubModule(k)
-      else return uielement[k] end
+  --inject "lazy loading" for submodules
+  mt.__index=function(_,k)
+    if submodules[k] then
+        return loadSubModule(k)
+    else
+        return nil -- if it's already in the module, __index is never called
     end
-    -- whoever gets it first (window vs application)
-    if not mt.__call then mt.__call=function(t,...) return t.find(...) end end
   end
+  -- whoever gets it first (window vs application)
+  if not mt.__call then mt.__call=function(t,...) return t.find(...) end end
 end
 
 return window
--- a/extensions/window/init.lua
+++ b/extensions/window/init.lua
@@ -743,11 +743,11 @@ function window.frontmostWindow()
 end
 
 for n,dir in pairs{['0']='East','North','West','South'}do
-  window['windowsTo'..dir]=function(self,...)
+  objectMT['windowsTo'..dir]=function(self,...)
     self=self or window.frontmostWindow()
     return self and windowsInDirection(self,n,...)
   end
-  window['focusWindow'..dir]=function(self,wins,...)
+  objectMT['focusWindow'..dir]=function(self,wins,...)
     self=self or window.frontmostWindow()
     if not self then return end
     if wins==true then -- legacy sameApp parameter
@@ -755,7 +755,7 @@ for n,dir in pairs{['0']='East','North','West','South'}do
     end
     return self and focus_first_valid_window(window['windowsTo'..dir](self,wins,...))
   end
-  window['moveOneScreen'..dir]=function(self,...) local s=self:screen() return self:moveToScreen(s['to'..dir](s),...) end
+  objectMT['moveOneScreen'..dir]=function(self,...) local s=self:screen() return self:moveToScreen(s['to'..dir](s),...) end
 end
 
 --- hs.window:focusWindowEast([candidateWindows[, frontmost[, strict]]]) -> boolean


asmagill:
@cmsj, is this still correct? https://github.com/Hammerspoon/hammerspoon/blob/master/extensions/window/init.lua#L946-L951

~~~lua
  if mt.__index==uielement then
    --inject "lazy loading" for submodules
    mt.__index=function(t,k)
      if t==window and submodules[k] then return loadSubModule(k)
      else return uielement[k] end
    end
~~~

`uielement` isn't defined in the file (locally or globally) and this causes an error (well a confusing one) if you try to access a non-existent `hs.window` function:

~~~lua
> hs.window.abc
...oon.app/Contents/Resources/extensions/hs/window/init.lua:950: attempt to index a nil value (global 'uielement')
stack traceback:
	...oon.app/Contents/Resources/extensions/hs/window/init.lua:950: in metamethod 'index'
	[string "return hs.window.abc"]:1: in main chunk
	[C]: in function 'xpcall'
	...app/Contents/Resources/extensions/hs/_coresetup/init.lua:514: in function <...app/Contents/Resources/extensions/hs/_coresetup/init.lua:494>
~~~

I think it should be the following, but want your opinion in case I'm missing something about your changes to window/uielement/application before adding it to this pull (or would you rather it be a separate pull?):

~~~lua
  --inject "lazy loading" for submodules
  mt.__index=function(_,k)
  if submodules[k] then
    return loadSubModule(k)
  else
    return nil if it actually exists in `window`, __index is never consulted
  end
~~~

latenitefilms:
@asmagill - Random thought... rather than doing this "lazy loading" in the individual extension, I wonder if this would be better handled in `_coresetup`? We currently use [`hs._coresetup.loader`](https://github.com/CommandPost/CommandPost/blob/develop/src/extensions/hs/_coresetup/loader.lua) in CommandPost, which may be relevant?

asmagill:
Possibly, but probably a bigger undertaking to migrate to that at the moment... I think we do something similar in a few other places and to add a new module (or at least its code) just to fix one of these without vetting the rest... maybe for a future release.

asmagill:
@cmsj, do we want to just go ahead and make the `hs.window` module __index metamethod check against the `hs.window` object metatable as well? It seems that more than one of the recent issues are related to people knowingly or unknowingly taking advantage of the fact that they were the same table in the past.

Pros:
* it makes *some* constructs a little easier to read: `hs.fnutils.ifilter(hs.window.allWindows(), hs.window.isStandard)` vs `hs.fnutils.ifilter(hs.window.allWindows(), function(o) return o:isStandard() end)`
* apparently some of our users (and modules?) may have been taking advantage of this anyways and will now be faced with errors to correct.

Cons:
* it confuses the distinction between function and method and gets into descriptions of what "self" is for a method ("Why does `hs.window.isStandard(obj)` require an argument but `obj:isStandard()` doesn't? Are they different in some way?")
* it begs the question why do we do this for `hs.window` and not *all* modules.

OTOH, I have done similar shenanigans in some of my own creations `__index` functions to make certain syntaxes easier or at least visually more appealing (accessing `hs.canvas` object elements as elements of the object's array or accessing actions and attributes in `hs.axuielement` as keys and methods, as two examples)

Thoughts?

cmsj:
@asmagill I do take your point about the confusion this change causes, but I think we'd be making a mistake if we keep supporting the mistakes that we made in the past. These core modules are already complicated enough, without having `__index` hacks to maintain a backwards compatibility we never promised.

If anything, I'd rather mirror the method names over to the module table and point them at a generic "you should be using an object method instead of this module function" type error.

asmagill:
Which is actually a harder "hack", in the sense that it would take more code... but I see your point.

Ok, at worst we'll see a short spate of issues that we can use a boiler plate answer for. At this point, however, we may have seen most of them, unless there are a number of people sticking with 0.9.78 until the hubbub dies down.

bgmort:
I'm one of those users who apparently was doing something wrong, and is now having to figure out why.  I was using the following:

    hs.fnutils.filter(hs.application.find(appName):allWindows(), hs.window.isStandard)

which apparently shouldn't have worked, but did, and now it doesn't.

I was able to figure it out, but it took a while to track down and fix, and I figured it was a new bug, not a fix, until I came in to report it and found this issue. Are you sure you don't want to maintain compatibility for everyone who hasn't been able to figure out (or hasn't noticed) their hs.window bugs yet?

latenitefilms:
@cmsj & @asmagill - Maybe we need a "pinned" issue explaining any known issues and major changes?

snarfed:
this bit me too. i'm on 0.9.81, my small config's `hs.window.filter` usage is broken, and i don't know how to fix it. it's been tricky to follow all the different issues here, but as far as i can tell, the conclusion is that i have latent bugs or otherwise unsupported lua code, which i now need to debug and fix? i'm not sure how though. @latenitefilms maybe your proposed pinned issue would include help? @bgmort any tips on how to debug?

here's an example stack trace i get on 0.9.81 with `focusWindowEast` (#2515 may be related):
```
ERROR:   LuaSkin: hs.hotkey callback: ...n.app/Contents/Resources/extensions/hs/window/filter.lua:2309: attempt to call a nil value (field '?')
stack traceback:
	...n.app/Contents/Resources/extensions/hs/window/filter.lua:2309: in method 'focusWindowEast'
	/Users/ryan/.hammerspoon/init.lua:9: in function 'focusRight'
```

code from my config is:
```
filt = hs.window.filter.default:setDefaultFilter{windowInCurrentSpace, windowVisible}

function focusRight()
    if hs.grid.get(hs.window.focusedWindow()).x == hs.grid.getGrid(hs.screen.mainScreen()).w - 1 then
       -- wrap around
       wins = filt:windowsToWest(nil, true, true)
       wins[1]:focus()
    else
      filt:focusWindowEast(nil, true, true)
    end
end
hs.hotkey.bind({"cmd"}, "L", focusRight)
```

thanks in advance!

Issue #2465, opened 2020-09-18
window/filter.lua throws "attempt to index a nil value" when trying to start a global watcher


sprak3000:
I am writing a spoon that iterates over all the windows on your screens and re-positions them.  This is the code loop that spawns the error:

```
    app:activate()
    local windows = hs.window.filter.new(appTitle):getWindows()

    for k, v in pairs(windows) do
        obj.logger.d('Positioning window ' .. v:id() .. ' of app ' .. appTitle)
        v:moveToScreen(screen)
        v:setFrame(frame, 0)
    end
```

Randomly, the attempt dies out leaving this error from the Hammerspoon console:

```
2020-09-17 21:39:39: ********
2020-09-17 21:39:39: 21:39:39 ERROR:   LuaSkin: hs.menubar:setClickCallback() callback error: ...n.app/Contents/Resources/extensions/hs/window/filter.lua:1561: attempt to index a nil value (field 'watcher')
stack traceback:
	...n.app/Contents/Resources/extensions/hs/window/filter.lua:1561: in upvalue 'startGlobalWatcher'
	...n.app/Contents/Resources/extensions/hs/window/filter.lua:1722: in upvalue 'start'
	...n.app/Contents/Resources/extensions/hs/window/filter.lua:1838: in method 'getWindows'
	...s/luis/.hammerspoon/Spoons/ArrangeDesktop.spoon/init.lua:38: in function 'ArrangeDesktop._positionApp'
	...s/luis/.hammerspoon/Spoons/ArrangeDesktop.spoon/init.lua:58: in function 'ArrangeDesktop.arrange'
	...s/luis/.hammerspoon/Spoons/ArrangeDesktop.spoon/init.lua:83: in function <...s/luis/.hammerspoon/Spoons/ArrangeDesktop.spoon/init.lua:83>
2020-09-17 21:39:39: ********
```

The line throwing this error is `global.watcher:start()` with `global.watcher` being set a few lines above by `global.watcher = appwatcher.new(appEvent)`. I tried going into the `appwatcher.new` code to see what happens there and why it might be returning nil at times. However, I'm not as familiar with the code in `watcher.m` to make a reasonable guess.

[...]

PR #2502, opened 2020-09-23
[READY] Fixed a regression in hs.eventtap.event:post([app])

### LINKED ISSUE #2498: 0.9.80 broke hs.eventtap.event:post()?
I have some code that basically does the following (the snippet in questions send a `;` keypress to the `YT Music` application which plays/pauses playback). Since upgrading to 0.9.80 it does not seem to work any more. The simplified example that follows does not either - no error thrown, it just doesn't seem to do anything. I've tried sending other keystrokes and to other applications with no apparent success.

```
> yt = hs.application.find("YT Music")
> e = hs.eventtap.event.newKeyEvent(";", true)
> e:post(yt)
hs.eventtap.event: Event type: 10 (0x6000019bc938)
```

latenitefilms:
- Closes #2498

### DIFF
--- a/extensions/eventtap/event.m
+++ b/extensions/eventtap/event.m
@@ -1,4 +1,5 @@
 #import "eventtap_event.h"
+#import "HSuicore.h"
 @import IOKit.hidsystem ;
 
 #define FLAGS_TAG "hs.eventtap.event.flags"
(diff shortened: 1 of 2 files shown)


asmagill:
So you didn't need to make any additional linkings in the subproject? That's good to know. I'll test this out later tonight and, barring any surprises, land it.

latenitefilms:
Nope, this works as-is. If I get a chance today, I'll try have a look around the other extensions and see if there's any other `hs.window`/`hs.application` stuff that's broken.

latenitefilms:
FWIW - I've done a search for `AXUIElementRef`, and I don't spot any other obvious problems. It's only really used in `HSuicore`, `hs.application`, `hs.axuielement`, `hs.uielement` and `hs.window`.

asmagill:
In `hs.webview` and `hs.console`, there is a method, `:hswindow()` but in both cases the relevant code is:

~~~objc
    CGWindowID windowID = (CGWindowID)[theWindow windowNumber];

    [skin requireModule:"hs.window"] ;
    lua_getfield(L, -1, "windowForID") ;
    lua_pushinteger(L, windowID) ;
    lua_call(L, 1, 1) ;
~~~

This is safe because it's leveraging `hs.window.windowForID` which means all of the relevant heavy lifting is already in `hs.window` and not in assumptions in `hs.webview` or `hs.console`.

While I still want to remove the code for these, since it's written in a way that won't cause problems, we can safely put that off for a while.

asmagill:
verified

asmagill:
Oh, @cmsj, my request re holding up 81 here has passed, though I do hope we can fix #2465 first.

PR #2505, opened 2020-09-23
Reduce HSuicore log spamming

asmagill:
@cmsj, do you need this log info in here anymore? Background apps don't have an NSAccessibilityFrontmostAttribute, and to be honest, all accessibility attributes  are technically optional anyways.

It causes a lot of console log spamming, especially if you use `hs.window.filter`

If you do think it's worth capturing, can we reduce the level (possibly even to breadcrumb?)? It would save some confusion.


cmsj:
I think we can live without it :)

PR #2506, opened 2020-09-23
[READY] Fixed bugs in Hammerspoon tests

latenitefilms:
- Fixed issue in the hs.fs testVolumes test

### DIFF
--- a/extensions/fs/test_fs.lua
+++ b/extensions/fs/test_fs.lua
@@ -317,7 +317,8 @@ function testVolumes()
       os.execute("diskutil rename "..ramdiskName.." "..ramdiskRename)
     end
     if event == hs.fs.volume.didRename then
-      newPath = info.path:match("(/Volumes/"..ramdiskRename..")/?$")
+      -- NOTE: in this case, `info` is returned as a NSURL object:
+      newPath = info.path and info.path.filePath:match("(/Volumes/"..ramdiskRename..")/?$")
       if not newPath then return end
       hs.fs.volume.eject(newPath)
       volumeWatcher:stop()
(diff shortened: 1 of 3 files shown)


latenitefilms:
@asmagill - You might want to idiot check the first commit (1a06ae8), just to make sure I'm not missing something stupid. Obviously this test must have worked correctly at some point - so I assume something has changed with how NSURL's are handled?

I'm still looking at some other failed tests in the meantime.

latenitefilms:
@asmagill - Here's an interesting observation. The `testTabs` test is failing because:

```lua
> hs.application.get("Safari")
hs.application: Safari (0x600003bd3178)

> hs.application.get("Safari"):mainWindow()
nil

> hs.inspect(hs.application.get("Safari"):allWindows())
{}
```

...it seems Safari 13.1.2 (14609.3.5.1.5) on Mojave isn't returning any window objects - at least not in this branch?

latenitefilms:
The good news is that all but one (testTabs) of the tests now pass.

Screen Shot 2020-09-23 at 4 13 15 pm

I'm assuming the problem with testTabs is not the test itself, but something in `hs.window`.

latenitefilms:
Ah, actually, `testTabs` is also related to accessibility permissions. I toggled Accessibility Permissions on and off and now I'm seeing:

```lua
> hs.inspect(hs.application.get("Safari"):allWindows())
{ <userdata 1> -- hs.window: Hammerspoon docs: hs.window (0x600003db8ef8), <userdata 2> -- hs.window:  (0x600003db8cf8) }
```

latenitefilms:
Ok, I give up on `testTabs` - it's failing because of lack of Accessibility Permissions, but for some reason, toggling them on and off in System Preferences isn't helping for this particular test.

cmsj:
@latenitefilms thanks for doing this. I'll see what my local Xcode Server CI makes of the tab tests (that VM runs with a lot of hacks to disable permission requests)

Issue #2507, opened 2020-09-23
latest version breaks me in at


maxandersen:
after #2481 being fixed I've now updated and past that issue but bumped into something that never broke before too:

my config hasn't changed but eventually I get this error:

```
2020-09-23 14:38:16: 14:38:16 ERROR:   LuaSkin: hs.hotkey callback: ...n.app/Contents/Resources/extensions/hs/window/tiling.lua:110: not a rect
stack traceback:
	[C]: in function 'error'
	...n.app/Contents/Resources/extensions/hs/geometry/init.lua:659: in function 'hs.geometry.toUnitRect'
	...n.app/Contents/Resources/extensions/hs/window/tiling.lua:110: in function 'hs.window.tiling.tileWindows'
	...n.app/Contents/Resources/extensions/hs/window/layout.lua:455: in upvalue 'performPendingActions'
	...n.app/Contents/Resources/extensions/hs/window/layout.lua:593: in function 'hs.window.layout.apply'
	...n.app/Contents/Resources/extensions/hs/window/layout.lua:685: in function 'hs.window.layout.resume'
	...n.app/Contents/Resources/extensions/hs/window/layout.lua:884: in function 'hs.window.layout.resumeAllInstances'
	/Users/max/code/dotfiles/.hammerspoon/init.lua:471: in function 'toggleLayout'

```

line init.lua:471 is an "else" line but I assume it hits:

    hs.window.layout.resumeAllInstances()
or
    hs.window.layout.pauseAllInstances()

maxandersen:
note, restart of hammerspoon things works fine - it just stops working eventually...not sure why/when yet

asmagill:
`hs.window.layout` uses `hs.window.filter`, so could you try the fix described in https://github.com/Hammerspoon/hammerspoon/issues/2465#issuecomment-697154463 (also a pending pull at https://github.com/Hammerspoon/hammerspoon/pull/2509 if you prefer) and see if that makes a difference?

maxandersen:
ive copied it into the file and restarted - now we wait to see if it fails again :)

maxandersen:
sorry clicked the worng button ;)

vritser:
same error

asmagill:
as in you're getting this same error? are you running the latest official release of Hammerspoon?

maxandersen:
FYI. I've not had the issue since I reported it and tried using the suggestion.

vritser:
[...]

PR #2509, opened 2020-09-23
guard from stopGlobalWatcher interferring with startGlobalWatcher through badly timed watcher __gc

asmagill:
Since this seems to be helping out with #2465, I figured I'd go ahead and put it here.

Comments welcome, as I'm still not totally confident in my understanding of `hs.window.filter`.

### DIFF
--- a/extensions/window/filter.lua
+++ b/extensions/window/filter.lua
@@ -1549,12 +1549,15 @@ local function startGlobalWatcher()
     preexistingWindowCreated[id]=time+id-999999
   end
   global.watcher = appwatcher.new(appEvent)
+  global.inStartGlobalWatcher = true
   local runningApps = application.runningApplications()
   log.f('registering %d running apps',#runningApps)
   for _,app in ipairs(runningApps) do
     startAppWatcher(app,app:name())
   end
   global.watcher:start()
+  global.inStartGlobalWatcher = nil
+
 end
 
 local function stopGlobalWatcher()
@@ -2065,7 +2068,8 @@ function WF:delete()
   self.log.i(self,'instance deleted')
   activeInstances[self]=nil spacesInstances[self]=nil applicationActiveInstances[self]=nil
   self.events={} self.filters={} self.windows={}
-  setmetatable(self,nil) stopGlobalWatcher()
+  setmetatable(self,nil)
+  if not global.inStartGlobalWatcher then stopGlobalWatcher() end
 end
 
 


cmsj:
I'm fine with this. Maybe @lowne would like to return and assist :D

latenitefilms:
I wonder if this problem is just due to the fact `hs.application` may be slightly slower now?

asmagill:
As far back as I recall, we've had the occasional random issue pop up with `hs.window.filter` that didn't always got satisfactorily resolved before just "disappearing on its own" or "couldn't be replicated"... one of the reasons I've personally avoided the module (and I usually found an easier way than learning a new module to do what I wanted)... it's conceivable that its a combination of changed collection metrics *and* speed differences and we've just been lucky with when garbage collectors happened to get invoked before... as I said in the analysis that started this, I can't *prove* 5.4 changed when collectors get invoked because the relevant verbiage in the docs is the same, but I never observed it while something else was running before, and while `debug.traceback` isn't my go-to debugging tool, I have used it like this before.

cmsj:
@asmagill should we merge this and quickly push out a 0.9.81?

latenitefilms:
I believe we're now using generational garbage collection now which does behave very differently:

```
In generational mode, the collector does frequent minor collections, which traverses only objects recently created. If after a minor collection the use of memory is still above a limit, the collector does a stop-the-world major collection, which traverses all objects. The generational mode uses two parameters: the minor multiplier and the the major multiplier.

The minor multiplier controls the frequency of minor collections. For a minor multiplier x, a new minor collection will be done when memory grows x% larger than the memory in use after the previous major collection. For instance, for a multiplier of 20, the collector will do a minor collection when the use of memory gets 20% larger than the use after the previous major collection. The default value is 20; the maximum value is 200.

The major multiplier controls the frequency of major collections. For a major multiplier x, a new major collection will be done when memory grows x% larger than the memory in use after the previous major collection. For instance, for a multiplier of 100, the collector will do a major collection when the use of memory gets larger than twice the use after the previous collection. The default value is 100; the maximum value is 1000.
```

cmsj:
@latenitefilms interesting.

For now we can choose to switch back to the mark&sweep collector, but knowing the Lua folk, I doubt we'll see that sticking around indefinitely.

asmagill:
@cmsj, Ideally I'd like @szymonkaliski's comments as he seems to be the most active one familiar with `hs.window.filter`, and it would be nice to see if this does in fact help with #2507, but we *are* starting to see duplicate issues and with this pull, 0.9.81 will definitely be better than people running 0.9.79 or 0.9.80, so I guess, yeah, I'm ok with that.

asmagill:
@latenitefilms The change (if it is indeed one and we just didn't observe closely enough at the right times before) is less about when things are marked for finalization and more about when the `__gc` methods are included in the parsing of lua commands...

There is no technical reason for the parser *not* to stick a call to some object's `__gc` right before a function returns... all that happens is a stack guard (or whatever method Lua uses to keep function call stacks isolated), the function is pushed on to the stack, followed by an object (table or userdata), `lua_(p)call` is invoked, and the guard is released -- the outside function would have no idea that it had ever happened if `debug.traceback` wasn't invoked at just the right time.

We just assumed, perhaps because of observational bias and not from a careful reading of the documentation, that it *didn't* happen this way. And maybe it didn't, in 5.3, though the docs indicated that we should code as if it did. And it's really only an issue when a finalizer can impact a global or up-value that a running function might depend on, so even if we got lucky in the past, it's only a very specific confluence of conditions where it would matter... `hs.window.filter` happens to be one of those.

asmagill:
So far, we've had nothing but positive feedback regarding this pull, so unless someone has objections, I'm going to land this later today.

PR #2511, opened 2020-09-23
[WIP] Updated hs.fs to use LuaFileSystem 1.8.0

### LINKED ISSUE #2346: Upgrade hs.fs with lfs 1.8.0
As `hs.fs` is at lest partially based on `lfs`, it needs to be updated as version 1.8.0 of lfs has been released and specifies compatibility with 5.4, which we expect to be moving to soon.

I've not tested the breaking code he gives below, but in any case, we should update for 5.4 if for no other reason.

> From the lua-l mailing list:
> 
> Hi list,
> 
> I am happy to announce LuaFileSystem 1.8.0, its first release in 3 years.
> 
> LuaFileSystem is a Lua library developed to complement the set of
> functions related to file systems offered by the standard Lua
> distribution. LuaFileSystem offers a portable way to access the
> underlying directory structure and file attributes.
> 
> https://keplerproject.github.io/luafilesystem/
> 
> What's new:
> 
> * lfs.link and lfs.symlinkattribute now work on Windows!
> * Fixed memory leak in case of realloc failure
> * MACOSX_DEPLOYMENT_TARGET is configurable in the Makefile
> * Fallback to _POSIX_PATH_MAX when MAXPATHLEN is not avaliable
> * lfs.lock uses CREATE_ALWAYS instead of CREATE_NEW on Windows
> 
> and last but not least:
> 
> * Lua 5.4 support, including __close handling in the lfs.dir()
> iterator. This requires no changes from the user, and ensures that the
> directory will be closed correctly even if you break from the loop.
> For example, with `uname -n` set below 10000, this script crashes on
> Lua < 5.4, and works in Lua 5.4:
> ~~~
>     local lfs = require("lfs")
>     for t = 1, 10000 do
>       local i = 0
>       for e in lfs.dir("/usr/lib") do
>          i = i + 1
>          if i == 100 then
>             break
>          end
>       end
>     end
> ~~~
> (This is not a forced example: I have hit this problem before when
> recursing over large directory trees and collecting files.)
> 
> The easiest way to install LuaFileSystem is to use LuaRocks:
> 
>    luarocks install luafilesystem
> 
> This release contains contributions by Dan Church, Stephen E. Baker,
> Kıvanç Çakmak, TsT, Peter Melnichenko, James McCoy, Joshua Root,
> Eroica, 云风 and myself.
> 
> Cheers,
> 
> -- Hisham
> _______________________________________________

latenitefilms:
- Closes #2346

### DIFF
--- a/extensions/fs/internal.m
+++ b/extensions/fs/internal.m
@@ -490,7 +490,16 @@ static int dir_iter_factory (lua_State *L) {
         lua_pushnil(L);
         lua_pushfstring(L, "cannot open %s: %s", path, strerror (errno));
     }
+
+    // Lua 5.4: use __close to close dir if you break the iterator
+    // SOURCE: https://github.com/keplerproject/luafilesystem/commit/842505b6a33d0b0e2445568ea42f2adbf3c4eb77
+    #if LUA_VERSION_NUM >= 504
+      lua_pushnil(L);
+      lua_pushvalue(L, -2);
+      return 4;
+    #else
     return 2;
+    #endif
 }
 
 
@@ -511,6 +520,13 @@ static int dir_create_meta (lua_State *L) {
     lua_setfield(L, -2, "__index");
     lua_pushcfunction (L, dir_close);
     lua_setfield (L, -2, "__gc");
+    
+    // Lua 5.4: use __close to close dir if you break the iterator
+    // SOURCE: https://github.com/keplerproject/luafilesystem/commit/842505b6a33d0b0e2445568ea42f2adbf3c4eb77
+    #if LUA_VERSION_NUM >= 504
+      lua_pushcfunction(L, dir_close);
+      lua_setfield(L, -2, "__close");
+    #endif
     return 1;
 }
 


latenitefilms:
@cmsj & @asmagill - I believe this is good to go, keeping in mind all I did was copy and paste between LuaFileSystem and `hs.fs`. Running the `hs.fs` tests all pass.

asmagill:
check https://github.com/keplerproject/luafilesystem/blob/842505b6a33d0b0e2445568ea42f2adbf3c4eb77/src/lfs.c#L315-L337 vs https://github.com/Hammerspoon/hammerspoon/blob/5850c6879869207b7cf703e028cc0c86cbbb3b05/extensions/fs/internal.m#L170-L183

While I suspect it was just truncated when originally copied in, I think we stuck in the 5.1 version and not the 5.2+ one... a little surprised this hasn't bitten us yet. @cmsj, I think you added this originally? Was there a reason?

latenitefilms:
Ah, interesting! I just stupidly assumed it was already up-to-date based on a quick glance at release dates. Will investigate and compare.

latenitefilms:
You're absolutely right, it looks like it's very out-of-date. It seems to have been synced around 16 Jan 2015 after [this commit](https://github.com/keplerproject/luafilesystem/commit/0a82aae451c5c95ceacb5fcb235f4d1ea964f5f5#diff-f6b7bcefa5f97fba6fa9e82b43357724). I'll try go through the file history and see if I can manually update.

cmsj:
Hmm, I have no idea why I did that. I now regret almost everything about the way I merged lfs 😫

asmagill:
Let me take a crack at lfs tonight... I want to try seeing if we can create a way to just drop in the source file without requiring cutting and pasting as I outlined in #2346... @cmsj mentioned that he was concerned about collisions if the user had also installed `lfs` via luarocks, but I think I can work around that.

asmagill:
Closing this in favor of #2526; reopoen or resubmit if you disagree

Issue #2104, opened 2019-05-15
Add ability for hs.eventtap to use the global queue


latenitefilms:
If I trigger the below code in the console:

```lua
hs.eventtap.event.newKeyEvent(hs.keycodes.map.shift, true):post()
```

...I expect it to hold down the SHIFT key until I trigger:

```lua
hs.eventtap.event.newKeyEvent(hs.keycodes.map.shift, false):post()
```

...however it doesn't seem to do anything?

I've also tried `hs.eventtap.event.newKeyEvent({}, hs.keycodes.map.shift, true):post()`.

Any ideas?

latenitefilms:
Any ideas what I'm doing wrong @asmagill ?

asmagill:
The event "queue" that `hs.eventtap.event` uses is not the global one, but rather one unique to Hammerspoon (I remember trying the global one at one point and it broke some things people were used to so didn't really pursue it further, though when I was working on an eventtap replacement (which might be worth reconsidering for 2.0) I did make this a choice you could set when creating the event).

Because of this, the "shift" key isn't "pressed" globally (i.e. events originating outside of Hammerspoon do not see this flag as having been set)

As an example, try the following in the console:

~~~lua
t = true ; d = hs.hotkey.bind({"cmd","alt"}, "S", nil, function() hs.eventtap.event.newKeyEvent(hs.keycodes.map.shift, t):post() ; t = not t ; end)
~~~

Now, press and release Cmd-Alt-S. Type something into the console (without the shift key pressed) and you'll see it in lower case, *BUT* try this in the console:

~~~lua
hs.eventtap.event.newKeyEvent(hs.keycodes.map.a, true):post() ; hs.eventtap.event.newKeyEvent(hs.keycodes.map.a, false):post()
~~~

And you'll see an upper case `A` appear in the console's input field.

To verify it's because of our hotkey, press and relase Cmd-Alt-S again, and then repeat the sample posts... this time, the input field receives a lower case `a`.

latenitefilms:
Ah, I see - thanks for the detailed explanation! Is there any workaround for this?

Is your eventtap replacement on GitHub?

latenitefilms:
@cmsj & @asmagill - This issue has cropped up again, so if you have any ideas let me know!

@asmagill - I haven't been able to find your `hs.eventtap` replacement anywhere - is it on GitHub?

[...]

PR #2512, opened 2020-09-24
Added hs.eventtap.event.newGesture()

### LINKED ISSUE #1434: Add hs.eventtap.event.newGesture() to simulate gestures
On my MacBook Pro, in Final Cut Pro, it's possible to use the Trackpad to "Zoom in or out" (i.e. Pinch with two fingers). However, what I'd like to be able to do is do the same thing using the scrollbar on my mouse (i.e. use the scroll wheel on my mouse to do exactly the same thing hat "pinch to zoom" currently does on my Trackpad).

Is there anyway to "simulate" the Trackpad's "Zoom in or out" function using `hs.eventtap`?

I'm assuming `hs.eventtap.event.types.NSEventTypeMagnify` is what I need, but I don't know how I actually "trigger" it.

Any ideas @asmagill or @cmsj ?

latenitefilms:
- Added `hs.eventtap.event.newGesture` with support for `beginSwipeLeft` and `endSwipeLeft`.
- Added `TouchEvents` to Hammerspoon
- Added swipe gestures for right, up and down.
- Currently these events only seem to work in Hammerspoon.
- Changed `kCGSessionEventTap` to `kCGHIDEventTap`, however this seems to have no impact.
- Will eventually close #1434


latenitefilms:
@asmagill - I finally got around to tackling #1434 myself.

The good news is that it works. The bad news is that it only seems to work within Hammerspoon.

For example, using this test code:

```lua
hs.hotkey.bind({"cmd", "alt", "ctrl"}, "q", function()
	local activeApp = hs.application.frontmostApplication()
	a = require("hs.eventtap.event").newGesture("beginSwipeDown")
	b = require("hs.eventtap.event").newGesture("endSwipeDown")
	print(string.format("Simulating a Swipe Down for: %s", activeApp))	
	a:post(activeApp)
	b:post(activeApp)	
	print(string.format("Simulating a Swipe Down"))	
	a:post()
	b:post()	
end)
```

I can press CMD+OPTION+CONTROL+Q to make the Hammerspoon Console window to scroll down, but it doesn't work in any other application.

I tried changing `kCGSessionEventTap` to `kCGHIDEventTap` but it hasn't helped.

This seems to be the same issue I ran into in #2104.

What's curious though is that [`SensibleSideButtons`](https://github.com/archagon/sensible-side-buttons) seems to work fine, and I can't see any real differences at all in the code.

Hopefully someone smarter than I can work it out. I wonder if @archagon has any suggestions?

latenitefilms:
Actually, it could just be because I had swipes disabled in "System Preferences > Trackpad > More Gestures", as I'm not able to trigger swipes in [TouchRecorder](https://github.com/calftrail/Touch/tree/master/TouchRecorder) either. Let me try some other gestures.

latenitefilms:
Interestingly, if I use this Lua code:

```
hs.hotkey.bind({"cmd", "alt", "ctrl"}, "q", function()
	print("swiping up")
	a = require("hs.eventtap.event").newGesture("beginSwipeUp")
	b = require("hs.eventtap.event").newGesture("endSwipeUp")
	a:post()
	b:post()
end)
```

...then press COMMAND+OPTION+CONTROL+Q, I do see "Swipe" get ticked in TouchRecorder, so it is actually working between apps, which is great.

asmagill:
Could be an issue of timing or place in application queue then.  Try posting `b` in a timer (e.g. `hs.timer.doAfter(hs.math.minFloat, function() b:post() end)`... may need to adjust delay or keep timer in upvalue to prevent collection if delay has to be significantly larger (with 5.3 for less than a second *usually* didn't matter  unless HS was slowed down enough that it really was more than a second, but I don't know about 5.4)

latenitefilms:
Note to self - this is also using the same code:

https://github.com/noah-nuebling/mac-mouse-fix

latenitefilms:
@asmagill - This all seems to be working correctly now! Woohoo!

Here's some test code:

```lua
hs.hotkey.bind({"cmd", "alt", "ctrl"}, "1", function()
	print("Time to Magnify!")
	a = require("hs.eventtap.event").newGesture("beginMagnify", 0)
	b = require("hs.eventtap.event").newGesture("endMagnify", 0.1)
	a:post()	
	hs.timer.doAfter(hs.math.minFloat, function() b:post() end)
end)
hs.hotkey.bind({"cmd", "alt", "ctrl"}, "2", function()
	print("Time to Swipe Down!")
	a = require("hs.eventtap.event").newGesture("beginSwipeDown", 0)
	b = require("hs.eventtap.event").newGesture("endSwipeDown", 3)
	a:post()
	hs.timer.doAfter(hs.math.minFloat, function() b:post() end)
end)
hs.hotkey.bind({"cmd", "alt", "ctrl"}, "3", function()
	print("Time to Rotate!")
	a = require("hs.eventtap.event").newGesture("beginRotate", 0)
	b = require("hs.eventtap.event").newGesture("endRotate", 3)
	a:post()
	hs.timer.doAfter(hs.math.minFloat, function() b:post() end)
end)
```

asmagill:
@latenitefilms Looking forward to trying this out later, but I have a few questions re rotate and magnify... your examples show `startXXX` with the optional value of 0 and the `endXXX` with a different value... is this a "start from" and "end at" kind of thing? What if the start event has a different number than 0? If the end event is negative does it occur in the opposite direction (i.e. "un-magnify" and "reverse rotation")? Do the numbers map to some sort of objective measure (e.g. for rotation by degrees or radians) or is it more of a "relative" thing, and if so, do we know relative to what?

latenitefilms:
To be honest, I have no idea off the top of my head. I'll do some testing and experimenting then update the documentation accordingly. I think, at least with magnify, it's a start value, end value kind of scenario, because smaller values make it magnify less. I've found Preview.app to be a good place to test.

latenitefilms:
I don't think it really matters (as there's already a lot of parse and semantic issues, as well as use of deprecations in Hammerspoon), but just incase you have any easy quick fixes...

This function uses some depreciated functions:

```objc
// AbsoluteToNanoseconds' is deprecated: first deprecated in macOS 10.8
// 'UpTime' is deprecated: first deprecated in macOS 10.8
static inline uint64_t tl_uptime() {
	AbsoluteTime uptimeAbs = AbsoluteToNanoseconds(UpTime());
	return ((uint64_t)uptimeAbs.hi << 32) + uptimeAbs.lo;
}
```

[This code](https://stackoverflow.com/a/4753909) seems like it might be a suitable replacement (with some modifications), but it goes a little over my head. Any recommendations?

Also, there's quite a few:

```objc
// Use of GNU empty initializer extension
AvgPosition avgTouch = {};
```

...is there an easy fix for these?

asmagill:
What exactly are they trying to get with that function? From the links you point to (and the Apple one they point to) I'm guessing a rough uptime in nano-seconds? (I say rough because I'm not aware of *any* method that measures nanoseconds that includes time asleep, though to be honest its been a couple of years since I searched).

This is what we do in `hs.timer` that is based on Apple's tech doc (though it will be interesting to see if this is going to require a change with the upcoming "Apple Silicon" platform):

https://github.com/Hammerspoon/hammerspoon/blob/master/extensions/timer/internal.m#L431-L451.

As to the gnu message, I think we used to end the tables of lua functions and methods with just `{}`, but switched to `{ NULL, NULL}` to suppress this or a similar message. Whether that works here or not will depend upon what members of an `AvgPosition` array contain... `static const luaL_Reg XXX[]` is an array of structures which contain two pointers (a `const char*` and a `* lua_CFunction`).

latenitefilms:
Legend, thanks! I'll have a play.

latenitefilms:
@asmagill - I've updated the documentation as best I can, and tested. I believe this is now good to go! Any questions let me know.

latenitefilms:
I have no idea what codecov/project is testing, and when I try view the site it just says:

```
Missing base report.
Unable to compare commits because the base of the pull request did not upload a coverage report.

Changes found in between c6d852c...4d46b06 (pseudo...base) which prevent comparing this pull request.
```

...so I'm just going to ignore whatever that failed test is.

@asmagill - Feel free to make any changes you want, and merge when ready.

latenitefilms:
@asmagill - Have you had a chance to look at this yet? No rush on merging it into Hammerspoon, but curious to get your general thoughts/ideas, because I'd love to merge it into CommandPost sooner rather than later.

asmagill:
@latenitefilms haven't forgotten -- this is on my list for today/tomorrow!

asmagill:
@latenitefilms Sorry it took so long...

All of your examples work for me except for the swipe down... what application did you try this in?

I'll try crafting some of my own examples now to further test, but figured we want to make sure any examples given in the docs definitely work since that's what people will often try first.

asmagill:
@latenitefilms FWIW, on an intel machine running Catalina, the synthetic swipe gesture seems to be ignored as well. Were you trying it in a specific application? I have observed some differences in how pressure events are received by eventtap that are dependent upon the "receiving" or "active" application, perhaps swipes are as well?

latenitefilms:
Sorry for the delayed reply!

From memory, I think all my tests were done with Safari, Photos and Pages.

asmagill:
I got swipe left and swipe right to work, but not up or down... I had to attach a Magic Mouse (the first generation mouse with the touch sensitive top) and enable swipes for it -- enabling them for the trackpad didn't seem to be sufficient. Maybe because the mouse was the only device that had a swipe option for single touch swiping? I was hoping that up/down would work as well, since the Magic Mouse scrolls with a single touch swipe up/down, but no dice.

Looking at the source for TouchEvents.c, I wonder if we need to provide a touches array (rather than the empty array you currently pass in) for swipes? I suspect that way you could specify how many fingers. Not sure what the entries in the array need to contain yet, though...

I'll play around some more. The other gestures work as advertised, though, even before I attached the mouse.

asmagill:
FWIW, this is the code I was testing with:

~~~lua
local event  = require("hs.eventtap.event")
local hotkey = require("hs.hotkey")
local timer  = require("hs.timer")
local math   = require("hs.math")

local sKeys = {
   h = "Left",
   j = "Down",
   k = "Up",
   l = "Right",
}

local mods = { "cmd", "shift", "alt", "ctrl" }

for key, dir in pairs(sKeys) do
    hotkey.bind(mods, key, function()
        print("Swipe "..dir)
        event.newGesture("beginSwipe"..dir, 0):post()
        timer.doAfter(math.minFloat, function() event.newGesture("endSwipe"..dir):post() end)
    end)
end
~~~

latenitefilms:
@asmagill - Apologies for the delayed reply! I'll try and review this properly over the next couple of weeks!

It was definitely working for me in Mojave - but I think I had to enable swipe gestures in the Trackpad preferences.

jonmackinder:
Following.
I want to add pinch zoom to my trackman marble mouse for use in Logic Pro. 
The Trackman is replacing my Magic Trackpad.

latenitefilms:
@asmagill - I finally got around to testing this out again. Apologies for the delay!

I can confirm that on macOS Mojave 10.14.6:

- I couldn't get [TouchRecorder](https://github.com/calftrail/Touch/tree/master/TouchRecorder) to detect any virtual gestures this time, and I have no idea why. Maybe a change elsewhere in Hammerspoon? It's detecting my physical Trackpad gestures correctly.
- Rotate in Safari (zooms into a page) & Preview work correctly.
- Swipe left/right works in Safari (changes tabs).
- Magnify works in Preview in Markup Mode.

I haven't been able to get anything to detect/trigger swipe up/down, however, swiping up and down seems to mostly be using for scrolling up and down, so I'm wondering if mouse position data needs to be included with the gesture for it to actually do anything?

cmsj:
@latenitefilms what's the status on this PR? Is  it working/ready?

latenitefilms:
I believe it's working/ready. The only thing @asmagill and I couldn't work out is triggering a swipe up/down, but that could just be because very few applications are using these gestures. I'd suggest we merge, then fix any issues if the community come back with any feedback.

cmsj:
Ignore this close/re-open, just testing something.

Issue #2521, opened 2020-09-26
`hs.reload` randomly crashes post-update


smackesey:
Not sure if this is something specific to my (fairly complicated) configuration, but I've found that reloading hammerspoon's config (either by clicking the button in the console window or via `hs.reload`) randomly crashes Hammerspoon (maybe 20% of the time?) since updating from 0.9.78. Including a Crash Report from Console.app for your convenience:

[Hammerspoon_2020-09-26-145550-1_stm-mbp-2016.crash.txt](https://github.com/Hammerspoon/hammerspoon/files/5287388/Hammerspoon_2020-09-26-145550-1_stm-mbp-2016.crash.txt)

asmagill:
Which version of Hammerspoon are you running (0.9.79 or 0.9.80)? There is a known crashing bug with `HSuicore.m` in 0.9.79 that should have been fixed in 0.9.80, so knowing the version will tell us if this has already been addressed or if this is something new.

If you're not sure, you can type `hs.processInfo.version` into the Hammerspoon console.

smackesey:
0.9.80

asmagill:
💩, that means its something new... or the last fix wasn't complete... it's only happening during a reload, correct? Do you happen to know what you're using in your code that utilizes a window watcher? `hs.window.filter` or something that uses it? a specific Spoon? one of your own custom callbacks?

From the crash log it doesn't look like the crash is occurring during the collecting and release of the outgoing Lua state, which is different... I'll take a closer look at `HSuicore` and see what might be able to persist through MJLuaDestroy (maybe an already queued callback? but the callback deref should scotch that...), as off the top of my head I can't think of anything that *should*.

smackesey:
Yes, only during a reload (that I've observed). Though, I'm not 100% confident that it wouldn't happen in other conditions, as I've recently been changing config and reloading a lot.

The only code I'm aware of I have that uses a window watcher (I think this qualifies?) is below. Both of these functions are called on reload:

```
-- ----- APP

-- Listen for changes in the active application or application termination
-- events. On either of these events, refresh the `windowTap` (so that it is now
-- listening on a new application). When a new application is activated,
-- trigger the window change callback.
function m:makeAppTap()
  self.appTap = hs.application.watcher.new(function (name, eventType, app)
    local activated = (eventType == hs.application.watcher.activated)
    if activated then
      self.windowTap:stop()
      self:makeWindowTap(app)
[...]

PR #2522, opened 2020-09-27
make retain and release of elementRef in HSuielement explicit

asmagill:
Hopefully addresses #2521 

Ensures that `elementRef` is retained and released when `HSuielement *` object is.


asmagill:
@cmsj please review and see if my logic makes sense. I'm not able to replicate the problem described in #2521, either with or without this, but I think my additions do make the code a little safer, but would like your opinion in case I missed something in my review of your code.

cmsj:
@asmagill agreed, this looks safer than my version, thanks! Feel free to merge :)

PR #2523, opened 2020-09-27
[READY] Skip hs.window Snapshots & Size tests in Travis

latenitefilms:
- These tests run fine locally on my machine but fail in Travis.
- I’m on a MacBook Pro (15-inch, 2017) running macOS Mojave 10.14.6.

### DIFF
--- a/Hammerspoon Tests/HSwindow.m
+++ b/Hammerspoon Tests/HSwindow.m
@@ -44,6 +44,7 @@
 }
 
 - (void)testSnapshots {
+    SKIP_IN_TRAVIS() // Added by @latenitefilms
     RUN_LUA_TEST()
 }
 
@@ -60,6 +61,7 @@
 }
 
 - (void)testSize {
+    SKIP_IN_TRAVIS() // Added by @latenitefilms
     RUN_LUA_TEST()
 }
 


latenitefilms:
FYI: This is what I had to add in #2512 to get around the Travis failing issues.

However, I do note that Travis is using Xcode 11.5 and macOS 10.15.4, so it's possible the reason the tests are failing is because of a change in macOS Catalina.

@asmagill - If you're running Catalina, do you want to see if the `hs.window` tests are working for you locally?

asmagill:
You're running these directly from within Xcode, correct?

And are you running it as the whole suite or as individually triggered tests? I know in the past when I've tried running the entire suite, there are always one or two tests that fail, but pass when I run them individually (and always a handful that fail no matter what, but I gather those are the one's you're trying to addressing with this pull).

latenitefilms:
Yes, directly in Xcode. I try the whole suite first, then if any fail, go back and try them individually. They all should pass if run individually.

asmagill:
Running Catalina 10.15.7 with Xcode 12.0.1 (12A7300), both of the tests you refer to in this pull pass.

`testTabs` (HSWindowTests) and `testFrontmostApplication` (HSApplicationTests) don't, but both of those are already marked `SKIP_IN_TRAVIS()`. Kind of curious why they fail on my machine, though... but that's a chore for another day.

latenitefilms:
In that case, I think this pull request is good to merge, as it'll mean Travis tests will be more useful for detecting other issues.

Curious that `testTabs` and `testFrontmostApplication` fail. I'll create another issue for that.

PR #2526, opened 2020-09-28
[WIP] Updated hs.fs to use LuaFileSystem 1.8.0 take 2

asmagill:
Should supersede #2511 

Update includes #2511 and:
* fixed missing `blocks` and `blksize` from `hs.fs.attributes` and `hs.fs.symlinkAttributes`
* added `target` attribute to `hs.fs.symlinkAttributes`
* modified some error messages to match format used in lfs 1.8.0
* updated `hs.fs.currentDir` to match lfs 1.8.0
* fixed `check_file` helper to properly use `luaL_stream` (as per lfs 1.8.0)

I do have a question about `hs.fs.dir` though... the example given in the docs is as follows:

~~~lua
       local iterFn, dirObj = hs.fs.dir("/Users/Guest/Documents")
       if iterFn then
          for file in iterFn, dirObj do
             print(file)
          end
       else
          print(string.format("The following error occurred: %s", dirObj))
       end
~~~

This is in the style required for 5.1, but since 5.2, you can also do:

~~~lua
for file in hs.fs.dir(path) do
    print(file)
end
~~~

Updating the comment is easy, but there's a catch... you'll no longer be able to check if the first argument is `nil` and then decide what to do with the error... you'll get this instead:

~~~
[string "for file in hs.fs.dir("~booger") do print(fil..."]:1: attempt to call a nil value
stack traceback:
   ...
~~~

With the stock `lfs`, they actually throw an error at that point (in fact they throw them at a number of places where we return `nil, errMsg`) and I tend to think, even with the other changes in how `hs.fs` handles errors vs stock `lfs` that we should too for this generator function -- it's more in line with what I think 5.2+ programmers expect from a generator function -- an error that says what's actually wrong (such as `cannot open /Users/amagill/.config/hammerspoon/~l: No such file or directory` instead of `attempt to call a nil value`.

Thoughts?

### DIFF
--- a/extensions/fs/init.lua
+++ b/extensions/fs/init.lua
@@ -2,9 +2,7 @@
 ---
 --- Access/inspect the filesystem
 ---
---- Home: http://keplerproject.github.io/luafilesystem/
----
---- This module is produced by the Kepler Project under the name "Lua File System"
+--- This module is partial superset of LuaFileSystem 1.8.0 (http://keplerproject.github.io/luafilesystem/). It has been modified to remove functions which do not apply to macOS filesystems and additional functions providing macOS specific filesystem information have been added.
 
 local module = require("hs.fs.internal")
 module.volume = require("hs.fs.volume")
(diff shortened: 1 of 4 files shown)


latenitefilms:
Quick thought - I wonder if we need to include all the checks for Lua versions in Hammerspoon given we're basically committed to using the latest and greatest Lua release?

asmagill:
I think most of them have already been removed. I do have a question regarding one of them in lfs (https://github.com/keplerproject/luafilesystem/issues/140), so we'll see if I get a response.

Most of our modules are our own, but this one in particular is based to a large degree on a code base that isn't... maybe, just as a prod when the next major Lua version comes out, we *invert* the checks -- if the lua version is > 504 break, just so we know that we should verify that there hasn't been an upstream change to accommodate the new version?

asmagill:
That presuposes we have an idea of what the changes in a new version would be... probably not then.

asmagill:
Looks like `hs.sqlite3` also uses `LUA_VERSION_NUM` checks... have we verified that there haven't been any major updates to it recently?

latenitefilms:
> Looks like `hs.sqlite3` also uses `LUA_VERSION_NUM` checks... have we verified that there haven't been any major updates to it recently?

I haven't used `hs.sqlite3` properly yet - only done some basic tests with it thus far.

asmagill:
**YAY** Travis finally doesn't hate me anymore! Well, for the moment at any rate.

@cmsj, I'm going to land this and then, unless @latenitefilms has any outstanding concerns or objections, I think we're ready for 0.9.81.

latenitefilms:
All sounds good to me! Release away!

Unless you want to get try and review `hs.eventtap.event.newGesture()` whilst we're waiting on @cmsj? I'm pretty keen to play with this in CommandPost, but no stress or pressure if you don't have time in the short term.

Issue #2527, opened 2020-09-28
Review testTabs and testFrontmostApplication tests on Catalina


latenitefilms:
As discussed [here](https://github.com/Hammerspoon/hammerspoon/pull/2523#issuecomment-699804260), the `testTabs` and `testFrontmostApplication` are failing for @asmagill on macOS Catalina 10.15.7 with Xcode 12.0.1 (12A7300), whereas they're passing for me on macOS Mojave 10.14.6 with Xcode 11.3.1. Travis CI already skips these tests.

asmagill:
Re `testTabs `, it's because Safari 14 handles tabs in a custom way... the existing code in `HSuicore.m` doesn't see them. Not sure yet of a work around (it's probably a different AXUIElement type we'll need to look for if the first check returns 0, not sure yet), but for now it's probably safe to say that if anyone is running Safari 14, the tab functionality of `hs.window` probably doesn't work.

Investigating.

latenitefilms:
Ah, interesting. I was using Safari 13 on my MacBook Pro, but I've just updated to Safari 14, so I'll see if I can reproduce too.

asmagill:
`testFrontmostApplication` appears to be a timing thing, at least when I manually try what the test does:

Given:

~~~lua
function testFrontmostApplication()
  hs.openConsole()
  local app = hs.application.frontmostApplication()
  app:activate()
  assertIsNotNil(app)
  assertTrue(app:isFrontmost())
  assertTrue(app:setFrontmost())

  return success()
end
~~~

I tried something like this:

~~~lua
> hs.hotkey.bind({"alt"}, "b", function()
  hs.openConsole()
  local app = hs.application.frontmostApplication()
  app:activate()
  print(app, app:isFrontmost(), app:setFrontmost())
end)
table: 0x606000b3c360
~~~

Now, if I tap `alt-b` when the Hammerspoon console is in the foreground, of course I get:
~~~
hs.application: Hammerspoon (0x606000941cf8)	true	true
~~~

But if I go to another application and *then* tap `alt-b`:
~~~
hs.application: Hammerspoon (0x60600029d238)	false	true
~~~

I can't recall the last time I ran the tests on my machine before the recent questions about them, so I can't say if this behavior is an artifact of changes we've made, because I'm running Catalina, because I'm using Xcode 12, or because my machine is older and thus a little slow compared to some.

latenitefilms:
I think it's safe to close this issue then. Thanks!

PR #2528, opened 2020-09-28
Safari 14 kludge for accessing Tabs

asmagill:
Safari 14 no longer keeps tabs in AXTabGroup element; rather in a regular AXGroup. This searches both.

Also fixes off by 1 for `hs.window:focusTab(#)`

Should partially address #2527, though not formally tested locally yet.

### DIFF
--- a/Hammerspoon/HSuicore.m
+++ b/Hammerspoon/HSuicore.m
@@ -518,6 +518,28 @@ static AXUIElementRef get_window_tabs(AXUIElementRef win) {
         }
     }
 
+    // Safari 14 puts them into an AXGroup, not an AXTabsGroup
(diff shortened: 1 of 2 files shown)


asmagill:
Should have added that while I haven't run the actual test suite yet (its really slow and I'm up *WAY* past my bed time), this works both with Safari 14 (which uses AXGroup) and with the Finder (which uses AXTabGroup).

latenitefilms:
Looks good to me! Merge away!

Tested on my MacBook Pro (15-inch, 2017) with macOS Mojave 10.14.6 (18G3020) running Xcode 11.3.1.

Screen Shot 2020-09-28 at 10 42 40 pm

PR #2529, opened 2020-09-28
[REVIEW] Updated hs.sqlite3 from 0.9.4 to 0.9.5

latenitefilms:
I manually compared the changes from [0.9.5](http://lua.sqlite.org/index.cgi/zip/lsqlite3_fsl09y.zip?uuid=fsl_9y) to the code in `hs.sqlite3` (0.9.4), and made the required changes.

Definitely worth checking @cmsj and/or @asmagill - although there's only a couple of minor changes, so should be pretty easy to review.

### DIFF
--- a/extensions/sqlite3/init.lua
+++ b/extensions/sqlite3/init.lua
@@ -3,7 +3,7 @@
 --- Interact with SQLite databases
 ---
 --- Notes:
----  * This module is LSQLite 0.9.4 as found at http://lua.sqlite.org/index.cgi/index
+---  * This module is LSQLite 0.9.5 as found at http://lua.sqlite.org/index.cgi/index
 ---  * It is unmodified apart from removing `db:load_extension()` as this feature is not available in Apple's libsqlite3.dylib
 ---  * For API documentation please see [http://lua.sqlite.org](http://lua.sqlite.org)
 return require("hs.sqlite3.lsqlite3")
--- a/extensions/sqlite3/lsqlite3.c
+++ b/extensions/sqlite3/lsqlite3.c
@@ -1,4 +1,4 @@
-// HAMMERSPOON NOTE: This file has been modified from the original 0.9.4 release. Be careful when updating it
+// HAMMERSPOON NOTE: This file has been modified from the original 0.9.5 release. Be careful when updating it.
 
 /************************************************************************
 * lsqlite3                                                              *
@@ -1182,10 +1182,7 @@ static int collwrapper(scc *co,int l1,const void *p1,
     lua_rawgeti(L,LUA_REGISTRYINDEX,co->ref);
     lua_pushlstring(L,p1,l1);
     lua_pushlstring(L,p2,l2);
-    if (lua_pcall(L,2,1,0)==0) {
-        double d=lua_tonumber(L,-1);
-        res=(int)d;
-    }
+	if (lua_pcall(L,2,1,0)==0) res=(int)lua_tonumber(L,-1);
     lua_pop(L,1);
     return res;
 }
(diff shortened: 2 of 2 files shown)


asmagill:
These are *really* the only changes? The collapsing of a a few lines the compiler probably optimizes anyways and adding an unused variable into a function?

Sorry for the time you spent, but that hardly seems worth being called an update... were there other changes to code in sections that we don't use (I seem to recall there was something we disabled because the Apple library wasn't compiled with support for it)?

latenitefilms:
Yes, there were some changes to the code in the sections that @cmsj has previous commented out, so I didn't bother bringing those changes across.

latenitefilms:
FYI - the release notes just say:

```lua
Version "0.9.5"

Since the "0.9.4-devel" release of this Lua library...

Lua 5.4 compatibility (backward compatible with Lua 5.3, 5.2, and Lua 5.1.5).
```

asmagill:
Hmm... If the code is commented out, but tracking the original source, I wonder if we should keep it up to date anyways? If we decide later we want to enable that section (if Apple changes how their library is built or if we decide its worth the trouble of including our own build of the sqlite3 library) I wouldn't want to have a mix of versions within the module code itself.

That, or remove the commented out code completely, maybe with a note that unneeded code was removed, so we *know* we have to go back to the source if we want to enable currently disabled portions.

latenitefilms:
Actually, looking at it again, the only difference is the type of comments anyway:

Screen Shot 2020-09-29 at 3 41 36 pm

asmagill:
I guess it depends upon how closely our `lsqlite3.c` tracks the module source... if it's an almost identical copy, then I think we should keep it as close to that as possible. If it's already the result of a lot of pruning and additions, like with `lfs`, then I guess removing the commented code, but leaving a note, makes more sense. At least to me.

asmagill:
Then I'm really not seeing how they can call this a new version, unless there were more changes in other supporting files that we don't include. Nothing in your original pull, and none of the comment changes, make one whit of difference to 5.4...

latenitefilms:
The only reason the comments are different in this case is because @cmsj has commented out this entire section (so he had to change `/* */` to `//`).

There's also lots of these sorts of changes (Hammerspoon is on the left):

Screen Shot 2020-09-29 at 3 45 38 pm

There's also a `sqlite3.h` file that's not included in Hammerspoon - so maybe that's had big changes - I'll check.

asmagill:
The casts on the HS side were most likely added to conform to our warning settings... unless you're saying they were in the earlier version as well? They look exactly like what I do when placating warnings, though...

latenitefilms:
Comparing `lsqlite3.c` in 0.9.4 to 0.9.5 there's actually no changes that affect Hammerspoon - so I reckon maybe the changes that I've got here are actually things that @cmsj removed when manually tweaking 0.9.4 (i.e. he probably removed `lua_Number n;`).

Both `sqlite3.c` and `sqlite3.h` has substantial changes though - however, we don't use these in Hammerspoon.

Long story short, I'm going to close this pull request, and add a new one that simply changes the version number.

asmagill:
Yeah, I reviewed it too and I'm in agreement -- most of the changes look like they are to the included static sqlite3 library (probably updating to keep it current with upstream sqlite3)  and to the rockspec, which *would* need minor tweaks to allow building under 5.4, but it looks like the "lua module" is basically unchanged.

asmagill:
Which is kind of disappointing in a way... the new "to-be-closed" variables could be of real use when doing database work... so it might be worth keeping an eye open, but for now I think we're good.

PR #2530, opened 2020-09-28
Added a script to generate autocomplete files using EmmyLua annotations

folke:
I've created a python script that uses `build/docs.json` to generate Hammerspoon API stubs using [EmmyLua annotations](https://emmylua.github.io/annotation.html)

These kinds of annotations are supported by the following lsp servers:
* [lua-language-server](https://github.com/sumneko/lua-language-server) (most complete)
* [EmmyLua-LanguageServer](https://github.com/EmmyLua/EmmyLua-LanguageServer)

Plugins exist for VScode, IntelliJ, vim, Emacs, ... to work with those lsp servers

Let me know your thoughts!

## ✅   TODO
- [x] cleaner parsing of function parameter signatures
- [x] better parsing of function return types
- [x] more type annotations for variables and constants
- [x] generate stubs for spoons
- [ ] create a wiki page on how to use this. Either by copying the `types` directory or including it as a library

## 📦   Building and using the Hammerspoon stubs
Checkout this branch and run the following:
```shell
$ make build/stubs
```

You now have two options to start using the stubs:
1) copy the generated directory to your `.hammerspoon` directory (or any other project where you need hs complation to work)
2) add the `build/stubs` folder as an external library. For **lua-language-server**, add the folllowing to your workspace settings in vscode:
```json
{
  "Lua.workspace.library": {
    "/Users/folke/projects/hammerspoon/build/stubs": true
  }
}
```
❗ Don't just symlink the `stubs` in your project directory! In at least **lua-language-server**, there's an issue with symlinked source roots and autocompletion won't work.

### Why not just add `/Applications/Hammerspoon.app/Contents/Resources/extensions/` as an external library to the lsp config?
Simply adding the hammerspoon `extensions` directory as an external library doesn't work for a couple of reasons:
* class instances, constructors and function return would not get proper autocompletion
* `_coresetup` populates the `hs[string]` table with all modules. LSP servers have a hard time dealing with this and think all functions/methods/variables anywhere are valid auto completions for every module. With the generated stubs, auto completion works for all HS functionality.

### How could this be distributed?
* possibly as an additional zip file as part of the Github release
* as a directory in the `Hammerspoon.app`.  For example `/Applications/Hammerspoon.app/Contents/Resources/stubs`

## 🎨   Screenshots
image
image
image
image

### Completion for Spoons
image


asmagill:
Interesting. I probably won't be able to test this for a few days (my current go to editor, BBEdit, doesn't support lsp servers, so I'm going to have to take a look at Nova or VSCode).

Just to make sure I understand, I only need to run the python script once to build the stubs each time the supplied `docs.json` file is changed, correct?

folke:
That is correct!

Ive also added an entry for it to the Makefile.

```shell
$ make build/emmy-lua
```

folke:
With the new improvements, auto completion works pretty much everywhere. For more info, see the updated description 🔝

folke:
Just added stub generating for spoons as well

asmagill:
Does the python script require any special packages to be installed or will it run with the stock macOS installation of python?

folke:
Standard python `/usr/bin/python` should work. There's no external dependencies on other packages either.

asmagill:
I will try and review this week (got sidetracked multiple times since I last posted here), but I am wondering if this might be better as a Spoon?

Since it can be built by anyone with the stock macOS python and the json file... I don't think there is anything that *requires* it to be part of a release build, but correct me if I'm wrong... a little file date checking and it a spoon could trigger a rebuild for you when Hammerspoon is updated.

At any rate, I'll give this a closer look soon.

folke:
That's actually a great idea. Since the `Spoons` directory is a subdirectory of the user's `.hammerspoon` directory, the `stubs` inside for example `AutoComplete.spoon` would work as is, without needing to define the library in the LSP config.

Another benefit is that auto completion would also work for non-official spoons, since we can generate the stubs on the fly when needed.

Is this something you would be interested to have as part of the [official spoons](https://github.com/Hammerspoon/Spoons/tree/master/Source)?

If so, what name would you suggest? `AutoComplete.spoon`?

cmsj:
@folke apologies for not being involved in this PR so far. I think having it as a Spoon is a great idea, and I would definitely accept it into the official repo.

Regarding the name - I wonder if we could come up with something a little more specific than `AutoComplete`, although I don't have any brilliant suggestions.

folke:
@cmsj ok, great. Will create a Spoon once I have some time. Thanks!

folke:
@cmsj I created a Spoon as discussed.

The implementation is now also fully in Lua, so Python is no longer needed.

taybart:
Any word on this PR?

PR #2535, opened 2020-09-29
[READY] Updated version number in hs.sqlite3

latenitefilms:
- No changes between 0.9.4 and 0.9.5

### DIFF
--- a/extensions/sqlite3/init.lua
+++ b/extensions/sqlite3/init.lua
@@ -3,7 +3,7 @@
 --- Interact with SQLite databases
 ---
 --- Notes:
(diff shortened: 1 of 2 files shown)


latenitefilms:
After all that... here you go @asmagill !

asmagill:
Sorry! It just looked wrong somehow that two unnecessary changes constituted a version update, but I think our final analysis was correct and that the "lua module" portion really is unchanged for now. Still, if anyone is watching, it shows that we are tracking it and if they do start to take advantage of new 5.4 features, they might eventually find their way in here.

asmagill:
`init.lua` needs the version number updated in the docs header as well, then I'll merge this. 😁

PR #2541, opened 2020-09-30
[REVIEW] Added hs.sound.getAudioEffectNames()

latenitefilms:
- Gets a table of installed Audio Units Effect names.


asmagill:
lgtm

Can we do anything with these effects (or will we be able to in the future) or is it merely informational?

latenitefilms:
> Can we do anything with these effects (or will we be able to in the future) or is it merely informational?

I use it in CommandPost to get a list of all the Audio Unit effects that are available on the Mac, so that I can list them in a `hs.chooser` and apply them in Final Cut Pro using the Accessibility Framework.

However, it would be possible to add Audio Unit support to `hs.sound` more generally later down the line for manipulating sounds.

PR #2542, opened 2020-09-30
[REVIEW] hs.httpserver can now recieve messages as NSData

### LINKED ISSUE #2387: Patch CocoaHTTPServer to better support websockets
As discussed in #2293

latenitefilms:
- Switched to using a CommandPost fork of CocoaHTTPServer, as the original repo is now dead.
- Updated the podfile and updated all pods.
- Added support for receiving NSData to CocoaHTTPServer.
- Originally discussed in #2293
- Closes #2387


latenitefilms:
@asmagill & @cmsj - As originally discussed in #2293, I've forked `CocoaHTTPServer` in my CommandPost account, and have applied a fix to allow the receiving of `NSData`. If you decide this is a good idea and worthwhile merging, I'd suggest we fork `CocoaHTTPServer` into the Hammerspoon account - and then if future fixes need to be applied to `CocoaHTTPServer`, the Hammerspoon maintainers "own" the code.

Whilst I agree long-term, it would be good to find a replacement for `CocoaHTTPServer` that's still actively maintained and updated, rather than a fork - that will require a major re-write of `hs.httpserver`, which I'm not sure anyone would be willing to take on-board at this point in time. As you may have noticed, I'm trying to close off all my existing Hammerspoon issues whilst I have a bit of spare time - but I don't think I have the skills or the knowledge to tackle reworking `hs.httpserver`.

You can use this example code for testing:

```lua
wsserver = hs.httpserver.new()
	:setCallback(function(requestType, path, headers, raw)
		print(string.format("requestType: %s", requestType))
		print(string.format("path: %s", path))
		print(string.format("headers: %s", hs.inspect(headers)))
		print(string.format("raw: %s", raw))
	end)
	:setPort(1234)
	:websocket("/ws", function(message)
		print(string.format("server message: %s", message))
	end)
	:start()

ws = hs.websocket.new("ws://localhost:1234/ws", function(event, message)
	print(string.format("client event: %s (%s)", event, message))
end)

hs.timer.doAfter(1, function()
	ws:send("sending from client to server")
	wsserver:send("sending from server to client")
end)
```

Any questions let me know!

Issue #2524, opened 2020-09-27
hs.window.filter.defaultCurrentSpace:getWindows() sometimes missing some windows


callum-oakley:
Since the update to 0.9.79 (and still on 0.9.80) I've noticed visible windows on the current space sometimes not appearing in the results of `hs.window.filter.defaultCurrentSpace:getWindows()`. I managed to reproduce this consistently by:

1. Going to a new space and opening the hammerspoon console:

```
> hs.inspect(hs.window.filter.defaultCurrentSpace:getWindows())
{ <userdata 1> -- hs.window: Hammerspoon Console (0x6000017f2cf8) }
```

2. Opening System Preferences (or any other app, doesn't seem to matter as long as it's not already open) via Spotlight. Hammerspoon console throws:

```
2020-09-27 19:38:22: 19:38:22 ERROR:   LuaSkin: Unable to fetch element attribute NSAccessibilityFrontmostAttribute for: System Preferences
```

Relevant?

3. Now I can see the preferences window, but it doesn't appear in `getWindows`:

```
> hs.inspect(hs.window.filter.defaultCurrentSpace:getWindows())
{ <userdata 1> -- hs.window: Hammerspoon Console (0x6000017f2cf8) }
```

4. If I change away from this space and back again, it's now there:

```
> hs.inspect(hs.window.filter.defaultCurrentSpace:getWindows())
{ <userdata 1> -- hs.window: Hammerspoon Console (0x6000017f2cf8), <userdata 2> -- hs.window: System Preferences (0x600001626078) }
```

I can't reproduce this on 0.9.78.

asmagill:
I am unable to replicate this behavior. Are you still seeing it with 0.9.81?

callum-oakley:
Unfortunately yes 😞 anything I can do to help debug?

Also I think the LuaSkin Accessibility error was a red herring, I don't have it any more on 0.9.81 and after removing and re-adding accessibility permissions (not sure which fixed it), but I still have the missing windows in getWindows.

I haven't looked much in to the code to see how hs.window.filter works, but it feels like an event is being missed somewhere when the new Preferences window first appears, because after doing pretty much any interaction (clicking on it and off it again, changing workspaces, etc) it then does appear in the list.

callum-oakley:
Screenshots:

[...]

PR #2552, opened 2020-10-05
[WIP] Fix hs.window.filter 0.9.81

asmagill:
Initial push should fix #2488 latest where hs.window methods were being treated as functions

Hoping to figure out #2524 as well, but so far no ideas... it seems like a timing thing -- the app watcher for the new app seems to stop checking for windows in the app after getting initial `activated` event and doesn't start again until the next `activated` event, so if the window is created *after* we receive the first `activated` event, it isn't seen until a forced recheck.

Any thoughts or help on #2524 greatly appreciated -- I am at the limit of my minimal understanding of that module.

### DIFF
--- a/extensions/window/filter.lua
+++ b/extensions/window/filter.lua
@@ -2303,10 +2303,10 @@ end
 
 for _,dir in ipairs{'East','North','West','South'}do
   WF['windowsTo'..dir]=function(self,win,...)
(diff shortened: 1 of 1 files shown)


asmagill:
duplicate of #2576 (and I prefer the solution there -- it's little more explicit)

Issue #2536, opened 2020-09-29
hs.axuielement:searchPath gone missing


jdtsmith:
I thought I should open a separate issue for this.  With `axuielement` migrating into the core and its API changed to emphasize asynchronous operation, a terribly useful bit of functionality has apparently gone missing: `searchPath`.  This method provided a very compact notation for "pruning" the breadth-first accessibility search layer by layer beneath a given accessibility object.  An example from [SplitView](https://github.com/jdtsmith/SplitView):

```lua
   local spaceAX=self.dockAX:searchPath({	-- a list of space controls for each screen
	 {role="AXGroup",identifier="mc"},
	 {role="AXGroup",identifier="mc.display"},
	 {role="AXGroup",identifier="mc.spaces"}})
```

A couple questions:

1. Is there an update to the [Queries](https://github.com/asmagill/hs._asm.axuielement/blob/master/Queries.md) guide (where `searchPath` is described and its performance advantages touted)?
1. Is there a reason `searchPath` was deprecated, or is it still to be implemented in the new API? 

I presume one could "roll their own" `searchPath` replacement as a driver of a coroutine which iterates through a path criteria table like the above, forms a depth 1 `elementSearch`, whose callback continues the coroutine with the element(s) found.  The coroutine then moves on to the next criterion in the path, yields a new `elementSearch` for each of the elements found at that layer, and so on, until the criteria table is exhausted and a final element is returned, or no element is found and an error is signaled.

latenitefilms:
I'll let @asmagill explain his logic behind it, but I believe that `searchPath` has basically been replaced by `elementSearch`, which is a lot faster, and also asynchronous as you note.

I'm pretty sure `searchPath` was all done in Lua-land, so you could bring it back yourself in your own code if you wanted:

https://github.com/asmagill/hs._asm.axuielement/blob/3892477d44bfe91bcf5cc29c6bf6910a75de3bca/init.lua#L748

However, you should be able to achieve the same thing as what you want to do in the `SplitView` example using `elementSearch`. If you have a look at the [`matchesCriteria`](http://www.hammerspoon.org/docs/hs.axuielement.html#matchesCriteria) documentation you'll see you can do a table of multiple criteria's, in a similar way that you have multiple search path roles above.

For anyone who stumbles across this issue, here's the example I posted in #2510:

```lua
hs.openConsole() -- Open the Hammerspoon Console
[...]

PR #2554, opened 2020-10-06
[WIP] Updates to hs.axuielement

asmagill:
Some planned, and not originally planned updates to hs.axuielement.

Bug fixes, documentation updates, additions...

Working towards addressing #2536, but rewriting Queries document as a sort of "How To" as I go and sticking in other helpers as they suggest themselves to me. Feel free to recommend other additions or suggestions if you have them as well; this will probably be a work-in-progress for a bit...


asmagill:
I have started on documentation for axuielement that may be of interest to those following this pull. While I haven't completed a proper solution for searchPath yet, some of the tests I've been running lent themselves to or suggested some areas that needed more documentation than is directly provided in the standard API documentation. If you want to take a look (it's a work in progress, so suggestions or corrections are welcome), you can check it out at the wiki of my fork of Hammerspoon at https://github.com/asmagill/hammerspoon/wiki/hs.axuielement-Overview -- while the "Searching for Specific Elements" document will probably end up being the one of most interest, it's also the one that is least documented at the moment (aside from parameterizedAttributes which I haven't even begun yet), but all of it needs at least a once over by someone else before being merged into the core wiki.

jdtsmith:
Took a look at the wiki, and the examples are appreciated.  I still think a higher level "pruned search" (ala alpha-beta pruning in early game engines) would be useful, especially for very wide/deep axuielement trees.   It might be nice to include an "any number of children" search primitive (all zsh's '**' glob matcher) which will descend any number of children while attempting to continue the search.  What are the largest trees you have seen in terms of number of nodes?  Probably browsers?

asmagill:
FWIW, a pruning search is still on the agenda.

No ETA at the moment, other than to say that I'm expecting to have some "down time" during the next month in which I can start finishing a number of projects that are currently "in progress".

asmagill:
Oh, and the worst tree of elements I've dealt with is a web browser with a lot of bookmarks... makes the AXMenuBar child take forever to fully enumerate.

These numbers are off the top of my head, but if I remember even slightly correctly, having 5 windows with an average of 5 tabs each and fully enumerating them, but skipping the menu bar, took me on average around 30-45 seconds, definitely under a minute... adding in the menubar made it more like 6 or 7 minutes. I'll run more exact numbers as it has been a while and I don't recall exactly, when I finish working on the pruning searcher and add it to the documentations.

jdtsmith:
Thanks.  The problem with unknown depth searches with varying criteria is that if you keep it async for speed, you quickly end up deep in endless callback hell.  

To solve this problem, I came up with [this little `searchPath` style experiment](https://gist.github.com/jdtsmith/84a89d41a1a4802ca3e2117f5249ddd7) for my actual example case (`dofile` it and Command-Up Arrow within 2 seconds).  It uses the approach I had mentioned before: create a table of `searchCriteriaFunction`'s, one for each level of search, which starts at some particular ax object.  Then iteratively execute a depth-1 search for all elements surviving the cut, collecting the last survivors and returning a table of them via a single, final callback.  To squash the callback madness, this uses a custom coroutine which is resumed by its _own_ callback (to `elementSearch`).  Probably this would be much faster if it was all in ObjC code. 

In my example the `searchCriteriaFunction`'s are simple id matches, but they could be arbitrarily complex.  It also wouldn't be too hard to add an "any-depth" (`**`) functionality: instead of kicking off an `elementSearch`: just immediately resume the search with _all_ children at that level (and take care to separately accumulate and stop searching on final matches, which may now come from arbitrarily deep levels). 

I do wonder if you have a recommendation for bridging the async->sync divide when calling a nested async search like this, e.g. when some sync function wants to "just wait for it" without locking up HS.  I can obviously set a `done` closure var in the callback and poll for that, but I don't want to block HS the whole while waiting.

I saw `applicationYield` but that seems to be for coroutines only.  (BTW, isn't `coroutine.yield` enough to give HS cycles back for its other work?)

asmagill:
Just a heads up -- going to try and work some more on this over Thanksgiving... @cmsj, are you planning to do a release before the weekend? If so, I can land this now as it does contain a couple of bug fixes to existing code... if you're planning to wait, then I do have some more additions I'm planning.

asmagill:
Note to self -- rewrite hs.window.filter (#2596) as test of this module and to see if additional helpers (beyond those already requested/planned) might be helpful.

asmagill:
Closing this in favor of #3149

PR #2561, opened 2020-10-12
Add support for top and bottom layout

luong-komorebi:
This will help users layout their screen better with a vertical screen.

### DIFF
--- a/extensions/layout/init.lua
+++ b/extensions/layout/init.lua
@@ -61,6 +61,56 @@ layout.right70 = geometry.rect(0.3, 0, 0.7, 1)
 --- A unit rect which will make a window occupy the right 75% of a screen
 layout.right75 = geometry.rect(0.25, 0, 0.75, 1)
 
+--- hs.layout.top25
+--- Constant
+--- A unit rect which will make a window occupy the top 25% of a screen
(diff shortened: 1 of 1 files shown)


asmagill:
I don't use `hs.layout`, so it will probably be easier for me to ask then to figure out how to test it: are you sure that `bot25` is correct? If I'm understanding how the unit rect's are being used, shouldn't it be `geometry.rect(0, 0.75, 1, 0.25)` instead of `geometry.rect(0, 0.25, 1, 0.25)`?

luong-komorebi:
@asmagill Thanks for the feedback. I m new to this as well as lua code so I hope you can understand my mistakes if any. 
I based on this documentation in the code
``` 
--- hs.geometry.rect(x, y, w, h) -> hs.geometry rect
--- Constructor
--- Convenience function for creating a rect-table
---
--- Parameters:
---  * x - A number containing the horizontal co-ordinate of the top-left point of the rect
---  * y - A number containing the vertical co-ordinate of the top-left point of the rect
---  * w - A number containing the width of the rect
---  * h - A number containing the height of the rect
``` 
and my own testing using my screen to figure out the number. 
For bot25, I agreee there must be some mistakes from my side.

PR #2606, opened 2020-11-24
Testing. Do not merge

cmsj:


cmsj:
@asmagill @latenitefilms FYI, this PR exists to test the commit I just pushed to master that uses GitHub Actions to do a test run. The intention here is to migrate away from Travis. The stretch goal here is to be able to do per-commit signed builds of master. The super stretch goal is to have a fully hosted process for doing releases.

Issue #2602, opened 2020-11-22
Failed creating window switcher


fancl20:
On Mac mini M1, 11.0.1

1. Create an init.lua with
```lua
switcher = hs.window.switcher.new()
```

2. Reload config
```
2020-11-22 21:43:18: -- Lazy extension loading enabled
2020-11-22 21:43:18: -- Loading ~/.hammerspoon/init.lua
2020-11-22 21:43:18: -- Loading extension: window
2020-11-22 21:43:18: -- Loading extensions: window.switcher
2020-11-22 21:43:18: 21:43:18 ERROR:   LuaSkin: hs.canvas:clipToPath must be a boolean
2020-11-22 21:43:18: 21:43:18 ERROR:   LuaSkin: hs.canvas:clipToPath must be a boolean
2020-11-22 21:43:18: 21:43:18 ERROR:   LuaSkin: hs.canvas:clipToPath must be a boolean
2020-11-22 21:43:18: 21:43:18 ERROR:   LuaSkin: hs.canvas:clipToPath must be a boolean
2020-11-22 21:43:18: *** ERROR: ...oon.app/Contents/Resources/extensions/hs/canvas/init.lua:525: bad argument #2 to 'canvasDefaultFor' (invalid argument type for clipToPath specified)
stack traceback:
	[C]: in method 'canvasDefaultFor'
	...oon.app/Contents/Resources/extensions/hs/canvas/init.lua:525: in metamethod 'newindex'
	...ntents/Resources/extensions/hs/drawing/canvasWrapper.lua:29: in function 'hs.drawing.canvasWrapper._image'
	(...tail calls...)
	...app/Contents/Resources/extensions/hs/window/switcher.lua:380: in upvalue 'setUiPrefs'
	...app/Contents/Resources/extensions/hs/window/switcher.lua:419: in function 'hs.window.switcher.new'
	/Users/user/.hammerspoon/init.lua:1: in main chunk
	[C]: in function 'xpcall'
	...app/Contents/Resources/extensions/hs/_coresetup/init.lua:702: in function 'hs._coresetup.setup'
	(...tail calls...)
```

asmagill:
Can't replicate on my end; what version of Hammerspoon are you running?

fancl20:
Version 0.9.82 (5614)

It's a clean install so the problem may not related to platform. Maybe I just forgot to meet some requirements.

fancl20:
[...]

PR #2614, opened 2020-12-08
[READY] Migrate away from @encode

asmagill:
It's a compiler directive that is architecture specific and causes issues (at least with BOOL) in universal builds when comparing against [NSNumber objCType]

While fixing the `BOOL` issue should address #2602 specifically, it's probably best to move away from it entirely just to be safe... compare like to like instead of two things that are *mostly* the same (i.e. `@encode` and `[NSNumber objCType]`.

I need to consider a way to specifically test the structure converters in LuaSkin (point, rect, etc.) as they aren't used often (they're mostly useful when such items are entries in an NSDictionary or NSArray -- for methods we create, it's usually more flexible to use the table provided and craft the specific structure we require ourselves).

@cmsj, since you use it with a c style structure in LuaSkinTests I *think* you should be safe -- from a little googling, most of the problems people have with @encode are with objective-c specific types and structures.

*edit: made 2nd to last paragraph clearer*

### DIFF
--- a/LuaSkin/LuaSkin/Skin.m
+++ b/LuaSkin/LuaSkin/Skin.m
@@ -1251,16 +1251,30 @@ nextarg:
     NSValue    *value    = obj;
     const char *objCType = [value objCType];
 
+    // @encode is a compiler directive that can give different results depending upon the
+    // architecture, so lets compare apples to apples:
+    static dispatch_once_t onceToken;
+    static const char *pointEncoding ;
+    static const char *sizeEncoding ;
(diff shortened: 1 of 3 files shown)


asmagill:
@cmsj this can be merged if there are no objections

asmagill:
@cmsj, I since you're merging items, please consider merging this unless you have questions/objections -- it's necessary for hs.canvas/drawing to work on M1 macs.

asmagill:
👍

PR #2624, opened 2020-12-14
[READY] LuaSkin object wrapper fix

asmagill:
Object wrapper can properly clear entries in NSDictionary now

Need to find good place/way to document its use and the `ls` functions in general, so holding off for the moment -- at present only `hs.doc` uses this, and it uses it in read-only mode so bug would never occur.

Planning to use this in future `hs.hash` updates, so giving it a review first.


asmagill:
I want this merged before I start in on some of the uses I have in mind for it, so even without documentation outside of source code, I consider it ready, pending any objections.

PR #2648, opened 2021-01-11
Enable 'make docs' support virtual envirnment like venv

deftsp:


cmsj:
So this is a tricky one - I've specifically tried to force all parts of the build/test/CI/release process to use the system Python provided by macOS, precisely because I don't want it to be some random other version of Python that happens to be present via a venv or Homebrew. basically I don't trust all of those things and I hate that my machine ends up littered with various versions of Python that may or may not be the same thing depending on which script/process I happen to use to do a release or that a CI job runs.

I realise it's not helpful for others, but for now I'm going to close this because my primary concern is making sure that CI and Releases work properly. Sorry!

Issue #2402, opened 2020-06-23
macOS 11 Big Sur support


cmsj:
So the good news is that Hammerspoon appears to work pretty much without any modification, on Big Sur (x86), at least for the APIs that I typically use.

Building it in the new Xcode seems to also mostly be ok, although bumping the target deployment from 10.12 to 10.13 introduces some new deprecations we'll need to take care of. Some of our Cocoapods are complaining about having deployment target versions that are too old, but they still seem to build ok.

Which brings me to the same question that I like to ask every year... do we have to keep supporting old releases? :) (I continue to not be a fan of supporting old releases, I would much rather do things like support the new Big Sur UI style now, than in 3 years time)

dasmurphy:
Just 2 cents from me. It would be awesome to get at least one release which works on 10.14. This is my current system due to 32bit support for some Apps. Currently i cannot upgrade, but i know that i will in future.

I don't know if anything would be possible to create with 10.14+ with Big Sur Style support. Any way this would be awesome.

What do you think you can do?

@cmsj Do you have access to the beta?

cmsj:
All of our releases since 10.14 was released, should work on 10.14. Right now the minimum supported release of macOS is 10.12 in Hammerspoon 0.9.78. We probably don't have 32bit support in the most recent releases though, since my machine will have been on 10.15 since it released and Xcode dropped 32bit support - is that what you meant?

latenitefilms:
Great news!

We’re still on Mojave on all our machines for the time being, as Catalina has so many bugs and issues in Film & TV land.

Based on initial reports, it sounds like Big Sur is a big step up from Catalina, and already more stable, so I can imagine we’ll probably skip Catalina and eventually go straight to Big Sur.

I’d like to keep supporting Mojave for the time being, until Big Sur’s been proven to be reliable after a few point updates.

dasmurphy:
> We probably don't have 32bit support in the most recent releases though, since my machine will have been on 10.15 since it released and Xcode dropped 32bit support - is that what you meant?

No, i'm using the current version, but i need 10.14 for some old 32bit Apps i need for work. That's why i'm currently waiting with the upgrade.

asmagill:
(thought I posted this already, but it's not showing up in the thread... assuming it was me forgetting to post it after previewing it, and not a github glitch, I'll do so again... if this or something similar appears twice, I appologize!)

- - -

[...]

PR #2655, opened 2021-01-20
New icons for Big Sur

maebert:
As mentioned in https://github.com/Hammerspoon/hammerspoon/issues/2402#issuecomment-761713629, the tilted icon sticks out next to the glossy new Big Sur candy. This is an attempt at creating an updated icon that lives well next to its macOS 11 brethren:

* Square base with rounded corners
* Plastic gloss and drop shadow for the main element
* Slightly thicker hammer

The 16px and 32px versions are rendered without gloss or shadows to make the lines a little clearer. Feedback welcome of course!

image


dasmurphy:
@maebert I would like to see the Hammer without the gloss and shadow, then it would be "ok" for me. I really don't like the new icon style for big sur. I really love that everyone can make an icon in his own style and design. Also that may be since i'm on an older system currently. I know the "same" design for all icons is much easier.

With the Hammerspoon icon there is not so much change and the version from you looks ok.

The second i think: Would it be possible to share the source? I think you designed it as an Vector, so it should be easy to export an SVG, AI or Affinity file. I think the source is important, since this is an open source project.

At last i think this is some good progress to the current icon. With some iterations this one would be very nice later...

cmsj:
@maebert thanks for looking at this! I quite like the look. I do wonder if we should make the stem of the hammerspoon shorter - it feels like the Apple icons leave more padding between the edge of the icon and the object inside.

cmsj:
Also it might be worth noting that the source files for our graphical assets live in https://github.com/Hammerspoon/Graphical-Assets and I would ideally prefer to keep some kind of source file in there that I can work with in the future if I need to (at the moment it's Affinity Designer, but SVG would be fine).

cmsj:
For now I've bodged together something myself in #2719 which will be in 0.9.83, but @maebert I'd love to work with you on getting it slightly more aligned with the Big Sur design guidelines. The Affinity Designer file ("Current Source") is in the Graphical Assets repo if that's of any use to you.

maebert:
Hi @cmsj , thanks for the vote of confidence. Super bogged down with work this week but I'll give this an iteration on the weekend!

cmsj:
Gonna close this for now, but please feel free to re-open if you want to revisit this :)

Issue #2663, opened 2021-01-27
Screens unnamed on Apple Silicon


markdrayton:
Hi!

I'm getting `(un-named screen)` for screen names (both internal and external) on my shiny new MacBook Air M1:

```
> for k, v in pairs(hs.screen.allScreens()) do print(k, v) end
2021-01-27 19:33:26: 1	hs.screen: (un-named screen) (0x600000951778)
2021-01-27 19:33:26: 2	hs.screen: (un-named screen) (0x600000951a38)
```

It doesn't matter if my Dell U2713HM is plugged in or not — everything is `(un-named screen)`.

This is on Big Sur 10.1 and Hammerspoon 0.9.82 (5614). I don't have much of an idea of how to debug this myself but I'm definitely willing to guinea pig anything for someone who does.

asmagill:
Confirmed. Running in Rosetta mode doesn't change the behavior either, so not sure if it's a M1 thing or a Big Sur thing.

I will have access later today to an Intel laptop running Big Sur and an Intel iMac running Catalina and will test on those to see if we can narrow the specifics.

cmsj:
It's got to be an M1 thing, my Intel Big Sur machine reports screen names.

cmsj:
I also feel like this isn't the first time this has come up, but there is essentially nothing we can do about it unless there is some other API that can be probed (which there must be, because macOS itself displays screen names in several places). I'd be curious to know what tools like ScreenResX report.

asmagill:
I've been having good luck with an IOKit search module as a possible `hs.battery` replacement (it doesn't work on the M1 macs either)... I'll see if it might offer an alternative for screen names as well.

markdrayton:
Yes, sorry, should have mentioned that Big Sur on Intel works fine.

markdrayton:
Related discussion:

https://developer.apple.com/forums/thread/666383?answerId=650068022#650068022

Unfortunately searching the internets for `IODPDevice` doesn't turn up anything useful 😞

dcai:
I'm having the same issue on m1 macbook, hope there will be a solution to this.

cmsj:
This should be resolved in the next release.

PR #2676, opened 2021-02-09
[Review] Update hs.screen:name for M1 macs

asmagill:
Uses NSScreen's localizedName attribute introduced in 10.15

Addresses #2663 

Apparently M1 macs no longer use IOKit for maintaining display information which means IODisplay* no longer provides useful information on these machines. This change to the macOS will likely also result in `hs.screen:get/setBrightness` and `hs.brightness` to be useless... At present I am not aware of any promising leads for a work-around.

### DIFF
--- a/extensions/screen/internal.m
+++ b/extensions/screen/internal.m
@@ -1278,22 +1278,25 @@ static int userdata_tostring(lua_State* L) {
     LuaSkin *skin = [LuaSkin sharedWithState:L];
     [skin checkArgs:LS_TUSERDATA, USERDATA_TAG, LS_TBREAK];
 
+    NSString *theName = @"(un-named screen)" ;
     NSScreen *screen = get_screen_arg(L, 1);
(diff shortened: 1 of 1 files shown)


asmagill:
Tested against M1 MacBook Pro running 11.2 and iMac running 10.15 (which should use new method as well as it's 10.15 -- but I can confirm on my machine the names match the ones returned by the previous code). A test against 10.14 would be appreciated, if anyone running development builds still uses it.

cmsj:
Great find! I suspect we may have to just mark the brightness calls as unavailable on M1 macs, and hope that Apple gives us some API in the future.

latenitefilms:
Tested on macOS 10.14.6:

```lua
> for _, screen in pairs(hs.screen.allScreens()) do
	hs.printf("name: %s", screen:name())
end
2021-02-10 17:57:24: -- Loading extension: screen
2021-02-10 17:57:24: name: LG UltraFine
2021-02-10 17:57:24: name: Color LCD
```

cmsj:
@asmagill FYI, CI should be fully fixed now, so if you want to get results, rebase your PRs onto `master` and re-push them 👍

PR #2677, opened 2021-02-09
Modernise hs.mouse and remove the ManyMouse dependency

cmsj:
I don't have tests for hs.mouse written yet, so I'll implement those as best as possible before landing this, but I also wanted to see what @asmagill thinks of the `LS_API` define. I'm not sure if I like it or not yet - the aim was to reduce the amount of boilerplate at the top of every single Lua API call to a single thing that just takes the `LS_T...` arguments and puts a `LuaSkin` object in the current context.


asmagill:
Re LS_API... I'm of two minds... I like reducing boilerplate, but I don't like hiding where a variable is being defined (i.e. `skin`) because it can mean scrolling back and forth or checking multiple files when trying to understand or debug new code... granted for most of us here, we're so used to `skin` in particular that this confusion isn't likely, but it's the principle... I always hate approaching new code when the original coder/team was macro happy...

cmsj:
@asmagill yeah that's exactly my reservation about the macro. I've just pushed a new revision that makes it "return" the LuaSkin object and while that's adding more boilerplate back in, I think I like it more.

asmagill:
Yeah, I like that better as well.

Issue #2593, opened 2020-11-13
hs.notify.deliveredNotifications() leaks memory


Pancia:
I've narrowed down increasing memory usage, up to ~20 GB, to `hs.notify.deliveredNotifications()`.
I used `debug.getregistry()` to find the offending table that just kept increasing in size, and it looks like the delivered notifications are the "same" but with different memory addresses each call...
I don't really understand obj-c and lua interop, but it seems like notifications pushed to lua never get released https://github.com/Hammerspoon/hammerspoon/blob/2671848973828d6bec08681d4bf107ae69af809d/extensions/notify/internal.m#L272

minimal repro case:
```
hs.timer.doEvery(1, function()
    hs.notify.deliveredNotifications()
end)
```

asmagill:
I can confirm that `hs.notify` is creating userdata objects that will never be collected (until forced by `hs.reload`).

This was probably the first module I created for Hammerspoon that wasn't just modifying an existing one from Mjolnir or Hydra... and it shows (to me at least) in that there are a lot of assumptions and code choices that I wouldn't make now...

It's probably going to be sometime late next week or so before I can make the time to look into a fix... the module really needs a complete overhaul.

latenitefilms:
@Pancia - awesome detective work! I was just wondering if you could provide some further information on how you used `debug.getregistry()` to find the offending table? I've been trying to hunt down various memory leaks in Hammerspoon for years now - but have never had much luck. Any tips or tricks?

Pancia:
@latenitefilms - Initially i used this https://github.com/yaukeywang/LuaMemorySnapshotDump, but i found it lacking for this particular example as it just told me there were a lot of `hs.notify` objects at `registry.45...`. So i just figured i could print the registry and take a look. It's not really automatable as i had to manually inspect it, but in this case the memory leak was in the GB's and was pretty easy to spot.
In this case even that was not enough, as it was not clear what was causing so many notification objects to remain in memory. So i went through all my code that had anything to do with notifications, and i eventually traced it by commenting out code, and narrowing it down to `hs.notify.deliveredNotifications()`. Specifically I printed out the length of that registry element, and saw it increasing when i called that function. 

In short, the most useful thing was that library, as i would not have really known what was leaking, or had the idea to look in the registry.

asmagill:
@latenitefilms FWIW, I've used something like the following before to examine the registry and look for tables that only increase and never decrease in size:

~~~lua
[...]

PR #2689, opened 2021-02-17
[WIP] Switch all refTables to UUIDs

cmsj:
A surprising number of tests are already passing with this. Gonna keep poking at it and see if I can get it to 100% passing.


latenitefilms:
I wonder if [this](https://github.com/Hammerspoon/hammerspoon/issues/2593#issuecomment-727828480) is something worth playing with at the same time?

> Which gives me an idea (not sure why I didn't think of it before, actually) about adding code in LuaSkin to add a __type field when it creates each module registry subtable, kind of like we do in userdata metatables (e.g. hs.getObjectMetatable("hs.hotkey").__type).

Might help with problem solving?

cmsj:
err, wow, I didn't expect *everything* to pass 😲

LuaSkin/LuaSkin/Skin.h
```diff
@@ -45,6 +45,9 @@ extern int luaopen_luaskin_internal(lua_State* L) ; // entry vector to luaskin.m
 // Define a break variable for the reference checker
 #define LS_RBREAK INT_MIN
 
+// Define a hacky fixed UUID refTable value to indicate LUA_REGISTRYINDEX
+#define LS_REGISTRYINDEX_REF @"11111111-1111-1111-1111-111111111111"
```
cmsj, line 49: I hate this, but I don't currently have a better idea.

cmsj: Well, I kinda do, and that would be to subclass NSUUID. Kinda wish instead of going through all these files and changing the type of refTable to NSUUID, I'd set up a typedef. I may well go back and do that.

LuaSkin/LuaSkin/Skin.h
```diff
@@ -330,9 +335,9 @@ NSString *specMaskToString(int spec);
 
  @param functions - A static array of mappings between Lua function names and C function pointers. This provides the public API of the Lua library
  @param metaFunctions - A static array of mappings between special meta Lua function names (such as <tt>__gc</tt>) and C function pointers
- @return A Lua reference to the table created for this library to store its own references
+ @return A UUID that can be used to retrieve the Lua reference to a per-module table for storage
```
cmsj, line 338: Missing docstring for libraryName

LuaSkin/LuaSkin/Skin.m
```diff
@@ -423,22 +428,25 @@ - (int)registerLibrary:(const luaL_Reg *)functions metaFunctions:(const luaL_Reg
     int tmpRefTable = luaL_ref(self.L, LUA_REGISTRYINDEX);
     NSAssert(tmpRefTable != LUA_REFNIL, @"Unexpected LUA_REFNIL registering library: %@", fname);
 
+    // FIXME: This doesn't seem to ever be used, and it conflicts with our new NSUUID driven refTable lookups. remove it?
```
cmsj, line 431: @asmagill something I did in this patchset is add a libraryName argument to the basic version of `registerLibrary` - am I reading the code right and this fname stuff is a hacky way to figure out the name, that I could drop as part of this?

latenitefilms:
When this eventually lands, will this mean that all external extensions (like `hs._asm.cfpreferences`, `hs._asm.undocumented.touchbar`, etc.) will all need to be updated to work?

cmsj:
> When this eventually lands, will this mean that all external extensions (like `hs._asm.cfpreferences`, `hs._asm.undocumented.touchbar`, etc.) will all need to be updated to work?

Almost certainly yes. Also this is still in the realm of *if* this lands - it's one hell of a disruptive change, and with our overall test coverage at slightly less than 30%, it would be very risky to push out in a release anytime soon.

LuaSkin/LuaSkin/Skin.m
asmagill: I think its used in two places -- throwing actually useful errors when two modules try to register the same NSObject helpers and adding a name to the actual registry table so that its easier to identify which module/file uses which table returned by `debug.getregistry`.

I suppose if you're rewriting the register methods so that passing in a string is *required* this code could be removed. This was just a way to do it without having to break any sort of compatibility with existing or third party modules.

asmagill:
@latenitefilms technically the section you refer to in #2593 has already been addressed. @cmsj's comments a little later about removing the `fname` code would be "slightly" cleaner but would break compatibility with existing modules by requiring an additional argument to be added to all included and third party modules. (by "cleaner" I mean making the label something like "hs.webview" -- or even something entirely personal -- rather than what is currently captured, i.e. "hs.webview.internal" -- my code identifies the actual shared library that the call originates from)

asmagill:
@cmsj what would be the potential implications/requirements of leaving the existing methods in place for compatibility but having them throw a warning like we did when switching from `[LuaSkin shared]` to `[LuaSkin shareWithState:L]` (which technically I think is still in place)? Would it even be possible? That way third party modules could be phased in as needed and not just suddenly *break*.

asmagill:
@cmsj, with such a disruptive change, might it be better as a foundational aspect of 2.0? I guess it kind of depends upon what, if any compromises, you're making to minimize the necessary changes to external modules (both included and third party).

cmsj:
@asmagill I'm ok with removing the extra argument to `registerLibrary`, I just kinda threw that in at the same time because it's always bugged me that I didn't add that argument in the first place, but the huge advantage of changing the type signature of all the existing Ref related methods is that it dramatically reduces the patch size - because apart from a few weird/wonky modules, the actual type of refTable is essentially irrelevant to the module itself. If we instead do this via new API, then every single Lua/LS API call that needs a reftable has to be changed, and a quick grep suggests that's about 600 extra changes in the diff.

I'm not thrilled about breaking all the existing modules, but I'm pretty sure that apart from your `undocumented` series, and any that CommandPost is carrying, there aren't really any others, so I think I can live with breaking the luaskin ABI (for the first time?) before 1.0.
I do see the argument for saving it for 2.0, but I really want to get to the bottom of why we have so many *weird* crashes before 1.0, and I think we have to be ok with getting "creative" about it, because I'm at least not smart enough to figure out what's going on at the moment.

latenitefilms:
FWIW - the best case scenario for me would be if @asmagill's WIP modules actually made it into the Hammerspoon `master` either undocumented or as "beta", as they're all extremely functional and awesome.

We're currently using:

- `hs._asm.cfpreferences`
- `hs._asm.undocumented.touchbar`
- `hs._asm.undocumented.touchdevice`
- `hs._asm.xml`

They've all been in use in CommandPost for several years now, and apart from a random crash with the Touch Bar extension that we've never been able to solve (which we think is just a bug in Apple's code, as BetterTouchTool works around similar issues) - they're all pretty rock solid. We could just add a big warning to the Touch Bar extension that it's using undocumented and unsupported API calls, so tread with care.

The benefit of adding them to the Hammerspoon `master` is that they get a bigger audience, which might help solve these outstanding issues anyway.

The XML extension would be very beneficial to the wider Hammerspoon community. The `hs._asm.cfpreferences` extension is something that's been discussed many times before about consolidating and improving all the preferences and plist support. So it's really just the Touch Bar and Touch Device extensions that are are slightly contentious.

cmsj:
On the subject of the touchbar modules - I continue to not own a Mac that has a touchbar, and if the speculation is true, Apple may stop including them in macbooks this year. If that does happen, I would probably lean towards never bringing the modules into the main repo.

latenitefilms:
Given the new 13‑inch MacBook Pro with Apple Silicon still has it, I really don't see the Touch Bar going away all together anytime soon, despite what the rumour mill says. It'll be really interesting to see what they do with the 16-inch MacBook Pro - but it would be very strange to have it on the 13-inch, and not the 16-inch. But if Apple do drop it all together, then problem solved! I'm going to hold onto my MacBook Pro (15-inch, 2017) and Mojave until the new 16-inch's ship.

asmagill:
Ok, if we're not trying to wrap up 1.0 "now" and are willing to wait a little longer for 2.0 (which I'm fine with... if I can keep up steam, I think these, and a couple of others I was recently working on, could be done in the next three months, give or take):

`cfpreferences`, `preferences`, and `plist`... yeah, that should happen. I still feel that they should be merged into one or at least rethought to better organize functionality, but I haven't had the time yet.

`xml`... I need to look at that again. When I first worked on it, I found that Apple's implementation was *way* more pedantic and convoluted than I had hoped... granted it should ultimately be fully compliant with the spec, but what I had really wanted was a simple parser to rip out some values from an XML source and ended up writing https://github.com/asmagill/hammerspoon-config-take2/blob/master/_Spoons/RokuControl.spoon/qduXML.lua.  That said, the Apple one would be more compliant, bi-directional (i.e. read-write), and the XQuery stuff looked interesting, I just never found a need strong enough to return to the module.

`touchdevice` is actually more about the trackpad than it is about the touchbar (though it can determine some basic info about the touchbar as well)... and I think if I introduce the `iokit` module I've been playing with lately that I can remove *most* of the private API requirements... plus I think there would be at least some interest (at least with those of us that still use X11 occasionally) to turn https://github.com/asmagill/hs._asm.undocumented.touchdevice/blob/master/Examples/middleButton.lua into a spoon.

- - -

However, I'm inclined to keep the touchbar module separate, pending figuring out some fundamental issues:

My biggest problems with `touchbar` are that the ability to use canvas objects within a touchbar is really kind of an ugly hack (that's more a fault of the way canvas is set up than the touchbar module itself, though... something I hope to address if I ever return to guitk/fenestra -- most likely in v2) and that the virtual touchbar portion (1) can't mimic the newer, shorter touchbar, and (2) leaks because trying to release the display stream after opening it crashes... all I have figured out how to do is stop it, then create a new one when reloaded, and (3) converts "touches" into mouse events, not actual touch events... the later seems to be addressed in the Duet implementation for iPad, and the first is possible with the XCode emulator, so... they should be *doable* I just don't know how; nor have I found any publicly available experimentation significantly beyond what we originally discovered.

As to the (dis?)continuation of the touchbar... I think I would have hated the first generation version which removed the Escape key; since the newer ones returned the Escape key, I've honestly not missed the physical function keys much... I use them so seldom. However, I think Apple's attempts to prove that people didn't really *want* a touch-screen kind of blew up on them... that, and keeping it limited to only *some* laptops...  if they seriously thought it was a superiour input device they should have also created a USB/BlueTooth version that could sit atop their other keyboards, rather than just keep it limited to a subset of their users.

cmsj:
I think we need to move the more general planning discussion off this PR, because I'll be closing it shortly. My plan is to split this out into PRs for the three things it now consists of:

1) Switch reftables from `int` to `LSRefTable` but keep that `typedef`'d to `int`. This is essentially a no-op change, but it makes it possible to change the type of reftables later without touching very many files.
2) Add the `libraryName` argument to `registerLibrary` because I think we should have that for future use. This is a relatively non-invasive change
3) Switch `LSRefTable` to the custom `LSRefTableUUID` class that's currently in this PR. This is the one that's risky and potentially breaky.

Unless anyone objects very strongly, I'll merge 1 and 2 relatively quickly, and then we can poke at 3 in isolation and see what form it should take, if any.

latenitefilms:
I generally like to try and keep CommandPost-App in-sync with the Hammerspoon master - does this mean I should avoid syncing until all three items are merged in, and @asmagill's extensions are updated?

latenitefilms:
Also, maybe it's worth enabling GitHub discussions for these more general chats, rather than using issues?

cmsj:
> I generally like to try and keep CommandPost-App in-sync with the Hammerspoon master - does this mean I should avoid syncing until all three items are merged in, and @asmagill's extensions are updated?

In theory 1) doesn't actually change the ABI, so that one would be fine. 2) does change the ABI, but it's just one extra argument to the `registerLibrary` call in each module. 3) is the one that will need a bit of work.

cmsj:
Given the later discussions, I'm going to close this PR since it definitely makes no sense to land it all as-is. It's been decomposed into:

 * #2693 
 * #2694 
 * #2695

Issue #2241, opened 2019-11-22
Cannot withdraw notifications from previous session


latenitefilms:
If I create a bunch of `hs.notify` notifications then restart Hammerspoon, I expect the below code should work to remove these notifications after restarting:

```lua
local deliveredNotifications = hs.notify.deliveredNotifications()
for _, notification in pairs(deliveredNotifications) do
    print(string.format("Withdrawing: %s", notification))
    notification:withdraw()
end
```

...however, sadly it doesn't.

I can see "hs.notify: creating new userdata for orphaned notification" in Console.app for each notification that's already on the screen.

`hs.notify.withdrawAll()` works as expected.

Any ideas @asmagill?

cmsj:
Slightly separately from this specific issue, and while it wouldn't help with crashes, I'm wondering if we should unconditionally withdraw all notifications when HS exits - or perhaps have a default shutdown callback that does it, so people can override the behaviour if they like littering their screen with orphaned notifications.

latenitefilms:
Ummm, I'm not sure about that. I think it's better to leave that up to the end-user to solve. In our case, I want Hammerspoon to "reconnect" to those existing notifications when restarting, but also have a way to detect when they've been closed when Hammerspoon wasn't running.

AFAIK this used to work correctly - so I think it's maybe a Mojave-related issue?

asmagill:
Will be addressed with #2691

latenitefilms:
Untested, but assuming this is fixed in #2691 ?

PR #2691, opened 2021-02-17
Rewrite hs.notify to remove memory leaks

asmagill:
Addresses #2593 

Long term we need to move to UNUserNotification but it's only 10.14+ (which is probably ok at this point) and fundamentally different so will require an entirely new module and approach.

This rewrite sticks with NSUserNotification but with the following changes under the hood (it *should* be syntacticly the same with maybe a little more argument type checking)
* userdata is now just the NSUserNotification object itself. No C-structure with bridged pointers anymore and state info is held in proxy object stored in module level mutable dictionary
* after sending/scheduling a notification, no changes are made to the notification itself (not even to its userInfo dictionary) to ensure that locally saved userdata and delivered userdata returned by hs.notify.deliveredNotifications/scheduledNotifications refer to the same macOS object
* no memory leaks -- registry no longer used to "save" userdata. Instead using selfRefCount via proxy object stored in module level mutable dictionary
* better checking for availability of private API methods

It needs testing, and I haven't really tested it that much against notifications that use the app's initial delegate (I think only `hs.showError` uses this, but need to check a little deeper) other than to verify that hs.notify.deliveredNotifications *does* report them... accessing details of these *may* cause crashes atm as it makes assumptions that aren't true for these specific notifications. I'll confirm and fix if necessary tomorrow (well later today).

@Pancia I have some ideas for tests I intend to run on this tomorrow, but honestly I haven't used hs.notify within my own configuration in quite a while, so if you're comfortable/able to build a local version of Hammerspoon against this pull I'd be interested in your feedback as well.

### DIFF
--- a/extensions/notify/internal.m
+++ b/extensions/notify/internal.m
@@ -248,8 +248,10 @@ static int notification_deliveredNotifications(lua_State *L) {
     // just in case pushNSUserNotification had to recreate our entries in ourNotificationSpecifics
     for (NSUserNotification *notification in deliveredNotifications) {
         NSString *gus = notification.userInfo[KEY_ID] ;
-        NSMutableDictionary *userInfo = ourNotificationSpecifics[gus] ;
-        userInfo[KEY_DELIVERED] = @(YES) ;
+        if (gus) {
+            NSMutableDictionary *userInfo = ourNotificationSpecifics[gus] ;
+            userInfo[KEY_DELIVERED] = @(YES) ;
+        }
     }
     return 1 ;
 }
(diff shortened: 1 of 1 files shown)


latenitefilms:
@asmagill - Any chance this could be resolved as part of these changes?

- Cannot withdraw notifications from previous session #2241

asmagill:
I *think* this should be taken care of as well since we never modify the notification object directly after it has been marked as dispatched anymore, but will make a point to test explicitly tomorrow.  Going to get some sleep now.

asmagill:
Latest push prevents methods from attempting to modify notifications not made by this module (i.e. `hs.showError` and `hs._notify`), though they can be withdrawn individually and via withdrawAll.

@latenitefilms I have confirmed that this will also address #2241

latenitefilms:
Amazing! Once this is merged into master I'll bring it across to CommandPost and see if it solves our issues. Thanks heaps!

cmsj:
Huge apologies for the amount of spam here.

Pancia:
Sorry I never got to testing this branch locally, been in a sort of time crunch the last week or so.

I updated hammerspoon as soon as I got the update notification, and haven't noticed anything wrong yet.
I'll report back if I notice the memory usage increasing again.

PR #2694, opened 2021-02-17
Add a libraryName argument to registerLibrary for future use

cmsj:


asmagill:
Food for thought:

I know that `[skin registerLibrary:"name" functions:moduleLib metaFunctions:nil];` reads better than `[skin registerLibrary:moduleLib metaFunctions:nil withName:"name"];` (or similar), but it basically breaks every existing module and requires an update to everything all at once...

I think it would be "nicer" to go the second route so we could leave `[skin registerLibrary:moduleLib metaFunctions:nil]` in and have it through a warning (like I do for `[LuaSkin shared]`) so that things don't immediately break, but we know that they need to be fixed.

(And, in the spirit of being totally upfront, I'll admit that since enough people use my third party modules, it would make it easier on me so I don't have to immediately maintain two versions of the packages when the transition occurs and can phase them in as I can get to them...)

Just my two cents!

cmsj:
That's a good point actually, I can just add the additional method, rework the old one and this patch doesn't have to change substantially. I'll work on that when I get a chance, thanks!

cmsj:
@asmagill any objections to this landing now?

LuaSkin/LuaSkin/Skin.m
```diff
@@ -403,6 +410,7 @@ - (int)registerLibrary:(const luaL_Reg *)functions metaFunctions:(const luaL_Reg
 
     NSString *fname = getCallerFileName() ;
 
+    // FIXME: Can this all be replaced just by using the libraryName argument?
```
asmagill, line 413: Yes, with one caveat: for now, at least, its possible that libraryName might not have a value for a given library... however since it's being applied to the reftable and for identification during debugging only, I suppose it not being present (i.e. not set) for a given table won't cause a problem.

asmagill:
See if you want to address inline comment for Skin.m that I just posted at this time, but otherwise, no objections.

LuaSkin/LuaSkin/Skin.m
cmsj: Ok, let's wait on that until CommandPost and you have updated all of your modules to use the new one.

PR #2695, opened 2021-02-17
Change LSRefTable to be UUID based

cmsj:


cmsj:
I don't think this really needs to exist anymore.

PR #2699, opened 2021-02-18
Guard C callbacks against LuaSkin having been reset.

cmsj:


cmsj:
@asmagill I'm going to merge this, but I have altered it to be non-fatal for now. I want to see how it shows up in the crash logs, once 0.9.83 is released.

PR #2722, opened 2021-03-04
Maintenance PR Experiment

cmsj:


cmsj:
The intention with this PR experiment is to see if it's possible to maintain a single PR for maintenance stuff (so build fixes, build system changes, Actions updates, etc.) that can be re-merged over time, rather than handle each tiny little thing as a separate PR.

PR #2732, opened 2021-03-05
Finding screen name with a "-" is broken

gwww:
Somewhere in the switch to 0.9.83 the screen name on my MBP changed from "Color LCD" to "Built-in Retina Display". The screen find command does not like the "-" in the screen name. Adding the true to the find turns off pattern matching. Substrings are still found, just turns off RE pattern matching (as far as I can tell from http://lua-users.org/wiki/StringLibraryTutorial)

### DIFF
--- a/extensions/screen/init.lua
+++ b/extensions/screen/init.lua
@@ -82,7 +82,7 @@ function screen.find(p)
         local sname=s:name()
         if s:getUUID() == p then
           r[#r+1]=s
-        elseif sname and sname:lower():find(p:lower()) then
+        elseif sname and sname:lower():find(p:lower(), 1, true) then
           r[#r+1]=s
         end
       end


gwww:
I have tested this on 0.9.85 by editing my local installation of hs.screen init.lua.

asmagill:
The name method now checks to see if the macOS recognizes the screenName parameter to NSScreen and if so honors it rather than doing an IOKit search that doesn't work on all (especially newer) machines. I think anyone running Catalina or newer may see a name change like this.

You are correct that a dash in the second parameter to string.find would be treated as a range indicator... and in this context it probably is simpler to use a plain text search rather than convert all hyphens to `%-`.

@cmsj are we doing approvals for PRs in the new process? Or is adding the appropriate tag to it sufficient?

zzamboni:
The behavior of `screen.find` is [documented](http://www.hammerspoon.org/docs/hs.screen.html#find) as interpreting the argument as a Lua pattern which gets matched against the screen name:

> a string pattern that matches (via string.match) the screen name as per hs.screen:name() (for convenience, the matching will be done on lowercased strings)

I don't see a good reason to change this behavior, since it provides more flexibility than plain string finding.

gwww:
I agree with @zzamboni, I withdraw this patch.

asmagill:
@zzamboni missed that... (note to self: make sure to re-read docs before commenting in the future!) I was thinking of it in terms of just comparing other captured screen names rather than the user trying to search for a specific one.

Agree with dropping this pull.

PR #2733, opened 2021-03-05
hs.window.desktop() now returns nil if Finder is not running

luckman212:
Found a small issue in `hs.window` where, if you have a function that depends on the special Desktop window— `hs.window.desktop()` it will fail when Finder.app is not running (either you quit it manually, or via AppleScript, or it crashed, etc.)

repro:
1. quit Finder
2. from hs console:
```lua
hs.window.desktop()
```

Error:
```
> hs.window.desktop()
...oon.app/Contents/Resources/extensions/hs/window/init.lua:51: attempt to index a nil value (local 'finder')
stack traceback:
	...oon.app/Contents/Resources/extensions/hs/window/init.lua:51: in function 'hs.window.desktop'
	(...tail calls...)
	[C]: in function 'xpcall'
	...app/Contents/Resources/extensions/hs/_coresetup/init.lua:514: in function <...app/Contents/Resources/extensions/hs/_coresetup/init.lua:494>
```

fix is in `extensions/hs/window/init.lua` - to check for nil before the iter:
```lua
if not finder then return false end
```

### DIFF
--- a/extensions/window/init.lua
+++ b/extensions/window/init.lua
@@ -38,7 +38,7 @@ window.animationDuration = 0.2
 ---  * None
 ---
 --- Returns:
----  * An `hs.window` object representing the desktop
+---  * An `hs.window` object representing the desktop, or nil if Finder is not running
 ---
 --- Notes:
 ---  * The desktop belongs to Finder.app: when Finder is the active application, you can focus the desktop by cycling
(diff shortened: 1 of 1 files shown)


cmsj:
I tweaked this so the function returns `nil` rather than `false` if Finder isn't running - I think that's more consistent with how we do this sort of thing elsewhere.

luckman212:
Ah yes, `nil` does make more sense. Thanks for the merge!

PR #2737, opened 2021-03-06
Fix some errors found via Xcode static analysis

cmsj:


cmsj:
@asmagill could you double check this for me? I'm pretty sure I reasoned them correctly, but I've been wrong before 😬

extensions/mouse/internal.m
```diff
@@ -114,7 +114,7 @@ -(io_service_t)createIOHIDSystem {
 }
 
 -(NSDictionary *)getIOHIDParamtersFromService:(io_service_t)service {
-    return (__bridge NSDictionary *)IORegistryEntryCreateCFProperty(service, CFSTR(kIOHIDParametersKey), kCFAllocatorDefault, kNilOptions);
+    return CFBridgingRelease(IORegistryEntryCreateCFProperty(service, CFSTR(kIOHIDParametersKey), kCFAllocatorDefault, kNilOptions));
```
asmagill, line 117: The constructs aren't quite identical -- `(__bridge NSDictionary *)` doesn't change ownership; `CFBridgingRelease` does.

I haven't looked at the rest of the code, so maybe we were wrong to begin with, but I did want to point out the change in semantics just in case you also need to add something elsewhere to clean it up.

asmagill:
I'd go ahead and remove the `NSEnableScreenUpdates` and `NSDisableScreenUpdates` from the `drawRect:` method of canvas... they don't do anything anymore and just annoy the compiler 😜

As to the two methods I added so users could manually trigger them, I don't know if anyone is actually using them... I'd either deprecate them or remove them entirely... at the very least update the docs to say that they don't do anything anymore and will go away at some point.

If you want me to do it, I'll try and get some time tomorrow or Monday -- today is already mostly shot for me.

extensions/mouse/internal.m
cmsj: From looking at the analysis, I think that change of ownership is needed - `IORegistryEntryCreateCFProperty()` is a Create under the CF ownership rules, so we are responsible for freeing its return, but we just bridge it to `NSDictionary` and nothing else, but by changing it to a `CFBridgingRelease()`, it's ARC's responsibility and thus returning it as an `NSDictionary` is fine.....

..... right? :)

asmagill: Sounds right to me.

PR #2754, opened 2021-03-10
A crash bug in hs.settings was fixed. [#2753](https://github.com/Hammerspoon/hammerspoon/issues/2753)

cmsj:


cmsj:
@asmagill could I get a quick sanity check on this please?

PR #2758, opened 2021-03-11
All Lua assert()ions are now enabled, and redirected to NSCAssert() for improved capture/logging

cmsj:


cmsj:
I'm going to rework this, I'm not fond of replacing macros with functions calls here, so we'll just go with adding the LUAI_ASSERT define.

PR #2765, opened 2021-03-12
Added an isData optional parameter to hs.websocket:send()

### LINKED ISSUE #2752: hs.websocket sends all messages as WebSocket Binary even when the message is a lua string
If any of you are using ProPresenter, you probably know it uses a simple websocket interface for remote control.

`hs.websocket` used to be able to talk to it, but it doesn't seem to be working anymore.

I've done a fair amount of testing, and it seems like `hs.websocket` is working fine with every websocket server I've tested it with, but it is not working with ProPresenter for some reason. My guess is that hammerspoon is sending something different in the headers that causes ProPresenter to ignore the data.

Another option is that hammerspoon is having a problem accessing the ProPresenter websocket endpoint. ProPresenter serves web sockets over a subdirectory. Here's the url for my case:

`ws://192.168.10.50:60157/stagedisplay`

Can anyone confirm this problem? Is there a way for me to inspect or modify the websocket headers sent from hammerspoon?

EDIT:
See below... I've determined the problem seems to be that `hs.websocket:send` always sends websocket messages as WebSocket Binary (opcode set to 0x0010) instead of Text (opcode set to 0x0001). Some programs (like ProPresenter) expect Text messages and only listen to text messages.

latenitefilms:
- Closes #2752

### DIFF
--- a/extensions/websocket/internal.m
+++ b/extensions/websocket/internal.m
@@ -183,14 +183,9 @@ static int websocket_send(lua_State *L) {
     HSWebSocketDelegate* ws = getWsUserData(L, 1);
 
     BOOL isData = (lua_gettop(L) > 2) ? (BOOL)(lua_toboolean(L, 3)) : YES ;
-    
-    if (isData) {
-        NSData *message = [skin toNSObjectAtIndex:2 withOptions: LS_NSLuaStringAsDataOnly];
-        [ws.webSocket send:message];
-    } else {
-        NSString *message = [skin toNSObjectAtIndex:2 withOptions: LS_NSPreserveLuaStringExactly];
-        [ws.webSocket send:message];
-    }
+
+    NSUInteger options = isData ? LS_NSLuaStringAsDataOnly : LS_NSPreserveLuaStringExactly;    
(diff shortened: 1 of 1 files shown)


latenitefilms:
@asmagill - Are you able to idiot check this?

@cmsj - I saw your message on the Google Groups, but pod headers we are using look different to the ones you linked to (i.e. there is no `sendString` and `sendData` methods).

asmagill (approved):
lgtm

cmsj:
> @cmsj - I saw your message on the Google Groups, but pod headers we are using look different to the ones you linked to (i.e. there is no `sendString` and `sendData` methods).

Ah, looks like that was added in the (now archived) GitHub repo for SocketRocket, but never released.

extensions/websocket/internal.m
```diff
@@ -163,21 +163,35 @@ static int websocket_new(lua_State *L) {
 
 /// hs.websocket:send(message) -> object
```
cmsj, line 164: Needs a [, isData]

cmsj (changes requested):

cmsj (approved):

PR #2767, opened 2021-03-12
Fixed potential crash in hs.serial when reloading Hammerspoon

### LINKED ISSUE #2763: Crash in hs.serial
This might be a bug in the underlying [ORSSerialPort](https://github.com/armadsen/ORSSerialPort) pod?

CC @armadsen in case he's interested.

Crash log: https://sentry.io/share/issue/57c1cd8f2df7404bb31ff5141a5b902a/

latenitefilms:
- Closes #2763

### DIFF
--- a/Hammerspoon Tests/HSserial.m
+++ b/Hammerspoon Tests/HSserial.m
@@ -1,0 +1,45 @@
+//
+//  hs.serial Tests
+//
+
+#import "HSTestCase.h"
+
+@interface HSserial : HSTestCase
+
+@end
+
+@implementation HSserial
+
+- (void)setUp {
+    [super setUpWithRequire:@"test_serial"];
+}
+
+- (void)tearDown {
+    [super tearDown];
+}
+
+- (void)testAvailablePortNames {
+    RUN_LUA_TEST()
+}
+
+- (void)testAvailablePortPaths {
+    RUN_LUA_TEST()
+}
+
+- (void)testNewFromName {
+    RUN_LUA_TEST()
+}
+
+- (void)testNewFromPath {
+    RUN_LUA_TEST()
+}
+
+- (void)testOpenAndClose {
+    RUN_LUA_TEST()
+}
+
+- (void)testAttributes {
+    RUN_LUA_TEST()
+}
+
+@end
--- a/scripts/docs/bin/build_docs.py
+++ b/scripts/docs/bin/build_docs.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python3
+#!/usr/bin/env python3
 # -*- coding: utf-8 -*-
 """Hammerspoon API Documentation Builder"""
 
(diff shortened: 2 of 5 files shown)


scripts/docs/bin/build_docs.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python3
+#!/usr/bin/env python3
```
cmsj, line 1: I know it makes everyone but my life harder to use the OS provided python3, but I want it to stay that way - I don't want to have to try and make sure that the same python environment is being used by makefiles, pip installs, local builds, CI builds, release scripts, etc.

Hammerspoon Tests/HSserial.m
```diff
@@ -0,0 +1,45 @@
...
+- (void)setUp {
+    [super setUpWithRequire:@"test_serial"];
+}
+
+- (void)tearDown {
+    [super tearDown];
+}
+
+- (void)testAvailablePortNames {
+    RUN_LUA_TEST()
```
cmsj, line 22: I think these are all going to need a SKIP_IN_GITHUB_ACTIONS() before the RUN_LUA_TEST()s. It's awesome that you've added tests, but it looks like they depend on /dev/cu.MALS existing, which I'm guessing is the hardware you're using this with.

I've never been able to figure out what it is, but my Mac Pro seems to have a /dev/cu.pci-serial2, so maybe we can add the skips and then in the Lua test file, figure out which of the two serial devices to use - at least that way the test will pass on both your machine and mind :)

cmsj (changes requested):

scripts/docs/bin/build_docs.py
latenitefilms: Ah, yes, you can revert this when merging. I'm still on Mojave, so I needed to make the change to make Hammerspoon build.

Hammerspoon Tests/HSserial.m
latenitefilms: `/dev/cu.MALS` should be built into macOS - do you not see it on your machine?

latenitefilms: I've changed to `Bluetooth-Incoming-Port`, which I'm HOPING is on most/all Mac's?

cmsj: yeah I have the bluetooth incoming one, but no MALS. Weird

asmagill: FWIW, I don't have the MALS one, either... what device does it correspond to?

latenitefilms: Maybe it's something to do with MacBook Pro's - connection to the Touch Bar?

asmagill: If so, then it's an intel thing as my M1 doesn't haven't it either (and it does have the touchbar). Oh well, I do think that the bluetooth one is probably more reliably available, so my curiosity about the device isn't that important 😜

PR #2772, opened 2021-03-14
A potential crash relating to hs.eventtap callback return value type errors, has been guarded against.

cmsj:


cmsj:
@asmagill sorry to keep pestering you about all these crasher issues - could you sanity check this please?

PR #2779, opened 2021-03-15
Fixed regression in hs.json which meant it sometimes returned a string instead of a table

### LINKED ISSUE #2778: hs.json.decode bug (introduced in 0.9.85?)
I have a bunch of code that relies on `hs.json.decode` that stopped working after the 0.9.85 update (still doesn't work in 0.9.86). I investigated a bit and couldn't figure out what was going on-- `hs.json.decode` seemed to sometimes be returning a string when it should've been returning a table. I was observing this by inserting `type(hs.json.decode(myJsonStr))` (which would return 'string' when it should've returned 'table') within the functions I was using, but I couldn't replicate this in the console.

In any case, I've temporarily fixed my issues by installing https://github.com/neoxic/lua-json and overwriting hs.json.decode in my init.lua:

```lua
local js = require('json')

local function decodeJson(jsonStr)
  local decoded, _ = js.decode(jsonStr)
  decoded['__array'] = nil
  return decoded
end

hs.json.decode = decodeJson
```

latenitefilms:
- Closes #2778

### DIFF
--- a/extensions/json/internal.m
+++ b/extensions/json/internal.m
@@ -96,15 +96,15 @@
 
 - (id)decodeFromFile:(NSString *)path {
     LuaSkin *skin = [LuaSkin sharedWithState:NULL];
-    NSError *error;
-    NSString *json = [NSString stringWithContentsOfFile:path encoding:NSUTF8StringEncoding error:&error];
-
+    NSError *error = nil;
+    NSData *json = [NSData dataWithContentsOfFile:path options:0 error:&error];
+    
     if (error) {
         [skin logError:[NSString stringWithFormat:@"Error reading JSON from file: %@", error.localizedDescription]];
         return nil;
     }
 
-    return [self decode:[json dataUsingEncoding:NSUTF8StringEncoding]];
+    return [self decode:json];
 }
 @end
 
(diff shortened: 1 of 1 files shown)


latenitefilms:
@cmsj - This passes the tests, but do you want to do a quick idiot check to make sure the changes are correct?

extensions/json/internal.m
```diff
@@ -105,7 +104,7 @@ - (id)decodeFromFile:(NSString *)path {
         return nil;
     }
 
-    return [self decode:json];
+    return [self decode:[json dataUsingEncoding:NSUTF8StringEncoding]];
```
cmsj, line 107: Maybe here instead of doing `NSString stringWithContentsOfFile` we should just do `NSData dataWithContentsOfFile`?

cmsj (changes requested):
One possible improvement, but overall lgtm. Assuming the tests still pass, I'm fine with merging this, although it would be nice to have some json data that would confuse the current version, so we can prove it's fixed.

latenitefilms:
I'll have a look at this tonight, and see if I can find a scenario that breaks it.

cmsj (approved):
lgtm

PR #2786, opened 2021-03-21
A possible crash in hs.webview:evaluateJavaScript() is fixed

cmsj:


cmsj:
@latenitefilms this fixes https://sentry.io/organizations/commandpost/issues/2287815197/?environment=production&project=5233380&referrer=alert_email

cmsj:
@asmagill FYI. I don't think this should cause any issues, but CommandPost got a crash where this callback was trying to happen off the main thread, so it seems reasonable to `dispatch_async()` it onto the main thread.

asmagill:
The only caveat to be aware of when using `dispatch_async()` to put something onto the main thread is that it becomes theoretically possible for the block to be invoked *after* a reload and thus run into the issue of confused lua states.

Yeah, the window from a dispatch is theoretically infinitesimal, but we think it's happened with other types of callbacks that queue to the  main thread, so... keep an eye on it... you may need to apply the experimental fix you've been working on for these edge cases (which I should review at some point so I understand what you're doing... is there a good example of one that you would recommend I review?)

cmsj:
@asmagill good point, I'll work on updating this to use the new protection mechanism.

hs.timer's callback function is likely the simplest place to see it in action.

PR #2789, opened 2021-03-22
A stability regression in hs.location was fixed

cmsj:


cmsj:
Holding this one until I write some tests that cover it.

asmagill:
As an alternative, you could have the helper return `[NSNull null]` instead of `@""` when the item is empty/non-existent -- It will satisfy the NSDictionary/NSArray that the objects are being stored in and LuaSkin will push this as `nil` which would cause the empty fields to just not be present instead of having an empty string.

Just an idea.

And does this mean Apple has significantly changed or deprecated the classes being used by `hs.location`? I haven't checked in a while but if there is a newer (and hopefully better) class, it may mean this module is ripe for a rewrite in the near future.

cmsj:
Good point about `NSNull`, although I wonder if that might be more hassle for users, since they'd have to check each item in the dictionary to see if it's a string or not.

It doesn't look like there has been massive change, they just deprecated that address dictionary for some reason.

asmagill:
Depends on your use case... If the intent is to print the data out, then yeah, an empty string simplifies that, but if you need to to check if a field contains useful information, `nil` is treated as false while an empty string is still true and you have to check lengths...

At any rate, it's not a big deal, just wanted to make sure you knew that `[NSNull null]` is in fact recognized by LuaSkin -- it comes up so seldom I sometimes forget about it myself.

cmsj:
I'm abandoning this for now, I merged in #2846 which reverts to the old API with deprecation warning suppression. I'm expecting it'll be some time until Apple actually removes the `addressDictionary` API, so we can address this later.

PR #2790, opened 2021-03-22
Alternate method for nil app obj detection in hs.window

luckman212:
belt and suspenders

### DIFF
--- a/extensions/window/init.lua
+++ b/extensions/window/init.lua
@@ -253,8 +253,10 @@ end
 --- Notes:
 ---  * This does not mean the user can see the window - it may be obscured by other windows, or it may be off the edge of the screen
 function objectMT.isVisible(self)
-  if not self:application() then return false end
-  return not self:application():isHidden() and not self:isMinimized()
+  if getmetatable(self).__type ~= 'hs.window' then return end
+  local parentApp = self:application()
+  if not parentApp then return false end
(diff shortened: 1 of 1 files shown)


luckman212:
@cmsj Sorry to bring this one back from the dead, but I got this today in my console, even with this patch applied:

image

In this case, PID 90218 was Hammerspoon itself, a bit odd.

image

Not sure what could be triggering this, but I added a couple of more checks and assign `self:application()` to an object to avoid calling the method twice in a row. So instead of

```lua
function objectMT.isVisible(self)
  if not self:application() then return false end
  return not self:application():isHidden() and not self:isMinimized()
end
```

I'm trying

```lua
function objectMT.isVisible(self)
  if getmetatable(self).__type ~= 'hs.window' then return end
  local parentApp = self:application()
  if not parentApp then return false end
  return not parentApp:isHidden() and not self:isMinimized()
end
```

Not sure if you agree.

luckman212:
@cmsj Just an anecdotal update - since these changes I haven't seen this error again.

luckman212:
@cmsj I see this didn't make the cut for 0.9.87 😞 
Any changes you want me to make?

cmsj:
@luckman212 all good thanks, 0.9.87 was just cut in a hurry :)

PR #2799, opened 2021-03-27
hs.alert.showWithImage()

pR0Ps:
A new function `hs.alert.showWithImage(message, image, ...)` was added that adds the ability to show an image in alerts.

This function supports the same optional parameters as `hs.alert.show` and allows for displaying app icons and other images next to alerts.

The alerts it creates look like this:

EDIT: see new image below

Screen Shot 2021-03-26 at 9 29 19 PM


pR0Ps:
Improved padding calculations (now based on text size by default so it scales nicely). New screenshot:

Screen Shot 2021-03-27 at 11 44 50 PM

Note that the padding isn't pixel perfect in this example since the bundle images returned from `hs.image.imageFromAppBundle` have some padding built into them already.

pR0Ps:
I see some tests failed but it doesn't look related to my changes. Is there anything I need to do for this PR?

cmsj:
@pR0Ps no you're good, I'll dig into it when I get some time

cmsj:
Thanks!

Issue #2793, opened 2021-03-25
hs.screen.find() no longer works


rectalogic:
As of 0.9.86 on macOS 10.15.7, `hs.screen.find()` no longer works.
```
> hs.screen.allScreens()[2]:name()
Built-in Retina Display
> hs.inspect(hs.screen.find("Built-in Retina Display"))
nil
```

asmagill:
Screen Find uses string.match for comparison. Try “Built%-in Retina Display”. If that doesn’t work, try seeing what the following reports (making this work on the M1 Macs under Big Sur required a change in the way the name is captured):

~~~lua
for i, v in ipairs(hs.screen.allScreens()) do
    print(v:name())
end
~~~

rectalogic:
Yes that works, thanks
```
> hs.screen("Built%-in Retina Display")
hs.screen: Built-in Retina Display (0x60600047d5f8)
```

PR #2800, opened 2021-03-29
Add hs.screen example for finding Built-in Retina Display

dacort:
As noted in #2793 finding "Built-in" display requires escaping the `-` character because it is a repetition character in `string.match`.

This adds an example to the docs because I feel like it's a fairly common task and those unfamiliar with Lua will get tripped up.


cmsj:
Thanks very much!

PR #2805, opened 2021-04-05
documentation fixes

Habbie:
This PR contains:
1. a commit that makes build_docs.py more portabl2
2. a formatting fix for hs.dialog.textPrompt

Many other bits of doc are broken like (2) fixes - please let me know if this fix is good or if you'd like some other format, then maybe I can go through more of the docs to fix this.

### DIFF
--- a/scripts/docs/bin/build_docs.py
+++ b/scripts/docs/bin/build_docs.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python3
+#!/usr/bin/env python3
 # -*- coding: utf-8 -*-
 """Hammerspoon API Documentation Builder"""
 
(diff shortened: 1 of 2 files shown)


cmsj:
@Habbie could we have this again, without the `/usr/bin/env` change, since that's now been handled by #2798

scripts/docs/bin/build_docs.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python3
+#!/usr/bin/env python3
```
cmsj, line 1: This part is no longer needed, it was already handled by another PR.

cmsj (changes requested):

PR #2810, opened 2021-04-07
Add links directly to HS source, to the generated docs

cmsj:


cmsj:
@asmagill this is currently only exposed in the HTML docs - would passing the info into JSON/Markdown be useful/interesting to you?

cmsj:
(the context for doing this is that I want docstring lint failures to emit annotations so a PR will get a proper comment against the line that is wrong, but I figured it would also be interesting to link directly to the source from docstrings)

Issue #2807, opened 2021-04-06
Add scale option to hs.canvas:imageFromCanvas()?


latenitefilms:
@asmagill - Thoughts on adding a `scale` option to `hs.canvas:imageFromCanvas()`, similar to what you find in `hs.image:encodeAsURLString()`?

> `scale` - an optional boolean, default false, which indicates that the image size (which macOS represents as points) should be scaled to pixels. For images that have Retina scale representations, this may result in an encoded image which is scaled down from the original source.

asmagill:
The source material for the image is entirely different, so not exactly sure where to start... what are you trying to accomplish with this? Might help to have a better idea of the actual problem first.

latenitefilms:
I'm sending `hs.image` icons to a control surface. They need to be 60x90 pixels, however for legacy reasons, some of our icons are stored as 90x90 pixels, so we're using `hs.canvas` to resize those using the below code:

```
local decodedImage = imageFromURL(encodedIcon)
local size = decodedImage and decodedImage:size()
if size and size.w ~= 60 or size.h ~= 90 then
    --------------------------------------------------------------------------------
    -- The Knob Icon isn't 60x90 pixels, so it must be a legacy icon that needs
    -- to be scaled:
    --------------------------------------------------------------------------------
    local v = canvas.new{x = 0, y = 0, w = 60, h = 90 }

    v[1] = {
        frame = { h = "100%", w = "100%", x = 0, y = 0 },
        fillColor = { alpha = 1, hex = "#000000" },
        type = "rectangle",
    }

    v[1] = {
      type = "image",
      image = decodedImage:croppedCopy({ x = 15, y = 0, h = 90, w = 60 }),
      frame = { x = 0, y = 0, h = 90, w = 60 },
    }

    local fixedImage = v:imageFromCanvas()

    local encoded = fixedImage:encodeAsURLString(true)
    decodedImage = imageFromURL(encoded)

    v:delete()
    v = nil -- luacheck: ignore
end
```

Currently I'm using `encodeAsURLString(true)` to get around a scaling issue.

dasmurphy:
I know this is a long shot and a not really good workaround, but how about making the image in accordance to the `scale`.

```lua
local scale = hs.screen.mainScreen():currentMode().scale

local v = canvas.new{x = 0, y = 0, w = 60/scale, h = 90/scale }
-- ... mode code...
-- also do this in the croppedCopy command
```

[...]

PR #2814, opened 2021-04-09
Add method to hs.image to create bitmap representation of image for external devices

asmagill:
Addresses #2807

### DIFF
--- a/extensions/image/internal.m
+++ b/extensions/image/internal.m
@@ -662,17 +662,17 @@ static int additionalImages(lua_State *L) {
 static int getExifFromPath(lua_State *L) {
     LuaSkin *skin = [LuaSkin sharedWithState:L] ;
     [skin checkArgs:LS_TSTRING, LS_TBREAK] ;
-    
+
     NSString* imagePath = [skin toNSObjectAtIndex:1];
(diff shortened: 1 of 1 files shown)


asmagill:
@cmsj, skimmed the Contributions document and as usual, I'm behind on tests... what kind of tests would you think this method should have? I suppose I could create an image (maybe via ASCII or by system name) and then make sure that the bitmrep doesn't equal it, but *can* we do something more? What kind of tests would you do if you had added this? I'll write them, I'm just not quite sure what would be reasonable here...

cmsj:
@asmagill even just checking that it returns an `hs.image` object would be a good first step. In the case of an image it's kinda hard to verify that the result is "correct" without looking at it, but maybe `hs.image:size()` would also make sense in a test that specifically constrains the size of the bitmap representation?

latenitefilms:
@cmsj - Any chance we could merge this, and then I could do the tests in another pull request?

cmsj:
Since this had a conflict with the new ascii method, I've merged this PR's branch by hand and pushed it as e9034e8a26442a406a1f21e99c594c4dc58293bf

PR #2815, opened 2021-04-10
Switch to using dynamic version/build numbers

cmsj:


cmsj:
@latenitefilms this is probably very relevant to you, depending on how you do releases in CommandPost.

The basic plan here is that instead of hard-coding the version and build numbers in Xcode, those stay permanently at "1.0" and "1" respectively, and then whenever a build is performed, the Info.plist inside `Hammerspoon.app` is modified to contain the latest tag as the version, and the total number of git commits as the build number.

One super important caveat here, is that there are two types of git tags and this only works with one of them, specifically, "annotated" tags. It turns out that for almost all of Hammerspoon's life I've been doing the release tags as the other kind, "lightweight" tags, which is super annoying because I now can't fix that, but I just pushed out 0.9.87 with an annotated tag, so I'm clear now, but you might need to pay special attention before merging HS master once this PR lands.

I've updated the release script to check that the tag it's using is an annotated one, but I've also removed the explicit version number argument to the script - instead it will just fetch the latest annotated tag and use that as the version number, so if I accidentally create a lightweight tag, the release process will fail because it checks if github already has a release with that number.

(to make an annotated tag vs a lightweight one, just do "git tag -a x.y.z" instead of "git tag x.y.z")

I hope that's all clear - feel free to ask any questions, or let me know if this somehow irrevocably breaks CP. My plan from here is to try and get GitHub CI to automatically spit out a signed build for each commit to master, with a version number that reflects the fact that it's not a proper release.

latenitefilms:
Haha, I was just looking at this pull request and thinking... "mmmmm, this will break things".

All good! All makes sense.

Currently CommandPost works across two repos - [CommandPost-App](https://github.com/CommandPost/CommandPost-App) is the Hammerspoon fork, which contains all the Objective-C stuff - which I try and keep in sync with Hammerspoon, then [CommandPost](https://github.com/CommandPost/CommandPost) is our main Lua code repo. I use a [release script](https://github.com/CommandPost/CommandPost/tree/develop/scripts) in the CommandPost repo, which is basically the same as Hammerspoon's, just slightly modified - and as part of the build process copies all the files from the CommandPost repo into the CommandPost-App application bundle.

I'll look into the tags, thanks for the heads up!

PR #2817, opened 2021-04-10
Bugfix for allowing a table argument to be optional.

cklein:
I'm writing a mixed extension where I'd like to pass some data from a config file to a lua function
that passes it to an Objective-C function (that middle step might be eliminated).

The function accepts two strings and two optional tables. I'm checking arguments in Objective-C like that.
```
[skin checkArgs:LS_TSTRING, LS_TSTRING, LS_TTABLE|LS_TOPTIONAL, LS_TTABLE|LS_TOPTIONAL, LS_TBREAK];
```

However when calling the function as `f("hello", "world")`, the following error message is shown,
indicating that `LS_TTABLE|LS_TOPTIONAL` might not work as expected:

```
...Contents/Resources/extensions/hs/myext/init.lua:12: ERROR: incorrect type 'nil' for argument 3 (expected table)
stack traceback:
    [C]: in function 'hs.myext._openWithAppAndConfig'
    ...Contents/Resources/extensions/hs/myext/init.lua:12: in function 'hs.myext.openWithAppAndConfig'
    /Users/cklein/.hammerspoon/init.lua:55: in function </Users/cklein/.hammerspoon/init.lua:48>
2021-04-11 01:12:28.243733+0200 Hammerspoon[33519:3082278] BREADCRUMB: (ERROR) hs.chooser:completionCallback: ...Contents/Resources/extensions/hs/myext/init.lua:12: ERROR: incorrect type 'nil' for argument 3 (expected table)
stack traceback:
    [C]: in function 'hs.myext._openWithAppAndConfig'
    ...Contents/Resources/extensions/hs/myext/init.lua:12: in function 'hs.myext.openWithAppAndConfig'
    /Users/cklein/.hammerspoon/init.lua:55: in function </Users/cklein/.hammerspoon/init.lua:48>
```

`!(spec & lsType)` also looks suspicious, or at least I don't understand it,
the two value sets don't seem to have anything to do with each other:

`lsType` is one of the following values:

```
#define LUA_TNONE       (-1)
#define LUA_TNIL        0
#define LUA_TBOOLEAN        1
#define LUA_TLIGHTUSERDATA  2
#define LUA_TNUMBER     3
#define LUA_TSTRING     4
#define LUA_TTABLE      5
#define LUA_TFUNCTION       6
#define LUA_TUSERDATA       7
#define LUA_TTHREAD     8
#define LUA_NUMTYPES        9
```

`spec` is a combination of the following values:

```
#define LS_TBREAK         1 << 0
#define LS_TOPTIONAL      1 << 1
#define LS_TNIL           1 << 2
#define LS_TBOOLEAN       1 << 3
#define LS_TNUMBER        1 << 4
#define LS_TSTRING        1 << 5
#define LS_TTABLE         1 << 6
#define LS_TFUNCTION      1 << 7
#define LS_TUSERDATA      1 << 8
#define LS_TNONE          1 << 9
#define LS_TANY           1 << 10
#define LS_TINTEGER       1 << 11
#define LS_TVARARG        1 << 12
#define LS_TTYPEDTABLE    1 << 13
#define LS_TWRAPPEDOBJECT 1 << 14
```

So it seems that only `lsType` divisible by 2 were allowed to be optional.
Note that I didn't remove that condition.

### DIFF
--- a/LuaSkin/LuaSkin/Skin.m
+++ b/LuaSkin/LuaSkin/Skin.m
@@ -734,7 +734,7 @@ catastrophe:
                 break;
         }
 
-        if (!(spec & LS_TANY) && !(spec & lsType) && !(spec & LS_TWRAPPEDOBJECT)) {
+        if (!(spec & LS_TANY) && !((lsType == LS_TNIL) && spec & LS_TOPTIONAL) && !(spec & lsType) && !(spec & LS_TWRAPPEDOBJECT)) {
             luaL_error(self.L, "ERROR: incorrect type '%s' for argument %d (expected %s)", luaL_typename(self.L, idx), idx, specMaskToString(spec).UTF8String);
         }
 nextarg:


asmagill:
In the C-API, there is a difference between an explicit `nil` and an *implicit* nil because there is no argument provided. The Lua scripting language doesn't make this distinction, but the C-API (which LuaSkin uses) does...

What's the lua code around `hs/myext/init.lua:12`? Is there an explicit `nil` being passed into the function instead of one of the optional tables?

cklein:
My config file looks like this:
```
local app = "/Applications/Visual Studio Code.app"
local path = "/Users/cklein/.hammerspoon/init.lua"
hs.launchservices.openWithAppAndConfig(path, app, nil, nil)
```

At first the last line looked like thi and I thought from what you wrote that adding the explicit nil would fix it:
```
hs.launchservices.openWithAppAndConfig(path, app)
```

`luaType` is 0 (`LUA_TNIL`), therefore `lsType` is set to 4 (`LS_TNIL`), `spec` is 66 (`LS_TTABLE | LS_TNIL`).

Having written the last part, changing the following line in the Objective-C file also fixes it: 
```
    // Before:
    // [skin checkArgs:LS_TSTRING, LS_TSTRING, LS_TTABLE|LS_TOPTIONAL, LS_TTABLE|LS_TOPTIONAL, LS_TBREAK];
    // Now:
    [skin checkArgs:LS_TSTRING, LS_TSTRING, LS_TTABLE|LS_TNIL, LS_TTABLE|LS_TNIL, LS_TBREAK];
    // This also works:
    // [skin checkArgs:LS_TSTRING, LS_TSTRING, LS_TTABLE|LS_TNIL, LS_TTABLE|LS_TNIL, LS_TBREAK];
```

Is this the expected behaviour? If yes, what does optional mean then?

The extension I'm working on is a single commit in https://github.com/cklein/hammerspoon/tree/extension-launchservices

asmagill:
If you set up the check arguments line as:

~~~objc
[skin checkArgs:LS_TSTRING, LS_TSTRING, LS_TTABLE|LS_TOPTIONAL, LS_TTABLE|LS_TOPTIONAL, LS_TBREAK];
~~~

Then only the following are legitimate:
* `function("abc", "def")`
* `function("abc", "def", { ... })`
* `function("abc", "def", { ... }, { ... })`
Passing in the arguments as variables will work, as long as they meet the agreed upon types... the following will *not* work, even though it would for a function defined in Lua:
~~~lua
local tbl1 = nil
local tbl2 = { 1,2,3 }
function("abc", "def", tbl1, tbl2) -- fails because tbl1 is an explicit `nil`
~~~

The reason why I wanted to see the lua code that was calling your C function was to see if something like the last example I gave was happening -- the error text "ERROR: incorrect type 'nil' for argument 3 (expected table) stack traceback:" suggests that this is what was occurring.

`LS_TOPTIONAL` indicates that this specific argument is optional; however, it is still place dependent, meaning that every argument *after* it must also be optional because it doesn't alter the index position that is being checked, and there is no way to send a `LUA_TNONE` type argument in that position. This also means that you can't "leave out" an "earlier" optional variable but specify a later one (e.g. if it's checking for `LS_TSTRING | LS_TOPTIONAL, LS_TTABLE | LS_TOPTIONAL`, then passing `function({1,2,3})` is invalid because a table is allowed as argument 2, not argument 1. But since they're both optional, you could leave *both* out: `function()` would be valid).

Most likely what you want is either:
* `[skin checkArgs:LS_TSTRING, LS_TSTRING, LS_TTABLE|LS_TNIL|LS_TOPTIONAL, LS_TTABLE|LS_TOPTIONAL, LS_TBREAK];`, if you want to allow a `nil` placeholder for the first table
* or `[skin checkArgs:LS_TSTRING, LS_TSTRING, LS_TTABLE|LS_TNIL|LS_TOPTIONAL, LS_TTABLE|LS_TNIL|LS_TOPTIONAL, LS_TBREAK];`, if you want to allow a `nil` placeholder for either/both tables.

cklein:
OK, thanks for explaining it. This was not really clear to me. I'm going to close the PR.

asmagill:
Practical examples of LuaSkin's usage for (usually) simplifying module building is something I'd like to document more, but it's kind of low on my list at the moment.

This is a good reminder that we *do* need it, though, and this is definitely one of the topics that should be addressed in depth in the Wiki.

cmsj:
Thanks @asmagill for writing the detailed explanation. I agree we should look at documenting this stuff better, it's *very* subtle.

PR #2819, opened 2021-04-11
A potential crash in hs.screenRecordingState() was fixed

### LINKED ISSUE #2812: Crash in hs.screenRecordingState()?
`Exception 1, Code 0, Subcode 8 > Attempted to dereference null pointer.`

https://sentry.io/share/issue/ca15be252b634d88813903b0495e8411/

I can't see anything obvious in [`isScreenRecordingEnabled()`](https://github.com/Hammerspoon/hammerspoon/blob/master/Hammerspoon/MJLua.m#L273)?

Thoughts @cmsj and/or @asmagill ?

latenitefilms:
- Closes #2812

### DIFF
--- a/Hammerspoon/MJLua.m
+++ b/Hammerspoon/MJLua.m
@@ -309,7 +309,9 @@ bool isScreenRecordingEnabled()
                     }
                 }
             }
-            CFRelease(windowList);
+            if (windowList) {
+                CFRelease(windowList);
+            }
         }
         return canRecordScreen;
     } else {


cmsj:
How about we use https://stackoverflow.com/a/65423834 instead seems dramatically simpler and both routes are 10.15 only anyway.

latenitefilms:
Mojave doesn't have the option for Screen Recording Permissions, so yeah, this is only required on Catalina & Big Sur. I'm still on Mojave, and only jump into Big Sur for building CommandPost (although I do have a Catalina partition for testing) - however, `hs.screenRecordingState()` is pretty important in CommandPost, because we use it when you first run CommandPost to ensure that the user understands why Screen Recording permissions are required (i.e. we need it for window positioning, not just "recording" the screen on Catalina & Big Sur).

Last time I checked `CGPreflightScreenCaptureAccess()` just crashed Catalina - so I feel like this is a Big Sur only feature, even though the [documentation says macOS 10.15+](https://developer.apple.com/documentation/coregraphics/3656523-cgpreflightscreencaptureaccess)?

This [comment](https://github.com/lwouis/alt-tab-macos/issues/704#issuecomment-762852689) seems to back up the crashing, although they seem to have a [simpler workaround](https://github.com/lwouis/alt-tab-macos/commit/d621ce5c5063f3506567b49050ab732920dc934c) - but I think this would also make macOS prompt for permissions, which would break the `shouldPrompt` option?

Maybe you could use `CGPreflightScreenCaptureAccess()` for Big Sur and above, but keep the existing function for Catalina?

cmsj:
@latenitefilms fair enough, we can merge this fix and think about switching to the other method later when our minimum supported versions allow.

PR #2827, opened 2021-04-17
Add a new log delegate method which can be used to explicitly capture an event with Sentry

cmsj:


cmsj:
This is currently untested, and is a response to the current state of crash reports for 0.9.87, where there have yet to be any coming from hs.timer or hs.eventtap C->Lua callbacks. There are some others, which I will add `checkLuaSkinInstance` to in due course, but I also want to get events logged in Sentry when we do spot an inconsistent LuaSkin instance, because we still need to eventually understand how that is coming to be.

I have a nebulous hypothesis that it might be due to some strong references somewhere, but I don't have any evidence or leads yet (@asmagill maybe that triggers some ideas with you?)

cmsj:
Tested and a Sentry event was successfully created: https://sentry.io/share/issue/3c2c45bf3690407396648cec2c9de93f/

Issue #2826, opened 2021-04-17
`hs.json.encode` stops working in coroutine


gineer01:
@cmsj I think the fix for #2673 breaks `hs.json.encode` when it's used in coroutine. Code to reproduce the problem:

```
> coroutine.wrap(hs.json.encode)({1,2})
table: 0x600000a57e00
```

This wraps hs.json.encode as a coroutine and runs it with the input {1,2}. The expected output is a string "[1,2]". The actual result is just the input table, without any change.

```
> test = {1,2}


> coroutine.wrap(hs.json.encode)(test)
table: 0x60000058cf00

> test
table: 0x60000058cf00

```

gineer01:
If my guess [1] is correct, `hs.json.encode` works on the wrong LuaState and Lua just returns the last param (i.e., the top of the current coroutine's stack) as the return value:
```
> coroutine.wrap(hs.json.encode)({1,2}, true)
true

> coroutine.wrap(hs.json.encode)({1,2}, false)
false
```

`hs.json.decode` in the same commit likely has the same issue. Similarly, I've found quite a few `[LuaSkin sharedWithState:NULL]` in Hammerspoon's code. They probably will have similar issues with coroutine?

[1] https://github.com/Hammerspoon/hammerspoon/commit/dfc95b23d9fc17827bcc0f72488fec4c27e82e23#r49634522

asmagill:
See my comment at https://github.com/Hammerspoon/hammerspoon/commit/dfc95b23d9fc17827bcc0f72488fec4c27e82e23#commitcomment-49653802

We're actually lucky this isn't causing a crash... coroutine crashes because of improper `lua_State*` (usually) can't be caught by the exception handler.

cmsj:
this should be fixed in 0.9.88, and if you want to test a development build with the fix, there is one at https://github.com/Hammerspoon/hammerspoon/actions/runs/764916499 (but be aware that it won't be able to upgrade automatically to the released version of 0.9.88, you'd have to do that manually, or wait for 0.9.89)

PR #2830, opened 2021-04-19
hs.json functions now work when called from a Lua coroutine

asmagill:
Addresses #2778 and #2826

### DIFF
--- a/extensions/json/internal.m
+++ b/extensions/json/internal.m
@@ -9,8 +9,8 @@
 @end
 
 @implementation HSjson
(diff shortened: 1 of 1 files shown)


cmsj:
damn, I should have checked the CI more carefully, `testReadWrite` is failing

cmsj:
I think d858e5f should fix it, but weirdly my local Xcode is failing to launch tests at the moment. My monitors are also being weird and my mouse pointer disappears in full screen apps. I think I'm due a reboot after 32 days 🤣

PR #2837, opened 2021-04-20
If hs.loadSpoon() is called twice for the same Spoon, the second call will now return the first spoon object instead of reloading it

### LINKED ISSUE #2835: hs.loadSpoon should check to see if spoon already loaded
Because of the way the `hs.loadSpoon` function is written, if you load a spoon twice then metadata gets modified again (probably benign), init gets called again (if it exists), and the docs (if any) get loaded a second time (which will cause warnings when help is next accessed).

The fix should be as simple as adding `if package.loaded[name] then return package.loaded[name] end` as the first line of the function in _coresetup, but I'm not in a position to do so and test at the moment... I'll try tomorrow if no one gets to this first.

cmsj:
Closes #2835﻿


asmagill:
If they choose to load the spoon as non-global, (i.e. pass in `false` as the second argument, this version will still duplicate executing `spoon:init()` and load the docs file (which in truth probably only affects developers and then only if they manually build the docs) a second time.

The only definitive check for all cases is to check if `package.loaded[name]` has a value or not.

PR #2854, opened 2021-04-29
Keep MJLuaLogDelegate minimally existant so we can log breadcrumbs of lua_State lifecycle

cmsj:


cmsj:
@asmagill any objections here? The idea is not to destroy and recreate the logging delegate, so it can still be used to capture breadcrumbs in Sentry when `createLuaState` and `destroyLuaState` are called in a reload.

asmagill:
If you're certain that the only messages will be breadcrumbs, this should be sufficient. If you think there might be other types of logging before the new lua state is fully loaded, you should consider adding some sort of check to https://github.com/Hammerspoon/hammerspoon/blob/master/Hammerspoon/HSLogger.m#L51-L60

cmsj:
@asmagill I think it's still ok because that function starts with a nullness check for `_L` and just sends things to breadcrumbs if there's no Lua state (although that's arguably not a very helpful thing to be doing, I might switch that out for NSLog)

asmagill:
Missed that. Ok, yeah, as long as _L is NULL until everything is loaded, that should be fine.

cmsj:
Thanks for the sanity check :)

PR #2863, opened 2021-05-04
Expand docstring linting, HTML docs now list Parameters/Returns/Notes/Examples separately.

cmsj:


cmsj:
@latenitefilms this may be of interest to you because it adds some stricter linting to docstrings, so your CP modules might need updating

@asmagill I'm fairly certain this doesn't affect the markdown output, but I am working towards some changes that will, because I want to add more structure to the output of `build_docs.py`, so at some point I'm going to do things like break down the individual Parameters lines so I can show the name of each parameter as code rather than regular text, etc.

PR #2867, opened 2021-05-06
A crash in hs.hotkey, relating to premature garbage collection, which was incorrectly fixed in 0.9.89, is now fixed

cmsj:


cmsj:
This is currently untested, and the approach is kinda hacky, but Carbon doesn't give us much option for passing opaque pointers around.

PR #2892, opened 2021-06-08
[WIP] Fixes and updates for Xcode13

cmsj:


cmsj:
This won't land until autumn when Xcode 13 releases, but I want to start collecting together a branch that builds cleanly with Xcode13 during the summer as new betas appear.

asmagill:
Does this affect compiling under Xcode 12?  I'll test myself in a couple of days but I'm in transit right now.

cmsj:
Yes, it already doesn’t build with 12 - the audio constants are only available in the new SDK 😫

asmagill:
Generally we've been able to wrap these... is this a case of they don't exist in Big Sur/Xcode 12, or that they've changed? We've handled both before and I can take a closer look this weekend if you like.

PR #2901, opened 2021-06-16
[pr-change] Canvas objects now respond to lua garbage collection

### LINKED ISSUE #2730: Auto collect hs.canvas objects?
Reading #2729 and remembering a discussion I had on here a couple of weeks ago about hs.canvas and collection specifically, I wonder if it's worth biting the bullet and changing canvas to auto-collect now rather than in some nebulous  2.0 (I hope more in the sense of *when* than in the sense of actuality)?

We may have to field some issues in cases where people have created drawings that they leave up all/most of the time because they may see some of them suddenly disappear... but I'm not sure how many cases that will be -- how many use cases can you think of where an image is created and *never* removed? Maybe the rounded-corners spoon... but any others I can think of also involve hiding them based on some logic which means that they have to be capturing the drawing in a variable already thus preventing collection.

It's not a completely simple fix (more tedious than hard) so I don't want to give an eta quite yet, but are there any objections if I proceed sooner rather than later?

asmagill:
Will resolve #2730

This is currently being tested by me but I'm putting it up in case others want to test it as well. I *think* I've accounted for all combinations of fade and hide, but will contrive more tests tonight.

The delete method now prints a deprecation warning (for the first 10 calls -- I figure this makes it likely that it will show up at least once in the console after a user's `init.lua` is done but doesn't spam the console forever) and then falls through to `hide`.

### DIFF
--- a/extensions/canvas/Canvas.h
+++ b/extensions/canvas/Canvas.h
@@ -6,7 +6,8 @@
 @end
 
 @interface HSCanvasView : NSView <NSDraggingDestination>
-@property int                 selfRef ;
+@property int                 selfRef ; // used during fadeOut to make sure collection doesn't interrupt
+@property int                 selfRefCount ;
 @property HSCanvasWindow     *wrapperWindow ;
 @property int                 mouseCallbackRef ;
 @property int                 draggingCallbackRef ;
--- a/extensions/canvas/internal.m
+++ b/extensions/canvas/internal.m
@@ -6,6 +6,8 @@ static const char *USERDATA_TAG = "hs.canvas" ;
 static LSRefTable refTable = LUA_NOREF;
 static BOOL defaultCustomSubRole = YES ;
 
+static int warnedAboutDelete = 0 ;
+
 // Can't have "static" or "constant" dynamic NSObjects like NSArray, so define in lua_open
 static NSDictionary *languageDictionary ;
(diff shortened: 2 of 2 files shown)


latenitefilms:
So once this lands, we basically just need to remove the `hs.canvas:delete()` method and replace it with a `canvas = nil`?

We have one bit of code in CommandPost that potentially generates a LOT of Canvas objects (we use it to basically check a screenshot of the screen to detect a change in the UI) [here](https://github.com/CommandPost/CommandPost/blob/develop/src/extensions/cp/apple/finalcutpro/viewer/ControlBar.lua#L268), I wonder if this change in garbage collection will cause any issues?

asmagill:
When it goes out of scope, it will be collected. If you want to potentially speed this along, then yes, set it to nil. For now there is no particular downside to using delete (other than some messages to the console) as long as you are also allowing the object to go out of scope. I figure we'll remove delete all together when 2.0 rolls around.

asmagill:
Yes. The only question is whether you want the `delete` method to print a deprecation warning to the console (it currently does for the first 10 times it is invoked) or just be a silent `nop`.

cmsj:
Merged in manually since GitHub can't handle `canvas/internal.m` having been renamed to `libcanvas.m`

PR #2908, opened 2021-06-19
Handle VOX not running in isRunning() - fixes #2906

### LINKED ISSUE #2906: hs.vox.isRunning() throws an error if VOX is not running
[hs.voxisRunning()](http://www.hammerspoon.org/docs/hs.vox.html#isRunning) throws an error if VOX is not running because `hs.application.get("VOX")` [returns nil](https://github.com/Hammerspoon/hammerspoon/blob/master/extensions/vox/init.lua#L309) resulting in a `attempt to index a nil value` error.

von:
Fix for case when VOX isn't running and hence hs.application.find("VOX") returns nil. Fixes #2906.

### DIFF
--- a/extensions/vox/init.lua
+++ b/extensions/vox/init.lua
@@ -8,6 +8,8 @@ local alert = require "hs.alert"
 local as = require "hs.applescript"
 local app = require "hs.application"
 
+local voxBundleId = "com.coppertino.Vox"
+
 -- Internal function to pass a command to Applescript.
 local function tell(cmd)
   local _cmd = 'tell application "vox" to ' .. cmd
@@ -306,7 +308,8 @@ end
 --- Returns:
 ---  * A boolean value indicating whether the vox application is running
 function vox.isRunning()
-  return app.get("VOX"):isRunning() ~= nil
+  local apps = app.applicationsForBundleID(voxBundleId)
+  return #apps > 0
 end
 
 return vox


latenitefilms (changes requested):
I wonder if `hs.application.applicationsForBundleID` should be used instead to reduce the risk of Hammerspoon getting the wrong application (ie if you have a Safari window open with the VOX website, etc)?

von:
> I wonder if `hs.application.applicationsForBundleID` should be used instead to reduce the risk of Hammerspoon getting the wrong application (ie if you have a Safari window open with the VOX website, etc)?

Makes sense. As I understand `hs.application.applicationsForBundleID` it returns a table with applications with the given BundleID, so the only check that is needed is for a non-empty table. I.e. there is no need to call `isRunning()` on an app from the table to confirm?

von:
> > I wonder if `hs.application.applicationsForBundleID` should be used instead to reduce the risk of Hammerspoon getting the wrong application (ie if you have a Safari window open with the VOX website, etc)?
> 
> Makes sense. As I understand `hs.application.applicationsForBundleID` it returns a table with applications with the given BundleID, so the only check that is needed is for a non-empty table. I.e. there is no need to call `isRunning()` on an app from the table to confirm?

Change pushed: f8fced0

cmsj:
Thanks!

PR #2911, opened 2021-06-27
hs.mouse.scrollDirection getter method name fix

### LINKED ISSUE #2910: hs.mouse.scrollDirection() only ever returns "normal"
### Versions tested
- Hammerspoon 0.9.90 (6030)
- macOS 11.4

### Steps to reproduce
1. In the Hammerspoon Console, execute `hs.mouse.scrollDirection()`
2. Open the Mouse System Preferences panel (with attached mouse)
3. Toggle the "Scroll direction: Natural" checkbox
4. Do step 1 again

### Expected behaviour
When "Scroll direction: Natural" is checked, `hs.mouse.scrollDirection()` should return "natural", and "normal" when it is unchecked.

### Notes
I believe this is a regression as this used to work for me fantastically well, but I'm not certain whether its due to a change in Hammerspoon, macOS or both. 

As an aside, I'd like to take the opportunity to express my deep gratitude and admiration for the Hammerspoon developers and community. I believe my script that this issue affected is one of the finest pieces of software I've ever written, which consistently made me smile every time it ran, and that is down to you fine developers.

asmagill:
Fixes HSMouse class getter for scroll direction.

Will close #2910

### DIFF
--- a/extensions/mouse/internal.m
+++ b/extensions/mouse/internal.m
@@ -12,14 +12,14 @@ static void enum_callback(void *ctx, IOReturn res, void *sender, IOHIDDeviceRef 
     if (res != kIOReturnSuccess) {
         return;
     }
-    
+
     NSString *vendor = (__bridge NSString *)IOHIDDeviceGetProperty(device, CFSTR(kIOHIDManufacturerKey));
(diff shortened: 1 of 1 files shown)


cmsj:
wow, big shame on me for that one. thanks!

PR #2916, opened 2021-07-11
Minor Documentation Clarification.

nathanvercaemert:
Based on the maps above, I believe this is a typo in the documentation.


cmsj:
Good spot, thanks!

PR #2922, opened 2021-07-24
hs.razer - Adds support for Open Razer devices

### LINKED ISSUE #2921: hs.razer - Add support for Razer devices
I've discussed this before in other issues (#2447, #2876), but something I'd love to tackle sooner rather than later is add support for the Razer Tartarus V2 to Hammerspoon, offering a similar extension to `hs.streamdeck`.

**Hardware:**
- https://www.razer.com/gaming-keypads/Razer-Tartarus-V2/RZ07-02270100-R3U1

**Useful Code:**
- https://github.com/1kc/librazermacos

**Related:**
- https://github.com/1kc/razer-macos
- https://github.com/FruityRazer/FruityKit
- https://github.com/kprinssu/osx-razer-blade
- https://github.com/openrazer/openrazer (Linux Only)
- https://github.com/z3ntu/razer_test (Works on macOS)
- https://github.com/CalcProgrammer1/openrazer-win32 (Windows Only)

latenitefilms:
- This is very much a work-in-progress. I'm only submitting a draft pull request to get help and support, and to alert others that might be interested in this little side-project.
- The aim is support changing the LEDs of all devices supported by Open Razer.
- Adding button callbacks is a bit more tricky, so I'm currently only planning to support the Razer Tartarus V2.
- Will eventually close #2921


extensions/razer/internal.m
```diff
@@ -0,0 +1,870 @@
...
+        //IOHIDManagerOpen((__bridge IOHIDManagerRef)self.ioHIDManager, kIOHIDOptionsTypeSeizeDevice);
+        
+        CFRunLoopRun();
+    }
+    return self;
+}
+
+#pragma mark - IOKit C callbacks
+
+static void keyboardCallback(void* context, IOReturn result, void* sender, IOHIDValueRef value)
```
latenitefilms, line 88: @asmagill - If you get a chance, are you able to help me with some basic structural things? For example, how can I trigger the Lua callback from within this function (i.e. how can I access the `_callbackRef` from here)?

extensions/razer/driver/razerkbd_driver.c
```diff
@@ -0,0 +1,1576 @@
...
+ *
+ * Returns a string
+ */
+ssize_t razer_attr_read_get_firmware_version(IOUSBDeviceInterface **usb_dev, char *buf)
+{
+    struct razer_report report = razer_chroma_standard_get_firmware_version();
+    struct razer_report response_report = razer_send_payload(usb_dev, &report);
+    
+    // MODIFIED BY CHRIS:
+    //return sprintf(buf, "v%d.%d\n", response_report.arguments[0], response_report.arguments[1]);
```
latenitefilms, line 262: @asmagill - This code was crashing. Any suggestions on how I can "join" these two unsigned char's?

extensions/razer/internal.m
latenitefilms: Ignore this... worked it out, I wasn't passing the `context` correctly.

extensions/razer/driver/razerkbd_driver.c
latenitefilms: Ignore me... I implemented a (lazy) workaround.

extensions/razer/internal.m
```diff
@@ -0,0 +1,1658 @@
...
+                           CGEventType type,
+                           CGEventRef event,
+                           void* refcon) {
+    
+    // Prevent a crash when doing garbage collection:
+    if (type == kCGEventTapDisabledByUserInput) {
+        return event;
+    }
+        
+    HSRazer *manager = (__bridge HSRazer *)refcon;
```
cmsj, line 122: I think you should check if `manager` is `nil`. It's probably not something that would cause a crash further down, but best to be safe I think.

extensions/razer/internal.m
```diff
@@ -0,0 +1,1658 @@
...
+            [obj doGC];
+            
+            obj.callbackRef = [skin luaUnref:refTable ref:obj.callbackRef];
+
+            // Disconnect Callback:
+            if (obj.callbackToken != nil) {
+                //[obj.razer close];
+                obj.callbackToken = nil;
+            }
+            obj = nil;
```
cmsj, line 1570: I would think `obj.nil` would want to go after `obj.lsCanary = tmplsCanary;` a few lines below?

cmsj (commented):
So this is a fairly large PR!

I'm overall not sure if we should merge this at all, but I'll come back to that. In the mean time, there's at least one Sentry change in here, which shouldn't be.

Secondly, if we're only going to support one device, maybe we should not import ~150 json files for devices that aren't supporter.

Finally, and most importantly, I see this includes third party library code that's GPL'd. This would be the first inclusion of GPL code into Hammerspoon and I think that warrants a wider discussion. I like the GPL as a license, but it would be a fairly substantial change for the project to link GPL'd code.

latenitefilms:
> In the mean time, there's at least one Sentry change in here, which shouldn't be.

Sorry, I'm still on Mojave on my main machine, so there's a couple of tweaks to Hammerspoon so that I can still build. This Sentry tweak is one of them.

> Secondly, if we're only going to support one device, maybe we should not import ~150 json files for devices that aren't supported?

All devices are currently "supported" in that if you have them connected, you can create a `hs.razer` object for them. Currently all keyboards work - I need to add methods for other device types such as mice. But the plan is basically to support everything that Open Razer can currently do. Button callbacks are a bit more tricky, because I basically just use `hidutil` to remap their buttons to "dummy" buttons, so macOS doesn't treat them as normal keyboards, so then I can use IOKIT to detect the keypresses. Once I finish, it should be easy to support other devices by just adding some details to the device's JSON file, but without having physical access to the device, it's not the kind of information you can find online.

> Finally, and most importantly, I see this includes third party library code that's GPL'd. This would be the first inclusion of GPL code into Hammerspoon and I think that warrants a wider discussion. I like the GPL as a license, but it would be a fairly substantial change for the project to link GPL'd code.

Basically all the third party code comes from [librazermacos](https://github.com/1kc/librazermacos), which uses [GNU General Public License v2.0](https://github.com/1kc/librazermacos/blob/master/LICENSE). However this repo is basically just a "port" of [Open Razer](https://github.com/openrazer/openrazer). Could `hs.razer` just use the same license?

Of course, it's no stress at all if you don't want to merge this into Hammerspoon - I have no idea if any Hammerspoon users would actually use Razer devices, as it's normally gamers that use these products. The only reason I even started looking into it, is a few CommandPost users have the Razer Tartarus V2, and Razer dropped Apple support in Big Sur.

extensions/razer/internal.m
latenitefilms: Actually, I'd love your advice on this anyway. I feel like there's something I'm very much misunderstanding anyway. I did try checking if `manager` is `nil`, but it still crashed during garbage collection because manager was a `NSZombie`. I feel like I'm missing something about using a bridge?

latenitefilms:
I'm going to close this pull request, and investigate another way to do this without using Open Razer.

PR #2948, opened 2021-09-08
hs.razer - Add support for Razer Tartarus V2

### LINKED ISSUE #2935: hs.razer - Add support for Razer Tartarus V2
Goals of this issue:

- Add support for the Razer Tartarus V2, for both button and scroll wheel callbacks, but also controlling the LED backlights.
- Keep the extension modular, so if other people want to add support for new Razer devices later down the line, this should be possible. However, new devices added should have full support - i.e. buttons and scroll wheel support, not just LEDs. To achieve this, we'll base this extension off how [hs.streamdeck](http://www.hammerspoon.org/docs/hs.streamdeck.html) works.
- Any code used must be MIT license compatible.
- The extension should be well documented, and easy to navigate.

latenitefilms:
- Adds support for the Razer Tartarus V2. 
- I'd like to try adding support for the Razer Nostromo too at some point.
- Closes #2935


latenitefilms:
@cmsj & @asmagill - I'm not sure this is something you want to include in Hammerspoon anyway, but if either of you have time to do a quick idiot check of this, that would be awesome. No urgent rush or pressure. Thanks heaps!

PR #2983, opened 2021-11-10
Significant refactor of build settings/process

### LINKED ISSUE #2972: Build system revamp
This is likely to be of most interest to @latenitefilms

While working on adding Shortcuts support to Hammerspoon, I have needed to do local builds that are fully signed, but not notarized and without symbols uploaded to Sentry, but use the `-dev` entitlements. This has been..... challenging, and frustrating.

Right now we have a weird mix of Xcode build phase scripts, librelease.sh, the Makefile and github actions helper scripts.

I think I'm going to rip all of this apart and refactor it down to:

 * Remove the makefile because it prevents environment variables being preserved into xcodebuild (and thus into the build phase scripts)
 * Rework the build-phase scripts so they are more heavily influenced by environment variables, but by default will focus mostly on just making a usable build from Xcode
 * Try to move things out of build-phase scripts and into xcconfig files where possible (e.g. signing could do that)
 * Rework the release script into something that stops trying to guess what it should be doing and just adds command line arguments for the various steps it does
 * Separate `release.sh` into `build.sh` and `release.sh`
 * Rework the github actions scripts to fit with the above

Arguments I'm currently thinking need to be on `build.sh` are:
 * --scheme (Hammerspoon or Release) }
 * --configuration (Debug or Release)      }- might want to even revisit if these two are still worth having vs xcconfig files.
 * --sign (which requires the presence of a signing token file)
 * --entitlements (Hammerspoon or Hammerspoon-dev)
 * --docs (build only the docs)

cmsj:
Closes #2972

- Convert Hammerspoon and extensions to xcconfig files for build settings
- Fix release builds to produce gatekeeper-valid bundles
- Remove Sentry upload build phase
- Makefile is now entirely covered by build.sh
- Bump minimum macOS version to 10.15
- Update Cocoapods to target 10.15
- Update in-build JSON docs generation to build.sh
- Actually finally implement Python requirements.txt assertion
- Implement dependency installation in build.sh
- Switch build.sh building to use an archive and export from that archive. Add a validation operation to build.sh. Stop manually copying extension dylibs and instead place them properly in the Frameworks folder using a proper Copy Files build phase. This means the app is currently broken - there are no init.lua files and even if there were, they wouldn't be able to find their internal.so's
- Remove/disable a couple of Build Phase scripts that are going to become unnecessary
- Split out Lua and dylib files into more correct app bundle paths, with lots of updates to cope with that
- Remove current LuaSkin dependency because it's weird and broken and somehow depending on a Debug build
- Re-embed LuaSkin properly into all build targets. Hopefully properly at least
- Rename all internal.m/watcher.m files to more meaningful names
- Add support to upload dSYMs after building
- Add release note support to build.sh and use GNU rm for its ability to block requests to rm / in case we have made a mistake in build.sh
- Migrate hs build settings into an xcconfig
- Remove old copy_extensions_to_bundle script and its disabled build phase
- Update docs to reflect latest build changes
- Migrate rebuild.sh to using scripts/build.sh
- Remove unneeded sign_bundle.sh
- Speculatively re-organise GitHub Actions CI for build.sh changes
- Fix up build.sh's build command when it's not supposed to use an archive


cmsj:
I'm pretty much ready to merge this, if anyone has any further feedback

latenitefilms:
Merge away, and we'll see what explodes when I merge into CommandPost!

asmagill:
I have no objections either.

For personal reasons, because I still have sometimes use of a Catalina limited machine and a Big Sur limited machine, intend to see what changes may be necessary to build and/or run on these older machines, but it will be easier to determine what additions or conditional changes are required once the new "normal" system is in place.

My time is somewhat limited at the moment, but I do hope to have more time to devote to Hammerspoon in a couple of weeks, so don't expect any additions or changes from me in the near term.

PR #2999, opened 2021-11-22
Add hs.camera and support for hs.audiodevice in-use status/notifications

### LINKED ISSUE #1255: Feature Request: Webcam events
As I work from home and jump in and out of video conference meetings all day, I'd like to sense when my camera turns on and off and set Do Not Disturb and Away statuses on various services automatically while I'm in a meeting. (I don't think there's an API that makes that possible yet, although I may be missing it.)

cmsj:
- Add hs.camera and some basic functions/methods to go with it, including a watcher for cameras being added/removed
- hs.audiodevice:watcherCallback() is now called for changes in the device's 'in use' state
- Flesh out hs.camera watchers

Closes #1255


cmsj:
If anyone wants to play with this:

```lua
cameras = hs.camera.allCameras()
print("Discovered cameras: "..hs.inspect(cameras))

hs.camera.setWatcherCallback(function(camera, event)
  print("Camera Device Event: "..tostring(camera:name()).." was "..tostring(event))

  if (event == "Added") then
    camera:setPropertyWatcherCallback(function(camera, property, scope, element)
        print("Camera Property Event: "..tostring(camera:name())..": "..tostring(property))
    end)
    camera:startPropertyWatcher()
  end
end)

hs.camera.startWatcher()
```

PR #3002, opened 2021-11-23
Allow LuaSkin's UTF8 cleaner to be called directly with a C string, instead of only being available via Lua

cmsj:


cmsj:
@asmagill any objections to this? It's not a meaningful change at the moment, but I'm looking at all of the way the Console/hs work and thinking about how it could all be better/simpler, probably by moving away from the current `evalfn` route and essentially having LuaSkin grow methods that can replicate Lua's own REPL.

asmagill:
Yeah, looks good to me.

PR #3006, opened 2021-11-28
expose: fix issue in screen layout

freewizard:
it seems adding new element to `screens` while looping thru it may cause some items processed twice.


cmsj:
Good catch, thanks!

PR #3033, opened 2021-12-15
Log a callstack when checkGCCanary detects an issue. Closes #2957

### LINKED ISSUE #2957: LuaSkin has caught an attempt to operate on an object that has been garbage collected
@cmsj - Is there a way to add a bit more detail to this LuaSkin error message? I'm seeing it every now and again after reloading, and I'm not really sure which extension is causing it.

cmsj:


cmsj:
@latenitefilms I suppose the question here is whether `logWarn` is the right choice. Perhaps the focus should be on capturing this in Sentry? It ought to be possible to explicitly log this to Sentry at runtime without a crash having to happen first.

We could also choose to crash by throwing an exception, which would force Sentry to catch a full backtrace and avoid the possibility of weird memory corruption later in the session?

latenitefilms:
Maybe it's better to just crash, and then we can try solve all those individual issues?

cmsj:
I think I probably agree with that. if the canary has failed, HS is no longer in a consistent state, and that stack trace will hopefully tell us exactly which objects are getting into that weird state, so we can find/fix/avoid the weirdness.

I'll rework this to throw an exception. thanks!

cmsj:
@latenitefilms switched to throwing an Exception. Is this relatively easy for you to reproduce? If not I can manually try and trip the error here.

PR #3072, opened 2022-01-12
fix docstring of `window:moveToUnit`

KappaDistributive:
moveToUnit expects a unitrect which in turn expects values in the
interval [0.0, 1.0]. The previous docstring erroneously suggested to use
integer values in the interval [0, 100].

### DIFF
--- a/extensions/window/window.lua
+++ b/extensions/window/window.lua
@@ -896,7 +896,7 @@ end
 ---  * The `hs.window` object
 ---
 --- Notes:
----  * An example, which would make a window fill the top-left quarter of the screen: `win:moveToUnit'[0,0,50,50]'`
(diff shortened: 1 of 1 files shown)


KappaDistributive:
> This pull request does not contain a valid label. Please add one of the following labels: `pr-fix, pr-change, pr-feature, pr-maintenance`

I don't know how to add a label to an existing PR. The menu on the right hand side doesn't let me select/edit a label, I don't see any other menu to do this and an internet search didn't yield an answer either.

latenitefilms:
@KappaDistributive - I believe only @cmsj or @asmagill have the power to add labels.

cmsj:
Thanks!

Issue #3076, opened 2022-01-15
hs.streamdeck does not support StreamDeck MK.2


jcutler:
The StreamDeck extension doesn't appear to recognize the [Stream Deck MK.2](https://www.elgato.com/en/stream-deck-mk2) which has a Product ID of `0x0080`.

jcutler:
[](url)

cmsj:
I would definitely like to get it supported. It is a little bit tricky without having a Mk2 on hand, but if you'd be willing to try out some test builds, I can try and add the support.

latenitefilms:
I'll try get my hands on a MkII.

cmsj:
> I'll try get my hands on a MkII.

I would, but I already have an Original, Mini and XL on my desk 🤣

jcutler:
> I would definitely like to get it supported. It is a little bit tricky without having a Mk2 on hand, but if you'd be willing to try out some test builds, I can try and add the support.

I would definitely try out some test builds if it would help.

gineer01:
@cmsj : I'm curious how a new device like Mk2 can be added. Looking at the source code for hs.streamdeck, it seems to reverse-engineer the protocol to talk to the Stream Deck device. What tool do you use to view what Stream Deck device send and receive? Thanks!

latenitefilms:
It looks like most of the hard work has been done here:

https://github.com/OpenMacroBoard/StreamDeckSharp/pull/36/commits/1ba07177dfe57913a28d1dc2031e638dc39d34ee

jcutler:
@latenitefilms @cmsj: I put together a PR (seems identical to other gen2 StreamDecks) with a couple of functionality caveats I'm not sure are expected or not: #3089

PR #3089, opened 2022-01-24
Add support for Elgato StreamDeck Mk2

jcutler:
#3076

Added support for Elgato StreamDeck Mk2. It appears from everywhere I can find that the spec is the same as other gen2 devices (see node-elgato-stream-deck [definition](https://github.com/Julusian/node-elgato-stream-deck/blob/master/packages/core/src/models/original-mk2.ts#L16) and [base class](https://github.com/Julusian/node-elgato-stream-deck/blob/master/packages/core/src/models/base-gen2.ts#L15), or StreamDeckSharp [definition](https://github.com/OpenMacroBoard/StreamDeckSharp/blob/master/src/StreamDeckSharp/Internals/StreamDeckMK2HardwareInfo.cs#L7) and [base class](https://github.com/OpenMacroBoard/StreamDeckSharp/blob/master/src/StreamDeckSharp/Internals/StreamDeckJpgHardwareBase.cs)).

Was able to confirm all functionality, with the following exceptions (not sure whether it is expected):
* `serialNumber` returns last 9 of 12 digit serial, eg `2K4B07621` instead of `DL42K4B07621` printed on the physical device
* `firmwareVersion` returns `nil`

### DIFF
--- a/Hammerspoon.xcodeproj/project.pbxproj
+++ b/Hammerspoon.xcodeproj/project.pbxproj
@@ -538,6 +538,8 @@
 		949CDC001990759B00906CCE /* ConsoleWindow.xib in Resources */ = {isa = PBXBuildFile; fileRef = 949CDBFF1990759B00906CCE /* ConsoleWindow.xib */; };
 		94A1E5481993AC5C003AEB26 /* MJFileUtils.m in Sources */ = {isa = PBXBuildFile; fileRef = 94A1E5471993AC5C003AEB26 /* MJFileUtils.m */; };
 		B8524F32201A896000844F3D /* libhid.m in Sources */ = {isa = PBXBuildFile; fileRef = B8524F25201A875600844F3D /* libhid.m */; };
+		BE28C754279E92D300D890D6 /* HSStreamDeckDeviceMk2.h in Headers */ = {isa = PBXBuildFile; fileRef = BE28C752279E92D300D890D6 /* HSStreamDeckDeviceMk2.h */; };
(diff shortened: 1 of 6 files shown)


cmsj:
I don't have the serial number quite nailed down yet, but https://github.com/Hammerspoon/hammerspoon/pull/3090 at least fixes the firmware version on the XL, which I think is basically the same as the Mk 2. The current abstraction of per-model is making a lot less sense now they're converging on a common hardware platform. Up to this point each model was kind of unique in weird ways, but maybe we can unify onto a new code structure that keeps some of this stuff unified?

cmsj:
Aha, I now know why we're losing the first few characters of the serial number. This will indeed take some re-architecting to fix.

jcutler:
> Aha, I now know why we're losing the first few characters of the serial number. This will indeed take some re-architecting to fix.

I'll check if the same changes work for the MK2.

I'm curious if the same issue exists (and is fixed the same way) on the Mini and V2, which other projects seem to suggest have a similar base spec.

jcutler:
@cmsj not sure your plans for rearchitecting. I'd imagine you'd want to subsume this PR and #3090 into something that covers it all? I'd be happy to validate any of it on the Mk2.

cmsj:
Likely yes. I'm working on a more unified version at the moment. I can test Original, Mini and XL, so with your Mk2 that only leaves us missing the Original v2, which maybe @latenitefilms has?

cmsj:
Closing this since we covered it elsewhere. Thanks again!

PR #3105, opened 2022-02-04
fix typo

luckman212:
`ccess` => `access`

https://github.com/Hammerspoon/hammerspoon/blob/b03d628f792eb88188962643c32a490f1174c10d/extensions/drawing/color/drawing_color.lua#L5


cmsj:
Thanks!

PR #3112, opened 2022-02-11
libmouse.m: update internal mouse name

aviau:
According to `hs.mouse.names()`, this is the name of the internal trackpad on my laptop.

```
> hs.inspect(hs.mouse.names())
2022-02-10 19:37:27: -- Loading extension: inspect
{ "Apple Inc.::Apple Internal Keyboard / Trackpad" }
```

Haven't rebuild hammerspoon to confirm that it works, but the snippet above gave me the confidence to open a PR.

Note that if I knew objective C better I'd suggest that we check if one of the name has `Apple Internal` in it.

### DIFF
--- a/extensions/mouse/libmouse.m
+++ b/extensions/mouse/libmouse.m
@@ -84,7 +84,7 @@ static void enum_callback(void *ctx, IOReturn res, void *sender, IOHIDDeviceRef 
 }
 
 -(BOOL)hasInternalMouse {
-    return [self.names containsObject:@"Apple Internal Keyboard / Trackpad"];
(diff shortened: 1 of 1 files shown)


aviau:
>Please add one of the following labels

Erm. I would if I  could.

cmsj:
Thanks for spotting this - looks like M1 MacBooks have changed the name of internal devices. I've implemented this more generically in #3123 so I'll close this PR.

cmsj:
(Actually for completeness' sake, this wasn't an Apple change, I broke this by including the Vendor name with the Product name in `hs.mouse.names()` and didn't update the internalness test to account for that)

aviau:
Nice, thanks!

PR #3136, opened 2022-03-01
Fix two code analysis issues that Xcode identified

cmsj:


cmsj:
Note to self: the window image change here needs to be more thoroughly tested before this merges.

Issue #3110, opened 2022-02-10
Package Manager for Spoons


ryansch:
I've created a simple package manager for spoons based on https://github.com/wbthomason/packer.nvim.

I'm looking for feedback on it's design and usefulness: https://github.com/ryansch/Packer.spoon

cmsj:
Looks interesting, I wonder if it might be worth mentioning in the README what some of the practical differences are to https://www.hammerspoon.org/Spoons/SpoonInstall.html ?

ryansch:
That's great feedback, thanks!

seamusdemora:
Speaking of package managers: I'd like to try `hammerspoon`, but it seems the only package manager that carries it is homebrew. I use MacPorts, and I'm afraid there's no single application that could ever lure me back into the homebrew quagmire :)  So - question is: Will `hammerspoon` ever be available through MacPorts?

PR #3139, opened 2022-03-03
feat: extra lua runtime lookup paths

tami5:
Saw https://github.com/Hammerspoon/hammerspoon/issues/3110 and remembered how (maybe coming from nvim) tedious and difficult to maintain my dotfiles with Spoon directory being required to part of my dotfiles.

For me, I solved the issue through adding Spoon directory to `.gitignore`, for others, some use submodules (very few) and some have a redundant copy.

Then came loading lua modules from luarocks ecosystem, e.g. [fzy] native and lua modules. I had to modify `package.cpath` and `package.path` manually to where I installed fzy.

In conclusion, I want to mention, how I tried to keep using the whole Spoon convention, published few extensions [^1] [^2] [^3]. But later I gave up and used the simple `require` and stop (sadly) using those extensions I made, thus stopped maintaining them.

This pr solves the aforementioned issues. Add ` "~/.local/share/hammerspoon/site"` to lua path lookup, in addition to  `"~/.local/share/hammerspoon/site/lib"` to lua cpath lookup

I'd suggest introducing a breaking change through stop considering `Spoon` directory and start pointing users to install package in the userruntime instead, as well as kill the whole wrapper over lua modules through loadSpoon, for the sake empowering authors and consumer to have one way of instructing and loading plugins but ... I'm getting ahead of self. though, I believe is worth considering though.

[^1]: https://github.com/tami5/Yabai.spoon
[^2]: https://github.com/tami5/HsLib.spoon
[^3]: https://github.com/tami5/Binder.spoon

[fzy]: https://luarocks.org/modules/sethwarn/fzy

cc @cmsj


asmagill:
While I'm of two minds about removing the Spoon support altogether (I don't use it myself very often, but understand the desire for something simple for new users to start with) I do like the idea of moving them out of the config directory (I use the `.gitignore` approach myself). I would support migrating them to `~/.local/share/hammerspoon/site` or similar... that's what `.local` is for after all.

Hammerspoon/setup.lua
```diff
@@ -1,10 +1,31 @@
...
-print("-- package.cpath: "..package.cpath)
+local paths = {
+  configdir .. "/?.lua",
+  configdir .. "/?/init.lua",
+  configdir .. "/Spoons/?.spoon/init.lua",
+  package.path,
+  modpath .. "/?.lua",
+  modpath .. "/?/init.lua",
+  userruntime .. "/?.lua",
+  userruntime .. "/?/init.lua",
```
asmagill, line 13: Suggestion: add `userruntime .. "/Spoons/?.spoon/init.lua",` at end of list as first step in possibly migrating Spoon installation directory.

Hammerspoon/setup.lua
```diff
@@ -1,10 +1,31 @@
...
+}
+
+local cpaths = {
```
asmagill, line 16: Suggestion: add `userruntime .. "/?.dylib",` at beginning of list to facilitate external module development.

asmagill:
My bad, I meant for the second one to be `configdir .. "/?.dylib",` to match the other pairs.

@cmsj and @latenitefilms I'd like your thoughts on this:

But that does beg the question... should third-party external modules be going into `hs.configdir` or into this new location by default? I can easily fix my makefiles and instructions to reflect this if desired.

If we do want to shift third-party modules into `~/.local/share/hammerspoon/site` and leave `hs.configdir` solely for configuration/run-time stuff, then we need for all of the `~/.local/share/hammerspoon/site/*` paths to occur *before* `package.[c]path` in the lists so that these third-party modules can continue to offer developers a way to override the built-in modules during testing.

To play devil's advocate, putting everything into `hs.configdir` does allow for a user to have *all* of their local Hammerspoon changes in one place (config files, Spoons, and modules), so I guess all three *should* be supported; the questions then become search order and which do we promote as "preferred"?

Thoughts?

latenitefilms:
My opinion is that they way things are working in the current public Hammerspoon release is totally fine. I wouldn’t break the way things currently work.

asmagill:
Well, in any event, the `cpath` I *meant* to request for parity is as follows:
~~~lua
local cpaths = {
  configdir .. "/?.dylib",
  configdir .. "/?.so",
  package.cpath,
  frameworkspath .. "/?.dylib",
  userruntime .. "/lib/?.dylib",
  userruntime .. "/lib/?.so",
}
~~~

We can discuss best practices and recommendations and breaking changes if/when we get around to thinking about 2.0 again.

tami5:
> Well, in any event, the `cpath` I _meant_ to request for parity is as follows:
> 
> ```lua
> local cpaths = {
>   configdir .. "/?.dylib",
>   configdir .. "/?.so",
>   package.cpath,
>   frameworkspath .. "/?.dylib",
>   userruntime .. "/lib/?.dylib",
>   userruntime .. "/lib/?.so",
> }
> ```
> 
> We can discuss best practices and recommendations and breaking changes if/when we get around to thinking about 2.0 again.

Oh I see, so userrunitme can override builtin lib?

asmagill:
The first match is taken, and since you're concat-ing them in table order, the user's local config dir can override the built in libraries, but the `~/.local/share/hammerspoon/site/*` stuff can't.

I patterned this after your ordering of `package.path`. 

And I guess the argument for keeping this ordering is that the latest-and-greatest, work-in-progress stuff goes into your regular config directory while `~/.local/share/hammerspoon/site/*` is more for third party stuff (and maybe eventually Spoons).

tami5:
> The first match is taken, and since you're concat-ing them in table order, the user's local config dir can override the built in libraries, but the `~/.local/share/hammerspoon/site/*` stuff can't.
> 
> I patterned this after your ordering of `package.path`.
> 
> And I guess the argument for keeping this ordering is that the latest-and-greatest, work-in-progress stuff goes into your regular config directory while `~/.local/share/hammerspoon/site/*` is more for third party stuff (and maybe eventually Spoons).

Make sense third party shouldn't be allowed to override builtins modules, unless the user manually rearrange his package path.

asmagill:
Ok, @cmsj, I like this now and think it should be merged. Objections?

cmsj:
lgtm, thanks all!

PR #3149, opened 2022-03-11
[READY] Updates to hs.axuielement

asmagill:
Closes #2554 and refactors changes for new module organizational layout.


asmagill:
@cmsj Where are we on the CI tests? Should I dig into this as part of my changes right now, or is this still an (undesired, but) expected behavior?

cmsj:
The unreliable ones are still unreliable, so can be ignored. I've been trying to figure out crashes more than CI instability.

asmagill:
There are still some additions I plan to add to axuielement, and I still want to look closer at CommandPost because they have some helper functions that could probably be useful additions, but it may be a bit before I can get back to this module, so lets land what we have and I can create a new pull later.

PR #3150, opened 2022-03-11
[WIP] Fix build warnings from Xcode 13

cmsj:


cmsj:
@asmagill most of this is pretty mechanical, but I wonder if you could glance at the changes to hs.canvas and hs.webview?

This doesn't resolve 100% of the build warnings, but the remaining ones after this PR are pretty gnarly and I'm not sure some of them are even resolvable.

asmagill:
Looks reasonable to me. The two modules could really use a full overhaul -- I have some ideas on make `canvas` more flexible and allow for more content types (think controls for making custom dialogs, etc.) and it looks like enough changes have been added to the `WK*` classes that `webkit` should be reviewed if not rewritten as well, but those are longer term projects.

cmsj:
Thanks @asmagill . I'm going to merge this one as-is and we can revisit the remaining build warnings in the future.

PR #3151, opened 2022-03-11
Fix hs.fs.volume test

cmsj:


cmsj:
Hmm, I'm not sure what is going on here. This passes before and after on my x86 mac, and on my ARM macbook, but not on an ARM Mini.

latenitefilms:
I think the problem is that `info.path` is just a string:

```lua
info: {
  name = "renamedDisk",
  oldName = "ramDisk",
  oldPath = "file:///Volumes/ramDisk/",
  path = "file:///Volumes/renamedDisk%205/"
}
```

...so it fails on:

```lua
newPath = info.path and info.path.filePath:match("(/Volumes/"..ramdiskRename..")/?$")
```

latenitefilms:
@cmsj - Random idea... does a specific extension need to load first, because you can use `NSURL's` in Lua-land?

PR #3152, opened 2022-03-12
[READY] Add support for loading dylib libraries from user config directory

asmagill:
This will allow external modules to be built and named in the same manner as they would be if/when they are migrated into the core application.

In addition, I've removed an extra `docs.json` registration in the `hs.bonjour` module (I have left the similar redundancy in `hs.axuielement` because it will be addressed when #3149 lands and I don't want to create a potential merge conflict).


cmsj:
seems like a good idea, although we might want to reformat this on top of #3139 since I think that one makes sense too - what do you think?

asmagill:
Ok, suggestions made in #3139 and I'll move the `bonjour` cleanup into #3149 since it's a one line change that is more in the name of cleaning out cruft than something that actually changes anything.

PR #3159, opened 2022-03-18
Annotate to the compiler that we know we are setting variables without using them, in specific circumstances

cmsj:


latenitefilms:
@cmsj - this might be relevant?

https://github.com/actions/virtual-environments/issues/3649#issuecomment-1068085030

cmsj:
Thanks to some help from the AppKit Abusers Slack I learned of the existence of `#if __has_warning()` so now this works on both Xcode 13.3 and earlier :)

PR #3162, opened 2022-03-18
Track Lua threads (ie coroutines) so LuaSkin is more aware of their lifecycle

cmsj:


asmagill:
I'm not sure we need this... outside of trying to address #3165 and similar, what use cases do you envision?

Looking at the source for `coroutine.status` in `lcorolib.c` in the lua source code, there looks to be a relatively simple way to check a thread's status if we do find that we have an arbitrary thread/state but I'm not sure why we would if we're coding properly...

(*Note: If you do decide to close this instead of merge, you appear to have made some unrelated updates to `noises` and `razer` that you might want to resubmit.*)

latenitefilms:
AFAIK - My understanding that this could be closed instead of merged (in favour of #3165), and the changes to noises and razer were to fix a bug that prevented Hammerspoon from building on the latest Xcode, which has now been fixed in #3159.

asmagill:
Just want to make sure @cmsj didn't have another use case in mind as well, but I agree.

cmsj:
I'm a little sad to throw the code away because this was a fun little exercise, but yeah it's obsoleted by #3165

Issue #3158, opened 2022-03-18
hs.http.asyncPost coroutine crashes Hammerspoon if garbage collected too soon


gineer01:
## Crash Info

hs.http.asyncPost works fine with coroutine most of the time, but if the network is slow (the response time is more than a few seconds), it crashes the Hammerspoon. 

The crash:
```
Process:               Hammerspoon [1770]
Path:                  /Applications/Hammerspoon.app/Contents/MacOS/Hammerspoon
Identifier:            org.hammerspoon.Hammerspoon
Version:               0.9.94 (6196)
Code Type:             X86-64 (Native)
Parent Process:        ??? [1]
Responsible:           Hammerspoon [1770]
User ID:               501

Date/Time:             2022-03-17 23:24:09.540 -0700
OS Version:            macOS 11.6.4 (20G417)
Report Version:        12
Bridge OS Version:     6.2 (19P744)
Anonymous UUID:        B53A6AB1-AEB1-4E7D-EA92-AC58D04D06A0

Sleep/Wake UUID:       EEAFD038-5151-4E15-92B3-CAF6F09C45DB

Time Awake Since Boot: 180000 seconds
Time Since Wake:       2500 seconds

System Integrity Protection: enabled

Crashed Thread:        0  Dispatch queue: com.apple.main-thread

Exception Type:        EXC_CRASH (SIGABRT)
Exception Codes:       0x0000000000000000, 0x0000000000000000
Exception Note:        EXC_CORPSE_NOTIFY

Application Specific Information:
abort() called

Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   libsystem_kernel.dylib            0x00007fff2040b92e __pthread_kill + 10
1   libsystem_pthread.dylib           0x00007fff2043a5bd pthread_kill + 263
2   libsystem_c.dylib                 0x00007fff2038f406 abort + 125
3   org.hammerspoon.Hammerspoon       0x000000010e8a13d0 -[SentryCrashExceptionApplication _crashOnException:] + 58
4   com.apple.AppKit                  0x00007fff230e4c07 -[NSApplication reportException:] + 732
5   org.hammerspoon.Hammerspoon       0x000000010e8a1369 -[SentryCrashExceptionApplication reportException:] + 371
6   com.apple.AppKit                  0x00007fff23198920 uncaughtErrorProc + 145
7   com.apple.CoreFoundation          0x00007fff206289d2 __handleUncaughtException + 716
[...]

Issue #3011, opened 2021-12-03
hs.http.asyncPost failure in coroutine crashes Hammerspoon


gineer01:
## Crash Info

hs.http.asyncPost works fine with coroutine most of the time, but if there's a failure (e.g., no network connection), it crashes the Hammerspoon. The crash:
```
4   org.hammerspoon.LuaSkin           0x000000010509f93b index2stack.cold.2 + 33
5   org.hammerspoon.LuaSkin           0x0000000105089ba3 index2stack + 90
6   org.hammerspoon.LuaSkin           0x0000000105089bc9 lua_rotate + 33
7   org.hammerspoon.LuaSkin           0x0000000105066cb0 -[LuaSkin protectedCallAndTraceback:nresults:] + 194
8   org.hammerspoon.LuaSkin           0x0000000105066d90 -[LuaSkin protectedCallAndError:nargs:nresults:] + 55
9   internal.so                       0x00000001098a7c4c -[connectionDelegate connection:didFailWithError:] + 423
```

## Steps to reproduce
To reproduce the problem, disconnect the Internet connection (e.g., turn off wifi) before running these codes:

Running hs.http.asyncPost without coroutine works fine:
> hs.http.asyncPost("https://google.com" , nil , nil, function(status,body) print(status,body) end )
2021-12-03 14:00:29: -1	Connection failed: The Internet connection appears to be offline. - https://google.com/

Running hs.http.asyncPost within coroutine crashes Hammerspoon:
> coroutine.wrap(hs.http.asyncPost)("https://google.com" , nil , nil, function(status,body) print(status,body) end )
--- Hammerspoon crashes

## Potential cause
I've investigated a bit and I think the bug is using an improper lua_State* (using NULL) at this line:
https://github.com/Hammerspoon/hammerspoon/blob/master/extensions/http/libhttp.m#L88

I'm guessing that `connectionDelegate` will need to be refactored so that `http_doAsyncRequest` can pass its lua_State* param to the `connectionDelegate` at this line.
https://github.com/Hammerspoon/hammerspoon/blob/master/extensions/http/libhttp.m#L207

gineer01:
@cmsj @asmagill : this bug is similar to #2826

cmsj:
Yeah that does seem pretty likely, and this pattern of "C callback passes NULL to `sharedWithState` pattern is *very* prevalent throughout Hammerspoon, unfortunately.
[...]

PR #3165, opened 2022-03-20
Fix hs.http callback crashes with Lua coroutines

cmsj:


asmagill:
Yeah, I think this is what is needed... Let me run some tests but I suspect the fix I was going to propose for `http` in https://github.com/Hammerspoon/hammerspoon/issues/3158#issuecomment-1073437596 is going to look a lot like this.

asmagill:
Ok, reviewing `libhttp.m` I don't see any other changes that need to be made, and verified thison the examples from both #3011 and #3158 and a few others of my own. This pull request is good to go.

And this is the correct approach to use; we should *never* be saving a lua_State for a callback anymore -- that's why coroutine's wouldn't work with earlier versions of Hammerspoon.

cmsj:
Thanks @asmagill

PR #3166, opened 2022-03-21
[REVIEW] Remove references to stored lua_State in objects and callbacks

asmagill:
Addresses `hs.noises`, `hs.screen.watcher`, and `hs.menubar`.

With the exception of noises, theses were just left-over crumbs that weren't actually being used by anything, just wasting a few bytes of memory.

WIP because I still need to review for the few modules that still use structures and for Carbon callbacks with `context` which may include bridged data for LuaSkin or lua_State. Hope to complete in the next day or so.


asmagill:
Ok, I searched for `_lua_stackguard_entry` as suggested by @cmsj, I checked every `void\s+\*` that is used as a context or information passing argument to the Carbon callbacks, and I looked for every `lua_State\s+\*` that occurred in a struct definition or class interface...

I *think* we're good... mostly I just found some left over stuff that we weren't actually using anymore, so it appears that `hs.http` and `hs.noises` were the outliers that didn't get properly updated.

So unless anyone can think of additional searches I should perform, I think this is ready for review and merging.

asmagill:
Oh, and I removed the `drawing.h` file that was still attached to `hs.drawing.color` since it wasn't actually being included... this means that `project.pbxproj` did get updated... so factor that into how/when you want to merge it.

cmsj:
Thanks for checking into this!

PR #3167, opened 2022-03-22
Add C->Lua safety checks to hs.canvas fadeOut:andDelete:withState

cmsj:
Should fix Sentry crash HAMMERSPOON-BX


cmsj:
@asmagill any objections?

PR #3177, opened 2022-04-02
Fixed crash in hs.image.getExifFromPath()

### LINKED ISSUE #3176: Crash - hs.image.getExifFromPath()
Sentry: https://sentry.io/share/issue/1ff2aa3ca8f1402c8458132bea657855/

How to make it crash:

```lua
hs.image.getExifFromPath("")
```

It looks like it's crashing [here](https://github.com/Hammerspoon/hammerspoon/blob/6eda506cd55395981376ae82fd2886bd64f0d315/extensions/image/libimage.m#L787):

```objc
CFRelease(imageSource);
```

Need to add a check to make sure `imageFileURL` returns a valid result.

latenitefilms:
- Closes #3176

### DIFF
--- a/Hammerspoon Tests/HSimage.m
+++ b/Hammerspoon Tests/HSimage.m
@@ -1,0 +1,23 @@
+#import "HSTestCase.h"
+
+@interface HSimage : HSTestCase
+
+@end
+
+@implementation HSimage
+
+- (void)setUp {
+  [super setUpWithRequire:@"test_image"];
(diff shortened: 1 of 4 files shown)


cmsj:
lgtm, although I now wonder what happens if you pass it a URL to a file that either doesn't exist, or isn't an image?

latenitefilms:
I tested passing in random text and it just returned nil, didn't try a valid path to a non-image, but I THINK if should return nil too - will test.

PR #3195, opened 2022-04-22
Added ability to save a hs.menubar position between reloads

### LINKED ISSUE #2878: [Request] Ability to have hs.menubar items retain their same position upon startup/hs.reload (or at least next to the "main" Hammerspoon menubar item) rather than always reverting back to the left-most item
Not sure this is even possible given the way Hammerspoon utilizes the APIs for hs.menubars/how the system behaves but just thought I'd raise the question. This is something that would benefit me and my personal workflow quite significantly as I utilize multiple hs.menubar items heavily and have a lot of non-HS items which means they aren't all always visible and so every time I reload, I have to move them all back to an always-visible section of menubar one by one.

Steps to reproduce:

1. Create and show a new hs.menubar item
2. Said menubar item will be placed at the end (left-most side) of all the existing menubar items
3. Cmd+click/drag the new item to a new position in the menubar
4. Perform hs.reload() or restart Hammerspoon enitrely
5. The hs.menubar item is placed back at the end of existing menubar items (unlike the "main" Hammerspoon menubar which, if you restarted Hammerspoon enitrely, will return to the position it occupied before the restart)

latenitefilms:
- Added an optional `autosaveName` parameter to `hs.menubar.new()`, which allows you to set a unique identifier for the menubar, allowing it to save and restore it's menubar position across reloads.
- Added a method to get and set the `autosaveName` -`hs.menubar:autosaveName()`.
- Removed `hs.menubar.newWithPriority()` as it no longer works on macOS Catalina (and the minimum system requirements for Hammerspoon are now Catalina & above).
- Closes #2878

### DIFF
--- a/extensions/menubar/libmenubar.m
+++ b/extensions/menubar/libmenubar.m
@@ -465,8 +465,7 @@ static int menubarNew(lua_State *L) {
 
 /// hs.menubar:autosaveName([name]) -> menubaritem | current-value
 /// Method
-/// Get or set the autosave name of the menubar. By defining an autosave name, macOS can restore
-/// the menubar position after reloads.
+/// Get or set the autosave name of the menubar. By defining an autosave name, macOS can restore the menubar position after reloads.
 ///
 /// Parameters:
 ///  * name - An optional string if you want to set the autosave name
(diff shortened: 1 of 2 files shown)


latenitefilms:
@cmsj & @asmagill - This should now be good to go unless you have any objections.

You can test by entering the following in the Console:

```lua
a = hs.menubar.new(true, "A"):setTitle("A")
b = hs.menubar.new(true, "B"):setTitle("B")
c = hs.menubar.new(true):autosaveName("C"):setTitle("C")
z = hs.menubar.new(true):setTitle("Z") -- this won't save position
```

...then change the menubar icon positions, restart/reload Hammerspon, run the code again, and the menubar items should be in the same position.

kusazh:
Hi, when will this be added to release?

PR #3196, opened 2022-04-23
[READY] Fix hs.spaces so it works when hs.spaces.screensHaveSeparateSpaces() is false

asmagill:
This still needs to be tested when a second monitor is attached and then I need to run through my tests again switching "Screens Have Separate Spaces" back on to make sure I didn't break anything.

Should be able to finalize tonight/tomorrow.


asmagill:
I hate CI

As far as I can tell the errors are unrelated, so this is ready to go unless anyone has objections.

latenitefilms:
Yeah, it looks like `hs.serial` is now that GitHub Actions is running on Monterey hardware - doh!

latenitefilms:
I've pushed to ignore these two tests in #3198

PR #3232, opened 2022-06-12
Allow the query to still be forwarded when no matching string exists

halfwit:
This is an example of how to implement this, it works in my local testing.
It adds a method, enableDefaultForQuery which can be passed a bool to allow the query to be returned to the final handler, even when it doesn't match an item on the choices list.

### DIFF
--- a/extensions/chooser/HSChooser.m
+++ b/extensions/chooser/HSChooser.m
@@ -396,7 +396,7 @@
         }
 
         _lua_stackguard_exit(skin.L);
-    } else if (self.enableDefaultForQuery != NO) {
+    } else if (self.enableDefaultForQuery != NO && self.completionCallbackRef != LUA_NOREF && self.completionCallbackRef != LUA_REFNIL) {
         // No row remaining in choices, return just query
         self.hasChosen = YES;
         LuaSkin *skin = [LuaSkin sharedWithState:NULL];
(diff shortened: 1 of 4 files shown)


extensions/chooser/HSChooser.m
```diff
@@ -393,7 +394,19 @@ - (void)tableView:(NSTableView *)tableView didClickedRow:(NSInteger)row {
             [skin pushNSObject:choice];
             [skin protectedCallAndError:@"hs.chooser:completionCallback" nargs:1 nresults:0];
         }
-        
+
+        _lua_stackguard_exit(skin.L);
+    } else if (self.enableDefaultForQuery != NO) {
```
cmsj, line 399: This should also check that `self.completionCallbackRef` is neither `LUA_NOREF` or `LUA_REFNIL`

cmsj (changes requested):
I like the look of this, and I also wonder if it might be worth adding the query string to the `choice` dictionary even when a valid choice exists? That doesn't need to be part of this PR, but we should do the `LUA_NOREF`/`LUA_REFNIL` checks before merging.

halfwit:
Sorry about the commit/revert, not sure what VSCode was doing there.

cmsj:
Thanks! Apologies for the slow merging

PR #3236, opened 2022-06-18
add DemoPro to SKIP_APPS_TRANSIENT_WINDOWS

kjslag:
add DemoPro to SKIP_APPS_TRANSIENT_WINDOWS since that app (which is on the app store) doesn't have windows


cmsj:
lgtm, thanks!

PR #3237, opened 2022-06-18
speedup window switching timers

kjslag:
Window switching with hs.window.switcher is noticeably slower than the default cmd-tab. This small patch decreases various timers to make the speed feel the same. Note: I decreased a timer that's supposed to prevent an el cap bug, but I couldn't test if this reintroduces the el cap issue.


extensions/window/window_switcher.lua
```diff
@@ -224,11 +224,11 @@ local function exit(self)
   --  if windows[selected]:application():bundleID()~='com.apple.finder' then
   --    windows[selected]:focus()
   --  else
-  timer.doAfter(0.15,function()windows[selected]:focus()end) -- el cap bugs out (desktop "floats" on top) if done directly
+  timer.doAfter(0.01,function()windows[selected]:focus()end) -- el cap bugs out (desktop "floats" on top) if done directly
```
cmsj, line 227: FWIW El Capitan is no longer supported by Hammerspoon, so we could see what happens if we remove the `timer.doAfter()` wrapper and just directly call `windows[selected]:focus()`

cmsj (changes requested):

extensions/window/window_switcher.lua
kjslag: I just tested. Directly calling it seems to work well for me. Should I change it?

cmsj: Apologies for not replying - yeah I think we should change it.

kjslag: Ok, I just committed the change

PR #3238, opened 2022-06-18
fix bug where window switcher does not go away when caps lock is on

kjslag:
When caps lock is on, the hs.window.switcher does not disappear. This small patch fixes that issue.


cmsj:
lgtm, thanks!

PR #3270, opened 2022-08-10
Fix hs.logger.* signatures in docs

MunifTanjim:
Fix misleading `:method` signature for `hs.logger` module. All of them are static functions.


cmsj:
Thanks!

PR #3298, opened 2022-10-13
feat(spaces): add force arg for moveWindowToSpace

okuuva:
Floating windows can be drawn over full screen windows so there should be a way to move them into full screen spaces as well.

Tested that this works with a floating window (kitty to be precise). Also tested moving a fullscreen window to a fullscreen (and normal) space and confirmed that it does not work. It does try to do it and return `true` but the source space just goes black and the window doesn't appear at the destination space (which is expected but wanted to check what happens). My extremely limited (read: none) knowledge in Obj C can get only so far, if we want to prevent even trying to move fullscreen windows even with `force = true` I need some help with that. I think it would be better to check if the window is in fullscreen rather than checking the source space type but don't know how to do that from within `spaces` extension or if it's even possible.

### DIFF
--- a/extensions/spaces/libspaces.m
+++ b/extensions/spaces/libspaces.m
@@ -131,26 +131,28 @@ static int spaces_windowsForSpace(lua_State *L) { // NOTE: wrapped in init.lua
     return 1 ;
 }
 
-/// hs.spaces.moveWindowToSpace(window, spaceID) -> true | nil, error
+/// hs.spaces.moveWindowToSpace(window, spaceID[, force]) -> true | nil, error
 /// Function
 /// Moves the window with the specified windowID to the space specified by spaceID.
 ///
 /// Parameters:
 ///  * `window`  - an integer specifying the ID of the window, or an `hs.window` object
 ///  * `spaceID` - an integer specifying the ID of the space
+///  * `force` - an optional boolean specifying whether the window should be tried to move even if the spaces aren't compatible
 ///
 /// Returns:
 ///  * true if the window was moved; otherwise nil and an error message.
 ///
 /// Notes:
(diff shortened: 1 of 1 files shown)


extensions/spaces/libspaces.m
```diff
@@ -131,26 +131,28 @@ static int spaces_windowsForSpace(lua_State *L) { // NOTE: wrapped in init.lua
     return 1 ;
 }
 
-/// hs.spaces.moveWindowToSpace(window, spaceID) -> true | nil, error
+/// hs.spaces.moveWindowToSpace(window, spaceID, force) -> true | nil, error
```
cmsj, line 134: because `force` is optional, the syntax here should really be:

```/// hs.spaces.moveWindowToSpace(window, spaceID[, force])```

File: extensions/spaces/libspaces.m [hunk 2]
```diff
@@ -131,26 +131,28 @@ static int spaces_windowsForSpace(lua_State *L) { // NOTE: wrapped in init.lua
...
 /// Function
 /// Moves the window with the specified windowID to the space specified by spaceID.
 ///
 /// Parameters:
 ///  * `window`  - an integer specifying the ID of the window, or an `hs.window` object
 ///  * `spaceID` - an integer specifying the ID of the space
+///  * `force` - a boolean specifying whether the window should be tried to move even if the spaces aren't compatible
```
cmsj, line 141: This should really say `an optional boolean`

cmsj (changes requested):

extensions/spaces/libspaces.m
okuuva: Fixed.

extensions/spaces/libspaces.m
okuuva: Fixed.

cmsj:
Thanks!

PR #3314, opened 2022-11-03
Added hs.serial:rts() and hs.serial:dtr()

### LINKED ISSUE #3310: hs.serial does not appear to receive serial data correctly
I am working on a custom Arduino-based macropad that takes status from Hammerspoon to update indicators and commands from the pad to update apps via Hammerspoon integrations and spoons. I am able to send serial data to the device but no messages are passed back to the callback. My setup is as follows:

**The setup:**
On the hardware side, I have a 32u4-based development board (Adafruit Itsy-Bitsy, 5V) with a test sketch that does a few things: reads serial and looks for "set" if that is followed by "on", it blinks the LED fast, otherwise it blinks it slowly. On switching, it prints back whether it was set fast or slow. It also has a button input that prints "button pressed".

On the Hammerspoon side, I have a simple setup that creates the serial port (hard coded path for testing), attaches a callback, and opens the port. It then attaches an app handler to look for when Hammerspoon is brought to the front to test serial recept bt the hardware. Lastly, it also has a timeout to close the port after 20 seconds so I don't need to quit Hammerspoon while testing in order to re-flash and use the other serial monitors.

**What happens:**
When I bring Hammerspoon console to the front, the LED blinks fast, when I switch to another app, it blinks slowly, indicating the remote hardware is receiving serial successfully. No messages are logged back or when I press the button.

**What I expect to happen:**
Whenever the apps switch, I expect to see a serial message logged. Whenever the button is pressed, I expect to see a serial message logged.

**Control:**
The serial messages work bidirectionally in all other serial consoles I have used.

Here is the complete Hammerspoon code:
```lua
local obj={}
obj.__index = obj

function serialWatcher(serialPortObject, callbackType, message, hexadecimalString)
  print("------------------------- serial event: " .. callbackType)
  if callbackType == "received" then
    print("message: " .. message)
  end
end

function sendToSerial(message)
  print("sending to serial:  " .. message)
  if obj.serial == nil or not obj.serial:isOpen() then
      print("not connected")
      return
  end
  obj.serial:sendData(message .. "\n")
end

function connectToUSB()
  print("----------------------------- connecting to USB")
  if obj.serial ~= nil and obj.serial:isOpen() then
      print("already connected")
      return
  end

  local path = "/dev/cu.usbmodem101"
  obj.serial = hs.serial.newFromPath(path)
  obj.serial:baudRate(115200)
  obj.serial:callback(serialWatcher)
  obj.serial:usesDTRDSRFlowControl(1) -- the docs say this should be boolean but only an int works, not sure if this is working
  obj.serial:usesRTSCTSFlowControl(1) -- the docs say this should be boolean but only an int works, not sure if this is working
  obj.serial:open()
end

function applicationWatcher(appName, eventType, appObject)
  if appName ~= nil and eventType == hs.application.watcher.activated then
      -- the following sends "set on" when the Hammerspoon console is brought forward
      sendToSerial("set " .. (appName == "Hammerspoon" and "on" or appName))
  end
end
obj.appWatcher = hs.application.watcher.new(applicationWatcher)
obj.appWatcher:start()

connectToUSB()

hs.timer.doAfter(20, function()
  print("------------------------ Closing serial")
  obj.serial:close()
end)

return obj
```

latenitefilms:
- Added methods to get and set the `RTS` and `DTR` values.
- Fixed some inconsistencies in the documentation.
- Fixed some typos where `LS_TNUMBER` was incorrectly used instead of `LS_TBOOLEAN`.
- Closes #3310

### DIFF
--- a/extensions/serial/libserial.m
+++ b/extensions/serial/libserial.m
@@ -106,6 +106,8 @@ static LSRefTable refTable = LUA_NOREF;
 @property BOOL                      usesDTRDSRFlowControl;
 @property BOOL                      usesDCDOutputFlowControl;
 @property BOOL                      allowsNonStandardBaudRates;
+@property BOOL                      rts;
+@property BOOL                      dtr;
 
 @end
 
@@ -134,6 +136,8 @@ static LSRefTable refTable = LUA_NOREF;
         _usesDTRDSRFlowControl      = NO;
         _usesDCDOutputFlowControl   = NO;
         _allowsNonStandardBaudRates = NO;
+        _rts                        = NO;
+        _dtr                        = NO;
     }
     return self;
 }
@@ -230,7 +234,7 @@ static LSRefTable refTable = LUA_NOREF;
 
 - (void)serialPort:(ORSSerialPort *)serialPort didReceivePacket:(NSData *)packetData matchingDescriptor:(ORSSerialPacketDescriptor *)descriptor
 {
-    // TODO: Impliment `ORSSerialPacketDescriptor` functionality
+    // TODO: Implement `ORSSerialPacketDescriptor` functionality
 }
 
 - (void)serialPortWasRemovedFromSystem:(ORSSerialPort *)serialPort;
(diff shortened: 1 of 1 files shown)


cmsj:
Thanks!

sethvoltz:
Ditto!

sethvoltz:
@cmsj what's the process to get a new release out? I see the latest release is March and I see there are a bunch of recent commits for GH tooling, are those the barrier to pushing up a new version?

Issue #3219, opened 2022-05-24
hs.styledtext and getString() for high plane Unicode characters


darth-cheney:
Hello,

I am trying to create a chooser that will help me select between different cuneiform unicode characters. The setup and display of these characters -- which are high up in the unicode plane, above 0x12000 -- works fine: the chooser displays and filters properly and there are no issues. However, when I "select" a choice, I am unable to extract the plain unicode string out of the choice itself. 
  
I have confirmed that I can get a hs.styledtext object back. And in the hammerspoon console it even prints this object with the correct character. But myText:getString() returns `nil` every time.
  
Here is an example:
```javascript
myString = "𒍠" // a unicode cuneiform character (might need fonts to display)
styled = hs.styledtext(myString)
print(styled:getString())
```

latenitefilms:
I'm not sure why that's not working, but as a potential workaround, could you just use a lookup table?

For example:

```lua
lookupTable = {
["item1"] = hs.styledtext.new("item1")
["item2"] = hs.styledtext.new("item2")
}
```

...and use the key in the `lookupTable` to return your actual styled text result?

asmagill:
After adding some logging into styledtext and trying your sample code, the problem appears to be because of encoding differences between UTF8 and UTF16...

As hex bytes, `myString`:

~~~lua
> hs.utf8.hexDump(myString)
00 : F0 92 8D A0                                      : ....
~~~

That entire sequence of 4 bytes is considered a *single* UTF8 character based solely on byte values (see lua's `utf8.charpattern`) -- I don't know yet if it's actually a valid single character according to the spec -- I'll have to research that.

However, macOS does everything in UTF16, and it considers the sequence as 2 UTF16 characters. There is a conversion function which attempts to map between the multi-byte-sized UTF8 sequences and the fixed-byte-sized of UTF16.

This causes the code which pulls the actual string out of the styled text object to fail because the range specified breaks a composite character.

The fix will require digging into the specs for UTF8 and UTF16 a bit deeper (ugh) so I don't have an eta, but I'll add it to my list and see what I can come up with.

SamuelSwartzberg:
[...]

PR #3356, opened 2023-01-13
Fix lua-objc mapping to handle high/low surrogate pairs in hs.styledtext

asmagill:
Addresses issues in #3219 

Completely rewrites lua-objc index mapping to properly handle UTF16 High/Low surrogate pairs

UTF8 (which Lua uses) can have Unicode characters of 1-4 bytes
UTF16 (which Objective-C/Swift use) are always 2 bytes, but for Unicode characters above U+10000 inclusive, 2 UTF16 characters (i.e. 4 bytes) are used.

This code assumes that no 1-byte UTF8 character can be a surrogate member... as far as I can tell this is a safe assumption, but I haven't found an explicit statement to that effect, so if anyone knows otherwise, please chime in.

### DIFF
--- a/extensions/styledtext/libstyledtext.m
+++ b/extensions/styledtext/libstyledtext.m
@@ -14,52 +14,48 @@ static LSRefTable refTable;
 // Lua string and the values are the corresponding character positions in the NSString.
 NSDictionary *luaByteToObjCharMap(NSString *theString) {
     NSMutableDictionary *luaByteToObjChar = [[NSMutableDictionary alloc] init];
-    NSData *rawString                     = [theString dataUsingEncoding:NSUTF8StringEncoding];
 
-    if (rawString) {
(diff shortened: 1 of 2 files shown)


cmsj:
@asmagill I'm not going to pretend to understand all the unicode stuff - are you happy for this to merge?

asmagill:
I left in a debugging function that I originally had commented out... it's undocumented, but might be helpful if someone does find a specific case where this fails, so I'm inclined to leave it as is. If you don't mind the undocumented function being left in, then yes, I think this is ready to be merged -- it hasn't broken anything on my setup and passed every test I've thought to come up with at it...

Though to be fair, I don't think a set of comprehensive unicode tests that covers absolutely all use cases exists... sometimes I feel like they make it up as they go! 😁

PR #3363, opened 2023-01-25
hs.fnutils.reduce takes an optional initial value

knu:
This is in line with the reduce function in other languages like JavaScript, Python, and Ruby.


cmsj:
Thanks!

PR #3371, opened 2023-01-31
Spaces now finds Dock instead of Docker

jmyounker:
`getDockElement()` didn't work when Docker was running.


jmyounker:
> Thanks!

You're very welcome!

PR #3398, opened 2023-02-28
Adjusted how we add text to the Console

### LINKED ISSUE #3389: Logging causes Hammerspoon slowdown over time
Co-maintainer of https://github.com/agzam/spacehammer

We've had reports over the past year or so about things like hs.alert opening very slowly and the hotkey modals taking a while to start and stop before switching to the next one.

A day or so ago accidentally discovered that issue was tied to logging upstream in Hammerspoon. The more that is logged over time, the slower hammerspoon becomes.

## Reproduction:

1. Create a binding that opens hs.alert and then create another binding that closes hs.alert
2. Fill up the log. Sharing my fennel test code below, but would recommend doing it asynchronously in a timer.
3. Before the logging starts, open and close the alert. Should happen nearly instantly.
4. After logging starts for a while, open and close the alert. Depending on how much info was logged, the UI will open and close a second or more after pressing the binding.

### Expected:

1. Ideally no matter how much info is logged, wouldn't expect a runtime performance decrease

### Actual

The more info that is logged, the slower Hammerspoon performs all around

Here's me opening the alert right after clearing the log:

Notice the time between pressing "m" and when my media modal menu opens.

- Pressing the same binding after artificially stuffing the log, the time between when "m" is pressed vs when the media modal opening has noticeably slowed down to about a second
- Clearing the hammerspoon log, then invoking the binding it's fast again.

My test code, specifics are not important here it just logs more data each iteration at a slower rate so the system is responsive. Use whatever you want to test.

```fennel
(local state
  {:loop-count 0
   :delay      10
   :entries    []})

(fn start!
  []
  (let [tick (fn [f]
               (hs.timer.doAfter
                 (/ state.delay 1000)
                (fn []
                  (print "debug-modal-delay"
                         (hs.inspect state))
                  (table.insert state.entries (.. "" (os.time)))
                  (set state.loop-count (+ state.loop-count 1))
                  (when (= (% state.loop-count 100) 0)
                    (set state.delay (* state.delay 2)))
                  (f f))))]
    (tick tick)))

{: start!}
```

Any thoughts as to the exact cause? Admittedly I'm not familiar with Obj-C\Swift but happy to help find and\or fix in any way I can.

latenitefilms:
- Closes #3389

### DIFF
--- a/Hammerspoon/MJConsoleWindowController.m
+++ b/Hammerspoon/MJConsoleWindowController.m
@@ -139,9 +139,11 @@ typedef NS_ENUM(NSUInteger, MJReplLineType) {
 
     NSDictionary* attrs = @{NSFontAttributeName: self.consoleFont, NSForegroundColorAttributeName: color};
     NSAttributedString* attrstr = [[NSAttributedString alloc] initWithString:str attributes:attrs];
-    [[self.outputView textStorage] performSelectorOnMainThread:@selector(appendAttributedString:)
-                                       withObject:attrstr
-                                    waitUntilDone:YES];
+    
+    [self.outputView.textStorage beginEditing];
+    [self.outputView.textStorage appendAttributedString:attrstr];
+    [self.outputView.textStorage endEditing];
+    [self.outputView didChangeText];
 }
 
 - (NSString*) run:(NSString*)command {


cmsj:
Closing this in favour of #3400

PR #3440, opened 2023-03-22
[Ready] hs.watchable -- added alwaysNotify to watcher

asmagill:
Should address #3437 

Also: callbacks can now be objects with __call metamethod
clarified some documentation
fixed useless debug metamethod on module

There were some changes to the watch constructor and callback method, so should be tested a bit more before merging, but initial tests look good and hasn't broken anything of mine yet...

### DIFF
--- a/extensions/watchable/watchable.lua
+++ b/extensions/watchable/watchable.lua
@@ -159,7 +159,7 @@ mt_watcher = {
 --- Method
 --- Get or set whether this watcher should be notified even when the new value is identical to the current value
 ---
---- Paramters:
+--- Parameters:
 ---  * `notify` - an optional boolean specifying whether or not the watchableObject should trigger a callback when the value of the watched path is set to its current value (i.e. it is set, but doesn't actually change value)
 ---
 --- Returns:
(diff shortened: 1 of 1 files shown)


asmagill:
If you don't want to do an actual build to test this yourself, it should be noted that these changes are to a single lua file... you can save https://raw.githubusercontent.com/Hammerspoon/hammerspoon/c5f6438e23fadd96b244592a9ae697c5628a5256/extensions/watchable/watchable.lua as `watchable.lua` and then copy it to your Hammerspoon insallation (usually in /Applications) with something like this:

~~~sh
cp watchable.lua /Applications/Hammerspoon.app/Contents/Resources/extensions/hs/watchable.lua
~~~

and then restart Hammerspoon.

I've never needed to adjust the accessibility settings after doing this, even though it is technically a change to the application bundle, so check that if you're concerned or see weird behavior.

asmagill:
Ok, looks like I have some tweaks to the doc strings to make, but the lua code will work if you want to test it as specified in my previous comment...

asmagill:
_call metamethods tested, so pending further suggestions, I'd say this is ready

Issue #2544, opened 2020-10-01
Adding file hashing functions to hs.hash


asmagill:
Rather than fill up #2514, it has come time to move this aspect of the discussion to its own thread.

The proposal is to add functions to provide hashes for individual files and for directories and their contents.

I think I'm leaning in @dge9's direction that when traversing a directory, a report of the files with individual hashes should be returned. Outside of the uses being considered in the above mentioned pull, the most common use I can think of would be to compare the hashes of files against those reported elsewhere (e.g. on a maintainers web site) and they will generally be listing a hash per file rather than relying on the file contents being merged in just the correct way before generating the hash.

As to the question of `.DS_Store`, what makes the most sense to me is add an argument which allows you to pass in a table of strings specifying files or subdirectories to ignore. I can imaging using this against, say a thumb drive, to verify its contents haven't changed, and there are about 4 (5? it's been a while) separate files I generally create on these to minimize the crap that Finder and Spotlight put on them.

The question is symbolic links... in order to resolve them and make sure that pathological loops can't lock things up, I had to add code to track the directories already visited which took just under 1.5 times the time that my original function did... both under a second, though, and that was traversing 515 directories and including 850 files.

Not sure what the changes to the report format may cause, because could theoretically be a pretty large string (or array of strings, not sure which yet) that has to be piped back into Lua, so this difference might be inconsequential... I'll know more tomorrow when I get a chance to work on the refactoring.

In the mean time, the main questions I have are:
1. objections to moving to the report of individual files approach?
2. do we include a flag to optionally attempt to resolve symbolic links and include them (and possibly their contents, if they point to a directory)? Or do we simply skip them?

latenitefilms:
Having the ability to define what file extensions should be ignored (or allowed) would be great. It would be awesome if you could tell it to only find `.lua` files, but on the flip side, find everything accept `.DS_Store` files, etc.

Even if the main function just reports individual files in a table, it would still be handy to have a "helper" function which returns a single hash.

I trust whatever you decide in regards to sym-links, however I'd be tempted to just ignore them. If the user wants hashes of sym-links, they can always do this work themselves before hashing?

latenitefilms:
Eventually it would be great to also support MD5, SHA1 and CRC32b checksums.

This might be of interest? 

https://github.com/jerolimov/NSHash

asmagill:
Note to self: ***DOCUMENT LUASKIN***!!!

[...]

Issue #2514, opened 2020-09-24
Harden Hammerspoon security model: configuration management


dge9:
Hi team, new-ish Hammerspoon user here. I’ve been running it for a few weeks and installed/coded a few Spoons to replace a number of small utilities I used to use, and I’m loving it, but some potential security vulnerabilities occurred to me around the handling of configuration and module loading. Obviously please correct me if there’s something I’ve missed or misunderstood.

The current design of Hammerspoon appears to provide numerous opportunities for the configuration to be modified by an otherwise unprivileged user process (or malicious actor if a machine is left unlocked), and thereby - silently - gain effective programmatic access to Accessibility controls, the clipboard, network, and any other privileges the user might have granted Hammerspoon e.g. the camera and microphone. It seems to me that an app with the privileges and functionality of Hammerspoon ought to prevent silent configuration changes via something like the standard macOS escalation dialog, in line with how apps like Safari/Keychain/etc. require a user’s password to view password entries.

Based on the current configuration model and Lua's package.path, three vectors occur to me:

Vector 1: Silent modification of config files in ~/.hammerspoon
Possible mitigation: Introduce a config locking mechanism similar to apps like Tunnelblick, where a “locked” version of the config is copied to a second location requiring escalation to write. Each time the configuration is reloaded, every file in .hammerspoon is then compared against its locked version, and if the config has changed the user is offered the option to a) revert to the saved configuration or b) re-lock the changed configuration, requiring the user to enter their password. For convenience while developing it should be possible to disable this for a time, perhaps via an option on the Hammerspoon status menu (maybe a binary “Lock configuration”/“Unlock configuration”, or via a submenu “Configuration reloading”>”Automatic”/“Manual”/“Locked”).

Vector 2: Shadowing extensions by introducing similarly-named files into /usr/local/[share,lib]/lua
Possible mitigation: Remove /usr/local from package.path by default. I assume this will break the setups of users using e.g. Homebrew modules, but users can manually add It back in through init.lua or a Preferences checkbox could be added with a suitable warning.

Vector 3: Modifying extensions directly inside Hammerspoon.app
[...]

PR #3483, opened 2023-05-06
Rewrite of hs.hash to simplify adding new hashes

asmagill:
Rewrite of hs.hash to simplify adding additional hash code.

In addition:
* adds SHA3_224, SHA3_256, SHA3_384, and SHA3_512
* adds CRC32 checksum
* adds hs.fs.fileListForPath to generate quick list of filenames in directory and subdirectories

Hash objects can now accept multiple inputs before calculating final value.
Maintains backwards compatibility with previous module.

### DIFF
--- a/Hammerspoon.xcodeproj/project.pbxproj
+++ b/Hammerspoon.xcodeproj/project.pbxproj
@@ -598,6 +598,11 @@
 		D6A318AB1DDD836800D9B328 /* libnetwork_ping.m in Sources */ = {isa = PBXBuildFile; fileRef = D6A3189B1DDD830A00D9B328 /* libnetwork_ping.m */; };
 		D6A318AC1DDD836800D9B328 /* SimplePing.m in Sources */ = {isa = PBXBuildFile; fileRef = D6A3189D1DDD830A00D9B328 /* SimplePing.m */; };
 		D6A318AD1DDD837400D9B328 /* SimplePing.h in Headers */ = {isa = PBXBuildFile; fileRef = D6A3189C1DDD830A00D9B328 /* SimplePing.h */; };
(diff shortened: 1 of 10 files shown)


asmagill:
These additions allow for a more robust discussion of the hashing aspects of #2514 and #2544.

As an example, the following generates a report I could save and refer back to for the Hammerspoon application bundle itself:

~~~lua
appBundlePathReport = function(h)
    local fs    = require("hs.fs")
    local hash  = require("hs.hash")
    local timer = require("hs.timer")

    local t = timer.secondsSinceEpoch()
    
    local appPath = hs.processInfo.bundlePath
    local hashFN  = h or "SHA3_256"

    local files, fileCount, dirCount  = fs.fileListForPath(appPath, {
        subdirs        = true,
        relativePath   = true,
        followSymlinks = false,
        ignore         = { "^\\..*$" },
    })

    local report = ""
    for i, v in ipairs(files) do
        report = report .. string.format("%3d. %s %s\n", i, hash.forFile(hashFN, appPath .. "/" .. v), v)
    end

    report = report .. [[

File count: ]] .. fileCount .. [[ in ]] .. dirCount .. [[ directories.
]]

    return report, hash[hashFN](report), timer.secondsSinceEpoch() - t
end
~~~

(For those interested, on my M1 Macbook Pro, using SHA3_256 takes just over half a second while using SHA256 takes 0.05 seconds -- SHA3 *is* where people are moving, but it's also slower...)

Further discussion should probably take place on one of those threads, but I wanted to put this out there as an example... true hardening will have to also address things like what directories to check, how hard to circumvent the check(s), how hard (without recompiling) to modify the checking code, etc. That's beyond the scope of this single pull, however.

PR #3484, opened 2023-05-06
Fix blockAlert documentation

obecker:
Change the style parameter in the example from "NSCriticalAlertStyle" to "critical". Also, according to the implementation (line 636), the default style will be "warning", not "informational".


obecker:
@cmsj @asmagill Hi, could you have a look please? Thanks

cmsj:
lgtm, thanks!

PR #3542, opened 2023-09-30
Update alert.lua: allow window object as parameter

muescha:
allow also a `hs.window` object as an parameter

### DIFF
--- a/extensions/alert/alert.lua
+++ b/extensions/alert/alert.lua
@@ -100,7 +100,7 @@ local showAlert = function(message, image, style, screenObj, duration)
 --        print(finspect(message:asTable()))
     end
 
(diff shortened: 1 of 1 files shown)


muescha:
optional: it is possible to change it just to `screenObj:frame()` because both `hs.window` and `hs.screen` have the `:frame()` method:

```lua
local screenFrame = screenObj:frame()
```

muescha:
- should I rename the `screenObj`?
- should also an `hs.geometry` parameter allowed?

muescha:
@cmsj 

> This pull request does not contain a valid label. Please add one of the following labels: `['pr-fix', 'pr-change', 'pr-feature', 'pr-maintenance']`

@cmsj a user can not set the label - so I can not fix the requested changes

PR #3573, opened 2023-12-20
Prototype for moving core functionality into a Swift framework

cmsj:


latenitefilms:
I love this!!! Will have a proper look/play soon!

cmsj:
This now covers three extensions - hs.math, hs.base64 and hs.camera.

hs.math is trivial, hs.base64 is also relatively simple, but showed a nice reduction in complexity (which, to be fair, we could have achieved using NSData without involving Swift), and hs.camera is a properly representative, complex extension with watchers.

hs.camera is not yet fully working - the isInUse KVO watcher doesn't seem to be working, but I'm not sure why yet. Possibly because the app needs Camera access permissions. Also I strongly dislike that one LuaSkin thing (callback lua ref) has now leaked into Hammertime, and one thing (camera object canary) is missing. Both of these call for switching the C-side userdata object to a struct that contains the canary, callback ref and camera object, rather than having the camera object be the userdata. This is something @asmagill have discussed previously and I've forgotten which of us preferred using structs and which preferred directly using ObjC objects as userdata, but whichever way round we were, this prototype indicates that structs are the only sane option if we want Hammertime to remain completely isolated from Lua (which we should want, because otherwise there's little point abstracting all the core functionality of the extensions into a framework).

Would welcome thoughts and reviews. I'll push to this a little more tomorrow to take care of the FIXMEs.

asmagill:
I am on the pro Object side because it simplifies allowing the use of one module's objects in another module without actually having to have any specific code to know how to handle the object -- it's always a (possibly distant) descendant of NSObject and can always be treated that way.

A structure could probably provide a similar "simplicity of mixing" if we insisted that the first three members in the structure were *always* the same and in the same order (the object,  the callback, and the canary, for example) with any additional unique pieces of information afterwards -- then the structure could be cast to the base type to get at the core object if required in another module.

cmsj:
@asmagill we could also make it an ObjC class, with an `id` for the real object and callbackref/canary properties. That way at least we're protected against the fickle nature of struct offsets if we need to change things later

cmsj:
Ok, so a hosting ObjC class turned out to not be great because then those have to be cached on the ObjC side as well as the Swift side having to cache its objects. Instead I added a raw pointer on the Swift side for the ObjC side to do what it likes with - in this case, storing a struct there which has the callbackRef and canary.

Oddly, the `CMIOObjectRemovePropertyListenerBlock()` call isn't working. It returns success, but the callback block is still called after an `hs.reload()`. It's sufficiently well guarded that it doesn't actually do anything (or crash), but I am totally baffled as to why it's happening.

That aside, I am now pretty happy with the state of hs.camera.

cmsj:
If anyone wants to play with the part that isn't working:

1. Add a breakpoint on `Camera.swift` line 213
2. Build & Run
3. Run the Lua below in the HS console:

``lua
cam = hs.camera.allCameras()[1]
print(cam:name())
cam:setPropertyWatcherCallback(function() print("isInUse changed") end)
cam:startPropertyWatcher()
```

If you now fire up something that uses the webcam, you'll see `isInUse changed` printed in the HS console.

Now run `hs.reload()` - at this point our Camera object has been garbage collected (which you can verify with a breakpoint on `libcamera.m` line 496 and some stepping.

Finally, either close or re-open the webcam-using app. You'll see the `Camera.swift` breakpoint is hit, even though it shouldn't be.

Issue #3607, opened 2024-02-29
Feature Request: I would like to know if a modal is in the state active


muescha:
I would like to know if a modal is in the state active

```lua
local ks = hs.hotkey.modal.new(
            mod,
            key,
            description);

ks:isActive()
ks.isActive -- call value direct
```

I can do a workaround like this:

```lua
function hotkeybindmodal(mod, key, description, startFn, exitFn)

    local active = false

    local ks = hs.hotkey.modal.new(
            mod,
            key,
            description);

    function ks:entered()
        debugInfo("Start Modal Mode for ".. description)
        startFn()
        active = true
    end
    function ks:exited()
        active = false
        exitFn()
        debugInfo("Exit Modal Mode for ".. description)
    end

    function ks:isActive()
        return active
    end

    return ks
end
```

and use it:
```lua
local ks = hotkeybindmodal(
            mod,
            key,
            description);
ks:enter()
print(ks:isActive())
```

muescha:
see:
- PR #3608: Update hotkey.lua: add isActive

PR #3608, opened 2024-02-29
Update hotkey.lua: add isActive

muescha:
add isActive

See:
- #3607 Feature Request: I would like to know if a modal is in the state active

### DIFF
--- a/extensions/hotkey/hotkey.lua
+++ b/extensions/hotkey/hotkey.lua
@@ -544,6 +544,7 @@ function hotkey.modal:enter()
   end
   for _,hk in ipairs(self.keys) do enable(hk,nil,true) end
   self:entered()
+  self.isActive = true
   return self
 end
 
@@ -564,6 +565,7 @@ function hotkey.modal:exit()
   if (self.k) then
     enable(self.k)
   end
+  self.isActive = false
   self:exited()
   log.d('Exited modal')
   return self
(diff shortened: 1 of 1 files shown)


muescha:
TODO:
- add docs if accepted
- should it be better as function like `htm:isActive()`?

cmsj:
@muescha ignore the CI tests, they're horribly broken at the moment. I think this feature makes sense to add, so if you'd like to go ahead with it, I like your idea that an `:isActive()` method would be a better choice overall (so the precise implementation is hidden and can be changed later). With docs and that method, this can merge.

muescha:
did you mean "hiding implementation" like this: 
- removing from metatable
- renaming value to `_isActive`
- creating `:isActive()` method

extensions/hotkey/hotkey.lua
```diff
@@ -589,7 +591,8 @@ end
 --- Notes:
 ---  * If `key` is nil, no global hotkey will be registered (all other parameters will be ignored)
 function hotkey.modal.new(mods, key, message)
-  local m = setmetatable({keys = {}}, hotkey.modal)
+    local m = setmetatable({keys = {}}, hotkey.modal)
```
Angel644-wep: Samsung a 356E

Angel644-wep (approved):

PR #3635, opened 2024-05-11
Add NSUserActivityTypeBrowsingWeb handling

mdbraber:
To handle Handoff from a native app to a (default) browser `NSUserActivityTypeBrowsingWeb` needs to be handled. If a default browser supports this, Handoff links that have a `webpageUrl` can be opened by the default browser.

More info: https://developer.apple.com/library/archive/documentation/UserExperience/Conceptual/Handoff/AdoptingHandoff/AdoptingHandoff.html

### DIFF
--- a/Hammerspoon/MJAppDelegate.m
+++ b/Hammerspoon/MJAppDelegate.m
@@ -88,6 +88,7 @@
             }
         }
 
+
         success = [[NSFileManager defaultManager] moveItemAtPath:fileAndPath toPath:dstSpoonFullPath error:&fileError];
         if (!success) {
             NSLog(@"Unable to move %@ to %@: %@", fileAndPath, spoonPath, fileError);
@@ -127,6 +128,17 @@
     }
 
     return YES;
+}
+
+- (BOOL)application:(NSApplication *)application
+    continueUserActivity:(NSUserActivity *)userActivity
+      restorationHandler:(nonnull void (^)(NSArray<id<NSUserActivityRestoring>> *_Nullable))restorationHandler
+{
+  //NSLog(@"Open URL in NSUserActivityTypeBrowsingWeb");
+  if ([userActivity.activityType isEqualToString:NSUserActivityTypeBrowsingWeb]) {
+    [[NSWorkspace sharedWorkspace] openURL:userActivity.webpageURL];
+  }
+  return YES;
 }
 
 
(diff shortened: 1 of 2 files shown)


mdbraber:
@latenitefilms not sure how to add a a label afterwards, but I guess this is a `pr-feature`

latenitefilms:
I believe only @cmsj and @asmagill  can add labels and approve pull requests.

asmagill:
I am out of town at the moment, but I will try and carve out some time here in the next couple of days to see about moving some of these forward.

mdbraber:
@asmagill any chance of getting this merged?

Hammerspoon/MJAppDelegate.m
```diff
@@ -129,6 +130,17 @@ - (BOOL)application:(NSApplication *)theApplication openFile:(NSString *)fileAnd
...
 
+- (BOOL)application:(NSApplication *)application
+    continueUserActivity:(NSUserActivity *)userActivity
+      restorationHandler:(nonnull void (^)(NSArray<id<NSUserActivityRestoring>> *_Nullable))restorationHandler
+{
+  //NSLog(@"Open URL in NSUserActivityTypeBrowsingWeb");
+  if ([userActivity.activityType isEqualToString:NSUserActivityTypeBrowsingWeb]) {
+    [[NSWorkspace sharedWorkspace] openURL:userActivity.webpageURL];
+  }
+  return YES;
```
cmsj, line 141: Should this be unconditionally returning `YES`? The [docs](https://developer.apple.com/documentation/appkit/nsapplicationdelegate/1428471-application) suggest that we probably ought to return `YES` only inside the `if` statement, and otherwise return `NO`, right?

cmsj (changes requested):

megalithic:
@mdbraber this is a fantastic update and would solve a major url handling complaint when using hammerspoon as the default browser. Thanks for tackling this!

megalithic:
@mdbraber any updates or thoughts on this?

Hammerspoon/MJAppDelegate.m
megalithic: @cmsj; if i were to correct this in a separate PR would that be ok? don't want to step on @mdbraber's toes, but also really want this in.

cmsj (approved):

cmsj:
@megalithic @mdbraber I made the change I think is correct and merged, it would be super helpful if someone could test the Development Build to check this is working. It should be available in an hour or so from: https://github.com/Hammerspoon/hammerspoon/actions/runs/10444554151

megalithic:
> @megalithic @mdbraber I made the change I think is correct and merged, it would be super helpful if someone could test the Development Build to check this is working. It should be available in an hour or so from: https://github.com/Hammerspoon/hammerspoon/actions/runs/10444554151

@cmsj thank you (also to @mdbraber); it works well! was able to handoff an ios brave browser tab to hammerspoon and the URL handlers for hammerspoon took over. 💯

mdbraber:
Great to see this merged while I was away :-)

Issue #3636, opened 2024-05-13
hs.spaces.moveWindowToSpace not working for me


akarsai:
the following code snippet used to work for me, but stopped working a few weeks ago. the snippet is responsible for finding the current terminal window, moving it to the active space, and activating it. unfortunately, the window is not moved by hammerspoon. the line`hs.spaces.moveWindowToSpace(app:mainWindow(), currentSpace)` is not working as it should, and returns `true` even though the window was not succesfully moved.

can anyone reproduce the issue and/or does anyone have any ideas on how to fix this? unfortunately, i do not exactly know what i have changed, it could be also be a macOS update that is at fault here. i use macOS 14.5 Beta (23F5064f).

```lua
-- quickly open kitty (terminal)
hs.hotkey.bind({"cmd"}, "g", function()
    local windowPosition = '0.2,0.4,0.8,0.9'
    
    -- Get current space 
    local currentSpace = hs.spaces.focusedSpace()
    
    -- Get kitty app
    local app = hs.application.get("kitty")
    
    -- If app already open:
    if app then
        
        -- If no main window, then open a new window
        if not app:mainWindow() then
            app:selectMenuItem("New OS Window", true)
            
            hs.timer.doAfter(.05,function ()
                app:mainWindow():raise()
                app:mainWindow():moveToUnit(windowPosition)
                end)
        
        -- If app is already in front, then hide it
        elseif app:isFrontmost() then
            app:hide()
        
        -- If there is a main window somewhere, bring it to current space and to front
        else
            -- First move the main window to the current space
            hs.spaces.moveWindowToSpace(app:mainWindow(), currentSpace)
            
            -- Activate the app
            app:activate()
            
            --Raise the main window and position correctly
            app:mainWindow():raise()
            app:mainWindow():moveToUnit(windowPosition)
        end
    
    -- If app not open, open it
    else
        hs.application.launchOrFocus("kitty")
        
        hs.timer.doAfter(.2,function () 
            app = hs.application.get("kitty") 
            app:mainWindow():raise()
            app:mainWindow():moveToUnit(windowPosition)
            end)
    end

end)
```

sidkang:
Facing the same issue, 14.4.1 worked, but after 14.5, not any more.

mattfloat:
It appears macOS 14.5 changed some functions to protected which could be the reason for our symptoms.
[...]

PR #3638, opened 2024-05-25
[Review] Fix moving window to space in Sonoma 14.5

asmagill:
Addresses #3636

### DIFF
--- a/extensions/spaces/libspaces.m
+++ b/extensions/spaces/libspaces.m
@@ -12,6 +12,14 @@ static int g_connection ;
 
 #pragma mark - Support Functions and Classes
 
+// https://github.com/koekeishiya/yabai/commit/7bacdd59bdf39b53012e024b442d61913095794e#diff-fab09245ac7f0bacbd4b3648bdd51eff6c41dc937f5dbaaf7459ff9ea70d3557R17-R27
+static bool workspace_is_macos_sonoma14_5_or_newer(void) {
+    NSOperatingSystemVersion os_version = [[NSProcessInfo processInfo] operatingSystemVersion];
+    if (os_version.majorVersion > 14) return true;
+    if (os_version.majorVersion == 14 && os_version.minorVersion >= 5) return true;
+    return false;
+}
+
 #pragma mark - Module Functions
 
 /// hs.spaces.screensHaveSeparateSpaces() -> bool
@@ -171,7 +179,15 @@ static int spaces_moveWindowToSpace(lua_State *L) { // NOTE: wrapped in init.lua
                 CFRelease(spacesList);
                 return 2 ;
             }
-            SLSMoveWindowsToManagedSpace(g_connection, (__bridge CFArrayRef)windows, sid) ;
+
+// https://github.com/koekeishiya/yabai/commit/98bbdbd1363f27d35f09338cded0de1ec010d830
+            if (workspace_is_macos_sonoma14_5_or_newer()) {
+                SLSSpaceSetCompatID(g_connection, sid, 0x79616265);
+                SLSSetWindowListWorkspace(g_connection, &wid, 1, 0x79616265);
+                SLSSpaceSetCompatID(g_connection, sid, 0x0);
+            } else {
+                SLSMoveWindowsToManagedSpace(g_connection, (__bridge CFArrayRef)windows, sid) ;
+            }
         }
         lua_pushboolean(L, true) ;
         CFRelease(spacesList) ;
--- a/extensions/spaces/private.h
+++ b/extensions/spaces/private.h
@@ -13,6 +13,8 @@ extern CFArrayRef SLSCopyWindowsWithOptionsAndTags(int cid, uint32_t owner, CFAr
 extern void SLSMoveWindowsToManagedSpace(int cid, CFArrayRef window_list, uint64_t sid);
 extern CFArrayRef SLSCopySpacesForWindows(int cid, int selector, CFArrayRef window_list);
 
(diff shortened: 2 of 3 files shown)


asmagill:
@cmsj the failure looks like a problem with the reporting? At any rate, unless I'm missing something I can ignore this I think... let me know if I'm wrong.

nzhl:
When will this be merged? Can't wait to see the fixed version !

dxgx:
I wait this merge every day :) Thanks for you work!

bethandtownes:
me too

nzhl:
is this repo deprecated ? It hasn't been updated for quite a while...

latenitefilms:
> is this repo deprecated ? It hasn't been updated for quite a while...

@nzhl - Definitely not deprecated, however, @cmsj (the maintainer) has been busy with other things, so sadly Hammerspoon hasn't been updated in a while.

@asmagill - I could always build & sign locally and share a local build here if helpful?

dxgx:
> > is this repo deprecated ? It hasn't been updated for quite a while...
> 
> @nzhl - Definitely not deprecated, however, @cmsj (the maintainer) has been busy with other things, so sadly Hammerspoon hasn't been updated in a while.
> 
> @asmagill - I could always build & sign locally and share a local build here if helpful?

@latenitefilms please do share build. Because my build is failing to organise windows to spaces. Thanks!

gartnera:
I have a build on my fork: https://github.com/gartnera/hammerspoon/releases/tag/0.10.0

Tested on MacOS Sonoma 14.5.

cunha:
I've been using the version packaged by @gartnera above. Although I can move windows across Spaces on the same screen without issue, moving windows across Spaces on different screens is not working (the window apparently gets moved but is not visible).

I am considering the error might be in my Lua code calling the Hammerspoon API. Can anyone share Lua code that works when moving windows across Spaces on different screens so I can compare?

irobertson:
I'm seeing similar issues to @cunha - moving a window to a space on a different screen results in the window being invisible. It appears that the cause is that the geometry of the window doesn't change, meaning that it's coordinates are still for the old screen. For now, I'm getting around this by calling window:moveToScreen in addition to calling moveWindowToSpace; I also have to add in a sleep of about a half a second between the two calls, or it ends up moving the window to the currently selected space on the screen containing the target space.

That said, even with the workaround needed, a _huge_ thanks to @asmagill for making this change. It has been driving me nuts since upgrading to 14.5 that I haven't been able to use hammer spoon to move windows around. Your change makes a significant improvement to my day!

dxgx:
> I'm seeing similar issues to @cunha - moving a window to a space on a different screen results in the window being invisible. It appears that the cause is that the geometry of the window doesn't change, meaning that it's coordinates are still for the old screen. For now, I'm getting around this by calling window:moveToScreen in addition to calling moveWindowToSpace; I also have to add in a sleep of about a half a second between the two calls, or it ends up moving the window to the currently selected space on the screen containing the target space.
> 
> That said, even with the workaround needed, a _huge_ thanks to @asmagill for making this change. It has been driving me nuts since upgrading to 14.5 that I haven't been able to use hammer spoon to move windows around. Your change makes a significant improvement to my day!

Yes i can confirm too. When i use multiple displays some windows that should appear in Mac book Built-in retina display appears hidden. Only "Window -> Move To Dell XXX" helps to "unhide" window

jjournet:
Hi guys,
since the upgrade to Sonoma 14.5, I can't move window through spaces, and it's a major hindrance to my workflow.

I understand that open source project rely on the time available to maintainers, so I would be willing to help out on this one if required. I'm not fluent in Objective-C and Lua, but I'm willing to learn enough to unblock the situation (and I'm very familier with Github workflows, so if that's the issue, I can help easily)

What would allow this to move forward ?

cunha:
The current workaround is to use the [build by gartnera](https://github.com/Hammerspoon/hammerspoon/pull/3638#issuecomment-2195783259), and then change the Lua code to show windows when they are moved to different screens (see discussion in the previous posts).

For reference, this is how my Lua code looks at the moment (it's hacky, but beats moving windows with the mouse):

```lua
local function MoveWindowToSpace(sp)
    prefix.exit()
    local win = hs.window.focusedWindow()      -- current window
    local screenID, spaceID = getScreenSpaceID(sp)
    print("moving window to space " .. spaceID)
    spaces.moveWindowToSpace(win:id(), spaceID)
    win:focus()  -- follow window to new space

    -- MacOS takes focus from us. Get it fucking back. This may cease to
    -- be an issue after we choose to "Reduce motion" in
    -- System preferences -> accessibility -> display.
    -- cunha@20240409 it's still an issue
    hs.timer.usleep(400000)
    -- moveToScreen may not be necessary after Sonoma fix is merged:
    -- https://github.com/Hammerspoon/hammerspoon/pull/3638#issuecomment-2209445157
    win:moveToScreen(screenID)
    win:focus()
end
```

jjournet:
Hello @cunha , thank you for the workaround, I'll perhaps implement it if I don't have any other option (I'm trying to avoid installing fork or branch releases as much as possible on my work Mac)

My question was probably unclear. What would it take to have this PR to be merged and delivered with the official Hammerspoon ?  I understand @asmagill fix is the solution, and he did an amazing job at fixing it, I just want to help if I can with any impediment in the PR that would prevent its merge.

latenitefilms:
I think the main bottleneck is that only @cmsj has the ability to push out a Hammerspoon release, and he's flat out with other things.

cmsj:
Sorry everyone, I know this has been super annoying!

PR #3665, opened 2024-08-10
Fixed websocket support in `hs.httpserver:start()`

### LINKED ISSUE #3468: Why is an HTTP callback required if I only want a WebSocket server in hs.httpserver?
Hello,

I'm using Hammerspoon to create a simple WebSocket server using the `hs.httpserver` module. However, I noticed that the module requires an HTTP callback function to be set, even if I'm only interested in WebSocket functionality.

Is there a specific reason for this requirement? I would expect that, since WebSocket and HTTP are separate protocols, it would be possible to set up a WebSocket server without having to provide a callback for normal HTTP requests.

Could you please clarify this requirement or consider updating the module to allow for WebSocket-only servers without the need for an HTTP callback?

Thank you for your time and assistance.

Best regards

latenitefilms:
- Closes #3468

### DIFF
--- a/extensions/httpserver/libhttpserver.m
+++ b/extensions/httpserver/libhttpserver.m
@@ -583,13 +583,8 @@ static int httpserver_start(lua_State *L) {
     LuaSkin *skin = [LuaSkin sharedWithState:L];
     HSHTTPServer *server = getUserData(L, 1);
 
-    if (server.wsCallback != LUA_NOREF) {
-        NSError *error = nil;
-        if (![server start:&error]) {
-            [skin logError:[NSString stringWithFormat:@"hs.httpserver:start() Unable to start object: %@", error]];
-        }
-    else if (server.fn == LUA_NOREF) {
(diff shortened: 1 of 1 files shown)


extensions/httpserver/libhttpserver.m
```diff
@@ -583,7 +583,12 @@ static int httpserver_start(lua_State *L) {
     LuaSkin *skin = [LuaSkin sharedWithState:L];
     HSHTTPServer *server = getUserData(L, 1);
 
-    if (server.fn == LUA_NOREF) {
+    if (server.wsCallback != LUA_NOREF) {
+        NSError *error = nil;
+        if (![server start:&error]) {
+            [skin logError:[NSString stringWithFormat:@"hs.httpserver:start() Unable to start object: %@", error]];
+        }
+    else if (server.fn == LUA_NOREF) {
```
cmsj, line 591: Rather than duplicating the `[server start:]` call, how about we just check if both `server.wsCallback` and `server.fn` are `LUA_NOREF` and use that to throw the "no callback set" error?

cmsj (changes requested):

PR #3667, opened 2024-08-12
Lua skin additions

asmagill:
Some proposed changes to LuaSkin:

I know we're planning a move to Swift soon, but in the meantime, this pull has the following:

1) the `pushNSObject:` methods will now check for a more specific match (i.e. `isMemberOfClass:`) and then fall back to `isKindOfClass:`.

2) the `log*:` methods will now accept varagrs in the `NSString stringWithFormat:` style. 
    * this allows `[LuaSkin logError:@"%s - callback error", USERDATA_TAG]`, as opposed to the current `[LuaSkin logError:[NSString stringWithFormat:@"%s - callback error", USERDATA_TAG]]`

### DIFF
--- a/LuaSkin/LuaSkin/Skin.h
+++ b/LuaSkin/LuaSkin/Skin.h
@@ -891,49 +891,49 @@ NSString *specMaskToString(int spec);
  @discussion This method is equivalent to invoking @link logAtLevel:withMessage: @/link with level @link LS_LOG_VERBOSE @/link
  @param theMessage the message to log
  */
-- (void)logVerbose:(NSString *)theMessage ;
+- (void)logVerbose:(NSString *)theMessage, ... ;
 
 /*!
  @abstract Log the specified message with LS_LOG_DEBUG level
  @discussion This method is equivalent to invoking @link logAtLevel:withMessage: @/link with level @link LS_LOG_DEBUG @/link
  @param theMessage the message to log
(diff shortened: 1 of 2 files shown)


cmsj:
Assuming CI "passes" with only its regular amount of failing tests, this lgtm.

cmsj:
The regular amount of tests failed, so I'm merging this.

asmagill:
Was just about to add comments in the internal `pushNSNumber:` method for when we migrate to Swift.

We currently do a ptr comparison against `kCFBooleanTrue` and `kCFBooleanFalse` to determine if an `NSNumber` is boolean (@encode for `BOOL` is the same as for `unsigned char`, so we can't use that).

Per https://stackoverflow.com/a/30223989, it's probably safer (and Swift friendly) to do something like:

~~~swift
func isBoolNumber(num:NSNumber) -> Bool
{
    let boolID = CFBooleanGetTypeID() // the type ID of CFBoolean
    let numID = CFGetTypeID(num) // the type ID of num
    return numID == boolID
}
~~~

They give an Obj-C version as well, but I don't want to change what's currently working atm.

If we have a place where we're sticking Swift related notes, let me know and I'll post this there.

cmsj:
We don't have such a thing yet, but we probably should. I wonder if a Project might make sense?

PR #3703, opened 2024-10-11
Fix `package.path`s under `userruntime` w/ unexpanded tilde

### LINKED ISSUE #3518: Tilde in `userruntime` variable used in `package.path` not expanded
Looks like a bunch of paths under `~/.local/share/hammerspoon/site` are added to `package.{c,}path` in `setup.lua`. See the current tree:

https://github.com/Hammerspoon/hammerspoon/blob/0ccc9d07641a660140d1d2f05b76f682b501a0e8/Hammerspoon/setup.lua#L3-L24

Unfortunately, using a directory in the path doesn't work, unless you add the same path with `/Users/$USER` instead of `~`. I think that first line which sets `userruntime` should be set to `os.getenv("HOME") .. "/.local/share/hammerspoon/site"` instead of `~/.local/share/hammerspoon/site`.

habibalamin:
Lua doesn't automatically expand tildes to `$HOME`, as this code seems to assume.

Fix #3518

I didn't mean to close #3702.

### DIFF
--- a/Hammerspoon/setup.lua
+++ b/Hammerspoon/setup.lua
@@ -1,6 +1,6 @@
 local modpath, frameworkspath, prettypath, fullpath, configdir, docstringspath, hasinitfile, autoload_extensions = ...
 
-local userruntime = "~/.local/share/hammerspoon/site"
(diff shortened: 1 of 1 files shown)


cmsj:
Thanks!

PR #3716, opened 2024-11-20
Add hs.audiodevice:thru() and hs.audiodevice:setThru(thru)

bburky:
```
hs.audiodevice:thru() -> bool or nil
hs.audiodevice:setThru(thru) -> bool
```

Get or set the play through (low-latency/direct monitoring) state of the the audio device via `kAudioDevicePropertyPlayThru`. This is the feature of some microphones where they can play their input directly to their output (e.g. headphone jack) so you can get low-latency feedback while recording.

* This only works on devices that have hardware support (often microphones with a built-in headphone jack)
* This setting corresponds to the "Thru" setting in Audio MIDI Setup
* I only have one device to test with, it's a combo input/output device. I'm unsure if they're ever split into multiple devices? I decompiled Audio MIDI Setup (before I got around to installing Xcode) and it does seem to do a conditional check, supporting both input and output devices, before setting thru.
   * I don't think we need separate `hs.audiodevice:setInputThru(state)` and `hs.audiodevice:setOutputThru(state)`?

(I have a USB microphone that annoyingly enables this every time is reconnected or I reboot, I want to script disabling it. This could also be useful for other users who like this setting, but perhaps want do something like assign it a keyboard shortcut.)

### DIFF
--- a/Hammerspoon Tests/HSaudiodevice.m
+++ b/Hammerspoon Tests/HSaudiodevice.m
@@ -144,6 +144,10 @@
     RUN_LUA_TEST()
 }
 
+- (void)testThru {
+    RUN_LUA_TEST()
+}
+
 - (void)testVolume {
     SKIP_IN_TRAVIS()
     RUN_LUA_TEST()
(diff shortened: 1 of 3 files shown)


bburky:
I guess I can add `kAudioDevicePropertyPlayThru`to `watchSelectors` for `hs.audiodevice:watcherCallback()` if that's valuable? The event name is `thru`. I tested it, let me know if I should add it too.

PR #3745, opened 2025-01-29
window_filter: track applications by PID

mogenson:
The `apps` and `pendingApps` lists in `window_filter.lua` are indexed by app name. The window filter will reject a new app if it has the same name as an existing app. This can be an issue because there's nothing in MacOS that prevents two apps from having the same name. For example, in Firefox, if you select "Launch profile in new browser", Firefox will open a new Firefox application with the same name, bundleID, and path. Two separate Firefox icons in the Dock will confirm that these are unique applications.

Switch the window filter to track applications by PID instead of name. Fetch the application PID in `startAppWatcher()` and stash it in the `App` object to reference later. Change the function signature of `windowEvent()` and `appWindowEvent()` to receive a PID instead of app name and lookup the `App` object by PID.

Both Firefox applications are recognized and tracked separately now. No regressions in window filter event processing have been observed.

### DIFF
--- a/extensions/window/window_filter.lua
+++ b/extensions/window/window_filter.lua
@@ -166,9 +166,10 @@ function windowfilter._showCandidates()
   local t={}
   for _,app in ipairs(running) do
     local appname = app:name()
+    local pid = app:pid()
     if appname and windowfilter.isGuiApp(appname) and #app:allWindows()==0
       and not windowfilter.ignoreInDefaultFilter[appname] and app:kind()>=0
-      and (not apps[appname] or not next(apps[appname].windows)) then
+      and (not apps[pid] or not next(apps[pid].windows)) then
       t[#t+1]=appname
     end
   end
@@ -277,9 +278,9 @@ function WF:isWindowAllowed(theWindow)
   --which allegedly is a window, but without id
   if not id then return end
   if activeInstances[self] then return self.windows[id] and true or false end
-  local appname,win=theWindow:application():name()
-  if apps[appname] then
-    for wid,w in pairs(apps[appname].windows) do
+  local pid,win=theWindow:application():pid()
+  if apps[pid] then
+    for wid,w in pairs(apps[pid].windows) do
       if wid==id then win=w break end
     end
   end
@@ -287,7 +288,7 @@ function WF:isWindowAllowed(theWindow)
     --    hs.assert(not global.watcher,'window not being tracked')
     self.log.d('window is not being tracked')
     win=Window.new(theWindow,id) --fixme
-    win.app={} win.app.name=appname
+    win.app={} win.app.name=theWindow:application():name()
     if self.trackSpacesFilters then
       win.isInCurrentSpace=false
       if not win.isVisible then win.isInCurrentSpace=true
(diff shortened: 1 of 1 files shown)


dmgerman:
I am testing it. So far so good. I have been having the issue that this PR solves with Emacs (having multiple instances) and with Bambu Studio (it runs new instances when selecting "New Window"). So far, it allows my Spoon to switch windows by title to work nicely. Of course, the app name and icon remains the same, but the window title changes.

great work.

cmsj:
Thanks @mogenson this looks pretty reasonable. Unfortunately the developer who wrote `hs.window.filter` no longer maintains it, and it's a pretty complex module, so I can't be completely sure that this isn't going to break something else.

If anyone would be willing to take on that module, it would be much appreciated!

dmgerman:
I commit to look after this file/module for any future bugs.

avegetablechicken:
Some applications quit with the event 'deactivated' instead of 'terminated', causing the `WF.filters`' key 'pid' not being refreshed. Before the patch, even if 'appname' was not refreshed as expected, it would not cause a problem because appname never changes. But it does cause a problem when using pid as the key of `WF.filters`.
For example, I register a window filter to autonomously hide or quit an app when its last window is destroyed. When the app restarts, the window filter registered before does not work any more.

dmgerman:
> Some applications quit with the event 'deactivated' instead of 'terminated', causing the `WF.filters`' key 'pid' not being refreshed. Before the patch, even if 'appname' was not refreshed as expected, it would not cause a problem because appname never changes. But it does cause a problem when using pid as the key of `WF.filters`. For example, I register a window filter to autonomously hide or quit an app when its last window is destroyed. When the app restarts, the window filter registered before does not work any more.

Can be specific about your use case? ie. when do you start the filter? Do you have example of such application?

avegetablechicken:
@dmgerman 
Below is my code to autonomously quit an app when it does not have a window.
```lua
local windowFilterAutoQuit = hs.window.filter.new(false)
for bundleID, cfg in pairs(appsAutoQuitWithNoWindows) do
  local func = function(appObject)
    windowFilterAutoQuit:setAppFilter(appObject:name(), cfg)
  end
  local appObject = findApplication(bundleID)
  if appObject ~= nil then
    func(appObject)
  else
    execOnLaunch(bundleID, func, true)
  end
end
windowFilterAutoQuit:subscribe(hs.window.filter.windowDestroyed,
  function(winObj)
    if winObj == nil or winObj:application() == nil then return end
    local bundleID = winObj:application():bundleID()
    processAppWithNoWindows(winObj:application(), true, appsWithNoWindowsDelay[bundleID])
  end)
```
Hammerspoon may only capture `event.deactivated` when the built-in Messages app (mine is in Chinese) quits. But the pid key is only cleared when `event.terminated`  is captured.

dmgerman:
Correct me if I am wrong. This is what you are doing/observing.

1. You start a window.filter when you start the application (execOnLaunch)

2. When the filter gets a filter.windowDestroyed you execute processAppWithNoWindows.

3. When you quit the application, the filter remains "active"

so is the problem that we will have a memory leak with these filters staying around even if the pid is no longer running?

or is there any other problem?

avegetablechicken:
> Correct me if I am wrong. This is what you are doing/observing.
> 
> 1. You start a window.filter when you start the application (execOnLaunch)
> 2. When the filter gets a filter.windowDestroyed you execute processAppWithNoWindows.
> 3. When you quit the application, the filter remains "active"
> 
> so is the problem that we will have a memory leak with these filters staying around even if the pid is no longer running?
> 
> or is there any other problem?

No other problem.

dmgerman:
Great. In that case, I think the solution will be, the next time a window filter is created, check that the pids are still valid (at the very least for that app).

dmgerman:
> BTW, Why not tracking applications by bundle identifier? I replaced all the pid in this commit with bundle id. it seems everything works as expected and I can now call `setAppFilter` with bundle identifier.

The window filter will stop working properly if you have two instances or more instances of the same application.

PR #3765, opened 2025-03-28
Allow to show window titles in hints

robewald:
Remove the showTitles argument from hints.displayHintsForDict. This was always nil and setting hints.showTitleThresh had no effect.  By removing this argument the hints.showTitleThresh parameter can be used again to show window titles.


robewald:
I don't seem to be able to add labels to this pull request.

PR #3778, opened 2025-05-17
Fix typo in logger's documentation

edkolev:


edkolev:
> This pull request does not contain a valid label. Please add one of the following labels: `['pr-fix', 'pr-change', 'pr-feature', 'pr-maintenance']`

Hm, I don't think I'm able to set a label.

latenitefilms:
I'm not sure this change is correct or makes sense?

Methods should use a column?

edkolev:
Hi @latenitefilms I'm fixing an error in the documentation here: https://www.hammerspoon.org/docs/hs.logger.html#new

The documentation, and the example in the documentation, suggests using dot instead of colon.

I'm fixing the following typo, rephrased:

Before: "do not use a **dot**, use a dot"
After: "do not use a **colon**, use a dot"

PR #3785, opened 2025-06-04
Miscellaneous updates to enable compilation

algal:
These are changes I made only in order to get the HEAD of master to compile. I believe they are behaviorally neutral, but I have not tested them in any depth.

But the app compiles and runs on my system (macOS 15.5 24F74, Xcode 16.4, brew/libgccjit 15.1.0, brew/gcc 15.10).

These are the changes:

- add `-Wno-missing-include-dirs` to silence a breaking warning for the target `noises`
- update the pod "Sentry" from 8.26.0 to 8.52.0, to silence errors due to a stricter compiler
- in target `osascript`, fixed a format specifier
- in target `stream deck`, refactored to an NS_ENUM and add a default switch case.
- in traget `razer`, fixed a format specifier
- in target `screen`, fixed format specifiers
- in target `midi`, **added crashing asserts** for previously unhandled switch cases, for unknown `MIKMIDICommandType`. This is one of the potentially dangerous changes, if code is accidentally relying on the non-existing default behavior now. Commit 2a571c5a
- in target `hid`, fixed format specifiers
- in target `hid`, **add crashing asserts** for previously unhandled switch case, for unknown op codes to `access_capslock`. Commit f982bb82.

As you can see, most of this is super boring and obvious.

The two risky commits are the ones which introduce asserts which will crash in case a switch gets a value which it now has no default handling for. I looked at these briefly and it seems fine. But it might deserve a glance from someone who knows that code, or else a more conservative approach by pulling the asserts. I'm not sure if asserts are compiled out in production builds in this code base?


latenitefilms:
@algal & @cmsj - Apologies, I have no idea what I've done, but I've seemed to completely destroyed this pull request by attempting to merge @algal's great work into my `CommandPost-App` repo. 🫠

Sorry!!

algal:
@latenitefilms @cmsj No worries. I don't understand but it looks like your commits are in my branch in my repo as well. 

I created a new branch which has my original commits, in case anyone wants to follow up on making use of them: https://github.com/algal/hammerspoon/commits/dev-compile_fixes2

PR #3812, opened 2025-08-12
Reuse [statusItem menu] if possible (fixes #1923)

### LINKED ISSUE #1923: Callling hs.menubar:setMenu while menu is visible should redraw with new menu
Currently, asynchronous calls to `setMenu` on an existing menu bar while the user has clicked the menubar item and has its menu open results in the menu being shown in a disabled state with no items, essentially a grey rectangle. This requires the user to click off to disappear the menu, then click the menubar again to display the new menu.

This user experience could be improved by detecting the opened state of the menu upon calls to `setMenu` so that the menu gets redrawn automatically after the previous one is destroyed and a new one is assigned to the menubar item.

HealsCodes:
This change allows calls to hs.menubar:setMenu() to update the menu even if it is currently visible.

The original code would delete the old menu and set a new NSMenu instance however macOS will not show a new menu while the old - now empty one - is still open.

### DIFF
--- a/extensions/menubar/libmenubar.m
+++ b/extensions/menubar/libmenubar.m
@@ -381,6 +381,22 @@ void erase_all_menu_parts(lua_State *L, NSStatusItem *statusItem) {
    return;
 }
 
+// Prepare an existing menu on a menubar item for reuse or create a new menu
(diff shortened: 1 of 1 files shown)


HealsCodes:
I'd love to Mr. GitLab-Bot but I'm not allowed to add labels - not even pr-fix.

PR #3827, opened 2025-10-09
Fix Lua stack corruption in window:application() when process terminated

delaneyb:
## Summary
Fixes intermittent crashes with assertion: "not enough elements in the stack" when calling hs.window:application() on windows from recently-terminated applications.

## Root Cause
- The binding used the LuaSkin shared singleton to push the return value. That singleton holds a mutable lua_State pointer (.L).
- At the crash point, the singleton's .L differed from the binding's L, so the return value was pushed onto a different lua_State than the caller was returning to.
- Evidence from runtime logs:
  - Before push: WINAPP before: L=0x600002ae2158, skin.L=0x15b00a208, app=nil
  - pushNSObject ran on L=0x15b00a208
  - After: WINAPP after: L=0x600002ae2158 top=0, pushed=1 → Lua asserted "not enough elements in the stack".
- The mutable .L can be changed by any code path using the shared singleton (e.g. logging or other sharedWithState calls). The logs show a concrete L mismatch at the moment app==nil; this change removes sensitivity to that for this binding.

## The Fix
- When app == nil (process terminated), push nil directly to the caller's L using lua_pushnil(L) instead of pushNSObject (which pushes to the singleton's .L).
- Clear the caller's stack first with lua_settop(L, 0) to avoid leaking the receiver (consistent with other window methods).

## Precedent in Codebase
- extensions/axuielement/common.m uses the same pattern:
  - if init returns nil → lua_pushnil(L); else → [skin pushNSObject:obj].

## Impact
- Minimal change; only affects the nil-return path.
- No behavior change for valid applications.
- Eliminates the observed state-mismatch crash during window iteration when the owning process has just terminated.

### DIFF
--- a/extensions/window/libwindow.m
+++ b/extensions/window/libwindow.m
@@ -493,7 +493,16 @@ static int window_application(lua_State* L) {
     [skin checkArgs:LS_TUSERDATA, USERDATA_TAG, LS_TBREAK];
     HSwindow *win = [skin toNSObjectAtIndex:1];
     HSapplication *app = [[HSapplication alloc] initWithPid:win.pid withState:L];
-    [skin pushNSObject:app];
+    // Ensure return value is the only item on the stack to avoid leaking the receiver in coroutines
+    lua_settop(L, 0);
+    // When app is nil (process terminated), push nil directly to the caller's L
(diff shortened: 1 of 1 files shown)


cmsj:
Thanks!

PR #3839, opened 2025-12-28
Add support for Stream Deck Neo, Module 6, Module 15, and Module 32

shensquared:
Add USB Product IDs and device matching for newer Elgato Stream Deck devices:
- Stream Deck Neo (0x009A) - 8 keys (4x2), uses Mini class
- Stream Deck Module 6 (0x00B8) - 6 keys (3x2), uses Mini class
- Stream Deck Module 15 (0x00B9) - 15 keys (5x3), uses MK2 class
- Stream Deck Module 32 (0x00BA) - 32 keys (8x4), uses XL class

Device PIDs from official Elgato HID documentation: https://docs.elgato.com/streamdeck/hid/

### DIFF
--- a/extensions/streamdeck/HSStreamDeckManager.m
+++ b/extensions/streamdeck/HSStreamDeckManager.m
@@ -164,6 +164,14 @@ static void HIDdisconnect(void *context, IOReturn result, void *sender, IOHIDDev
                                           productIDKey: @USB_PID_STREAMDECK_PLUS};
         NSDictionary *matchPedal      = @{vendorIDKey:  @USB_VID_ELGATO,
                                           productIDKey: @USB_PID_STREAMDECK_PEDAL};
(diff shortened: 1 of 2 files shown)


latenitefilms:
@shensquared - Have you tested this on a Streamdeck Neo? I've already merged this code into CommandPost, but apparently there's some issues with the Neo - so not sure if this is a bug in Objective-C Extension or my Lua. I don't have the hardware, so can't easily test in the real-world.

@cmsj - What do I need to do to merge this in?

Willamin:
I tested this out with a Neo and found these issues:
* I couldn't set button images or colors
* With the button callback, I could only see events for buttons index 4, 5, and 6, which corresponded to the first three buttons of the first row
* The `buttonLayout()` returned `3 2` instead of the correct `4 2`.

PR #3841, opened 2026-01-06
Fix broken URL for Quartz Event Reference in eventtap docs

jacix:
eventtap docs point to https://developer.apple.com/hints/reference/coregraphics/quartz_event_services which is broken.  New location is https://developer.apple.com/documentation/coregraphics/quartz-event-services.


jacix:
I don't have permissions to add a label, so would one of the maintainers add label ['pr-fix']?  Thanks!

latenitefilms:
@cmsj - Not sure why the tests are failing, but this should be safe to merge as it's only updating documentation anyway

PR #3845, opened 2026-02-01
Fix memory leak from recursive print warning flood in hs.ipc

dmgerman:
The printReplacement recursion guard was using log.w() to warn about recursion, which itself triggers print, causing a cascade of warnings about the recursion. A single hs -c command could generate thousand of warning lines (which make the memory print of hammerspoon to grow extremely fast).

Changes:
- Add warnedRecursionInstances table to track warned CLI instances
- Warn only once per instance using originalPrint (not log.w)
- Clean up tracking tables when instances unregister


vladdoster:
@dmgerman,

Any update on this?

PR #3860, opened 2026-03-31
Add NSCalendarsFullAccessUsageDescription to Info.plist

### LINKED ISSUE #3803: Cannot read calendar with applescript
Hello,

I'm trying to make a module to read the event in apple calendars starting with `MEET` to add a google meet link to them. However impossible to read the calendar with apple script and hammerspoon never request permissions. 
```
local ok, result = hs.osascript.applescript([[
    tell application "Calendar"
        return count of calendars
    end tell
]])
print("OK:", ok, "Result:", result)
```
This request will return `false` and `nil` but it will request control permission after reset `org.hammerspoon.Hammerspoon`. So it ask for permission, but still fails. This command works in the script app of course.

```
local ok, result = hs.osascript.applescript([[
    tell application "Calendar"
        activate
        return "Hello mom"
    end tell
]])
print("OK:", ok, "Result:", result)
```

But this request will works ! It will focus the app and return Hello mom.

All that seems to be a permission problem that hammerspoon never ask for. The permission asked for control will not show up in "Privacy & Security" as other app like ghostty do

souljedi:
## Problem

Without `NSCalendarsFullAccessUsageDescription` in `Info.plist`, macOS silently refuses to prompt the user for Calendar access. Hammerspoon never appears in System Settings → Privacy & Security → Calendars, and all calendar-data reads (via AppleScript or EventKit) fail silently.

This was reported in #3803.

## Fix

Add the two missing privacy usage description keys, following the same wording pattern as the existing `NSCameraUsageDescription`, `NSMicrophoneUsageDescription`, etc. entries:

- `NSCalendarsFullAccessUsageDescription` — required on macOS 14+ for full EventKit access
- `NSCalendarsUsageDescription` — legacy key, honoured on macOS ≤ 13

## Testing

1. Apply the patch and build
2. Launch Hammerspoon
3. Run in the console:
   ```lua
   hs.osascript.applescript('tell application "Calendar" to get name of every calendar')
   ```
4. macOS should now show the Calendar permission dialog for Hammerspoon

Closes #3803

### DIFF
--- a/Hammerspoon/Hammerspoon-Info.plist
+++ b/Hammerspoon/Hammerspoon-Info.plist
@@ -214,6 +214,10 @@
 	<string>Hammerspoon needs to send events to automate applications.</string>
 	<key>NSAppleScriptEnabled</key>
 	<true/>
+	<key>NSCalendarsFullAccessUsageDescription</key>
+	<string>Some Hammerspoon extensions (such as hs.calendar) require full access to your calendars.</string>
+	<key>NSCalendarsUsageDescription</key>
+	<string>Some Hammerspoon extensions (such as hs.calendar) require access to your calendars.</string>
 	<key>NSCameraUsageDescription</key>
 	<string>Some Hammerspoon extensions require access to the camera.</string>
 	<key>NSHumanReadableCopyright</key>


souljedi:
Is there anything i should change? It looks like the timeout for some of the tests may have been to short?

PR #3866, opened 2026-04-13
Add getPropertyFloat and setPropertyFloat to hs.eventtap.event

tbodt:
Some undocumented event properties are double-typed. In order to use them with Hammerspoon, the API needs some way to explicitly say that you want to set a property as a float.


tbodt:
Based on this, I plan to extend hs.spaces with code to instantly switch spaces by posting trackpad gestures.

tbodt:
Missed #3859 :P if that is merged please close this as duplicate

Issue #3868, opened 2026-04-24
hs.window.filter.hasNoWindows triggers inappropriately


asp55:
The "hasNoWindows" pseudo-event triggers after every windowRejected event (rejection/destruction) so long as the filter previously had an event that set "hasWindow" to true.

Based on the description

> Pseudo-event for `hs.window.filter:subscribe()`: the windowfilter now rejects all windows

I'm pretty sure it should only fire when there are 0 windows remaining that match the filter.

dmgerman:
I am in the process of writing a new implementation is in file window_filter_new.lua The goal is to replace window_filter with this new implementation.

In theory it is a drop-in-replacement and it should fix the bug you are observing.

I am already running it instead of window_filter. Please, if you use window_filter, give it a try, and report any bugs. In my opinion it is more memory efficient and snappier than the original one, and more importantly, it should be 100% API compatible (and while doing that, I discovered--well claude discovered--a bug in rejectRegions).

The code is in this branch:

https://github.com/dmgerman/hammerspoon/tree/new-window-filter

Clone it and copy window_filter_new.lua to a new location and do at the beginning of your init.lua

   hs.window.filter = dofile('/path/to/window_filter_new.lua')
If you use it, keep an eye on the repo. I'll be updating it as errors are found. I have already added luadoc strings to clarify the new configuration parameters.

asp55:
@dmgerman interesting. I'll definitely check it out.

I also submitted a [PR that resolves this bug](https://github.com/Hammerspoon/hammerspoon/pull/3871) with 1 line change.

PR #3869, opened 2026-04-24
Replaced `wf.hasWindow` with `not next(wf.windows)` as test to determ…

asp55:
Replaced `wf.hasWindow` with `not next(wf.windows)` as test to determine whether or not to emit `hasNoWindows`

Now hasNoWindows will only emit when the last window in the window filter has been rejected.

Fix for [hs.window.filter.hasNoWindows triggers inappropriately](https://github.com/Hammerspoon/hammerspoon/issues/3868)


asp55:
Renaming branch for hygiene on my side.

PR #3871, opened 2026-05-01
Replaced `wf.hasWindow` with `not next(wf.windows)` as test

asp55:
Replaced `wf.hasWindow` with `not next(wf.windows)` as test to determine whether or not to emit `hasNoWindows`

Now hasNoWindows will only emit when the last window in the window filter has been rejected.

Fix for https://github.com/Hammerspoon/hammerspoon/issues/3868


asp55:
I think this should be labeled 'pr-fix', but I think labeling is limited to those with write access.

PR #3876, opened 2026-07-08
hs.urlevent: match event names case-insensitively (fixes #3837)

### LINKED ISSUE #3837: Registering a URL handler running macOS Tahoe does not seem to work
I have an `init.lua` file containing the following:

```
hs.urlevent.bind("myEvent", function(url)
		    hs.alert.show("URL Event received!")
end)
```

When I issue 
```
open -g hammerspoon://myEvent
```

I get the following message in the console:

```
2025-12-26 17:28:07: 17:28:07 ** Warning:  urlevent: Received hs.urlevent event with no registered callback:myevent
```

I unfortunately changed two things at the same time: Upgraded to Tahoe (26.2) and to hammerspoon 1.1.0. This was working fine in Sequoia with the previous version of hammerspoon.

Upon further investigation, it seems that if I lowercase the `eventName` like so:

```
hs.urlevent.bind("myevent", function(url)
		    hs.alert.show("URL Event received!")
end)
```
then both `open -g hammerspoon://myEvent` and `open -g hammerspoon://myevent` work. Any ideas why this is?

TowyTowy:
## Summary

Since #3691 (commit `5395f34`), the ObjC layer lowercases `url.host` before delivering it to the Lua callback, so the event name reaching `urlEventCallback` is always lowercase. But `hs.urlevent.bind()` stored the callback under the user's *original-case* event name, so `hs.urlevent.bind("myEvent", …)` could never match the delivered `myevent`, logging *"Received hs.urlevent event with no registered callback"*. This showed up on macOS Tahoe / Hammerspoon 1.1.0, the first release carrying #3691. Matches @cmsj's suggestion in the issue thread to "lowercase the name at binding time too."

## Change

Normalize event names to lowercase in both `bind()` (storage) and the dispatch lookup, so bindings match regardless of the case used in either the config or the URL. Removal via `bind(name, nil)` is normalized the same way. Docstring updated to note the case-insensitive matching.

## Repro (before this PR)

```lua
hs.urlevent.bind("myEvent", function() hs.alert.show("fired") end)
```

`open -g hammerspoon://myEvent` → *"...no registered callback: myevent"*. Lowercasing the bind name was the only workaround.

## Testing

Verified by loading the module (pre- and post-fix) with stubbed dependencies and simulating a delivered (host-lowercased) event: mixed-case binds now fire for both mixed- and lower-case URLs; lowercase binds keep working; `bind(name, nil)` unbinds across cases. A full XCTest for `urlevent` would require new `HSurlevent.m` scaffolding plus URL-event injection (not currently present); happy to add it if desired.

Fixes #3837

### DIFF
--- a/extensions/urlevent/urlevent.lua
+++ b/extensions/urlevent/urlevent.lua
@@ -71,6 +71,9 @@ local function urlEventCallback(scheme, event, params, fullURL, senderPID)
             log.wf("Something called a " .. hsScheme .. ":// URL without an action")
             return
         end
+        -- Event names are matched case-insensitively (see hs.urlevent.bind()), so
+        -- normalise the incoming event name before looking up its callback
+        event = string.lower(event)
         if not callbacks[event] then
             log.wf("Received hs.urlevent event with no registered callback:"..event)
         else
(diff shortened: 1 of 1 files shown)


cmsj:
Good catch, thanks!

PR #3877, opened 2026-07-08
Fix window animations overshooting on mixed-scale displays (#3852)

### LINKED ISSUE #3852: window:moveToScreen doesn't always preserve proportions
Hi! For a few years I've been using the following code to "flip" windows from one screen to another. Recently (last few weeks or perhaps even a couple of months?) an new issue has appeared.

How to reproduce:

1. On the larger of two screens, position a window to occupy the entire right hand half of the screen.
2. Call `flipWindow()`

Expected behaviour: window now occupies the right half of the smaller screen.
Actual behaviour: window is in the right location, but extends off the right hand side of the smaller screen.

Starting with the window filling the entire *left* half of the larger screen, the behaviour is correct.

I wonder if this this is Tahoe specific. Can I provide any further info?

### "Flip window" code

```lua
local flipWindow = function()
    local window = hs.window.focusedWindow()
    local otherScreen = hs.fnutils.find(hs.screen.allScreens(), function(s)
                                           return s ~= window:screen()
    end)
    if otherScreen ~= nil then
       -- window:moveToScreen(otherScreen)  <- originally this worked for me, but even the version below doesn't work now
       window:moveToScreen(otherScreen, false, true)
    end
end
```

### Platform & versions

I'm on Tahoe (sigh) 26.3.1 and Hammerspoon 1.1.1.

Displays have these sizes:
Image

TowyTowy:
Fixes #3852.

On multi-display setups where the displays have different scale factors, an animated `window:moveToScreen()` / `setFrame()` overshoots the right and bottom edges of the target screen. The window renders larger than requested even though `:size()` reports the correct value. Passing `duration = 0` (no animation) always lands correctly.

Cause: 04b4c3b and 540faee (both 2025-12-17) replaced the old side-effect-free Lua `_setFrame` (`setSize; setTopLeft; setSize`) with a C `_setFrame` that disables and re-enables `AXEnhancedUserInterface` around the move. The animation loop in `extensions/window/window.lua` calls `_setFrame` on every 17 ms tick, so a single animated move now toggles `AXEnhancedUserInterface` roughly a dozen times while the window crosses a scale boundary. That per-tick churn re-applies the enhanced-UI scale transform to intermediate frames and leaves the window rendered too large. The `duration = 0` path toggles it once, which is why it works.

Fix: intermediate animation frames now use the lightweight three-step resize without touching `AXEnhancedUserInterface` (the pre-December behavior, fine for throwaway interpolation frames), and only the final frame uses the EUI-managed `_setFrame`, identical to the `duration = 0` path. Side benefit: far fewer AX round-trips per animation.

Testing: verified by code trace and the reporter's measurements (their target-frame math is provably correct, so the fault is isolated to the animated apply step). I only have a single display here, so confirmation from someone with differently-scaled displays would be appreciated; the reporter's recipe reproduces it.

An alternative would be disabling EUI once at animation start and restoring at the end, but that needs state tracking across ticks and stopAnimation interruptions; this version is self-contained.

### DIFF
--- a/extensions/window/window.lua
+++ b/extensions/window/window.lua
@@ -275,16 +275,20 @@ local function animate()
   local time = timer.secondsSinceEpoch()
   for id,anim in pairs(animations) do
     local r = quadOut(time,anim.time,anim.duration)
-    local f = {}
+    local win = anim.window
     if r>=1 then
-      f=anim.endFrame
       animations[id] = nil
+      -- final frame: use the AXEnhancedUserInterface-managed _setFrame() for pixel-correct placement
+      win:_setFrame(anim.endFrame)
     else
+      local f = {}
       for _,k in pairs{'x','y','w','h'} do
         f[k] = anim.startFrame[k] + (anim.endFrame[k]-anim.startFrame[k])*r
       end
+      -- intermediate frames: plain size,position,size resize *without* toggling AXEnhancedUserInterface;
+      -- toggling it on every animation tick corrupts the window geometry on mixed-scale displays (#3852)
+      win:_setSize(f) win:_setTopLeft(f) win:_setSize(f)
     end
-    anim.window:_setFrame(f)
   end
   if not next(animations) then animTimer:setNextTrigger(DISTANT_FUTURE) end
 end


TowyTowy:
Thanks - this is a bug fix (resolves #3852), so `pr-fix` looks like the right label. I can't add it from a fork; could a maintainer apply it to clear the label check?

The red **Build and Test** run isn't related to this change. It fails in `scripts/github-ci-pre.sh` -> `brew bundle install`, before anything compiles:

```
Tapping a7ex/formulae
Refusing to load formula a7ex/formulae/xcresultparser from untrusted tap a7ex/formulae.
```

That's Homebrew's new untrusted-tap policy blocking the `xcresultparser` install; the `TestResults.xcresult`/codecov errors are downstream of it. The same failure is currently hitting `master` too, so it's a CI-environment issue rather than anything in this PR (which only touches `extensions/window/window.lua`). Happy to rebase once CI is unblocked.

(This change was prepared with AI assistance, as noted in the commit trailer.)

PR #3880, opened 2026-07-28
ci: pass the commit author name through env in the webhook notification

kobihikri:
Hi, and thanks for Hammerspoon.

In `.github/workflows/ci_testbuild.yml`, the "Send notification (Push)" step builds a JSON payload for the CI webhook. You already handle the risky part of it correctly — the commit message goes through `toJSON(...)`:

```
* ${{ toJSON(github.event.head_commit.message)}}
```

The commit **author name** on the same line does not:

```
"message": "CI for ${{ github.event.compare }} (${{ github.event.head_commit.author.name }}):\n ...
```

A commit author name is chosen by whoever made the commit — it is just `user.name` from their git config — so it arrives as untrusted text. The whole `-d` argument is single-quoted, and Actions splices the value into that string before bash parses the line, so a name containing a single quote ends the quoting early and the rest of the name is read as shell. That command also contains `${{ secrets.CI_WEBHOOK_SECRET }}` and `${{ secrets.CI_WEBHOOK_URL }}`, which is why I thought it worth reporting rather than leaving be.

To be fair about reachability: the step is `if: github.event_name == 'push' && github.repository == 'hammerspoon/hammerspoon'`, so this is a commit that has already landed on your repo — someone has to merge it. It is not open to a drive-by.

The change passes the author name through `env:` and references it as a shell variable:

```yaml
env:
  AUTHOR_NAME: ${{ github.event.head_commit.author.name }}
run: |
  ... "message": "CI for ${{ github.event.compare }} ('"$AUTHOR_NAME"'):\n ...
```

The `'"$AUTHOR_NAME"'` quoting closes the single-quoted section, expands the variable inside double quotes, and reopens it, so the value can no longer be treated as shell.

One thing I want to be upfront about: this fixes the shell side, not JSON escaping. An author name containing a double quote would still produce awkward JSON — but that is true today too, and I did not want to quietly rewrite your payload construction. If you would rather have it built properly, `jq -n --arg author "$AUTHOR_NAME" ...` would do it, and I am happy to send that version instead.

The "Send notification (PR)" step below has the same shape; I have left it alone in this PR so the change stays small, and can extend it if you would like.

Disclosure: I used AI assistance to help spot this and prepare the change, and I read the step and its conditions myself.

### DIFF
--- a/.github/workflows/ci_testbuild.yml
+++ b/.github/workflows/ci_testbuild.yml
@@ -93,8 +93,10 @@ jobs:
 
       - name: Send notification (Push)
         if: ${{ always() && github.event_name == 'push' && github.repository == 'hammerspoon/hammerspoon' }}
+        env:
+          AUTHOR_NAME: ${{ github.event.head_commit.author.name }}
         run: |
-          /usr/bin/curl -H "X-CI-WebHook: true" -H "Content-Type: application/json" -d '{"secret": "${{ secrets.CI_WEBHOOK_SECRET }}", "repository": "hammerspoon", "workflow": "CI (Push)", "message": "CI for ${{ github.event.compare }} (${{ github.event.head_commit.author.name }}):\n  * ${{ toJSON(github.event.head_commit.message)}}\n  * ${{ steps.test.outputs.test_result }}"}' "${{ secrets.CI_WEBHOOK_URL }}"
+          /usr/bin/curl -H "X-CI-WebHook: true" -H "Content-Type: application/json" -d '{"secret": "${{ secrets.CI_WEBHOOK_SECRET }}", "repository": "hammerspoon", "workflow": "CI (Push)", "message": "CI for ${{ github.event.compare }} ('"$AUTHOR_NAME"'):\n  * ${{ toJSON(github.event.head_commit.message)}}\n  * ${{ steps.test.outputs.test_result }}"}' "${{ secrets.CI_WEBHOOK_URL }}"
 
       - name: Send notification (PR)
         if: ${{ always() && github.event_name == 'pull_request' && github.repository == 'hammerspoon/hammerspoon' && env.CI_WEBHOOK_URL != '' }}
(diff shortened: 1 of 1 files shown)


cmsj:
Nice catch, thanks. I'm happy to take the most correct version of this, if that's ok? :D

kobihikri:
Thanks — yes please, and that's the better call. Pushed in d70abb2.

Both notification steps now build their payload with `jq` instead of string interpolation, so the shell quoting and the JSON escaping are both handled properly rather than just the first one:

```yaml
env:
  COMPARE_URL: ${{ github.event.compare }}
  AUTHOR_NAME: ${{ github.event.head_commit.author.name }}
  COMMIT_MESSAGE: ${{ github.event.head_commit.message }}
  TEST_RESULT: ${{ steps.test.outputs.test_result }}
run: |
  MESSAGE=$(printf 'CI for %s (%s):\n  * %s\n  * %s' "$COMPARE_URL" "$AUTHOR_NAME" "$COMMIT_MESSAGE" "$TEST_RESULT")
  jq -n --arg secret "$CI_WEBHOOK_SECRET" --arg message "$MESSAGE" \
    '{secret: $secret, repository: "hammerspoon", workflow: "CI (Push)", message: $message}' \
    | /usr/bin/curl -H "X-CI-WebHook: true" -H "Content-Type: application/json" -d @- "$CI_WEBHOOK_URL"
```

A few notes on what changed and why:

- I did the **PR** step as well, not just the push one. It has the same shape via `${{ github.event.pull_request.title }}`, which is also free text.
- `toJSON(...)` on the commit message is gone, because `jq --arg` now does that job — and it does it inside the string rather than emitting a second set of quotes in the middle of one.
- The payload goes to curl via `-d @-` on stdin, so the secret is no longer sitting in the process's argument list.
- The message format is unchanged: same `CI for <url> (<author>):` line and the same two `  * ` bullets.

I checked it against an author name containing both an apostrophe and double quotes plus a multi-line commit message, and the output is valid JSON with everything escaped.

One thing I can't do from here: the bot is asking for one of `pr-fix` / `pr-change` / `pr-feature` / `pr-maintenance`, and I don't have permission to set labels. `pr-maintenance` looks like the right one if you'd add it.

Disclosure, same as before: I used AI assistance (Claude Opus 5) to prepare this, and I tested the payload output myself.

Issue #3684, opened 2024-09-05
Crash on Sonoma due to missing Info.plist key


kriomant:
After upgrading Mac to Sonoma Hammerspoon crashes on startup.
Problem report says:

```
Termination Reason:    Namespace TCC, Code 0 
This app has crashed because it attempted to access privacy-sensitive data without a usage description. The app's Info.plist must contain an NSBluetoothAlwaysUsageDescription key with a string value explaining to the user how the app uses this data.
```

Adding this key to Info.plist helps. Hammerspoon starts.
I have problems enabling accessibility access, but this may be non-related issue.

latenitefilms:
Huh, you must be one of the few Hammerspoon users using Bluetooth for stuff! Good spotting!

latenitefilms:
Probably a few more ones we need to add from here too:

https://developer.apple.com/documentation/bundleresources/information_property_list/protected_resources

Issue #793, opened 2016-02-13
Feature Request: Add an hs.bluetooth module


EranAvidor:

Habbie:
Hello, what features are you thinking of in this module?

EranAvidor:
Hi, it should behave something like the wifi module
basic functionality should support toggling ON/OFF, get list of devices..
how does it sound?

reestr:
Perhaps the ability to enumerate the connected devices.  Also a watcher to trigger events as bluetooth devices are connected/disconnected.

chdiza:
At the very least the ON/OFF mechanism would be nice.  Even people like me who don't use Bluetooth can benefit from an easy way to ensure that it's turned off and not draining the battery.  I've been using [this](https://github.com/toy/blueutil), but it seems like it's dead or near-dead.

cmsj:
The issue here is that there is no public API for interacting with Bluetooth hardware in OSX. You might like to look at https://github.com/asmagill/hammerspoon_asm.undocumented/tree/master/bluetooth

EranAvidor:
@cmsj thanks

cmsj:
I'm going to close this out for now, since there's not much we can safely/reasonably do in the main tree. If that changes, we'll jump on it :)

retrography:
I realize this thread is old, but it finishes with an upbeat promise from @cmsj! So, here you are:

I have just found out there is [this command line tool](https://github.com/toy/blueutil) for working with Bluetooth devices using macOS's private APIs. The program is not new, but the new version is pretty powerful and its methods may be worth considering. [Here](https://gist.github.com/ysimonson/fea48ee8a68ed2cbac12473e87134f58) you can find a test config. I use it to connect/disconnect the bluetooth devices on my work desk (e.g. mouse and keyboard) each time I plug/unplug my external monitor. Very useful.

I have noticed that now `hs.battery` relies on private bluetooth API. Then why not a full extension for BT?

drn:
Thanks for pointing out [toy/blueutil](https://github.com/toy/blueutil)!

In case it's helpful for anyone else, you can easily hook into existing CLIs from HS bindings via `hs.execute`. You'll need to supply full paths to the CLIs, but it allows replication of the same functionality as you'd get with a native bluetooth hammerspoon module.

Something like:
```
function bluetooth.toggle()
  local output = hs.execute("/usr/local/bin/blueutil --connect 00-00-00-00-00-00")
[...]

PR #3881, opened 2026-08-03
Add NSBluetoothAlwaysUsageDescription to Info.plist

akalsey:
macOS requires a usage description string before an app may touch the Bluetooth APIs. Without one the system does not prompt for consent — it terminates the offending process outright, so Hammerspoon can never appear in System Settings → Privacy & Security → Bluetooth and can never be granted access.

This shows up two ways:

- Hammerspoon crashing on launch with `Termination Reason: Namespace TCC, Code 0`, whose crash report names this exact key (#3684).
- Configs that shell out to Bluetooth tools such as `blueutil`, where the child process is killed with SIGABRT. TCC attributes the access to the responsible process, so the spawned tool inherits Hammerspoon's inability to be granted Bluetooth.

Verified on macOS 26.5.2: a `blueutil` invocation from `hs.task` dies with exit 134 and no consent prompt, and `tccutil reset BluetoothAlways org.hammerspoon.Hammerspoon` does not help because there is nothing to prompt with. The same binary launched from a bundle that *does* declare the key prompts normally and works.

Adding the key does not grant access or trigger a prompt on its own — it only makes consent possible when something actually uses Bluetooth. This is deliberately not a proposal for an `hs.bluetooth` module (#793); it is just the one string that lets users grant the permission themselves.

Refs #3684

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Hammerspoon/Hammerspoon-Info.plist
+++ b/Hammerspoon/Hammerspoon-Info.plist
@@ -214,6 +214,8 @@
 	<string>Hammerspoon needs to send events to automate applications.</string>
 	<key>NSAppleScriptEnabled</key>
 	<true/>
+	<key>NSBluetoothAlwaysUsageDescription</key>
+	<string>Some Hammerspoon configurations require access to Bluetooth (for example, to connect and disconnect Bluetooth devices).</string>
 	<key>NSCameraUsageDescription</key>
 	<string>Some Hammerspoon extensions require access to the camera.</string>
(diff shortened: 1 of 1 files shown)


akalsey:
I don't have permission to add labels as an external contributor. `pr-fix` seems the right one here. This fixes the launch crash reported in #3684 rather than adding a feature. Happy for a maintainer to apply it or correct me.

PR #3896, opened 2026-09-13
Add IOHIDEvent module

tbodt:
Fiddling with dock control events in an event tap broke in macos 27. Turns out this is because when the CGEvent is translated from a kernel-level IOHIDEvent, it actually remembers the whole original IOHIDEvent, and macos 27 Dock.app decided to start ignoring the nicely translated CGEvent fields and check the IOHIDEvent directly. So it becomes necessary to be aware of IOHIDEvents. I decided to write a new lua module around the IOKit API because the alternative is directly messing with the CGEvent serialization format...


tbodt:
I'm using this module something like
```
tap = hs.eventtap.new({types.dockControl}, function(ev)
    if ev:getProperty(props.eventSourceUnixProcessID) ~= 0 then return end
    local phase = ev:getProperty(props.eventGesturePhase)
    if phase == gesturePhases.canceled then
        ev:setProperty(props.eventGesturePhase, gesturePhases.ended)
        phase = gesturePhases.ended
    end
    if phase == gesturePhases.ended then
        ev:setProperty(props.eventGestureSwipeVelocityX, ev:getProperty(props.eventGestureSwipeVelocityX) * 10000)
        ev:setProperty(props.eventGestureSwipeVelocityY, ev:getProperty(props.eventGestureSwipeVelocityY) * 10000)
        local hidevent = ev:getIOHIDEvent()
        print('before\n', hidevent:dump())
        if hidevent then
            local velEvent = hidevent:getChildren()[1]
            if velEvent then
                print(velEvent:getType())
            end
            if velEvent and velEvent:getType() == 9 then
                local velX = (9<<16)+0
                local velY = (9<<16)+1
                velEvent:setFloatProperty(velX, velEvent:getFloatProperty(velX) * 30)
                velEvent:setFloatProperty(velY, velEvent:getFloatProperty(velY) * 30)
            end
        end
        print('after\n', hidevent:dump())
        return true, {ev}
    end
end)
tap:start()
```

PR #3899, opened 2026-09-20
hs.socket: allow listening on a specific interface

### LINKED ISSUE #3634: hs.socket and hs.socket.udp listens on all interfaces
neither TCP or UDP server sockets allow binding to a specific interface or address, eg https://github.com/Hammerspoon/hammerspoon/blob/master/extensions/socket/libsocket_udp.m#L209

it would be nice to support listening on a specific interface, eg communication on a bridge interface with a VM without exposing the server to the outside world.

anandghegde:
Fixes #3634.

Both `hs.socket` and `hs.socket.udp` server sockets currently bind to every interface, with no way to narrow that down. As @cmsj noted on the issue, CocoaAsyncSocket already exposes the APIs for this:

> CocoaAsyncSocket supports binding to either interface or address, and I agree we should do this.

### What changed

`listen()` on both extensions takes an optional `interface` string:

```lua
-- only reachable from this machine
server = hs.socket.server(9001, "loopback", callback)

-- only on the bridge interface to a VM
hs.socket.udp.new(callback):listen(9001, "bridge100")
```

The interface can be a name (`"en1"`, `"lo0"`), the corresponding IP address (`"192.168.4.34"`), or the special strings `"localhost"`/`"loopback"` — whatever `GCDAsyncSocket` accepts.

* `socket_listen()` now calls `-acceptOnInterface:port:error:` instead of `-acceptOnPort:error:`
* `socketudp_listen()` now calls `-bindToPort:interface:error:` instead of `-bindToPort:error:`
* `hs.socket.server()` / `hs.socket.udp.server()` take the same optional argument between the port and the callback

### Backwards compatibility

Passing no interface is unchanged. In CocoaAsyncSocket, `-acceptOnPort:error:` is literally `return [self acceptOnInterface:nil port:port error:errPtr];`, and `-bindToPort:error:` is `return [self bindToPort:port interface:nil error:errPtr];`, so the nil path is the same code as before.

The `server()` constructors shuffle their arguments when the second one is a function, so `hs.socket.server(port, fn)` keeps working.

Unix domain sockets aren't bound to an interface, so `hs.socket:listen(path, interface)` logs a warning and ignores the interface rather than failing.

### Tests

Five new tests in `test_socket.lua` / `test_udpsocket.lua`, registered in `Hammerspoon Tests/HSsocket.m`:

* `testTcpListenerSocketInterfaceAttributes` / `testUdpListenerSocketInterfaceAttributes` — a listener bound to `"loopback"` reports `localHost == "127.0.0.1"` rather than `"0.0.0.0"`
* `testTcpListenerSocketInterfaceConnect` — a client can still connect to a loopback-bound server and the server reads what it wrote
* `testTcpListenerSocketBadInterface` / `testUdpListenerSocketBadInterface` — an unknown interface returns `nil` and leaves the socket unbound; a Unix domain socket ignores the interface

Local run of the `HSsocket` suite: 46 passed, 1 failed. The failure is `testUdpReusePort`, which also fails on an unmodified `master` here — it broadcasts to `255.255.255.255`, which this machine doesn't allow.


anandghegde:
CI is asking for one of the `pr-fix` / `pr-change` / `pr-feature` / `pr-maintenance` labels, and I can't set labels as an outside contributor. This one is `pr-feature` — could a maintainer add it so the check goes green? Happy to rebase if anything has moved in the meantime.

